PR #6, opened 2016-09-28
Fix Xcode project after rename

soffes:
This also restores Carthage support

### DIFF
--- a/SwiftProtobufRuntime.xcodeproj/project.pbxproj
+++ b/SwiftProtobufRuntime.xcodeproj/project.pbxproj
@@ -7,234 +7,238 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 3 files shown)


allevato:
I made a similar change in #20 before realizing this was also here. It looks like you deleted-and-readded the source files instead of pointing the existing group at the new directory location, so the .pbxproj changes are much bigger because it regenerated all the file IDs.

In the interest of keeping a clean commit log, I suggest merging #20 instead, and then we can take another PR that adds the change for the extension-only APIs. Sound good?

thomasvl:
I'm going to go ahead and close this since I think most of the original intent is now landed.  From the one comment, it sounds like Carthage support might be busted, but I'm not sure what is needed to fix that.

soffes:
I can confirm Carthage works in 0.9.24

PR #20, opened 2016-09-29
Fix Xcode project after module renaming.

### LINKED ISSUE #19: Xcode project needs to be fixed after reorganizing source files
Since the plugin was merged and the directory structure changed, the Xcode project also needs to be updated. It should have its module/targets changed from "Protobuf" to "SwiftProtobuf" as well.

allevato:
The Sources/Tests had been moved without updating the project, and this
change also renames the targets to SwiftProtobuf to match the Swift
package description.

Fixes #19.


allevato:
Oops, I didn't notice that @soffes already made a similar PR (#6) for this. However, the .pbxproj changes in that one are much larger (too large to be viewed by Github), which is a bit odd. It's probably worth keeping the extension-only-API change in a separate PR as well, instead of rolling it into the same one.

thomasvl:
fyi - you can click on the individual comments in a PR to see them, so the second commit in that PR is pretty small (to support extensions).  But the initial commit for project cleanup is pretty large.

tbkka:
Looks good.   I'll merge this and then we can look at #6 to see what else is going on there.

Issue #10, opened 2016-09-28
Revisit isEmpty methods


thomasvl:
Currently generating isEmpty methods, but they might actually not be needed.  It doesn't seem like the other languages need them, and if folks build their protos into modules/frameworks, they might not deadstrip, bloating things up.  So if they aren't really needed (maybe for the current isEqualTo (and that might not be fully correct)), dropping them is likely the correct call.

jcanizales:
\<Standard reminder that adding stuff later if there's demand is always possible, but removing isn't.\>

tbkka:
I think the generated `isEmpty` is only still being used in one case:  the equality test on messages uses it to ensure that a message with no allocated storage class is considered equal to one where the storage class is allocated but "empty."

An alternative approach would be to always allocate the backing storage object.  Then this case never arises.  Empty objects (allocated but with no fields set) should be uncommon, so I don't see a big drawback.  It would also simplify the `uniqueStorage()` implementation.

PR #21, opened 2016-09-30
Issue #10: remove isEmpty

tbkka:
This removes the isEmpty property from the generated code (and quite a few other places).

There were a couple of uses of this still in the tree that had to be edited:
  * It was used in some equality checks to treat an unmodified message struct as equivalent to a missing message struct.  This seemed to be confusing, so is dropped.
  * The internal generated code ensured that a not-yet-allocated storage object was treated as equal to an allocated-but-unmodified storage object.  This was necessary to make the message structs behave identically whether they used a storage object or not.  This was changed so that the storage object is no longer optional; it's always allocated so its lack never needs to be considered.

This is a big diff primarily because of all the regenerated reference protos.  It's split into separate commits for easier review.


thomasvl:
Forgot to add, thanks for splitting this up; it was much easier to read. And since github finally support _squash and merge_, it can still ands as one unit.

Reference/conformance/conformance.pb.swift
```diff
@@ -2018,352 +1886,338 @@ public struct Conformance_TestAllTypes: ProtobufGeneratedMessage {
...
-        }
-        return isEmpty // empty storage == nil storage
-      } else if let os = other._storage {
-        return os.isEmpty // nil storage == empty storage
-      }
-      return true // Both nil, both empty
+      return _storage === other._storage || _storage.isEqualTo(other: other._storage)
     }
 
     private mutating func _uniqueStorage() -> _StorageClass {
```
tbkka: Good point.

tbkka:
I'm happy updating the documentation separately.

I tried to comment each separate commit clearly, so I think it would be fine without the squash.

thomasvl:
Re squash - My default is usually to ensure any commit is always a valid building state.  If someone pointed at some of these hash, it seems like things might not build (mismatches between generated code and tests, etc.).

Anyway.  Looks good.  I'll let you merge it which ever way you want.

tbkka:
Another follow-up:  There's a lot of duplication between the "Reference" folder (which was conceived as a way to test the plugin by itself separately from the runtime library) and the generated parts of the "Tests" folder.

There's certainly value in verifying the exact generated output -- I've caught a fair number of bugs that way -- so I'm reluctant to just delete the "Reference" folder and the associated `test-plugin` Makefile target.

thomasvl:
Yup, but having the Reference directory to help see changes to the generator is really useful.  Maybe stick a README.md in there also saying why it exists and consumers of the library can safely ignore it?

jcanizales:
Re squash: If one does it manually, it used to obliterate the review history of the pull request (as comments were attached to specific commits). Do you know if that's still a problem, or if GitHub's new "automatically squash" feature solves it in some way? I haven't squashed PRs since forever for this reason, so I don't know.

thomasvl:
https://github.com/google/protobuf/pull/1970 was landed with a squash. They default to hiding since the commits are "gone", but you can still expand out the full history to see things.

tbkka:
Seems to work nicely if you're a little careful:  the squashed commit

https://github.com/apple/swift-protobuf/commit/89bdab722ba396160d7c7df92cd31031a3944c6b

references this PR in the commit message, so people can find details if they need them.

jcanizales:
From clicking around google/protobuf#1970, it looks like GitHub is storing the snippet of context for each hidden comment, but won't let you peek any further above (and clicking the "View full outdated diff" link unsurprisingly leads to a nice 404). It's an improvement that they don't delete the comments anymore, although still not what I'd like personally :)

thomasvl:
I wonder if my dropping the branch causes that.  Otherwise, the source of the PR would still have the full files.

Issue #8, opened 2016-09-28
Merge runtime and plugin repositories


tbkka:
Having the runtime library and the plugin program in separate repositories is proving quite awkward.

Here's the proposal:
  * Move the plugin source and tests into `swift-protobuf-runtime`
  * Make sure everything builds, runs, etc. correctly
  * Deprecate the `swift-protobuf-plugin` repository -- delete the contents and change README to explain that the plugin is in a new place now.
  * Rename `swift-protobuf-runtime` repository to simply `swift-protobuf`

Since this is a pretty disruptive change, I'd like to do this pretty soon.  Please comment if you have any concerns about this.

tbkka:
I've copied all of the plugin bits into the runtime repository and spent a couple of hours fixing up the Makefile and trying to make sure everything works.

Next step:  I'd like some feedback from other people that the combined repository actually works for them (that I didn't miss anything).

Once that's done, I propose:
  * Asking the admins to rename `swift-protobuf-runtime` to `swift-protobuf`
  * Removing all code from `swift-protobuf-plugin`
  * Editing `README.md` in `swift-protobuf-plugin` to point to the combined repository.
(In particular, `swift-protobuf-plugin` seems to be the repository that people are pointing to in blog posts, etc, so it would be nice to keep that there as a redirect.)

allevato:
The combined repo is working for me, after updating the Xcode project in #20. Since they're in the same repo now I'd like to add a command-line target for `protoc-gen-swift` as well, but that's not a blocker.

Let's move on renaming the repo ASAP so that we don't collect debt in our own forks (I don't use Github on multi-person projects enough to know what happens if I have a fork and its origin gets renamed).

jcanizales:
I suspect we'll have to recreate our personal GitHub forks. Once that's done, it's just a matter of removing and adding again the remotes configured in your local clone, to point to the correct URLs. Your local branches shouldn't be affected at all, given that the history tree of the remote repo will be the same under the new name.

tbkka:
The `swift-protobuf-runtime` repository has been renamed.

allevato:
Thanks!

It looks like we don't need to recreate our forks—Github detected the change and knew that "allevato/swift-protobuf-runtime" was still a fork of "apple/swift-protobuf", so I just renamed my fork to match and then did a couple `git remote set-url` to update my local repo to match.

tbkka:
[...]

PR #25, opened 2016-09-30
Update Podspec

radex:
In line with changes made in #8 + update version tag (the `version` field in the Podspec requires a bump every time a new version tag is released)


tbkka:
Someday, we're going to have build a proper release script that can copy the version number into all the right places:  the Podspec file, the protoc-gen-swift copyright/version banner, etc.

PR #27, opened 2016-09-30
Create a performance testing script.

allevato:
This script will let us measure the performance (build time, execution
time, and binary size) of generated protos and the runtime library using
a message with various numbers and types of fields (provided via the
command line).

It builds and runs against the runtime and plugin in the parent
directory, so local changes can be easily tested and compared.

### DIFF
--- a/Performance/.gitignore
+++ b/Performance/.gitignore
@@ -1,4 +1,2 @@
 Generated/
 Results/
-Sources/*+Generated.swift
-Sources/*.pb.swift
--- a/Performance/.gitignore
+++ b/Performance/.gitignore
@@ -1,0 +1,4 @@
+Generated/
+Results/
+Sources/*+Generated.swift
+Sources/*.pb.swift
--- a/Performance/.gitignore
+++ b/Performance/.gitignore
@@ -1,2 +1,2 @@
-Generated/
-Results/
+_generated/
+_results/
--- a/Performance/perf_runner.sh
+++ b/Performance/perf_runner.sh
@@ -1,0 +1,212 @@
+#!/bin/bash
+
+# SwiftProtobuf/Performance/perf_runner.swift - Performance test runner
+#
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See http://swift.org/LICENSE.txt for license information
+# See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+#
+# -----------------------------------------------------------------------------
+#
+# This script creates a .proto file with the requested number of fields of a
+# particular type and generates a test harness that populates it, serializes
+# it, and then parses it back out. Then it generates the Swift source for the
+# .proto file, builds the harness, and runs it under Instruments to collect
+# timing samples.
+#
+# The results are stored in the Performance/Results directory. The instruments
+# traces are named based on the number and type of fields; multiple runs with
+# the same inputs will append to the existing trace file if it is present.
+#
+# In addition to the collected samples, this script also prints the time it
+# takes to compile/link the harness and the byte size of the harness before and
+# after stripping.
+#
+# -----------------------------------------------------------------------------
+
+set -eu
+
+function usage() {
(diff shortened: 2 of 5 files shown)


Performance/.gitignore
```diff
@@ -0,0 +1,4 @@
+Generated/
+Results/
+Sources/*+Generated.swift
```
thomasvl: Will generating things in here cause problems if they are laying around when someone builds in general?

allevato: These files are generated in /Performance/Generated, so what goes on in here shouldn't have any effect on general building where it only looks in /Sources.

And actually that line 3 in the .gitignore can be removed, it's a remnant of another attempt that didn't work where I tried to build the performance tester with Swift PM as well.

Performance/perf_runner.sh
```diff
@@ -0,0 +1,212 @@
...
+Options:
+    -p <true|false>: Adds a packed option to each field.
+    -s[23]:          Generate proto2 or proto3 syntax. proto3 is
+                     the default.
+EOF
+  exit 1
+}
+
+function mkdir_if_necessary() {
+  mkdir "$@" 2>/dev/null || true
```
tbkka: Next time, you might try `mkdir -p` which succeeds if the directory already exists.

Performance/perf_runner.sh
```diff
@@ -0,0 +1,212 @@
...
+fi
+
+readonly field_count=$1
+readonly field_type=$2
+readonly script_dir="$(dirname $0)"
+
+# Make sure the runtime and plug-in are up to date first.
+pushd "$script_dir/.." >/dev/null
+swift build -c release
+popd >/dev/null
```
tbkka: Rather than pushd/popd, for one-liners like this, I usually prefer to use a sub shell:
```
(cd "$script_dir/.."; swift build -c release)
```

Performance/perf_runner.sh
```diff
@@ -0,0 +1,212 @@
...
+results="$script_dir/Results/$field_count fields of $field_type"
+harness="$script_dir/Generated/harness"
+
+echo "Generating test proto with $field_count fields..."
+generate_test_proto "$field_count" "$field_type"
+
+echo "Generating test harness..."
+generate_perf_harness "$field_count" "$field_type"
+
+protoc --plugin="$script_dir/../.build/release/protoc-gen-swift" \
```
tbkka: In my experimenting, this did not seem to fail when the specified plugin didn't exist but there was a `protoc-gen-swift` somewhere in the current `PATH`.

In the top-level `Makefile`, I instead copied `protoc-gen-swift` to `protoc-gen-swiftX` and then invoked `--swiftX_out` to ensure that the local built copy will be used and the protoc command will fail if it couldn't be used for any reason.

Performance/perf_runner.sh
```diff
@@ -0,0 +1,212 @@
...
+    "$gen_harness_path" \
+    "$script_dir/Harness.swift" \
+    "$script_dir/Generated/message.pb.swift" \
+    "$script_dir/main.swift" \
+)
+echo
+
+echo "Running test harness in Instruments..."
+instruments -t "Time Profiler" -D "$results" "$harness"
+open "$results.trace"
```
tbkka: Nice touch!

Performance/.gitignore
```diff
@@ -0,0 +1,2 @@
+Generated/
+Results/
```
tbkka: Thanks for segregating built content into separate directories like this.

Recently, I've started naming directories like this with a leading underscore, e.g. `_generated` or `_results`.  That helps distinguish them from directories that are part of the repo.

Performance/perf_runner.sh
allevato: Done.

Performance/perf_runner.sh
allevato: Good one. I don't do as much shell scripting as I probably should, so tips like this are always appreciated.

Performance/perf_runner.sh
allevato: Good point. I'll add in logic that does the same copying approach here.

Performance/.gitignore
allevato: Seems reasonable, done.

PR #33, opened 2016-10-03
Fix import module for Tests due to the module rename

ontouchstart:
See https://github.com/apple/swift-protobuf/commit/6b1a0b0a270f31b93c3d4255d94f8d8f2f9e1ae0


allevato:
This looks good. Thanks!

PR #43, opened 2016-10-04
Binary serialization/decoding performance improvements

allevato:
This PR contains the following changes:

* Precalculate the space required to serialize a message in a separate pass before writing it, so the entire buffer can be allocated once up front.
* Avoid very costly `[UInt8]`-to-`Data` conversions and vice versa by rewriting the encoder/decoder to manipulate underlying memory directly.
* Add helpers to concentrate zig-zag encoding and varint size calculation into one place.

In a basic test of 100 `fixed32` fields, these changes were able to speed up serialization by 3500% (!) and decoding by about 16%.

### DIFF
--- a/Sources/SwiftProtobuf/ProtobufBinaryEncoding.swift
+++ b/Sources/SwiftProtobuf/ProtobufBinaryEncoding.swift
@@ -54,7 +54,7 @@ public struct ProtobufBinaryEncodingVisitor: ProtobufVisitor {
         encoder.startField(tagType: protoFieldNumber * 8 + 2)
         var packedSize = 0
         for v in value {
-            packedSize += try S.encodedSize(of: v)
+            packedSize += try S.encodedSizeWithoutTag(of: v)
         }
         encoder.putVarInt(value: packedSize)
         for v in value {
@@ -99,7 +99,7 @@ public struct ProtobufBinaryEncodingVisitor: ProtobufVisitor {
             encoder.startField(tagType: protoFieldNumber * 8 + 2)
             let keyTagSize = Varint.encodedSize(of: UInt32(truncatingBitPattern: 1 << 3))
             let valueTagSize = Varint.encodedSize(of: UInt32(truncatingBitPattern: 2 << 3))
-            let entrySize = try keyTagSize + KeyType.encodedSize(of: k) + valueTagSize + ValueType.encodedSize(of: v)
+            let entrySize = try keyTagSize + KeyType.encodedSizeWithoutTag(of: k) + valueTagSize + ValueType.encodedSizeWithoutTag(of: v)
             encoder.putVarInt(value: entrySize)
             encoder.startField(tagType: 8 + KeyType.protobufWireType())
             KeyType.serializeProtobufValue(encoder: &encoder, value: k)
--- a/Sources/SwiftProtobuf/ProtobufBinaryEncoding.swift
+++ b/Sources/SwiftProtobuf/ProtobufBinaryEncoding.swift
@@ -19,11 +19,10 @@ import Foundation
 import Swift
 
 public struct ProtobufBinaryEncodingVisitor: ProtobufVisitor {
-    private var encoder = ProtobufBinaryEncoder()
+    private var encoder: ProtobufBinaryEncoder
 
-    public var buffer: [UInt8] {return encoder.buffer}
-
-    public init(message: ProtobufMessageBase) throws {
+    public init(message: ProtobufMessageBase, pointer: UnsafeMutablePointer<UInt8>) throws {
+        encoder = ProtobufBinaryEncoder(pointer: pointer)
         try withAbstractVisitor {(visitor: inout ProtobufVisitor) in
             try message.traverse(visitor: &visitor)
         }
@@ -32,7 +31,7 @@ public struct ProtobufBinaryEncodingVisitor: ProtobufVisitor {
     public mutating func withAbstractVisitor(clause: (inout ProtobufVisitor) throws ->()) throws {
         var visitor: ProtobufVisitor = self
         try clause(&visitor)
-        encoder.buffer = (visitor as! ProtobufBinaryEncodingVisitor).encoder.buffer
+        encoder = (visitor as! ProtobufBinaryEncodingVisitor).encoder
     }
 
     mutating public func visitUnknown(bytes: [UInt8]) {
@@ -53,11 +52,14 @@ public struct ProtobufBinaryEncodingVisitor: ProtobufVisitor {
 
     mutating public func visitPackedField<S: ProtobufTypeProperties>(fieldType: S.Type, value: [S.BaseType], protoFieldNumber: Int, protoFieldName: String, jsonFieldName: String, swiftFieldName: String) throws {
         encoder.startField(tagType: protoFieldNumber * 8 + 2)
-        var subencoder = ProtobufBinaryEncoder()
+        var packedSize = 0
         for v in value {
-            try S.serializeProtobufValue(encoder: &subencoder, value: v)
+            packedSize += try S.encodedSize(of: v)
         }
-        encoder.putBytesValue(value: subencoder.buffer)
+        encoder.putVarInt(value: packedSize)
+        for v in value {
+            try S.serializeProtobufValue(encoder: &encoder, value: v)
+        }
     }
 
     mutating public func visitSingularMessageField<M: ProtobufMessage>(value: M, protoFieldNumber: Int, protoFieldName: String, jsonFieldName: String, swiftFieldName: String) throws {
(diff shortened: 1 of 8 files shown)


Sources/SwiftProtobuf/ProtobufBinarySizeVisitor.swift
```diff
@@ -0,0 +1,119 @@
...
+        let tagSize = Varint.encodedSize(of: UInt32(truncatingBitPattern: protoFieldNumber << 3))
+        let keyTagSize = Varint.encodedSize(of: UInt32(truncatingBitPattern: 1 << 3))
+        let valueTagSize = Varint.encodedSize(of: UInt32(truncatingBitPattern: 2 << 3))
+        for (k,v) in value {
+            let entrySize = try keyTagSize + KeyType.encodedSize(of: k) + valueTagSize + ValueType.encodedSize(of: v)
+            serializedSize += entrySize + Varint.encodedSize(of: Int64(entrySize))
+        }
+        serializedSize += value.count * tagSize
+    }
+}
```
thomasvl: Does it make sense to expose an extension on message to provide serializedSize?  The other languages tend to expose this as it can be useful for developers.

Sources/SwiftProtobuf/ProtobufBinaryTypes.swift
```diff
@@ -606,6 +666,12 @@ public extension ProtobufString {
         }
         throw ProtobufDecodingError.invalidUTF8
     }
+
+    public static func encodedSize(of value: String) -> Int {
+        let stringWithNul = value.utf8CString
```
thomasvl: Is there a perf difference between this an using value.utf8.count?

Sources/SwiftProtobuf/ProtobufBinarySizeVisitor.swift
allevato: This PR adds `serializedSize()` on messages [here](https://github.com/apple/swift-protobuf/pull/43/files#diff-fcfd4c47681aa25d602aab7dba779084R753), which does *not* include the length varint before it. There's also the `encodedSize()` method in the various binary type protocols that does include it, but that's not exposed to clients. Did you mean the first or the second one?

Sources/SwiftProtobuf/ProtobufBinaryTypes.swift
allevato: I haven't looked into that yet—I copied the logic currently being used in `ProtobufBinaryEncodingVisitor` for consistency. Later I plan to dig more into strings, since they're our biggest serialization bottleneck right now, especially compared to Objective-C.

thomasvl: Want dump this idea into an issue so we don't drop it?  (I wasn't sure the if `String.UTF8View` would be cheaper since this seems to make a cstring just to get the length.

Sources/SwiftProtobuf/ProtobufBinarySizeVisitor.swift
allevato: `encodedSize()` is an implementation detail; users don't see it. For each of the `ProtobufBinary*` types, it answers the question "how much space does this thing need on the wire after the tag is written", so in the case of messages it has to include that length prefix. The size visitor calls it abstractly to get the size of each field, whether it's an int, a string, a message, etc.

Sources/SwiftProtobuf/ProtobufBinaryEncoding.swift
```diff
@@ -53,11 +52,14 @@ public struct ProtobufBinaryEncodingVisitor: ProtobufVisitor {
 
     mutating public func visitPackedField<S: ProtobufTypeProperties>(fieldType: S.Type, value: [S.BaseType], protoFieldNumber: Int, protoFieldName: String, jsonFieldName: String, swiftFieldName: String) throws {
         encoder.startField(tagType: protoFieldNumber * 8 + 2)
-        var subencoder = ProtobufBinaryEncoder()
```
tbkka: Oh, good.  You managed to avoid the sub encoder here.

Sources/SwiftProtobuf/ProtobufBinarySizeVisitor.swift
thomasvl: If it is internal, should we put some prefix on it?  Anything that looks like a getter/setter off Message has the potential to collide with proto fields.

Sources/SwiftProtobuf/ProtobufBinaryTypes.swift
allevato: Filed #50.

Sources/SwiftProtobuf/ProtobufBinarySizeVisitor.swift
allevato: Let me look into whether we can use access controls to solve this instead. If we make the necessary protocols internal to the module, we may be able to work it out in such a way that clients outside the module wouldn't be able to see it.

If that doesn't work, prefixing is the best alternative.

Sources/SwiftProtobuf/ProtobufBinaryTypes.swift
```diff
@@ -36,6 +36,13 @@ public protocol ProtobufBinaryCodableType: ProtobufTypePropertiesBase {
 
     // Special interface for decoding a value of this type as a map value.
     static func decodeProtobufMapValue(decoder: inout ProtobufFieldDecoder, value: inout BaseType?) throws
+
+    /// Returns the number of byte required to encode `value` on the wire.
+    ///
+    /// Note that for length-delimited data (such as strings, bytes, and messages), the returned
+    /// size includes the space required for the length prefix. For messages, this is a subtle
+    /// distinction from the `serializedSize()` method, which does *not* include the length prefix.
+    static func encodedSize(of value: BaseType) throws -> Int
```
tbkka: I like how `encodedSize(of:)` reads, but there's a hard naming problem here.

There are several places where we need to distinguish:
* A full protobuf **field** (including leading tag)
* A protobuf **field body** (including size varint, where appropriate)
* A protobuf **field contents** (not including size varint) -- this is typically the same as the encoded object, so doesn't necessarily need a separate term.

It would be great to have a clear convention about how to name these that we can use throughout the library.

What do you think about `fieldBodySize(of:)` or `encodedFieldBodySize(of:)`?

Sources/SwiftProtobuf/ProtobufBinaryTypes.swift
```diff
@@ -137,6 +145,10 @@ public extension ProtobufFloat {
         }
         return true
     }
+
+    public static func encodedSize(of value: Float) -> Int {
+        return MemoryLayout<Float>.size
```
tbkka: Good!  I found `MemoryLayout<>` just recently and have been meaning to go through and adopt it.

Sources/SwiftProtobuf/ProtobufBinaryTypes.swift
```diff
@@ -679,6 +750,7 @@ public protocol ProtobufBinaryMessageBase: ProtobufMessageBase {
     // Serialize to protobuf
     func serializeProtobuf() throws -> Data
     func serializeProtobufBytes() throws -> [UInt8]
+    func serializedSize() throws -> Int
```
tbkka: For consistency, this should be `serializedProtobufSize()`  (Yes, I know you think `Protobuf` here is implied.  ;-)

Sources/SwiftProtobuf/ProtobufBinaryTypes.swift
```diff
@@ -679,6 +750,7 @@ public protocol ProtobufBinaryMessageBase: ProtobufMessageBase {
     // Serialize to protobuf
     func serializeProtobuf() throws -> Data
     func serializeProtobufBytes() throws -> [UInt8]
```
tbkka: Related:  I've been planning to drop whichever of `Data` or `[UInt8]` is significantly slower.  If you're confident that the `Data` approach is really going to be the winner, we could start stripping out the `[UInt8]` versions.

(If we found that `Data` was slower, we would still need to support `Data` at the API boundary for interoperability reasons, but could stick with `[UInt8]` internally.)

tbkka:
You should go ahead and merge this -- I made a couple of naming suggestions which you can take or leave as you see fit.

Sources/SwiftProtobuf/ProtobufBinaryTypes.swift
allevato: It's easy enough to modify the performance harness to try each one, so I'll do that as part of #44 (which I filed to ask the very same question 😄). My hope is that `Data` is the better one because it's what other file/stream APIs are going to return to clients, and since `Data` to `[UInt8]` conversion is very costly, they would have to stay as two separate code paths anyway.

Sources/SwiftProtobuf/ProtobufBinaryTypes.swift
allevato: I'll rename it for now, for consistency, and pick that fight again later 😉

Sources/SwiftProtobuf/ProtobufBinaryTypes.swift
allevato: The other protobuf implementations fairly consistently use some form of `NoTag` in the names to describe this (i.e., the size of whatever would come after the tag, including the length if present), so in the interest of improving a reader's ability to map concepts between the Swift library and others, maybe `encodedSizeWithoutTag(of:)` would be best.

Sources/SwiftProtobuf/ProtobufBinarySizeVisitor.swift
jcanizales: Nice!

What would be the difference, if any, of not using a generic there:
```swift
public func callTheInternalMethod(_ p: PublicProtocol) {
  p.internalMethod()
}
```

allevato: That also works in this specific case, because the protocol isn't an existential type; i.e., it doesn't have any associated type requirements. If it did, it could only be used be used as a generic type constraint, so we'd have to go with the `<P: PublicProtocol>` variant.

It would be interesting to look at the affects on code size between the two in the situations when they're both well-formed; I would hope `swiftc` is good enough to not create extra specializations when they obviously aren't needed.

Issue #51, opened 2016-10-05
Lock down visibility of as much of the runtime as possible


allevato:
A lot of types in the runtime library are public when they probably don't need to be. By making certain implementation details (especially extensions on public protocols) internal instead of public, we can be sure that those names won't collide with generated message members that happen to share those names. In other words, we significantly reduce the number of collisions we have to worry about, _and_ make sure that the user-facing API of a generated message is as small as possible.

thomasvl:
Tagging as API changing since changing visibility adds/removes symbols.

allevato:
Also, for things that have to be public as implementation details, move them into an extension of the `Internal` struct if possible (i.e., they're not generic or protocols).

thomasvl:
I believe I've completely this, but before closing this out; @tbkka & @allevato can you both make a sweep of all the files/types to confirm what is public makes sense to you?

tbkka:
I've gone through a `grep public` of the library code and it looks pretty good.

allevato:
Ditto; I think we're done here for 1.0.

PR #55, opened 2016-10-05
Restrict visibility of some types to internal.

allevato:
Namely, the encoding/description/hash/mirror visitors can be internal since they are only used
by extensions within the runtime.

There's more work to be done here; hopefully more parts of the decoders
can be internalized as well, but there are protocol/generic constraint
relationships that need to be pulled apart first. That will have to
happen as part of the larger segregation of JSON/text from binary
effort.

Relates to #51.


Sources/SwiftProtobuf/ProtobufUtility.swift
```diff
@@ -17,7 +17,7 @@
 
 import Swift
 
-public func ==(lhs: [[UInt8]], rhs: [[UInt8]]) -> Bool {
+func ==(lhs: [[UInt8]], rhs: [[UInt8]]) -> Bool {
```
tbkka: FYI:  It's possible these `[[UInt8]]` helpers can just be dropped.  They're mostly leftovers from when `bytes` was being rendered as `[UInt8]` but now we're mapping `bytes` to `Data`, so these may no longer be used at all.

Sources/SwiftProtobuf/ProtobufUtility.swift
```diff
@@ -27,11 +27,11 @@ public func ==(lhs: [[UInt8]], rhs: [[UInt8]]) -> Bool {
     return true
 }
 
-public func !=(lhs: [[UInt8]], rhs: [[UInt8]]) -> Bool {
+func !=(lhs: [[UInt8]], rhs: [[UInt8]]) -> Bool {
     return !(lhs == rhs)
 }
 
-public func ==<T: Equatable>(lhs: Dictionary<T, [UInt8]>, rhs: Dictionary<T, [UInt8]>) -> Bool {
+func ==<T: Equatable>(lhs: Dictionary<T, [UInt8]>, rhs: Dictionary<T, [UInt8]>) -> Bool {
```
tbkka: Ditto here:  `map<T, bytes>` now gets rendered as `Dictionary<T, Data>` instead of `Dictionary<T,[UInt8]>`, so these helpers are probably no longer needed at all.

Sources/SwiftProtobuf/ProtobufUtility.swift
allevato: Yes, it looks like these can be safely deleted now.

Sources/SwiftProtobuf/ProtobufUtility.swift
allevato: Ditto.

allevato:
Right—I'm not too concerned about users wanting to override those and missing the visitors because of the reasons you mentioned (and AFAIK it rarely comes up in practice, especially on an individual message level). Plus, since the visitors are structs, there are no real useful extension points; they can't be inherited and overridden piecemeal anyway, so it's all or nothing.

Issue #57, opened 2016-10-05
Clean up Xcode project


thomasvl:
Something up, both schemes seem to list macOS, iOS, and watchOS targets for running/testing.  So something setting was is confusing Xcode.  It would be nice to figure that out and get a project that acts a little more normal.

thomasvl:
I think I might have figured out part of the issues, I'll try to clean things up a bit.

PR #61, opened 2016-10-06
Issue57 project cleanup round1

thomasvl:
There is still more I'll look at cleaning up/syncing between the platforms, but this seems like a reasonable first cut, and actually provides a project that responds a little better since it isn't trying to deal with mismatched platforms.

Progress on #57

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,7 +120,22 @@ PLUGIN_PROTOS= \
 	regenerate-test-protos \
 	test \
 	test-plugin \
-	test-runtime
+	test-runtime \
+	test-xcode \
+	test-xcode-debug \
+	test-xcode-release \
+	test-xcode-iOS \
+	test-xcode-iOS-debug \
+	test-xcode-iOS-release \
+	test-xcode-macOS \
+	test-xcode-macOS-debug \
+	test-xcode-macOS-release
+
+.NOTPARALLEL: \
+	test-xcode-iOS-debug \
+	test-xcode-iOS-release \
+	test-xcode-macOS-debug \
+	test-xcode-macOS-release
 
 default: build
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -235,3 +250,40 @@ regenerate-test-protos: ${PROTOC_GEN_SWIFTX}
 	for t in ${TEST_PROTOS}; do \
 		${PROTOC} --plugin=${PROTOC_GEN_SWIFTX} --swiftX_out=Tests/SwiftProtobufTests -I Protos Protos/$$t; \
 	done;
+
+
+# Helpers to put the Xcode project through all modes.
+
+# Grouping targets
+test-xcode: test-xcode-iOS test-xcode-macOS
```
tbkka: Adding `test-xcode` is a nice touch.  Thanks for doing that.

Makefile
```diff
@@ -235,3 +250,40 @@ regenerate-test-protos: ${PROTOC_GEN_SWIFTX}
...
+	  -scheme SwiftProtobufRuntime_iOS \
+	  -configuration Release \
+	  -destination "platform=iOS Simulator,name=iPhone 6s,OS=latest" \
+	  test
+
+test-xcode-macOS-debug:
+	xcodebuild -project SwiftProtobufRuntime.xcodeproj \
+	  -scheme SwiftProtobufRuntime_macOS \
+	  -configuration debug \
+	  build test
```
thomasvl: no, test and build are different setups in the schemes, nothing says they overlap.  build more directly maps to what would run in the UI.  At the moment, the targets end up the same, but this makes it a little more future proof for macos.

Makefile
thomasvl: np.  not sure if you saw (I'm catching up on email), i backed this all out because i realized i broke clean builds.  Xcode was locally caching something.  I'll reland once I sort it all out.

Issue #13, opened 2016-09-28
Determine if the hand-written well-known types are necessary


allevato:
SwiftProtobuf has three hand-written implementations of the types in three of Google's well-known type files: Any, Struct, and Wrappers. These pose difficulties for future maintenance if they must be kept in sync by hand instead of generated.

We need to determine if there are advantages to keeping these hand-written, and if the advantages are significant enough to keep doing it compared to generating them. For example, can any custom behavior just be added via extensions instead?

tbkka:
Wrappers could easily be implemented as a generated base + extensions for the custom JSON behaviors.  Someone recently asked me about "easy starter projects" -- I think that would be one.

I suspect Struct and even Any could also be done that way, though Any might require some adjustments to the rest of the library to expose the right customization points.

allevato:
The `NullValue` enum in `struct.proto` is going to make this a little trickier. Since it's an enum and the `json` property is generated directly on the type, there's no way to have an extension replace it. (Aside: Even if we extract JSON support into separate extensions, we can't introduce our own hand-written extension on the same type without it causing an ambiguity.)

I haven't thought *too* deeply about how to solve this yet, but the options that come to mind feel like hacks:

1. Add a method to `ProtoEnum` like `remapJSONName(_ value: Self, _ name: String)` and have the generated code call it, passing it the JSON string for the case. The default implementation in the protocol extension would just return the string verbatim. A hand-written extension on concrete enum types could then override it and return something else (in this case, `"null"`).
2. Just add a special case for `Google_Protobuf_NullValue` in the JSON serialization code.
3. Other options...?

Any option will pick will likely have a small cost in terms of runtime library size, generated code size, and execution speed. Hopefully they're negligible; in terms of execution speed at least, surely the rest of the string processing will dominate whatever we add.

tbkka:
> Since it's an enum and the json property is generated directly on the type, there's no way to have an extension replace it.

The technique I used to allow overriding generated methods on messages could work here.  In the library, define:
```
extension ProtobufEnum {
   var json: String {return _protoc_generated_json}
}
```

The generator then emits a definition for `_protoc_generated_json` instead of defining `json` directly.  Then we can override `json` in an extension (struct/enum extensions can override protocol extensions).

thomasvl:
@allevato is working on this for Struct.  Once that lands, I've got an idea for how we might be able to do this for Any.

thomasvl:
#387 removed the Any version.  This is now done.

[...]

PR #65, opened 2016-10-07
Replace hand-written wrappers.proto messages.

allevato:
All tests currently pass with this change.

Part of #13.


SwiftProtobufRuntime.xcodeproj/project.pbxproj
```diff
@@ -262,15 +261,16 @@
...
+		AAEA526E1DA6EF83003F318F /* Google_Protobuf_Struct.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Google_Protobuf_Struct.swift; sourceTree = "<group>"; };
 		__PBXFileRef_Package.swift /* Package.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Package.swift; sourceTree = "<group>"; };
 		__PBXFileRef_ProtobufTestSuite_Info.plist /* ProtobufTestSuite_Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; name = ProtobufTestSuite_Info.plist; path = SwiftProtobufRuntime.xcodeproj/ProtobufTestSuite_Info.plist; sourceTree = SOURCE_ROOT; };
 		__PBXFileRef_Protobuf_Info.plist /* Protobuf_Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; name = Protobuf_Info.plist; path = SwiftProtobufRuntime.xcodeproj/Protobuf_Info.plist; sourceTree = SOURCE_ROOT; };
-		__PBXFileRef_Sources/Protobuf/Google_Protobuf_Any.swift /* Google_Protobuf_Any.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Google_Protobuf_Any.swift; sourceTree = "<group>"; };
```
thomasvl: Did you mean for Any and Struct to also get removed?

allevato: Thanks for catching that. I added them back (I tried replacing all the WKTs at once, then realized that was a bad idea) but it caused some churn in the pbxproj. Pushed a commit to remove the churn.

tbkka:
I'm curious:  What did this do to the overall size of the compiled module?

I recall that my original thinking behind basing these on a common generic implementation was in the hopes it would reduce the overall code size.

allevato:
Before this commit, the release version of libSwiftProtobuf.dylib on my system was 3,005,612 bytes. After this commit, it's 3,035,608 bytes. So there's a small increase, but it would better to get that savings back elsewhere because maintaining the hand-generated protos will be a huge pain going forward (especially right now, as we make changes to the API).

Once we get CI set up, seeing reports of the dylib size delta for each PR would be *amazing*.

PR #67, opened 2016-10-10
Xcode project cleanup round 2

### LINKED ISSUE #57: Clean up Xcode project

thomasvl:
Fixes #57 

Everything is done this time around as small sub commits incase there are issues after the fact.  Also makes it more clear what was done and why.


Makefile
```diff
@@ -235,3 +250,40 @@ regenerate-test-protos: ${PROTOC_GEN_SWIFTX}
...
+	  -scheme SwiftProtobufRuntime_iOS \
+	  -configuration Release \
+	  -destination "platform=iOS Simulator,name=iPhone 6s,OS=latest" \
+	  test
+
+test-xcode-macOS-debug:
+	xcodebuild -project SwiftProtobufRuntime.xcodeproj \
+	  -scheme SwiftProtobufRuntime_macOS \
+	  -configuration debug \
+	  build test
```
tbkka: Is there a reason the macOS test targets use "build test" and the iOS ones use only "test" here?

SwiftProtobufRuntime.xcodeproj/project.pbxproj
```diff
@@ -961,309 +959,190 @@
...
-				ONLY_ACTIVE_ARCH = YES;
-				PRODUCT_BUNDLE_IDENTIFIER = "Apple.ProtobufTestSuite-iOS";
+				LD_RUNPATH_SEARCH_PATHS = "@loader_path/Frameworks";
+				PRODUCT_BUNDLE_IDENTIFIER = com.apple.protobuf.SwiftProtobufRuntimeTests;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = iphoneos;
-				SWIFT_ACTIVE_COMPILATION_CONDITIONS = DEBUG;
-				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
-				SWIFT_VERSION = 3.0;
+				TARGETED_DEVICE_FAMILY = "1,2";
```
tbkka: One concern:   Testing only iOS and macOS won't cover any 32-bit platforms.  Maybe we should add watchOS tests?

Makefile
thomasvl: Cause Xcode.  `xcodebuild` will honor the `build` action a macOS framework, but it won't for an iOS framework.  I tend to try to use both just incase we add other things to the actions in the future.

SwiftProtobufRuntime.xcodeproj/project.pbxproj
thomasvl: I've got another branch queued up to add 32bit iOS destinations to the tests also.  The initial Xcode project didn't really support it, so I wanted to land these cleanups before changing things in the project.  It's on my other machine, so hopefully I'll get a chance to send out the PR later today.

PR #68, opened 2016-10-11
Allow iOS 9 for the iOS unittest.

thomasvl:
- Drop the project settings
- Add coverage to the makefile to get 64bit & 32bit coverage.


allevato:
👍

PR #71, opened 2016-10-11
Fix linker errors in release builds.

### LINKED ISSUE #70: Release builds busted
It looks like something about the wrappers extensions.  Then generate a pile of linker errors at the moment.  Debug works.

allevato:
When using the `ProtobufWrapper.init(_ value: WrappedType.BaseType`)
initializer from externally linked code, the linker dies in release
builds with errors that it cannot find the protocol witness table for
the initializers on the concrete types. This *may* be a Swift bug;
either the compiler should not allow the mismatched visibility on the
conformances, or the linker should allow the call.

For now we duplicate the implementation throughout the concrete types
instead, despite the small amount of bloat this adds.

Fixes #70.


allevato:
Comment added.

PR #74, opened 2016-10-12
Makefile cleanup/additions.

thomasvl:
- Wrap comments to 80c (was a mix)
- Update some docs.
- Add test-all/test-everything that also invokes the Xcode tests.
- Add variables that can be set on the command line to tweak the xcodebuild
  invocation:
  - Enabling the analyzer.
  - Skipping the optimizer (to speed things up).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,16 +5,20 @@
 #   Build the runtime library and plugin
 # make test
 #   Build everything, run both plugin and library tests:
-#   Plugin test verifies that plugin output matches the "Reference" files exactly
+#   Plugin test verifies that plugin output matches the "Reference" files
+#      exactly
 #   Library test exercises most features of the generated code
 # make regenerate
 #   Recompile all the necessary protos
 #   (requires protoc in path)
+# make test-xcode[-NAME]:
+#   Runs the tests in the Xcode project in the requested mode(s).
 #
-# Caution:  'test' does not 'regenerate', so if you've made changes to the
-# code generation, you'll need to do more than just 'test':
+# Caution: 'test' does not 'regenerate', so if you've made changes to the code
+# generation, you'll need to do more than just 'test':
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -114,6 +116,22 @@ PLUGIN_PROTOS= \
...
 	swift-options.proto
 
+XCODE_BUILD_EXTRAS =
+# Invoke make with XCODE_SKIP_OPTIMIZER=1 to suppress the optimizer when
+# building the Xcode projects. For Release builds, this is a non trivial speed
+# up for compilation
+XCODE_SKIP_OPTIMIZER=0
+ifeq "$(XCODE_SKIP_OPTIMIZER)" "1"
+  XCODE_BUILD_EXTRAS += SWIFT_OPTIMIZATION_LEVEL=-Onone
+endif
```
thomasvl: We could flip option name and default here so we normally skip the optimizer to speed things up.

Makefile
```diff
@@ -114,6 +116,22 @@ PLUGIN_PROTOS= \
...
+
+# Invoke make with XCODE_ANALYZE=1 to enable the analyzer while building the
+# Xcode projects.
+XCODE_ANALYZE=0
+ifeq "$(XCODE_ANALYZE)" "1"
+  XCODE_BUILD_EXTRAS += RUN_CLANG_STATIC_ANALYZER=YES CLANG_STATIC_ANALYZER_MODE=deep
+endif
```
thomasvl: Same, we could change things so the default it to do this and one has to skip it instead.

thomasvl:
fyi - rebased this off master and updated the other invokes of `xcodebuild` added since.

thomasvl:
Going ahead and landing this, we can revisit/flip the defaults for the two things in a follow up cl.

PR #79, opened 2016-10-12
Add wire format and tag helper types.

allevato:
The `WireFormat` and `FieldTag` types encapsulate behaviors around those protobuf concepts, to un-scatter all the bit-twiddling currently in the code.

### DIFF
--- a/Sources/SwiftProtobuf/FieldTag.swift
+++ b/Sources/SwiftProtobuf/FieldTag.swift
@@ -17,6 +17,11 @@
 
 /// Encapsulates the number and wire format of a field, which together form the
 /// "tag".
+///
+/// This type also validates tags in that it will never allow a tag with an
+/// improper field number (such as zero) or wire format (such as 6 or 7) to
+/// exist. In other words, a `FieldTag`'s properties never need to be tested
+/// for validity because they are guaranteed correct at initialization time.
 internal struct FieldTag: RawRepresentable {
 
   typealias RawValue = UInt32
@@ -50,8 +55,10 @@ internal struct FieldTag: RawRepresentable {
   /// Note that if the raw value given here is not a valid tag (for example, it
   /// has an invalid wire format), this initializer will fail.
   init?(rawValue: UInt32) {
-    // Verify that the wire format is valid and fail if it is not.
-    guard let _ = WireFormat(rawValue: UInt8(rawValue % 8)) else {
+    // Verify that the field number and wire format are valid and fail if they
+    // are not.
+    guard rawValue & ~0x07 != 0,
+      let _ = WireFormat(rawValue: UInt8(rawValue % 8)) else {
       return nil
     }
     self.rawValue = rawValue
--- a/Sources/SwiftProtobuf/FieldTag.swift
+++ b/Sources/SwiftProtobuf/FieldTag.swift
@@ -1,0 +1,65 @@
+// SwiftProtobuf/Sources/SwiftProtobuf/FieldTag.swift - Describes a binary field tag
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2016 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+// -----------------------------------------------------------------------------
+///
+/// Types related to binary encoded tags (field numbers and wire formats).
+///
+// -----------------------------------------------------------------------------
+
+
+/// Encapsulates the number and wire format of a field, which together form the
+/// "tag".
+internal struct FieldTag: RawRepresentable {
+
+  typealias RawValue = UInt32
+
+  /// The raw numeric value of the tag, which contains both the field number and
+  /// wire format.
+  let rawValue: UInt32
+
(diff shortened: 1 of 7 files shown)


Sources/SwiftProtobuf/WireFormat.swift
```diff
@@ -0,0 +1,31 @@
...
+///
+// -----------------------------------------------------------------------------
+
+
+// TODO(#51): This type is currently public because there are protocols that use
+// it that cannot currently be easily made internal. Revisit the visibility of
+// this type once we've figured out if we can lock those other protocols down.
+
+/// Denotes the wire format by which a value is encoded in binary form.
+public enum WireFormat: UInt8 {
```
tbkka: Swift generally leans toward just using Int for numeric types (more like Java in this respect, less like C).   Since WireFormat here is not directly tied to a particular in-memory representation, I would just use Int for the base type.

Sources/SwiftProtobuf/FieldTag.swift
```diff
@@ -0,0 +1,65 @@
+// SwiftProtobuf/Sources/SwiftProtobuf/FieldTag.swift - Describes a binary field tag
```
tbkka: Since FieldTag and WireFormat are specific to protobuf binary coding, I would prefer the source files be named accordingly.

Sources/SwiftProtobuf/FieldTag.swift
```diff
@@ -0,0 +1,65 @@
...
+/// Types related to binary encoded tags (field numbers and wire formats).
+///
+// -----------------------------------------------------------------------------
+
+
+/// Encapsulates the number and wire format of a field, which together form the
+/// "tag".
+internal struct FieldTag: RawRepresentable {
+
+  typealias RawValue = UInt32
```
thomasvl: Not used?

Sources/SwiftProtobuf/ProtobufBinaryDecoding.swift
```diff
@@ -682,14 +674,16 @@ private class ProtobufScanner {
 
     // Parse index/type marker that starts each field.
     // This also bookmarks the start of field for a possible skip().
-    func getTagType() throws -> Int? {
+    func getTag() throws -> FieldTag? {
         fieldStartP = p
         fieldStartAvailable = available
         if let t = try getRawVarint() {
             if t > 7 && t < UInt64(UInt32.max) {
```
thomasvl: Any way to drop the 7 magic here?  This is basically checking that the field number wasn't zero.  Should FieldTag instead valid the field number along with the wire format?

Sources/SwiftProtobuf/WireFormat.swift
thomasvl: If WireFormat ends up in any objects as instance data, then it means the instances are all a fair amount larger (especially on 64bit).  So likely not an issue here, but something to keep in mind for anything else that might end up in a GeneratedMessage.

Sources/SwiftProtobuf/FieldTag.swift
allevato: This kind of goes back to the whole `protobufData:` vs. `data:` debate. Binary encoding *is* the fundamental representation of protobufs—the text and JSON encodings are extras. Adding "Binary" or some other qualifier to these just adds noise without much other benefit. It should also be noted that the other language implementations use similar terms unqualified, and it's probably better not to diverge from those so that it's easier for users of one library to map concepts across easily.

Sources/SwiftProtobuf/FieldTag.swift
allevato: It's an associated type requirement of `RawRepresentable`.

Sources/SwiftProtobuf/WireFormat.swift
thomasvl: Actually, C is like what you were suggesting for Swift, one tends to use `int` to get the default sized integer when it doesn't matter.

The value of being explicit (in all languages) is when you are writing code to encode/decode a wire format, the sizes of the values helps document the code.

Sources/SwiftProtobuf/ProtobufBinaryDecoding.swift
allevato: Done. I had to keep the `t < UInt64(UInt32.max)` case because it would require `FieldTag` to accept larger than 32-bit values, which I would consider a design flaw. A better way to handle that would be to have the scanner use a separate `getRawVarint32` method, as other language implementations do, and have *that* throw if the value encountered is actually larger than 32-bits.

thomasvl: I guess the other option would be to accept an int64 to init and put the check there.

Sources/SwiftProtobuf/ProtobufBinaryDecoding.swift
```diff
@@ -659,7 +659,7 @@ private class ProtobufScanner {
         fieldStartP = p
         fieldStartAvailable = available
         if let t = try getRawVarint() {
-            if t > 7 && t < UInt64(UInt32.max) {
+            if t < UInt64(UInt32.max) {
```
jcanizales: nit: make indentation 2 spaces?

allevato: @thomasvl and/or I will start doing that in separate commits soon.

PR #80, opened 2016-10-12
Tweaks the plugin about the WKTs.

thomasvl:
The WKTs can be picked off by their proto package. So use this (like the other
language generators do) to pick them off and remove the SwiftProtobuf import as
they WKTs will be bundled with the runtime.

### DIFF
--- a/Sources/protoc-gen-swift/FileGenerator.swift
+++ b/Sources/protoc-gen-swift/FileGenerator.swift
@@ -122,6 +122,16 @@ extension Google_Protobuf_FileDescriptorProto {
         }
     }
 
+    var baseFilename: String {
+        return splitPath(pathname: name ?? "").base
+    }
+
+    var isWellKnownType : Bool {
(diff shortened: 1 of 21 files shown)


Sources/protoc-gen-swift/FileGenerator.swift
```diff
@@ -259,10 +267,19 @@ class FileGenerator {
...
-        p.print(
-            "import Foundation\n",
-            "import SwiftProtobuf\n",
-            "\n")
+        // The well known types don't need to import SwiftProtobuf, so catch
+        // them via their package (just make sure it isn't descriptor since that
+        // isn't needed/shipped, it would be outside the library.
+        if isWellKnownType {
+          p.print(
+              "import Foundation\n",
```
allevato: Maybe move the `import Foundation` line out above this conditional, so that if we ever decided to add more imports (I'm not sure what they would be though), we avoid duplication?

PR #81, opened 2016-10-12
Add coding style guidelines.

### LINKED ISSUE #52: Code style/conventions (indentation, wrapping, etc.)
Now that we have multiple people contributing to the repo, it would be a good idea to agree to a few code style/conventions so we can try to stay consistent across the repo (and because I just realized that I inadvertently checked in some tabWidth changes to the .pbxproj because I had to adjust the files I was working in to meet existing indentation).

I'd like to propose the following:

* 2 space indentation (matches Swift stdlib source)
* 100 character line length (Xcode's default wrapping isn't ideal, IMO. Swift stdlib uses 80, but I find that to be too short with Swift's descriptive naming conventions).
* One type per file, with the filename matching the type name, where reasonable. This one can be a little flexible, especially for small related types; for example, it's probably fine to have all the `ProtobufBinaryTypes` types stay in a single file, rather than a handful of very small source files.
* In a similar vein, if a file primarily contains an extension that adds protocol conformance to a type, name it `Type+Protocol.swift` so the content/purpose is easily glanceable. (As a new reader of the code, I've found myself bouncing between multiple files trying to hunt down certain extensions because the filenames don't directly map to the things in them.)

Feel free to disagree and/or add your own, and we can have a discussion and resolve this issue with the final decisions.

We don't necessarily have to reformat the entire repo at once, but as we make changes, it might be nice to do a "slow burn" and cover what we can.

Sadly, the `swift-format` tool that was added to the `swift` driver doesn't look like it made it into the Swift release that's in Xcode 8.0; it would have been nice to use it to automate things.

Thoughts?

allevato:
Comments/suggestions welcome before we lock down the first version of this.

Fixes #52.

### DIFF
--- a/Documentation/STYLE_GUIDELINES.md
+++ b/Documentation/STYLE_GUIDELINES.md
@@ -1,0 +1,84 @@
+# Style Guidelines
+
+If you're interested in contributing to the Swift Protobuf project, welcome!
+Please make sure that your APIs are named and code is formatted according to
+these guidelines so that we can ensure a consistent look-and-feel across the
+entire codebase.
+
+The following guidelines are non-exhaustive.
+
+## Formatting
+
+* Indentation: 2 spaces (do not use tabs)
+* Line length: 80 characters
+
+When in doubt about how to format a particularly tricky construct (such as
+a method with a large number of generic types and/or arguments), we suggest
+two pieces of advice:
+
+1. Look to the
+   [Swift standard library](https://github.com/apple/swift/tree/master/stdlib/public/core)
+   for guidance.
+1. Don't fight Xcode's auto-indenting unless doing so would make the
+   formatting look horrible. Xcode has some baked-in assumptions about how
+   Swift code should be formatted and fighting it will make your life harder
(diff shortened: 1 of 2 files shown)


Documentation/STYLE_GUIDELINES.md
```diff
@@ -0,0 +1,84 @@
...
+  would contain a type named `Foo`).
+  * In some cases, however, many small related types may be combined into a
+    single .swift file for convenience. In that case, name the file based on a
+    plural noun that describes the grouping. For example, the
+    `ProtobufBinaryTypes.swift` file contains multiple small related types.
+* Source files that primarily contain an extension that adds protocol
+  conformance to a type should be named `Type+Protocol.swift` so the content
+  and purpose of the file is easily glanceable based on the name and so that it
+  is easy to find the source file where a particular concept is implemented.
+  * Mirroring
```
jcanizales: Something's missing here.

allevato: Indeed it is. Another PR incoming.

Issue #83, opened 2016-10-13
File generation strips directories (unlike other languages)


thomasvl:
When using large numbers of protos, it isn't uncommon to end up with a directory structure of them like:

```
dir1/
  foo.proto
  settings.proto
dir2/
  bar.proto
  settings.proto
```

The current generator strips all the passed directories, so generating all these at once would result in collisions.

Since the Swift Package Manager likes flat directories, the best solution is likely to add a compiler option to control the output file naming, supporting:

1. keeping the directory structure
1. flattening the structure by combining the directory names with underscores (for the package manager)
1. stripping the directory structure completely (for anyone with simple enough cases where they don't want any structure.

tbkka:
Ignoring the package name when generating filenames was a very deliberate decision on my part.

I've many times in the past tried to use protoc with `make` or Xcode build rules and found it very frustrating.   Having the generator build a single output file in the current directory from a single input file makes it easy to do simple things with common build systems.   (This is also why the plugin does not mangle filenames like the Obj-C generator does or build multiple output files for a single input like the Java generator does.)

Tools like Bazel with specialized knowledge of proto semantics may be necessary for complex cases, but are overkill for common cases where people have just one or two protos in a project.

I would not be opposed to having alternate mechanisms that can generate more structured output, but I'm reluctant to make simple cases complicated by default.

tbkka:
Re-reading, I now realize that the proposal was to use the input _directory name_ when generating the output, not the input _package name_.  The directory name is visible to build tools, so I could agree to that.

PR #85, opened 2016-10-13
Add options parsing and returning errors to protoc

thomasvl:
Checkpoint on the way working on https://github.com/apple/swift-protobuf/issues/83.

### DIFF
--- a/Sources/protoc-gen-swift/main.swift
+++ b/Sources/protoc-gen-swift/main.swift
@@ -24,8 +24,8 @@
 import Foundation
 import PluginLibrary
 
-enum MyError: Error {
-  case failure
+enum GenerationError: Error {
+  case readFailure
 }
 
 func help(progname: String) {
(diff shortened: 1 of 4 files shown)


Sources/protoc-gen-swift/main.swift
```diff
@@ -26,6 +26,8 @@ import PluginLibrary
 
 enum GenerationError: Error {
   case readFailure
+  /// Raise when parsing the parameter sting and found an unknown key
```
allevato: s/sting/string/

Sources/protoc-gen-swift/main.swift
```diff
@@ -78,10 +78,17 @@ if justVersion {
...
+  let response: CodeGeneratorResponse
+  do {
+    let rawRequest = try Stdin.readall()
+    let request = try CodeGeneratorRequest(protobuf: rawRequest, extensions: SwiftOptions_Extensions)
+    let context = try Context(request: request)
+    response = context.generateResponse()
+  } catch GenerationError.readFailure {
+    response = CodeGeneratorResponse(error: "Failed to read the input")
+  } catch {
+    response = CodeGeneratorResponse(error: "Internal Error")
```
allevato: Maybe `catch let e` here and include `\(e)` in the error message string so we can see something about what the error was?

thomasvl: Done.

Issue #86, opened 2016-10-13
Remove the convenience initializers on generated messages


allevato:
The generated convenience initializers that take default values are a great idea in theory, because they let users quickly create a message using whatever combination of fields they wish to initialize. However, they come with a hidden but **large** performance cost.

Default function arguments in Swift are implemented by creating a small shim for each argument, which loads the default value onto the stack. When such a function is called, these shims are called for any argument not provided by the user, and then the function body is executed.

This implies two things:
1. Code is generated for each default argument, increasing the binary size
2. Calling a function with *N* arguments that have defaults incurs the execution cost of loading *all* of the arguments, even if only one is provided.

In other words, if you have a message with 100 fields, you generate 100 shims for those default arguments; but *worse*, the execution cost of initializing the message with one argument is the same as the cost of initializing it with *all 100* arguments. Even in release mode with heavy optimization, this cannot be avoided; see [this gist](https://gist.github.com/allevato/2a22bc877c5aec45fb76a53f1fe080aa#file-big_initializer-s-L834) for a snippet of the Hopper disassembly of a 100-field message where only `field89: 0x600b34` was passed in.

We can also look at the binary size effects of these initializers, using our performance harness:

 | With convenience init | Without convenience init
--- | --- | ---
Harness size, bytes (before stripping) | 415,544 | 250,352
Harness size, bytes (after stripping) | 152,992 | 144,536
Runtime .dylib size (release) | 3,044,060 | 3,012,252

Since the runtime .dylib size includes the well-known types, there's an effect there as well. The difference in harness size is very stark before stripping (a 165KB savings by removing *one initializer*!), but in that case, stripping eliminated much of the overhead.

## Proposed alternative

Being able to create a message with values at initialization time has its benefits; maybe you want to create one inline when you call another function to avoid a temporary variable, or you want to keep your variables immutable whenever possible and having to set properties after initialization time prevents that.

This is a topic that's been debated frequently on swift-evolution, because this isn't a protobuf-specific problem; indeed, all value types are affected. A very well-received alternative is to create a helper function that creates the value and takes a closure where initialization is performed. Unlike Swift value types in general, all of our protos conform to a common protocol, so we can provide such a helper easily across the board:

```swift
extension ProtobufMessage {
  public static func with(initializer: (inout Self) -> ()) -> Self {
    var message = Self()
    initializer(&message)
    return message
  }
}
```

Then, at the usage site, you can do:

```swift
let foo = My_Generated_Message.with {
[...]

PR #87, opened 2016-10-13
Remove generated convenience initializers.

allevato:
See #86 for rationale.


allevato:
Note: For `Duration` and `Timestamp`, I manually added convenience initializers back via extensions. For `Timestamp` in particular, there were both `secondsFromEpoch:nanos:` and `seconds:nanos:` (generated) with no discernible difference, so I only renamed the former to match the latter. I also added some convenience initializers to `CodeGeneratorResponse(.File)` used in the plugin to keep those usage sites clean.

Sources/SwiftProtobuf/Google_Protobuf_Duration_Extensions.swift
```diff
@@ -157,6 +157,11 @@ public extension Google_Protobuf_Duration {
 
 extension Google_Protobuf_Duration: ExpressibleByFloatLiteral {
     public typealias FloatLiteralType = Double
+    public init(seconds: Int64, nanos: Int32 = 0) {
```
tbkka: There's already an empty initializer, so default seconds doesn't help here.

thomasvl: So we don't want support for Google_Protobuf_Duration(nanos:12) ?

tbkka: I don't see a problem with asking people to write out `Google_Protobuf_Duration(seconds: 0, nano: 12)` in that case.

We could change it later.

thomasvl: Should we change Timestamp to match?  That as a default on each of them.

allevato: I'll add one for `Duration` for completeness. I added the one to `Timestamp` because it was used in a test and it seemed useful generally.

PR #91, opened 2016-10-14
Update for google/protobuf master changes

thomasvl:


Protos/unittest_swift_naming.proto
```diff
@@ -720,7 +720,12 @@ enum EnumFieldNames {
     __FUNCTION__ = 63;
     __LINE__ = 64;
     _ = 65;
-    __ = 1065;
```
tbkka: It would be nice to have a separate enum for the double-underscore case.

thomasvl: I'll make a second.

Just want to confirm, nothing uses this enum, so there isn't anything else I should update as part of adding the second enum?

allevato: LGTM

PR #94, opened 2016-10-14
Support an generator parameter for output FileNaming

thomasvl:
The available values for "FileNaming" are:

* "FullPath" - (default) - Like all other language, "foo/bar/baz.proto" makes
  "foo/bar/baz.pb.swift.
* "PathToUnderscores" - To help with things like the Swift package manager where
  someone might want all the files in one directory; "foo/bar/baz.proto" makes
  "foo_bar_baz.pb.swift".
* "DropPath" - Drop the path from the input and just write all files into the
  output directory; "foo/bar/baz.proto" makes "baz.pb.swift".

Fixes https://github.com/apple/swift-protobuf/issues/83

### DIFF
--- a/Sources/protoc-gen-swift/main.swift
+++ b/Sources/protoc-gen-swift/main.swift
@@ -27,8 +27,10 @@ import PluginLibrary
 enum GenerationError: Error {
   /// Raised for any errors reading the input
   case readFailure
-  /// Raise when parsing the parameter string and found an unknown key
+  /// Raised when parsing the parameter string and found an unknown key
   case unknownParameter(name: String)
+  /// Raised when a parameter was giving an invalid value
+  case invalidParameterValue(name: String, value: String)
 }
 
 func help(progname: String) {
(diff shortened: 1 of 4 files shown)


Sources/protoc-gen-swift/main.swift
```diff
@@ -27,8 +27,10 @@ import PluginLibrary
 enum GenerationError: Error {
   /// Raised for any errors reading the input
   case readFailure
-  /// Raise when parsing the parameter string and found an unknown key
+  /// Raised when parsing the parameter string and found an unknown key
   case unknownParameter(name: String)
+  /// Raised when a parameter was giving an invalid value
+  case invalidParameterValue(name: String, value: String)
```
jcanizales: `, acceptedValues: [String]`? To make the error message more actionable.

thomasvl:
I'll do a follow up to add to the docs how to pass the parameters to the plugin to change generation.

Sources/protoc-gen-swift/main.swift
thomasvl: I debated that one, but I wasn't sure if there was an easy way to collect that from the `enum` to avoid having a second list to update.  I'm sorta hoping the docs alone will cover it.

jcanizales: Apparently there's not automatic way, so Internet suggests adding this into the enum: `static let allValues = [value1, value2, ...]`, to be able and do `MyEnum.allValues`. It wouldn't avoid the repetition, although it'd keep it together with the original list.

thomasvl: The throwing site we'd still need shim code to transform the enums into an array of strings.  And better docs still seems like the idea solution to this sorta problem.  So I'm leaning towards not doing this at this point and just adding the docs instead.

thomasvl:
> I'll do a follow up to add to the docs how to pass the parameters to the plugin to
> change generation.

Looks like the plugin docs haven't been update for the project shuffle, so I guess all the docs in a different PR since there is more to get done.

Issue #22, opened 2016-09-30
Fetching a oneof doesn't return the default value (when not set)


thomasvl:
The fields in a oneof are just fields.  So fetching it should still return the default value even when the oneof was set to a different element.

This also allows proto authors to modify a proto by moving a field into a oneof in the future.  From a wire format pov, nothing changes.

tbkka:
I think your comment only applies to proto3?

In proto2, oneof fields are optionals, and we map proto2 optionals to Swift optionals, so the default value in that case is nil.

Oh.  I suppose proto2 optionals can have explicit default values?  I don't think I've seen that case, but that would make sense as well.

thomasvl:
You can also have a message in the oneof, so returning the default there to dot into the default fields also is useful.

tbkka:
> You can also have a message in the oneof, so returning the default there to dot into the default fields also is useful.

Yes, good point.  Are there examples of this in the `google/protobuf/unittest_*` protos?  If so, then it should be easy to fill in some tests for these cases.

thomasvl:
Yup - https://github.com/google/protobuf/blob/master/src/google/protobuf/unittest.proto#L177

thomasvl:
PR #119 added tests to confirm this got fixed in PR #102

PR #102, opened 2016-10-17
Generate non-optional properties for all fields.

### LINKED ISSUE #24: Fetching a optional/required field doesn't return the default value
Like #22, but for fields not in a oneof, fetching an optional/required field should return the default value is it hasn't been set yet.  In proto2 syntax, the default isn't the _zero_, so this becomes even more important.

allevato:
This updates the generator to generate non-optional properties for all fields, along with `has` properties and `clear` methods. Sadly this will increase the generated code size a little bit, but the alternative was not proto-behaviorally correct, so hopefully we can make that increase up elsewhere. In the future I hope that we can make this Swiftier again with resettable properties, property behaviors, or one of the other similarly proposed ideas on swift-evolution.

There's still some cleanup to be done here—the decoder can be simplified now that we're not using the full set of methods, but this change started getting unwieldy enough and that kind of cleanup can be done after the fact.

The commits in this PR have been broken up by file directory to make them easier to scan through. They will not build independently and will be squashed before merging.

Fixes #24.


thomasvl:
My guess is this will also fix #22, but we can close that out after the fact once we add some tests to go with it.

Sources/protoc-gen-swift/FileGenerator.swift
```diff
@@ -247,7 +250,7 @@ class FileGenerator {
     }
 
     func generateOutputFile(printer p: inout CodePrinter, context: Context) {
-        let inputFilename = descriptor.name ?? "<No name>";
+        let inputFilename = descriptor.hasName ? descriptor.name : "<No name>"
```
thomasvl: When I was doing the output names, I was debating stuff like this.  Protoc really never should send us a descriptor without names, so it might make sense to eventually do a sweep for stuff like this and change the code to just abort.  That's what most of the C++ generators do when they get input that doesn't make sense.

allevato:
@tbkka, please take a look at this as soon as you can; since it's a fairly significant change from your original design I want your sign-off. Because of the scope of this change I'm effectively blocked unless I want to pile up more conflicts (and this change is necessary for some of the other cleanup/enhancements that I want to do). Thanks!

Reference/google/protobuf/compiler/plugin.pb.swift
```diff
@@ -79,7 +79,17 @@ public struct Google_Protobuf_Compiler_CodeGeneratorRequest: ProtobufGeneratedMe
...
+  public var parameter: String {
+    get {return _parameter ?? ""}
+    set {_parameter = newValue}
+  }
+  public var hasParameter: Bool {
+    return _parameter != nil
+  }
+  public mutating func clearParameter() {
+    return _parameter = nil
+  }
```
tbkka: I'm still thinking about this.

## Arguments for this change

If I understand correctly, you have two key arguments for this change:

### 1. The spec comment about default values.

In particular, the proto2 spec says:
> If the default value is not specified for an optional element, a type-specific default value is used

I'm not especially persuaded by this:
 *  C++ and Java lack any general concept of an "unset property".  As a result, it's necessary for the spec to describe the value that should be seen if a property is not set (since properties in those languages can never be truly unset).
 * Python and Javascript do have a notion of an "unset property" but the syntax for testing them is awkward.
 * Swift `Optional` does provide a robust concept of an "unset variable" and provides trivial syntax (via '??' syntax) for working with possibly-unset values.

(Yes, I've weakened my own argument by attaching it to a `string` property here, since C++/Java reference types are nullable.)

### 2. Consistency with specified defaults

The concern here is whether these are truly the same:
```
    optional string parameter = 99;
```
and
```
    optional string parameter = 99 [default=""];
```
If you consider the first to just be shorthand for the second, then these are the same and we should generate the same results.  I suspect the most expert users of protobuf for C++ and Java do consider these to be the same and would be surprised if we generated different code for them.  Of course, such experts are unlikely to change one of these to the other.

## Arguments against this change

### 3. Generated code size

This change adds four methods (getter, setter, `has`, and `clear`) that are entirely unnecessary in the earlier form.

### 4. Explicit defaults are the less-common case

In helping protobuf users here, I've only seen a handful of uses of `default`.   Other examples: the proto2 language guide only shows two fields with explicit defaults (out of several dozen example fields),  descriptor.proto has 89 optional fields with only two defaults (both enum-valued).  The protobuf/src/google/protobuf has almost 1600 optional fields of which only 100 have explicit defaults.

### 5. Natural mapping of `optional` to `Optional`

Since `Optional` types are natural and well-supported in Swift, it seems odd to go to such lengths to avoid them.  The default values in the spec (zero, "", false, etc.) should not need to be explicitly provided by us.

So I fear that we're complicating the common case, generating code that will be confusing to Swift users, all in order to make an uncommon idiom a little easier to understand.

Reference/google/protobuf/unittest_custom_options.pb.swift
```diff
@@ -1013,7 +1093,56 @@ public struct ProtobufUnittest_ComplexOptionType3: ProtobufGeneratedMessage {
     "complexoptiontype5": 2,
   ]}
 
-  var unknown = ProtobufUnknownStorage()
+  private class _StorageClass {
```
tbkka: Why is this now generating a storage class?  This seems like unnecessary code bloat.

Reference/google/protobuf/unittest_custom_options.pb.swift
```diff
@@ -1367,36 +1539,66 @@ public struct ProtobufUnittest_Aggregate: ProtobufGeneratedMessage {
 
   struct Extensions {
 
-    static let Google_Protobuf_FileOptions_nested = ProtobufGenericMessageExtension<ProtobufOptionalMessageField<ProtobufUnittest_Aggregate>, Google_Protobuf_FileOptions>(protoFieldNumber: 15476903, protoFieldName: "nested", jsonFieldName: "nested", swiftFieldName: "ProtobufUnittest_Aggregate_nested", defaultValue: nil)
+    static let Google_Protobuf_FileOptions_nested = ProtobufGenericMessageExtension<ProtobufOptionalMessageField<ProtobufUnittest_Aggregate>, Google_Protobuf_FileOptions>(protoFieldNumber: 15476903, protoFieldName: "nested", jsonFieldName: "nested", swiftFieldName: "ProtobufUnittest_Aggregate_nested", defaultValue: ProtobufUnittest_Aggregate())
```
tbkka: We could also consider dropping the `defaultValue` parameter from `ProtobufGenericMessageExtension` since the default is already implied from the types.

Reference/google/protobuf/unittest_custom_options.pb.swift
```diff
@@ -1684,14 +1906,24 @@ public struct ProtobufUnittest_OldOptionType: ProtobufGeneratedMessage {
...
+    return _value = nil
+  }
 
   public init() {}
 
   public mutating func _protoc_generated_decodeField(setter: inout ProtobufFieldDecoder, protoFieldNumber: Int) throws -> Bool {
     let handled: Bool
     switch protoFieldNumber {
-    case 1: handled = try setter.decodeRequiredField(fieldType: ProtobufUnittest_OldOptionType.TestEnum.self, value: &value)
+    case 1: handled = try setter.decodeOptionalField(fieldType: ProtobufUnittest_OldOptionType.TestEnum.self, value: &_value)
```
tbkka: In `unittest_custom_options.proto`, the field `OldOptionType.value` is a required field.

Reference/google/protobuf/unittest_custom_options.pb.swift
```diff
@@ -1951,327 +2193,603 @@ public func ==(lhs: ProtobufUnittest_TestMessageWithCustomOptions.OneOf_AnOneof,
 }
 
 extension Google_Protobuf_MessageOptions {
-  public var ProtobufUnittest_ComplexOptionType2_ComplexOptionType4_complexOpt4: ProtobufUnittest_ComplexOptionType2.ComplexOptionType4? {
-    get {return getExtensionValue(ext: ProtobufUnittest_ComplexOptionType2.ComplexOptionType4.Extensions.Google_Protobuf_MessageOptions_complexOpt4)}
+  public var ProtobufUnittest_ComplexOptionType2_ComplexOptionType4_complexOpt4: ProtobufUnittest_ComplexOptionType2.ComplexOptionType4 {
+    get {return getExtensionValue(ext: ProtobufUnittest_ComplexOptionType2.ComplexOptionType4.Extensions.Google_Protobuf_MessageOptions_complexOpt4) ?? ProtobufUnittest_ComplexOptionType2.ComplexOptionType4()}
```
tbkka: This might be redundant:  I believe `getExtensionValue` already returns the default (which you provided above), so the extra `??` here seems unnecessary.  Especially if it results in redundant initializer invocations.

Reference/conformance/conformance.pb.swift
```diff
@@ -1929,11 +1903,23 @@ public struct Conformance_TestAllTypes: ProtobufGeneratedMessage {
...
+  }
+  public mutating func clearOptionalNestedMessage() {
+    return _storage._optionalNestedMessage = nil
+  }
 
   public var optionalForeignMessage: Conformance_ForeignMessage {
     get {return _storage._optionalForeignMessage ?? Conformance_ForeignMessage()}
     set {_uniqueStorage()._optionalForeignMessage = newValue}
   }
+  public var hasOptionalForeignMessage: Bool {
```
allevato: Agreed, the dot-into behavior takes advantage of Swift's copy-on-write behavior in a really unique and clean way. Other languages effectively have this as well (Objective-C provides it automatically, C++ through the `mutable` methods, and so forth), but Swift's is probably the most elegant both in its user facing API and in its implementation because of the features of the language.

Reference/conformance/conformance.pb.swift
```diff
@@ -2162,67 +2154,51 @@ public struct Conformance_TestAllTypes: ProtobufGeneratedMessage {
...
-        _uniqueStorage()._oneofField = .oneofNestedMessage(newValue)
-      } else {
-        _uniqueStorage()._oneofField = .None
-      }
+      _uniqueStorage()._oneofField = .oneofNestedMessage(newValue)
     }
   }
 
-  public var oneofString: String? {
+  public var oneofString: String {
```
allevato: This was intentional; users should rely on the case of the `oneof` to determine which value it has, rather than querying the individual fields. This is a different use case from asking the field directly for its value and returning the default if the case is different, which should still be allowed. They can also set the oneof case property itself to `None` to clear it (although we could debate whether we should also provide a `clear` method for that, since that extra case may not be discoverable enough).

It's worth noting that C++ doesn't provide `has` methods for the individual oneof fields either (the documentation is wrong in this regard).

Reference/google/protobuf/unittest_custom_options.pb.swift
allevato: Because one of its fields is a message type. The generator wasn't completely handling this correctly before.

Think of it this way: messages are value types, and `Optional<>` is a value type (an enum with non-`indirect` cases). Therefore, they cannot directly or indirectly have a circular reference back to the original struct. But it's possible to imagine a series of proto message definitions like this:

```protobuf
message A { B x = 1; }
message B { C f = 1; }
message C { D f = 1; }
...
message Y { Z f = 1; }
message Z { A f = 1; }
```

...which would fail if there was no heap storage in the middle to break the cycle, whether the generated properties were optional or not.

Reference/google/protobuf/unittest_custom_options.pb.swift
allevato: Decoding of individual fields doesn't need to distinguish between the optional-/required-ness of the field; this should be done with an `isInitialized` check at the end of decoding (and only then in the non-partial parsing case).

I could have made this clearer in the original PR comment; one side effect of changing the optionality of the properties in this PR was that `decodeRequiredField` is no longer required and I could use `decodeOptionalField` everywhere. After this change we should delete `decodeRequiredField` and rename `decodeOptionalField` to `decodeSingularField`, but I wanted to keep this PR as targeted as possible.

Reference/google/protobuf/unittest_custom_options.pb.swift
allevato: I'm not sure I follow you here: being proto2, extensions can have explicit default values as well ( see https://github.com/google/protobuf/blob/master/src/google/protobuf/unittest_custom_options.proto#L65-L70). How are they implied by the types?

Reference/google/protobuf/unittest_custom_options.pb.swift
allevato: Ah, this was a side-effect of `ProtobufOptionalField.ValueType` still being optional (`BaseType?`). I need to look more deeply into that—I think you're right that if I fix that up, I can get rid of the unnecessary `??`.

Reference/google/protobuf/unittest_custom_options.pb.swift
tbkka: Oops.  I misread the code and somehow thought that `MessageExtension` was only for message-valued extensions.  You're right that we do need the explicit default support for other field types.

Reference/google/protobuf/compiler/plugin.pb.swift
tbkka: Thanks for your careful response.  I'll continue to think about it.

To be clear, my only question at this point is with proto2 optional fields that do not have an explicit default.  I agree with you about the other cases.  In particular, I agree that your generation of proto2 optional fields with explicit defaults is the only appropriate way to go.  I agree that we need `has` methods for message-typed fields.

A couple of your comments above intrigue me:

> ... even if we have to wait for a future version of Swift to improve on this ...

I haven't given much thought to this, but do you think we have enough versioning capabilities to make this kind of breaking change to our API in the future?

> It's reasonable to omit has/clear for proto3 because the defaults are well known...

Actually, it's reasonable to omit them because proto3 does not guarantee access to field presence information.  Proto2 does guarantee this.

(As you quoted just above, the defaults for proto2 are also well-known.)

> The large number of force-unwraps that I removed in the generator code after I updated descriptor.pb.swift to use non-optional properties is a sign to me that the user-facing API was not ideal to work with.

In my opinion, this is your strongest argument.  My primary concern is with usability in real code.  I'll have to take a careful look at those changes.

allevato: > I haven't given much thought to this, but do you think we have enough versioning capabilities to make this kind of breaking change to our API in the future?

Not yet, no. It's been in the back of my mind, though. If we look at Objective-C and C++, they use a combination of preprocessor directives and runtime logic to verify whether the generated protos match the runtime library they're being compiled/linked against. We're a bit more limited in what we can do in that regard because of the lack of a preprocessor, but I've experimented with some protocol conformance tricks that we can use to create a type that only compiles when a specific version of the library is imported.

Sources/protoc-gen-swift/ExtensionGenerator.swift
```diff
@@ -105,7 +109,7 @@ struct ExtensionGenerator {
         }
         let scope = declaringMessageName == nil ? "" : "static "
         let traitsType = descriptor.getTraitsType(context: context)
-        p.print("\(scope)let \(swiftRelativeExtensionName) = ProtobufGenericMessageExtension<\(extensionFieldType)<\(traitsType)>, \(swiftExtendedMessageName)>(protoFieldNumber: \(descriptor.number!), protoFieldName: \"\(descriptor.name!)\", jsonFieldName: \"\(descriptor.jsonName!)\", swiftFieldName: \"\(swiftFieldName)\", defaultValue: \(defaultValue))\n")
+        p.print("\(scope)let \(swiftRelativeExtensionName) = ProtobufGenericMessageExtension<\(extensionFieldType)<\(traitsType)>, \(swiftExtendedMessageName)>(protoFieldNumber: \(descriptor.number), protoFieldName: \"\(descriptor.name)\", jsonFieldName: \"\(descriptor.jsonName)\", swiftFieldName: \"\(swiftFieldName)\", defaultValue: \(defaultValue))\n")
```
tbkka: Can `descriptor.name!` and `descriptor.number!` lose the force-unwrap?

allevato: They did; am I missing something? :)

Reference/google/protobuf/compiler/plugin.pb.swift
tbkka: > The large number of force-unwraps that I removed in the generator code after I updated descriptor.pb.swift to use non-optional properties is a sign to me that the user-facing API was not ideal to work with.

Wow.  Your "large number" is a lot more than I expected.   After playing with 'grep' a bit, it looks like there are really very few force-unwraps left in the generator codebase, and it looks like it would not take much work to reduce it further.

allevato:
Thanks for a spirited review! Landing it now.

PR #114, opened 2016-10-19
Add a protoc version check into the build helpers.

thomasvl:
This doesn't help end developers using the plugin, but it atleast helps
validate that the generated files checked in are being generated with a current
enough protoc for their JSON support to be correct.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,7 +201,7 @@ all: build
 #  but we only update the file when it changes to avoid extra builds.)
 # (Someday, 'swift test' will learn how to auto-discover test cases on Linux,
 # at which time this will no longer be needed.)
-build ${PROTOC_GEN_SWIFT}:
+build:
 	@${AWK} -f CollectTests.awk Tests/SwiftProtobufTests/Test_*.swift > Tests/LinuxMain.swift.new
 	@if ! cmp -s Tests/LinuxMain.swift.new Tests/LinuxMain.swift; then \
 		cp Tests/LinuxMain.swift.new Tests/LinuxMain.swift; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -210,6 +210,18 @@ build ${PROTOC_GEN_SWIFT}:
...
+# This will get run by any other rule that tries to use the plugin.  This hacks
+# in a check during the library build to ensure that the protoc on the local
+# system is one we expect. This was inspired by the findings that lead to
+#   https://github.com/apple/swift-protobuf/issues/111
+# This likely will need follow up to eventually do a pattern match to ensure it
+# is 3.1 or higher.
+${PROTOC_GEN_SWIFT}: build
+	@if [ "$(shell ${PROTOC} --version)" != "libprotoc 3.1.0" ]; then \
+	  echo "WARNING: Unexpected version of protoc: $(shell ${PROTOC} --version)"; \
+	  echo "WARNING: The JSON support in generate files may not be correct."; \
```
allevato: s/generate/generated/

Makefile
```diff
@@ -210,6 +210,18 @@ build ${PROTOC_GEN_SWIFT}:
...
 	${SWIFT} build
 
+# This will get run by any other rule that tries to use the plugin.  This hacks
+# in a check during the library build to ensure that the protoc on the local
+# system is one we expect. This was inspired by the findings that lead to
+#   https://github.com/apple/swift-protobuf/issues/111
+# This likely will need follow up to eventually do a pattern match to ensure it
+# is 3.1 or higher.
+${PROTOC_GEN_SWIFT}: build
+	@if [ "$(shell ${PROTOC} --version)" != "libprotoc 3.1.0" ]; then \
```
allevato: Does this version number change infrequently enough that it's safe to punt on a more general pattern match until later?

thomasvl: Not sure.  There were a few months between the 3.0.0 release and 3.1.0 releases.  There were some 3.0.x releases, but that number didn't always make it to the master branch.

The other catch, is I'm think the only shell we can rely on make spawning in /bin/sh, and test (`[`) doesn't have a regex matcher either.  So when this needs beefing up, it make get ugly and be yet another reason to move from a Makefile to a script for wrapping the build/test/regenerate process.

Makefile
thomasvl: Done (amended commit to include fix)

Makefile
allevato: Changing the != check to a grep for a few different version number patterns might not be that bad? I'll leave it up to you whether you want to try it now or file an issue and clean it up later.

tbkka: grep returns success if nothing was selected, so you can use it with an 'if' in /bin/sh, using constructs like this:
```
     if ${PROTOC} --version | grep pattern ; then ...
```

PR #115, opened 2016-10-19
Simplify ProtoFieldDecoder.

### LINKED ISSUE #112: Simplify ProtobufFieldDecoder
Now that we've unified property generation in #102, there should be some opportunities to simplify `ProtobufFieldDecoder`:

* `decodeRequired[Message|Group]Field` can all go away; it's not used in generated code anymore.
* We should merge the various `decodeOptional...` and `decodeSingular...` pairs. This will be a little trickier for the scalar case because the `Optional` variant takes an `Optional<>` `inout` argument while the `Singular` variant takes a non-optional. The distinction matters because under-the-hood we generate optional storage properties in proto2 and non-optional ones in proto3.

For the second bullet, it might just be better to rename `decodeOptionalField` to `decodeSingularField` and have them both be overloads of the same name. That simplifies the generator code with one less conditional to choose the name, and we might even be able to implement one in terms of the other in a protocol extension.

allevato:
Fixes #112.


allevato:
Does encoding support that right now? `ProtobufVisitor` currently doesn't make a distinction between required and optional fields, and I don't see anything in `ProtobufDebugDescription` that tries to distinguish either.

PR #116, opened 2016-10-19
Add access option to change generation

thomasvl:


Sources/protoc-gen-swift/ExtensionGenerator.swift
```diff
@@ -119,7 +121,7 @@ struct ExtensionGenerator {
         if comments != "" {
             p.print(comments)
         }
-        p.print("public var \(swiftFieldName): \(apiType) {\n")
+        p.print("\(generatorOptions.visibilitySourceSnippet)var \(swiftFieldName): \(apiType) {\n")
```
thomasvl: @tbkka the extensions where the ones I'm the least sure about, are these correct?

thomasvl: @tbkka bump.  Look ok?

Issue #36, opened 2016-10-03
Support protobuf text format serialization and deserialization


BradLarson:
I've been able to pull in binary protobuf data using init(protobuf:), and I've seen init(json:) to bring in JSON, but I can't seem to find an input initializer for text format protobuf data. 

Specifically, I'm trying to pull in the network definition files from the Caffe framework, which are specified as .prototxt files ([an example here](https://github.com/BVLC/caffe/tree/master/models/bvlc_alexnet)). I have all the types from the protocol buffer compiler, run against their caffe.proto definition, and can pull in the .caffemodel binary protobuf from that page. I just can't figure out how to bring in the text format protobuf network data there.

My apologies for asking this as a question, but the documentation and code didn't make it clear if this was present or if I had overlooked something.

jcanizales:
Yeah it's kind of hidden, but [the text format isn't implemented yet](https://github.com/apple/swift-protobuf/blob/master/Documentation/PLUGIN.md#todo). IIRC, you can transform a file from text to binary and vice versa by calling `protoc` with some flags (`protoc --help` should specify it).

tbkka:
It is not yet implemented, but the current code was designed to make it straightforward to add this sort of capability.

For serialization, the easiest place to start would be from the `ProtobufBinaryEncoding.swift` source file.   Deserialization is slightly trickier because you have to map the proto text field names to proto numbers before you can invoke the `decodeField` method to set the value.  The JSON decoder might be an easier place to start.

thomasvl:
Out of curiosity, what are you looking to use it for?

It was never really meant as a format for data exchanges, but can be useful in unittests/etc. but even there using protoc to turn text into binary for a unittest has the advantage of the apps not having to ship the parsing code, keeping them smaller.

BradLarson:
@thomasvl - The specific case that I'm looking to support is neural network definition files. The Caffe framework, one of the more common ways to define and train convolutional neural networks, uses text format protobuf files to define neural network architectures. Binary protobuf format files hold the final trained network weights for that architecture, and I can read those without a problem.

The text format is used to provide a human-readable and human-modifiable description of a network architecture. I'd like to be able to pull that in as well as be able to write it out for use in anything that employs the Caffe framework. It could allow for an easy way to instantiate, design, and run neural networks on Mac, iOS, and Linux.

I'll see if I can build on the starting points mentioned above for serialization / deserialization.

BradLarson:
[...]

PR #117, opened 2016-10-20
Adding text format serialization / deserialization

BradLarson:
Per issue #36, this adds protobuf text format serialization and deserialization. To load a text format protobuf into a message, you use

```
let message = try Message(text:protobufText)
```

and to serialize it out to text, you use

```
try message.serializeText()
```

I'll admit, this is a bit of a hack, in that I copied over the JSON serialization / deserialization files and modified those until they matched the protobuf text format. A ton of code is therefore replicated between these files and could be abstracted out to a shared set of text/JSON input and output code.

My unit tests are rudimentary, but they exercise the data types from the Proto3TestAllTypes class, which seem to cover the available data types. I'm sure that I'm not handling all error cases correctly. 

I've tested this against a series of complex .prototxt text format files from the Caffe framework, and all of them appear to load correctly.


Sources/SwiftProtobuf/ProtobufTextEncoding.swift
```diff
@@ -0,0 +1,291 @@
...
+    }
+    
+    mutating func putBoolValue(value: Bool, quote: Bool) {
+        if quote {
+            append(text: value ? "\"true\"" : "\"false\"")
+        } else {
+            append(text: value ? "true" : "false")
+        }
+    }
+    mutating func putStringValue(value: String) {
```
tbkka: This was copied from the JSON support and uses JSON conventions.   Does protobuf text use `\u` escaping, or does it use UTF-8,  or...?

Sources/SwiftProtobuf/ProtobufTextEncoding.swift
```diff
@@ -0,0 +1,291 @@
...
+    }
+    mutating func putFloatValue(value: Float, quote: Bool) {
+        putDoubleValue(value: Double(value), quote: quote)
+    }
+    mutating func putDoubleValue(value: Double, quote: Bool) {
+        if value.isNaN {
+            append(text: "\"NaN\"")
+        } else if !value.isFinite {
+            if value < 0 {
+                append(text: "\"-Infinity\"")
```
tbkka: The "-Infinity" and similar cases need to be tested against Google's C++ implementation to make sure this is the correct coding.

Sources/SwiftProtobuf/ProtobufTextEncoding.swift
```diff
@@ -0,0 +1,291 @@
...
+    }
+    public mutating func endObject(tabLevel: Int) {
+        for _ in 0..<tabLevel {
+            append(text:"  ")
+        }
+
+        append(text: "}")
+    }
+    mutating func putNullValue() {
+        append(text: "null")
```
tbkka: Does protobuf text encoding use "null"?

Sources/SwiftProtobuf/ProtobufTextDecoding.swift
```diff
@@ -0,0 +1,1229 @@
...
+}
+
+
+public enum ProtobufTextToken: Equatable, ProtobufFieldDecoder {
+    case colon
+    case comma
+    case beginObject
+    case endObject
+    case beginArray
+    case endArray
```
tbkka: Does protobuf text format use `[...]` array notation at all?

Sources/SwiftProtobuf/ProtobufTextDecoding.swift
```diff
@@ -0,0 +1,1229 @@
...
+//   = "4.12e2" => "412" (adjust decimal point)
+//   = "1.0000" => "1" (drop extraneous trailing zeros)
+//
+// Note: This does reject sequences that are "obviously" out
+// of the range of a 64-bit integer, but that's just to avoid
+// crazy cases like trying to build million-character string for
+// "1e1000000".  The client code is responsible for real range
+// checking.
+//
+private func normalizeIntString(_ s: String) -> [Character]? {
```
tbkka: This could likely be shared with JSON support, since these numeric conventions are very common.

Sources/SwiftProtobuf/ProtobufTextDecoding.swift
```diff
@@ -0,0 +1,1229 @@
...
+                case .endArray:
+                    return true
+                default:
+                    throw ProtobufDecodingError.malformedJSON
+                }
+            }
+        }
+    }
+    
+    mutating func decodeRepeatedGroupField<G: ProtobufMessage>(fieldType: G.Type, value: inout [G]) throws -> Bool {
```
tbkka: Does protobufText handle groups differently from messages?  If not, this can simply defer to `decodeRepeatedMessageField()`

allevato: There's a subtle difference with regard to name capitalization: https://github.com/google/protobuf/blob/master/src/google/protobuf/text_format.cc#L445

Sources/SwiftProtobuf/ProtobufTextDecoding.swift
allevato: If we share them, we'll need to make sure that the numbers parsed by JSON format and text format are equivalent. Text format numeric parsing can be found starting here: https://github.com/google/protobuf/blob/master/src/google/protobuf/text_format.cc#L900

Sources/SwiftProtobuf/ProtobufTextDecoding.swift
allevato: Repeated fields can be repeated in text format, but there is also a shorthand bracket syntax: https://github.com/google/protobuf/blob/master/src/google/protobuf/text_format.cc#L527

Sources/SwiftProtobuf/ProtobufTextEncoding.swift
```diff
@@ -0,0 +1,291 @@
...
+    mutating public func withAbstractVisitor(clause: (inout ProtobufVisitor) throws -> ()) throws {
+        var visitor: ProtobufVisitor = self
+        try clause(&visitor)
+        encoder.text = (visitor as! ProtobufTextEncodingVisitor).encoder.text
+    }
+    
+    
+    mutating public func visitUnknown(bytes: [UInt8]) {
+        // JSON encoding has no provision for carrying proto2 unknown fields
+        // TODO: Does this apply to the text format, as well?
```
allevato: Text format supports unknown fields (rather, the printer has a flag that lets you control whether they're printed or not). They get printed with their field number: https://github.com/google/protobuf/blob/master/src/google/protobuf/text_format.cc#L1928

Sources/SwiftProtobuf/ProtobufTextEncoding.swift
allevato: The C++ code calls this `CEscapeAndAppend` function, which prints a C-escaped UTF-8 version of the string: https://github.com/google/protobuf/blob/fd046f6263fb17383cafdbb25c361e3451c31105/src/google/protobuf/stubs/strutil.cc#L579

Tests/SwiftProtobufTests/Helpers.swift
```diff
@@ -104,6 +104,29 @@ extension PBTestHelpers where MessageTestType: ProtobufMessage & Equatable {
...
+        var configured = empty
+        configure(&configured)
+        XCTAssert(configured != empty, "Object should not be equal to empty object", file: file, line: line)
+        do {
+            let encoded = try configured.serializeText()
+            
+            XCTAssert(expected == encoded, "Did not encode correctly: got \(encoded)", file: file, line: line)
+            do {
+                let decoded = try MessageTestType(text: encoded)
+//                print("==== Begin Decoding ====")
```
allevato: Delete commented out debugging code.

Sources/SwiftProtobuf/ProtobufTextEncoding.swift
allevato: No, `null` is only used by JSON format for certain special cases of the well-known types. None of those special cases apply for text format—WKTs are just serialized as ordinary messages with their fields.

BradLarson:
Thanks for the comments. I'll see if I can clean up the easier elements first (mostly copy-paste items left over from the JSON parsing code), then move on to expansion of the tests. The JSON tests and some of the commented code are artifacts left over from my experimentation, and I simply forgot to pull them out when I reviewed this earlier.

This was just the minimum viable implementation to do what I needed, so there's a lot to clean up inside it. I'll see what I can do to improve it per these suggestions.

BradLarson:
Sorry it took me so long to respond, I had a few other things come up in the meantime.  I've updated the pull request per many of the review comments above:

- Renamed the Test_Text.swift file to Text_Text_proto3.swift to give room for proto2 tests.
- Removed the leftover JSON tests that I'd been using for debugging purposes.
- Split the text encoding tests into one per field type, and added some failure cases.
- Added tests for repeated items.
- Shorthand bracket syntax for repeated fields is supported, and added a test for that.
- Removed identified commented-out code.
- Removed left-over debugging statements.
- Removed unnecessary (_) pattern matches in case statements. 

These are issues that I haven't addressed yet, due to the limited time I've had to work on this:

- Failure test cases are not exhaustive.
- Not sure what to do about the errors that refer to JSON error cases.
- Not sure if I should combine the JSON and text numeric parsers, haven't dug into this: https://github.com/google/protobuf/blob/master/src/google/protobuf/text_format.cc#L900
- Haven't addressed the differences in group name capitalization from here: https://github.com/google/protobuf/blob/master/src/google/protobuf/text_format.cc#L445
- Don't handle text format unknown fields at present: https://github.com/google/protobuf/blob/master/src/google/protobuf/text_format.cc#L1928
- I don't verify string escaping per here: https://github.com/google/protobuf/blob/fd046f6263fb17383cafdbb25c361e3451c31105/src/google/protobuf/stubs/strutil.cc#L579
- The "-Infinity" and similar cases need to be tested against Google's C++ implementation to make sure this is the correct coding.

If these are still blockers, I can keep working on the remaining issues.

tbkka:
I looked through the C++ unit tests some and have a few thoughts:

* I see no short-term need for duplicating the excessive number of formatting options supported by the C++ version.  Our encoder does not need to support UTF-8 and non-UTF8, "short" vs. "long" vs. "single line mode", configurable indent level, etc.

* Our text format decoder probably should support all of these, of course.  The C++ unit tests give a lot of examples we can use to build our own unit tests.

* Especially given that our unknown field support for protobuf binary is still evolving, I don't feel an immediate need to handle that for protobuf text format.

thomasvl:
For ObjC TextFormat generation I did utf8 only (Apple's tool chains generally do better with utf so there didn't seem to be a need to do non utf8); I did support an optional initial indent, but that was in part because I reused TextFormat generation for `-description`.

If we are going to support TextFormat parsing, we do need to support parsing all forms as they could come from any other implementation (or be hand written).

tbkka:
@BradLarson:  I just sent you a PR with a number of new tests that I cribbed from the C++ unit test suite.  Let me know if this is helpful...

tbkka:
I'm going to reconcile this with #139 today and hopefully fix some of the other outstanding issues with it.

tbkka:
@BradLarson:  I just pushed a bunch of work into my branch of this.  In particular:

* I restructured everything to match the new file organization (most files define just one type, filenames mostly match the type of that name, many of the `Protobuf` prefixes are gone)
* I rewrote the token struct and most of the scanner to better match protobuf text conventions (e.g., protobuf text has both an 'identifier' literal and a 'string' literal)
* I copied a bunch more tests from the protoc C++ test suite (lots more to go!)

With the above, I think I have most of the repeated field handling, octal and hex numeric literals, and a number of other things working.  I haven't yet looked at groups, maps, extensions or the proto3 Well-Known Types.

Let me know what you think...

tbkka:
Oh, yeah:  I also got `bytes` encoding/decoding working pretty completely.

Sources/SwiftProtobuf/ProtobufTextDecoding.swift
tbkka: Text format never codes integers using exponential notation, so this isn't needed there.

BradLarson:
@tbkka - My apologies on being slow on the response. Had a few things that drew me away at work. I'll see if I can't review all this tomorrow.

tbkka:
Although still incomplete, I think #169 is getting close.  Any feedback greatly appreciated.

Sources/SwiftProtobuf/ProtobufTextDecoding.swift
tbkka: I addressed this in #169 by changing the name that protoc-gen-swift puts into the fieldNames list.  If this is a group field, it now puts the group type name instead of the field name.  This matches the behavior of C++ protobuf text format.

I'm not sure why protoc puts a different field name into the descriptor than the name used by the text encoding.  That's a very odd mismatch.

Sources/SwiftProtobuf/ProtobufTextDecoding.swift
tbkka: In #169, I completely rebuilt the numeric handling.  In particular, this was needed to correctly deal with octal and hexadecimal integer formats:  text format allows them for integer fields, rejects them for floating-point fields.

Sources/SwiftProtobuf/ProtobufTextDecoding.swift
tbkka: My #169 implements the bracket syntax.  It even handles that together with combining string fields and varying quotes, e.g.
```
some_field: ["abc", "def" 'ghi' "jkl", 'mno']
```
decodes to `["abc", "defghijkl", "mno"]`

Sources/SwiftProtobuf/ProtobufTextEncoding.swift
tbkka: I've not implemented unknown field support in #169 yet.

BradLarson:
Given the work that tbkka's done on this over a #169, I'll close this out in favor of that pull request.

PR #123, opened 2016-10-27
Update for the current state of version matching:

thomasvl:
Try to better lead folks down the path of getting a matched plugin and library
sources.


thomasvl:
@gmosx fyi.

PR #124, opened 2016-10-28
Drop backwards compatibility shims.

thomasvl:


tbkka:
Thank you for cleaning those up.

Issue #129, opened 2016-10-31
Data vs. [Uint8] consistency?


thomasvl:
Unknown fields uses the arrays internally, but we've changed most of the binary serialization apis over to use Data.  Should we be a little more consistent on what we use where?

Data seems to be slightly better in that there could be cases where `func enumerateBytes(_ block: (UnsafeRawPointer, NSRange, UnsafeMutablePointer<ObjCBool>) -> Void)` could be used to avoid forcing the information into a continuous block.

tbkka:
As part of #134, I changed the unknown handling path to use `Data` instead of `[UInt8]`

Issue #128, opened 2016-10-31
Rework the setFromXyz methods in ProtobufBinaryTypes.swift for unknown fields


thomasvl:
From @tbkka 

> Rework all the setFromXyz methods in [ProtobufBinaryTypes] to explicitly call
> `decoder.markFieldUnknown()` to archive an unknown field which in turn means we can drop the
> `Bool` return. This would be a much clearer way to handle unknown fields, I think.  It does ripple
> through the entire decode stack, though.

This would remove the need for the return true/false.  The one other thing to remember is for proto2 syntax, they data goes in unknowns as suggested, but for proto3 syntax, the data would be dropped.

tbkka:
Addressed in #134:  Instead of explicitly telling the decoder that a field is unknown, it turned out to be more straightforward for the decoder to track whether the field value is actually requested.  This still removes the return true/false through much of this path and isolates unknown field handling in the protobuf binary decoder, which is the only place that actually needs this support.

Issue #127, opened 2016-10-31
Proto2 syntax, packed repeated enum fields are lossy for unknown values


thomasvl:
Proto2 syntax enum fields (singular or repeated), should move any unknown values into the unknown fields support so the data isn't lost and will reflect back if the message is serialized back out.  Currently the data is lost:

https://github.com/apple/swift-protobuf/blob/master/Sources/SwiftProtobuf/ProtobufBinaryTypes.swift#L735

```swift
 public static func setFromProtobufBuffer(buffer: UnsafeBufferPointer<UInt8>, value: inout [Self]) throws -> Bool {
        var decoder = ProtobufBinaryDecoder(protobufPointer: buffer)
        while let t = try decoder.decodeInt32() {
            if let e = Self(rawValue:Int(t)) {
                value.append(e)
            }
        }
        return true
    }
```

From @tbkka:

> The inner ‘if’ here needs an ‘else’ where it does something with the unrecognized enum value.
> I think the best approach is going to be to encode a new field and push that constructed field
> onto the unknown storage via the decoder object.  That requires the decoder object to have a
> reference to the unknown field storage.  That should be a small shift.

tbkka:
Fixed in #134

Issue #130, opened 2016-10-31
Provide something for parsing a binary stream of protobuf data


thomasvl:
#23 was where we decided to drop the initial implementation, but some form of generic support for parsing the binary streams without needing message objects has come up again in context of the proto2 syntax unknown fields support.  That currently keeps the data as one blob, so having a simple way to parse that data to extract information would be useful.

allevato:
I've been thinking about this for a while, so I'm going to jot down my train of thought design-wise here.

Other language implementations provide some notion of an *unknown field set* that encapsulates proto2 unknown fields. An *unknown field set* is conceptually a mapping from field numbers to *unknown field* objects, which encapsulate multiple lists of values based on their wire formats—varints (as 64-bit unsigned integers), fixed 64-bit unsigned ints, fixed 32-bit unsigned ints, length-delimited blobs, and nested unknown field sets representing groups. The raw bytes corresponding to the unknown fields must be parsed to provide this interface, and they are then reëncoded when written back on the wire. This ends up being a small performance penalty if the unknown fields are never touched by the user (the majority of cases).

The reason each unknown field set has lists of values for each wire format is because you can have repeated fields. You can also imagine a weird situation where maybe you've concatenated two messages, one with an unpacked repeated field and one with a packed repeated field, where if an older client parsed it, you'd end up with *two* wire formats in the unknown field set: a list of the unpacked values (say, varints), and a list containing the length-delimited blob of the packed values. That's why we have to support multiple wire formats for a single field.

But you can even imagine a degenerate case where someone has crafted a binary message with every possible wire format for the same field:

`[field 1: varint=12345][field 1: fixed32=67890][field 1: fixed64=98765][field 1: length-delimited=0xF00DFACE...][field 1: group=...]`

Even though no well-formed message should ever have that on the wire, the parser would be able to handle it and write them back so that the original contents are preserved.

It should also be noted that the unknown field sets are mutable—users are free to add new fields or mutate the lists of existing values, and those changes will get written out as expected.

The current behavior in SwiftProtobuf of not parsing the unknown fields is a nice feature performance-wise, but we need to add the ability for users to retrieve the parsed values and to modify them. Instead of just concatenating all the unknown bytes at parse-time like we do now, we can do some on-demand parsing by representing each unknown field value internally as a pair of `(bytes, value)` (the exception being length-delimited data which can just be the bytes). Then,

* When an unknown field is encountered, add its bytes to an unknown field value `(bytes, nil)`
[...]

PR #134, opened 2016-11-09
Rework unknown handling

tbkka:
This addresses Issues #127, #128, and #129:

  * It fixes the handling of unrecognized enum values in packed repeated enum fields

  * It simplifies the overall handling of unknown fields:  The generated decoders are now entirely ignorant of unknown field handling issues.  The only generated code that deals with unknown fields is the `unknown` property itself and the traversal logic which iterates that property.  In particular, the boolean `handled` values are entirely gone.

  * It migrates unknown handling to use `Data` instead of `[UInt8]` for consistency with the rest of the system.

  * It now tags generated messages as Proto2 or Proto3.


Reference/conformance/conformance.pb.swift
```diff
@@ -555,15 +547,11 @@ struct Conformance_ConformanceResponse: ProtobufGeneratedMessage {
...
 
-  public mutating func _protoc_generated_decodeField(setter: inout ProtobufFieldDecoder, protoFieldNumber: Int) throws -> Bool {
-    let handled: Bool
+  public mutating func _protoc_generated_decodeField(setter: inout ProtobufFieldDecoder, protoFieldNumber: Int) throws {
     switch protoFieldNumber {
-    case 1, 6, 2, 3, 4, 5:
-      handled = try result.decodeField(setter: &setter, protoFieldNumber: protoFieldNumber)
-    default:
-      handled = false
+    case 1, 6, 2, 3, 4, 5: try result.decodeField(setter: &setter, protoFieldNumber: protoFieldNumber)
```
thomasvl: Is there value in sorting these?  (and making them ranges when they are?)

Sources/SwiftProtobuf/ProtobufBinaryTypes.swift
```diff
@@ -25,14 +25,15 @@ public protocol ProtobufBinaryCodableType: ProtobufTypePropertiesBase {
 
     static func setFromProtobufVarint(varint: UInt64, value: inout BaseType?) throws -> Bool
     static func setFromProtobufVarint(varint: UInt64, value: inout [BaseType]) throws -> Bool
```
thomasvl: Hard to tell skimming the diffs, why do some still return Bools and others throw?

tbkka: All of them throw if the input cannot be parsed for any reason.

Three of these have special requirements from the unknown enum handling.  It may be possible to rework that, but at least that detail is entirely isolated to the ProtobufBinary*.swift sources.

tbkka: Look at `ProtobufFieldWireTypeVarint` in ProtobufBinaryDecoding.swift to see how these specific returns are used, and `extension ProtobufEnum` in ProtobufBinaryTypes.swift to see the only place that actually sets them.

The other specific variation is in `setFromProtobufBuffer` which passes in a `Data` object to receive any leftovers.

The variation among the `setFrom` methods here is not ideal, but I've not come up with an approach I like better.  This does fix the unknown enum handling and it does isolate this issue, so I think it's an improvement.  If you see a way to make it cleaner, please let me know.

Sources/SwiftProtobuf/ProtobufBinaryTypes.swift
```diff
@@ -789,14 +777,21 @@ public extension ProtobufBinaryMessageBase {
...
     init(protobuf: Data) throws {
         try self.init(protobuf: protobuf, extensions: nil)
     }
 
+    init(protobufBuffer: UnsafeBufferPointer<UInt8>) throws {
+        try self.init(protobufBuffer: protobufBuffer, extensions: nil)
+    }
+}
+
+// For Proto3 messages, we need to preserve the unknown fields when we're done
```
allevato: I'm not crazy about these protocols, especially when the implementation is identical for both except for the one line that handles unknowns in proto2 decoding below. Especially since proto3 messages don't have extensions; it doesn't make much sense to have those arguments there.

I wonder if there's another place where we can inject the unknown blob to avoid needing these... Of course, that doesn't solve the problem of extension sets being available to proto3 messages, which could only be solved by having some sort of protocol division...

tbkka: The alternative is basically to generate `decodeIntoSelf`.  I considered that, but...

1. I'm trying to reduce boilerplate in the generated code.

2. There is a certain sense to having message structs explicitly tagged as being Proto2 or Proto3, since there are real semantic differences between the two.

allevato: (1) is definitely a concern. There might be a happy medium; keep `decodeIntoSelf` and then have it call something like `_protoc_gen_takeUnknowns` that gets generated to do nothing in proto3 (perhaps even a protocol extension default implementation) and gets generated to set the `unknown` property in proto3. It's still slightly more generated code, but less.

For (2), I'm not sure how much value there is in letting clients determine whether a message is proto2 or proto3. That would arguably be the role of the descriptor—if we were generating them—and the syntax is technically a property of the *file*, not of the *message*, even though it affects how the behavior of the message.

At least two of the language implementations (Obj-C and Java) don't bother distinguishing between proto2 and proto3 at all in the public message API (they just return null/empty extension sets and unknown stores for proto3, for example), so it's questionable whether we'd want to support something that doesn't appear to have a strong need. If there *were* cases where someone needed this information in a generic fashion, they can already do that using a visitor.

tbkka: > At least two of the language implementations (Obj-C and Java) don't bother distinguishing between proto2 and proto3 at all in the public message API (they just return null/empty extension sets and unknown stores for proto3, for example),...

That's partly a difference in language style:  Obj-C and Java also both publish hash value and equality tests for all objects, unlike Swift.  Swift tends to prefer publishing this kind of distinction at compile time instead of run time.

thomasvl: For ObjC, I we didn't add a different base class because most ObjC code that deals with generic messages would still use the common base between them for passing messages.  With the dynamic runtime, adding a second class would actually make things a bit more complex on the runtime size.  So just letting the methods return empty sets felt a little more ObjC-ish (just like messaging nil, etc.) and keep things simple.

There also was very little interesting in proto3 syntax (over proto2) when the work was initially being done.  The unknown support for mobile is actually pretty useful as it means a server can get feed extra data, clients to store it; and then when the users update their app, the data is there.  With proto3, developers have to write extra logic to go refetch things if they could now have extra info they care about.

tbkka: > We already have the ProtobufExtensibleMessage protocol that extends ProtobufMessage, so I wonder if we can leverage that further by putting extension-set-specific initializers/methods there instead of creating more protocols.

I have some ideas along these lines that I want to explore after I land this change.

tbkka:
@allevato, @thomasvl: Any objections?

Refactoring the protocols is next on my list; I need to actually try a couple of the approaches and see where they lead.  I could tinker with the `decodeIntoSelf` factoring now (and the related proto2/proto3 distinctions), but I suspect it would be ripped up again in that work.

thomasvl:
I'll defer to @allevato, aside from the quick scan the other night, I haven't gotten to go back and read through this.  Sorry. 🙁

allevato:
LGTM.

I've been thinking about how to extend this to give us an actual unknown field storage type that lets users parse out the individual fields into more usable types (like varints, data blobs, fixed ints, etc.) while trying to preserve the performance of not parsing them if they're never requested. Such a class could also be used as a "raw message" like the one we used to have. Is that something you had planned to do? If not, I'm happy to poke at it once this is in.

thomasvl:
@allevato that is captured in issue #130

allevato:
Right—I guess my question was, are you already thinking about the implementation because I don't want to duplicate any work you might be doing. :)

tbkka:
I was planning to look next at the protocol refactoring (which may involve a first stab at splitting out the JSON/Text support metadata) and maybe help @BradLarson with the text format support.  In particular, the protocol refactoring is among the few areas that are likely to change the generated code.

I agree that it makes sense to tie ProtobufUnknown with the raw message support, and I really like the idea of making the actual parsing as lazy as possible, primarily to save memory for common cases where people never look at the unknown fields.  I'm not sure whether ProtobufUnknown _is_ the raw message class or not; that will depend on how the implementation looks as it comes together.  It may be that ProtobufUnknown remains just a simple container and the raw message handling is a parser class that can be attached to any byte source; that would also give the stream parsing I know you've been interested in.  I can't wait to see what you come up with!

Issue #60, opened 2016-10-06
Revisit type names (protocols, structs, etc.)


thomasvl:
Spun out of @soffes comments on https://github.com/apple/swift-protobuf/commit/6b1a0b0a270f31b93c3d4255d94f8d8f2f9e1ae0

Since all the types will be scoped to the package `SwiftProtobuf`, do we really need to repeat "Protobuf" in all of the types?

`SwiftProtobuf.ProtobufGeneratedMessage` vs just  `SwiftProtobuf.GeneratedMessage`
etc.

Taking it a step further, `Generated` message are actually the common case, so we might want to just make the protocol `Message` and use qualifiers for the other types.  There by shortening the common names folks will see.

allevato:
We can get rid of the `Generated` message when we remove the hand-generated WKTs, and just have a single protocol for messages. However, `Message` feels a bit too generic, since Swift module imports are an all-or-nothing prospect currently. Anyone using `Message` to mean anything in their own client code would be forced to explicitly disambiguate.

I don't anticipate that users would need to touch the protocol that much on its own anyway—they'll stick to using the generated types. In that case, a name like `ProtobufMessage` serves to keep the global namespace clean when this module is imported along with others.

Likewise, things like `ProtobufFloat` have to be public because they're referenced in generated code, but in that case, since they're non-generic types, we could nest them inside a caseless enum to keep them separated (and to let users know they don't need to touch them directly in their own code).

tbkka:
The protocol people *should* be using (for generic functions that accept "any protobuf serializable message") is `ProtobufMessage`.    `ProtobufGeneratedMessage` is purely to collect some things that the generated code needs but aren't otherwise part of the "official message API".  The simplest transition might be to rename the other `*Message` protocols with an underscore prefix to mark them as private, or rename them `ProtobufPrivate*` to clarify.

Another way to clarify this:
   * Rename `ProtobufGeneratedMessage` to `_ProtobufGeneratedMessageSupport`
   * Change the plugin so that messages conform to `ProtobufMessage` and to `_ProtobufGeneratedMessageSupport` 

That might make the relative significance of `ProtobufMessage` clearer.

tbkka:
[...]

Issue #17, opened 2016-09-28
Can we reduce the number of protocols?


allevato:
A single generated message currently conforms to a large number of protocols: `ProtobufGeneratedMessage`, `ProtobufAbstractMessage`, `ProtobufMessage`, `ProtobufMessageBase`, `ProtobufBinaryMessageBase`, `ProtobufJSONMessageBase`, `ProtobufTraversable`, and a handful of standard library protocols.

For the first four in particular, it's not immediately clear from the names alone what the different responsibilities of each one are. (Some of them relate to the differences between the hand-written well-known types and other generated messages, so if we can get rid of the hand-written ones in #13, we can get an easy reduction there.)

Similarly, is there value in keeping `ProtobufBinaryMessageBase` separate? All messages should be binary codeable, since that's defined by the protocol buffer spec. Since it's unlikely that you'd want a non-proto-message type to implement `ProtobufBinaryMessageBase`, its members can be folded into one of the other message protocols.

Likewise, `ProtobufTraversable` doesn't need to stand alone—it provides only one method, and is only implemented by the message types (once #16 is fixed, removing the message/group distinction). That method can be moved into one of the other message protocols.

Reducing this API service will improve the size footprint of the runtime library and make it easier to maintain/understand.

tbkka:
Few notes on the current status quo as I work through some of this:

I don't think there's a technical obstacle to collapsing `ProtobufMessage`, `Protobuf*MessageBase`, and `ProtobufTraversable`.  I'm going to see what that looks like.

I also don't think there's a technical reason preventing `ProtobufAbstractMessage` (the general implementation base) and `ProtobufGeneratedMessage` (the implementation base for generated messages) from being combined.  I'll play with that as well.

However, `ProtobufMessage` and `ProtobufAbstractMessage` cannot be combined.  `ProtobufMessage` is used as a protocol type so cannot have `isEqualTo` (which has a `Self` constraint).  But right now all messages (generated or otherwise) must have `isEqualTo` in order to handle type-erased equality (needed by extensions).  Essentially, `ProtobufAbstractMessage` is the implementation base for all messages, while `ProtobufMessage` is the protocol type.  Maybe there's a better name for `ProtobufAbstractMessage`?

Another interesting constraint:  I've talked to a few people who tried to figure out the base message types by examining the generated code.  So I really want the messages to directly specify `ProtobufMessage` as their first conformance.  As pointed out above, that's insufficient; there must be others, but I think it's important that one appear first.

[...]

PR #139, opened 2016-11-17
Tbkka grand renaming

tbkka:
This renames a lot of types along the lines suggested in Issue #60 "Revisit Type Names", simplifies some of the protocol hierarchies along the lines suggested in Issue #17 "Can we reduce the number of protocols?", breaks up some of the larger source files, renames files to match their contents, and eliminates some duplicate code along the way.

This is admittedly pretty disruptive: Since the type name change and Message hierarchy refactoring was already churning a lot of stuff, I thought I would just dig in and try to get a lot of similar changes in now.


Sources/SwiftProtobuf/Map.swift
```diff
@@ -17,7 +17,7 @@
 import Foundation
 import Swift
 
-public struct ProtobufMap<KeyType: MapKeyType, ValueType: MapValueType>
+public struct Map<KeyType: MapKeyType, ValueType: MapValueType>
```
allevato: In general I'm supportive of simplifying these names (I'm still iffy about `Message`, but we can see if it causes problems for people downstream), but this one feels wrong to me. Technically this type is not a "map", but more like a type descriptor of sorts. I feel like we need to capture that in some way in the name (especially since the analogous types, like `ProtobufInt32`, will need some kind of suffix on their names to work if you want to remove `Protobuf` there).

`Descriptor` would be the wrong word to use here though, since that's a loaded term in protobuf. `MapField`/`Int32Field`/etc?

allevato: I think leaving `Proto` on these types would be fine, since as you say, they're metadata related specifically to the protobuf nature of these types. They're also unlikely to be used by clients unless they're doing something advanced like writing their own visitor; most users aren't going to see them, so the slightly more "technical" name doesn't really uglify code.

tbkka:
@thomasvl, @allevato :  Any concerns before I merge this?

I know the file renaming and shuffling is a little disruptive, so if you really want to squeeze something else in first (and make me do the merge instead of you ;-), let me know.

allevato:
Looks good to me—I'm OOO today so this would be an ideal time to merge it :)

soffes:
Just updated to this! So great to rename `ProtobufJSONMessageBase` to just `Message` everywhere 🎉 Thanks so much for doing this @tbkka! ❤️

PR #140, opened 2016-11-17
Better enum case stemming

### LINKED ISSUE #138: Enum case stemming can result in invalid Swift
I like how `protoc-gen-swift` generates enum cases which are "stemmed", meaning that you remove the first word if the case name if it's the same as the enum name.

However, if I have a proto file which looks like this:

```
enum Clock {
    CLOCK_24_HOUR = 1;
    CLOCK_12_HOUR = 2;
}
```

(which I in fact do in my project), then I get an output file which looks like this:

```swift
public enum Clock: ProtobufEnum {
  public typealias RawValue = Int
  case 24Hour // = 1
  case 12Hour // = 2
  //...
```

...which is not valid Swift.

I think the right answer is to see if any of the stemmed case names are invalid Swift, and then not stem any of them. Working on a PR but wanted to file this to get it on your radar.

dpassage:
This fixes #138.

I tried to match the style of the files I was editing, which were different. Also, I only check for identifiers starting with `[0-9]` and literal `_`; if the new identifier ends up being a reserved word, other code in the generator already quotes it with backticks.

Thanks for putting as much work and thought into this as you do!


Sources/protoc-gen-swift/StringUtils.swift
```diff
@@ -274,3 +274,12 @@ func stringToEscapedStringLiteral(_ s: String) -> String {
   }
   return out + "\""
 }
+
+func isValidSwiftIdentifier(_ s: String) -> Bool {
```
allevato: The name of this function is a bit misleading; the comment makes it clearer why this works, but I would fear that future readers/users might think this function does what the name says in *all* cases. What about `isProtoIdentifierValidSwiftIdentifier`, which captures the notion that the input is assumed to already be a valid protobuf identifier?

tbkka: I would like to use `isValidSwiftIdentifier` throughout the generator.  How hard would it be to expand it to actually do what it says?

The full spec is not small, but it's not inherently complicated, either.  It boils down to "first character must be this and the rest must be that":

 https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/zzSummaryOfTheGrammar.html

allevato: If you go this route, consider precompiling a (rather large) [`NSRegularExpression`](https://developer.apple.com/reference/foundation/nsregularexpression) as a private global and matching against that. It looks like it handles the higher Unicode code points the way it should with `\U`, so the hard part is just getting everything written down in the expression.

dpassage: Working on that point, thanks!

dpassage:
OK! `StringUtils` now has 3 validation functions:

* `isValidSwiftIdentifier`, which matches the full production from the Swift Programming Language for identifiers
* `isUsableSwiftIdentifier`, which excludes things we can't use in declarations, i.e., `_` and `$0`-`$9`
* `isValidProtobufIdentifier`, which matches valid identifiers for `.proto` files.

In the process, I added unit tests against `protoc-gen-swift` to cover these. I did run `swift package generate-xcodeproj`, but the diff was nuts - are there options you used in generating the one in the project?

In any case, hopefully this addresses the concerns.

dpassage:
@allevato @tbkka Can you take another look at this PR? Thanks!

kientzle:
The tests look yummy.

Do you have any way to test this on Linux?  I'm not sure that NSRegularExpression exists there.  If not, this may have to be coded in a somewhat more explicit fashion, building on basic functions like:
```
func isValidSwiftIdentifierCharacter(_ c: Character) -> Bool {
  switch c {
  case "a"..."z", "A"..."Z", "_": return true
  case "\u00a8"..."\u00aa": return true
  /* ... etc ... */
  default: return false
  }
}

func isValidSwiftIdentifier(_ s: String) -> Bool {
  for c in s.characters {
    if !isValidSwiftIdentifierCharacter(c) {
      return false
    }
  }
  return true
}
```

dpassage:
I don't have an immediate way to run this on Linux, no. If someone could help with validating my PR there I'd appreciate it, otherwise I can get it going...

dpassage:
Sure 'nuff, no `NSRegularExpression` on linux, working on it...

dpassage:
OK, so I've updated the `StringUtils` so that it compiles on Linux. I also updated some of the tests so they `import Foundation`, as `Data` isn't in the standard library.

In any case, I'm at the point in this where I'm doing more work to make things work on Linux than to address this bug in the `swift-protobuf` compiler. I'm happy to help with that effort; is that something that needs to happen in order to merge this bug fix?

tbkka:
I'm working through some other issues right now, but should have time to help with this later this week.

dpassage:
Great, thanks!

tbkka:
Now that #178 has been merged, you should be able to simplify this a lot.  Thanks for your patience while we work this out.

dpassage:
No, please, thank you!

dpassage:
Closing in favor of a new branch...

PR #142, opened 2016-11-21
Clean-up binary field skipping.

allevato:
This code was never updated to replace raw math with `FieldTag` when
that type was introduced.


tbkka:
Looks good!

PR #143, opened 2016-11-22
Update tag version to latest one

kitasuke:
`0.9.23` should be `0.9.24`.


thomasvl:
The docs says to use what ever one you pick, so it is more meant as an example.

PR #144, opened 2016-11-26
Assign access control by visibility option

kitasuke:
It would be nice to use same access control for properties and functions in a struct unless these are not defined in protocol. Since default value of `Visibility` is set to `Internal`, we might have public properties and functions in internal struct which looks unnecessary for some case. However, I don't know any background about why `Visibility` option applies to only struct itself, so it might be better to leave it as it is though.

### DIFF
--- a/Reference/conformance/conformance.pb.swift
+++ b/Reference/conformance/conformance.pb.swift
@@ -60,17 +60,17 @@ import SwiftProtobuf
 //      iOS apps, where fork/stdin/stdout are not available.
 
 enum Conformance_WireFormat: ProtobufEnum {
-  public typealias RawValue = Int
+  typealias RawValue = Int
   case unspecified // = 0
   case protobuf // = 1
   case json_ // = 2
   case UNRECOGNIZED(Int)
 
-  public init() {
+  init() {
     self = .unspecified
   }
 
-  public init?(rawValue: Int) {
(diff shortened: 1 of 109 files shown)


allevato:
This looks reasonable; if a struct is internal, there's no reason for its members to be more visible than that.

Since this PR changes code generation, can you please add commits that regenerate the test protos and reference protos to verify that everything builds and that tests pass after this change? See the [documentation in the Makefile](https://github.com/apple/swift-protobuf/blob/master/Makefile#L17-L23). Thanks!

kientzle:
This looks good, as long as it passes tests after regenerating.

kitasuke:
Thank you for your feedback! I followed the introduction and committed the test protos and reference protos. I checked that tests passed successfully, but is this correct way?

tbkka:
Thank you!  I'll go ahead and merge this.

NachoSoto:
I recently upgraded and all my types default to `internal`. Which option can I use to control visibility? I tried `swift_visibility` but no luck.

NachoSoto:
Never mind, I just found it [in the plugin docs](https://github.com/apple/swift-protobuf/blob/master/Documentation/PLUGIN.md). For anyone else looking, you need to set the `Visibility` option in the plugin, not inside each `.proto` file.

PR #145, opened 2016-11-28
Update generated names for Extensions

thomasvl:


Sources/PluginLibrary/swift-options.pb.swift
```diff
@@ -10,21 +10,21 @@ import Foundation
 import SwiftProtobuf
 
 
-let Google_Protobuf_FileOptions_appleSwiftPrefix = ProtobufGenericMessageExtension<ProtobufOptionalField<ProtobufString>, Google_Protobuf_FileOptions>(protoFieldNumber: 50138, protoFieldName: "apple_swift_prefix", jsonFieldName: "appleSwiftPrefix", swiftFieldName: "appleSwiftPrefix", defaultValue: "")
+let Extensions_appleSwiftPrefix = ProtobufGenericMessageExtension<ProtobufOptionalField<ProtobufString>, Google_Protobuf_FileOptions>(protoFieldNumber: 50138, protoFieldName: "apple_swift_prefix", jsonFieldName: "appleSwiftPrefix", swiftFieldName: "appleSwiftPrefix", defaultValue: "")
```
thomasvl: The swiftFieldName on these is a little odd; I'm not really sure what we want.  The top level ones did end up changing with how I've coded this.  I guess the main use of it will be for mirror, in which case, maybe both forms should make it more clear this is an extension, maybe use some notation more like TextFormat does for extensions?  Right now this looks a bit like a normal field which seems wrong.

I also thought extensions weren't supported in json, so I'm not sure why that's needed on these.

allevato:
This looks like the right behavior to me for extension naming; I'll give @tbkka a chance to look at it as well.

Sources/PluginLibrary/swift-options.pb.swift
kientzle: I implemented full JSON support for all of proto2 (including extensions, groups, etc) just to see what it would require.  Turns out to be pretty straightforward, actually.  We should probably rip that back out again to avoid confusion.

thomasvl: Spun up #146 to hopefully capture the JSON part.  I'm still not sure about what should be used for the `swiftFieldName` part.

allevato: `swiftFieldName` should probably match the property added to the struct, even if that means it starts with an uppercase letter (because of the package prefixing).

We have to choose between either (1) package-qualified property names that match the name mangling scheme, which means they'll start with an uppercase, or (2) introducing yet another name mangling scheme to make them start lowercase. I'd choose (1) just because we don't need yet another mapping. As long as those property names won't ever collide with generated type names, we should be ok.

No matter how we slice it, extensions are going to be ugly, so let's not twist ourselves too much about it.

thomasvl: Ha, I picked the wrong thing to look at when thinking about this initially.  `swift-options.proto` has no proto package and no Swift prefix, so that is why this remained _simple_ name.  Just checked all generated code for extensions in the `Reference` directory, they names are fully scoped.  There is one other test file with no proto package or prefix, so the names didn't change there either (as expected).  So I think we're good here.

thomasvl:
Assuming no other comments, I'll squash/commit this tomorrow.

PR #147, opened 2016-11-29
Add an instruction for usage with Carthage

kitasuke:
It looks like that an installation of this library by Carthage is already supported, so I added an instruction for that.


thomasvl:
Thanks!

Issue #97, opened 2016-10-15
Equality is inconsistent with other languages


allevato:
For proto2 messages, the current Swift implementation considers an unset field to be equal to one explicitly set to its default value. This is not equality, but [equivalence (see C++)](https://github.com/google/protobuf/blob/fd046f6263fb17383cafdbb25c361e3451c31105/src/google/protobuf/util/message_differencer.h#L120-L123). Equality distinguishes between an unset field and one explicitly set to its default value.

This should be a quick fix; it simplifies the generator slightly since we don't need to worry about what the default value is, and the tests will need to be modified where they exercise the current behavior.

thomasvl:
Since hash is done via Sources/SwiftProtobuf/HashVisitor.swift the fact that isEqual is special casing default values likely means we're breaking the contract between isEqual and hash - in that the defaults could make isEqual end up true while the hash values are actually different.

tbkka:
We could augment the `test{Format}{Encode|Decode}` macros so that whenever they assert equality, they also assert hash equality.  That would give us some coverage of the hash contract for free on top of our existing tests.  Of course, tailored hash tests would also be a nice addition.

thomasvl:
I still need to do more complete tests for these, but I'm going to go ahead and close this since we believe the methods themselves are fixed.

PR #149, opened 2016-11-30
Remove field names from visitor methods.

### LINKED ISSUE #9: Split strings off visitor pattern
Look at moving the string args off the visitor method (just field number and values).  Then update the things that need the strings can use the maps between strings:field numbers to do the lookup instead.

The string based things will end up a little slower, but it should help shrink the generated code sizes as well as speed up the binary format (since it won't have to pass unused arguments all the time).

allevato:
This PR moves field names into a mapping provided by a new protocol.

The name mapping detects duplicates (proto name and JSON name the same) and only generates one copy if they are. (I didn't bother doing the same for Swift field name because I'd like to just eliminate it completely soon. But I'm not doing that here to keep this focused.)

A future PR will implement a *transformability* check on names. Very few JSON names in real world protos are actually different than what `protoc` would generate. The code generator should use the same logic to determine if that JSON name can be auto-derived; if so, we can eliminate more strings by transforming them at runtime instead.

Also in the future, we can slice this protocol off the main structs and move it to an extension in separate files. That will let users omit it if they don't want them.

Fixes #9.


Sources/SwiftProtobuf/FieldNameMap.swift
```diff
@@ -0,0 +1,105 @@
...
+  /// A "bundle" of names for a particular field. We use different "packed"
+  /// representations to minimize the amount of string data that we store in the
+  /// binary.
+  ///
+  /// TODO: Remove the Swift field name since it's only used for debugging
+  /// purposes and use the proto name instead at those usage sites.
+  public enum Names {
+
+    /// The proto name and the JSON name are the same string
+    case same(proto: String, swift: String)
```
tbkka: Scanning the generated code, most of the time, all three of these seem to be identical.  Should there be a case for that as well?

Sources/SwiftProtobuf/FieldNameMap.swift
```diff
@@ -0,0 +1,105 @@
...
+  public init() {}
+
+  /// Creates a new bidirectional field name/number mapping from the dictionary
+  /// literal.
+  public init(dictionaryLiteral elements: (Int, Names)...) {
+    for (number, name) in elements {
+      numberToNameMap[number] = name
+      protoToNumberMap[name.protoName] = number
+      jsonToNumberMap[name.jsonName] = number
+    }
```
tbkka: I'm nervous about eagerly building all these maps.  Would it save data space to build them lazily?

Sources/SwiftProtobuf/ProtoNameProviding.swift
```diff
@@ -0,0 +1,53 @@
...
+
+/// Messages conform to this protocol to provide the proto/text and JSON field
+/// names for their fields. This allows these names to be pulled out into
+/// extensions in separate files so that users can omit them in release builds
+/// (reducing bloat and minimizing leaks of field names).
+public protocol ProtoNameProviding {
+
+  /// The mapping between field numbers and proto/JSON field names defined in
+  /// the conforming message type.
+  static var _protobuf_fieldNames: FieldNameMap { get }
```
tbkka: Elsewhere, the code generator uses `_protoc_generated` as a prefix for generated items.

That was done to make it easy to override, but it might make things clearer here as well.

Sources/SwiftProtobuf/ProtoNameProviding.swift
```diff
@@ -0,0 +1,53 @@
...
+
+  /// Returns the field name bundle for the field with the given number.
+  ///
+  /// The default implementation looks up the field in the static name map,
+  /// which is sufficient for proto3. For proto2 extensions, making this an
+  /// instance method allows generated messages to override it and ask their
+  /// extension sets for names as well.
+  func _protobuf_fieldNames(withNumber number: Int) -> FieldNameMap.Names?
```
tbkka: I like the `_protobuf_` prefix here.  We should probably use that elsewhere for internal methods.

Sources/SwiftProtobuf/ProtobufJSONEncoding.swift
```diff
@@ -21,16 +21,32 @@ struct ProtobufJSONEncodingVisitor: ProtobufVisitor {
...
         }
     }
 
     mutating func withAbstractVisitor(clause: (inout ProtobufVisitor) throws -> ()) throws {
         encoder.startObject()
+
+        // TODO: This is a bit of a hack that exists as a workaround to make the
+        // hand-written Any serialization work with the new design. We need to
+        // generate those WKTs instead of maintaining the hand-written ones,
+        // handle the special cases differently, and then remove this.
```
tbkka: I've started thinking of 'Any JSON' serialization as its own distinct serialization different from 'regular JSON'.  The two obviously have a lot in common, but they're not the same.

Sources/SwiftProtobuf/ProtobufJSONTypes.swift
```diff
@@ -818,12 +818,7 @@ public extension ProtobufJSONMessageBase {
     }
 
     func serializeAnyJSON() throws -> String {
-        var jsonVisitor = ProtobufJSONEncodingVisitor()
-        try jsonVisitor.withAbstractVisitor {(visitor: inout ProtobufVisitor) in
-            try visitor.visitSingularField(fieldType: ProtobufString.self, value: anyTypeURL, protoFieldNumber: 1, protoFieldName: "type_url", jsonFieldName: "@type", swiftFieldName: "typeURL")
-            try traverse(visitor: &visitor)
-        }
-        return jsonVisitor.result
+        return try ProtobufJSONEncodingVisitor(message: self, anyTypeURL: anyTypeURL).result
```
tbkka: As above, I've started thinking of `AnyJSON` as a distinct serialization.  The original code here leveraged the JSON encoder to support most of the fields, but didn't ask the regular JSON encoder to handle the special `@type` member.

Reference/google/protobuf/any.pb.swift
```diff
@@ -108,18 +108,14 @@ import Foundation
...
-    "typeUrl": 1,
-    "value": 2,
-  ]}
-  public var protoFieldNames: [String: Int] {return [
-    "type_url": 1,
-    "value": 2,
-  ]}
+  public static let _protobuf_fieldNames: FieldNameMap = [
+    1: .unique(proto: "type_url", json: "typeUrl", swift: "typeURL"),
+    2: .same(proto: "value", swift: "value"),
```
tbkka: I know you've been agitating to drop the swift names, but if all three are usually the same, there may not be much gain by dropping them.  (And I think there are small but positive advantages for mirror and debug description to use the  swift field names instead of the proto names.)

Sources/SwiftProtobuf/FieldNameMap.swift
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /// Returns the name bundle for the field with the given number, or `nil` if
+  /// there is no match.
+  public func fieldNames(withNumber number: Int) -> Names? {
```
tbkka: `fieldNames(withNumber:)` is a very Objective-C naming style.  I would personally prefer `fieldNames(for:)` or `fieldNames(protoFieldNumber:)` or `fieldNames(forProtoFieldNumber:)` or `fieldNames(forNumber:)`.

Sources/SwiftProtobuf/ProtoNameProviding.swift
allevato: My thinking exactly—I was going to file an issue to audit the properties/methods of any public protocols that we inject into generated messages and add a prefix like `_protobuf_`, but I figured I'd float it here first. Glad you agree!

Reference/google/protobuf/any.pb.swift
allevato: I'm curious what you think about #149 then. The reason I'm itching to drop the Swift names is because they serve no purpose other than debugging—they're only used when generating the `debugDescription` and the mirror, and aren't used as input at all.

All three strings are unlikely to be the same for `more_than_one_word` proto field names, but the Swift identifier is likely to be identical to the JSON one in many cases, so your observation about them not adding much overhead in most cases is probably true. That being said, I don't think we mangle the names *so much* when we generate properties that users would be terribly confused if we used the proto names instead of the Swift names in `debugDescription` and `Mirror`. I want to toss out everything that we don't have a strong need for to keep things slim.

Sources/SwiftProtobuf/FieldNameMap.swift
allevato: (This will depend on the resolution of the other Swift field name comments.)

Sources/SwiftProtobuf/ProtoNameProviding.swift
allevato: Do you think it's worth distinguishing between `_protobuf_` below and `_protoc_generated_`? I wonder if we should just use `_protobuf_` everywhere.

Sources/SwiftProtobuf/FieldNameMap.swift
tbkka: Good point that the `FieldNameMap` itself is already lazily built.  We can revisit this later to see if there's a win here.

Sources/SwiftProtobuf/FieldNameMap.swift
tbkka: That all makes sense.

Sources/SwiftProtobuf/ProtoNameProviding.swift
tbkka: I think it makes the various protocols easier to understand to have `generated` in the name of the things that are actually built by the code generator.  Just seeing the names, it's a lot clearer:
```
   static var _protoc_generated_fieldNames: FieldNameMap
   func _protobuf_fieldNames(for:)
   func _protobuf_fieldByName()
```

Reference/google/protobuf/any.pb.swift
tbkka: Mirrors are used by things other than debugging, of course, though I hope that people won't use our mirror to tie our objects into some other JSON serialization library.  😱

As for #149, I'm a little torn to be honest.  I agree that slimming things down is good, but I also place a high value on features (in case you hadn't noticed).  And the approach you've taken here suggests that simply deduplicating the JSON and Swift names could eliminate a lot -- maybe even most -- of the remaining overhead.

Sources/SwiftProtobuf/ProtobufJSONEncoding.swift
allevato: The way that JSON for `Any` and the other well-known types have so far been handled here was the most challenging thing to get working with the new design, because of the handwritten protos and special cases—hence the hack above. I don't plan to let that hack live for very long, but I didn't want to overload this PR with other work.

To fix this for good, we need to get rid of the handwritten protos and generate them like everything else. The side effect of doing that is that we can no longer customize the JSON representation in the type itself, because even Swift extensions can't override behaviors. So instead, we'd have to do something like the [Java implementation](https://github.com/google/protobuf/blob/master/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java#L564), which just handles those special cases within the JSON printer itself.

While I can definitely appreciate the desire to make things generalized and extensible, in this case it's added a lot of complexity to the API that would be good to clean up. The protobuf spec says [this finite set of specific types is given special JSON treatment](https://developers.google.com/protocol-buffers/docs/proto3#json) and nothing more, so it's *better* to be restrictive and limit customization because we don't want to (a) increase our maintenance burden and (b) make it possible for anyone to write types that arbitrarily don't conform to the spec.

Sources/SwiftProtobuf/ProtobufJSONTypes.swift
allevato: ^ See above.

Sources/SwiftProtobuf/ProtoNameProviding.swift
allevato: There's a bit of a rub here: some things could end up being `_protobuf` because they're part of an implementation-detail protocol, but then some but not all generated messages might include customized implementations, which if we followed the rules to the letter, would want to be named `_protoc_generated` instead. (For example, `_protobuf_fieldNames(for:)` has a default implementation in a protocol extension, but proto2 messages generate their own override so that they can touch the extension set.)

Reference/google/protobuf/any.pb.swift
allevato: Features are nice if they provide high value, but it's hard for me to see "the field is rendered as `fooBar` instead of `foo_bar`" as a high value feature, at least not worth the complexity it would add to the name bundles (what if proto name, JSON name, and Swift name are all different? What if proto name and Swift name are the same but JSON name is different? What if...). And aside from the debugger and playgrounds, are there any other compelling use cases for `Mirror` that would be harmed by using the proto names instead of the Swift ones? (To be honest, I've never been a fan of that API. What looks at first glance like it's supposed to be a general reflection API turns out to be nothing more than a value snapshot renderer that hooks into limited places in the IDE.)

One of the major design goals for protocol buffers in general has been to make sure we're not shipping anything that users don't truly need—that's why, for example, the Objective-C implementation doesn't generate descriptors. We still don't have the luxury of effectively-infinite space on user's phones like we do on servers, and I want to stay very conscious of that since mobile is a major audience for this library. Sure, if we look at individual strings here and there it's easy to say "that's only a few bytes, what's the big deal?" but if we add up a lot of those, we start to get real savings.

So if it seems like I'm harping on about a handful of characters, it's more of a general design philosophy than picking at specific nits. IMO, we really must be focusing on performance and size right now instead of minor features in this early period before we're doing official releases and we have the "we will break you" warning banner up. It's **much** harder to take away or change things once people are depending on them than it is to add new things later.

allevato:
Just pushed the naming changes.

Some performance results to go with this change (`./perf_runner.sh 100 int32`):

The generated code size shrunk, taking the harness down from 242604 to 238412 bytes unstripped, and from 136332 to 132068 bytes stripped. 4K for about 100 fields.

The execution time shrunk slightly, from 35ms inside `Harness.run` to 33ms, so basically statistical noise. Not surprising, since the harness doesn't exercise the JSON code paths. (I should update it to do so and run the tests again.) We were hoping that removing all of the field names from the visitor calls would give more of a speed up, but it's likely not that dramatic because pushing static string pointers onto the stack is very cheap.

Sources/PluginLibrary/plugin.pb.swift
```diff
@@ -118,13 +113,13 @@ public struct Google_Protobuf_Compiler_CodeGeneratorRequest: ProtobufGeneratedMe
 
   public func _protoc_generated_traverse(visitor: inout ProtobufVisitor) throws {
     if !fileToGenerate.isEmpty {
-      try visitor.visitRepeatedField(fieldType: ProtobufString.self, value: fileToGenerate, protoFieldNumber: 1, protoFieldName: "file_to_generate", jsonFieldName: "fileToGenerate", swiftFieldName: "fileToGenerate")
+      try visitor.visitRepeatedField(fieldType: ProtobufString.self, value: fileToGenerate, protoFieldNumber: 1)
```
thomasvl: Given we're visiting proto fields, s/protoFieldNumber/fieldNumber:/ ?

Sources/SwiftProtobuf/FieldNameMap.swift
```diff
@@ -0,0 +1,105 @@
...
+
+  /// The mapping from JSON names to field numbers.
+  private var jsonToNumberMap: [String: Int] = [:]
+
+  /// Creates a new empty field name/number mapping.
+  public init() {}
+
+  /// Creates a new bidirectional field name/number mapping from the dictionary
+  /// literal.
+  public init(dictionaryLiteral elements: (Int, Names)...) {
```
thomasvl: Stupid question - How do these literal end up in the binary?  i.e. - would making this a tuple or just put the field number in Names so this is an array be any smaller/better?

Reference/google/protobuf/any.pb.swift
thomasvl: anecdotal from ObjC protobuf's history - -description used to print out the objc names as it logged the message.  That output would make it into emails/bug reports/etc, and would some times confuse the folks that worked server side or on other platforms as the fields didn't fully line up to what they expected.  When we switch to using the TextFormat names, all the time spent saying, "no, that's what is in the message, they are just the objc names" went away as the protos were what all developers expected since the format is programming language agnostic.  So I sorta see #149 as helping apply that same logic.

Sources/SwiftProtobuf/FieldNameMap.swift
thomasvl: Re reading this, I guess they are tuples?  So I guess the compiler is transforming the source into an array of tuples?

Sources/PluginLibrary/plugin.pb.swift
```diff
@@ -118,13 +113,13 @@ public struct Google_Protobuf_Compiler_CodeGeneratorRequest: ProtobufGeneratedMe
...
     }
     if let v = _parameter {
-      try visitor.visitSingularField(fieldType: ProtobufString.self, value: v, protoFieldNumber: 2, protoFieldName: "parameter", jsonFieldName: "parameter", swiftFieldName: "parameter")
+      try visitor.visitSingularField(fieldType: ProtobufString.self, value: v, protoFieldNumber: 2)
     }
     if !protoFile.isEmpty {
-      try visitor.visitRepeatedMessageField(value: protoFile, protoFieldNumber: 15, protoFieldName: "proto_file", jsonFieldName: "protoFile", swiftFieldName: "protoFile")
+      try visitor.visitRepeatedMessageField(value: protoFile, protoFieldNumber: 15)
```
allevato: I've had that in the back of my head for a while but was planning to do it in a separate PR, but I guess since I'm touching everything already... why not?

Sources/SwiftProtobuf/ProtobufJSONDecoding.swift
```diff
@@ -121,10 +121,12 @@ public struct ProtobufJSONDecoder {
     // TODO: ProtobufMessage here should be ProtobufFieldDecodableType to encompass both groups and messages
     // then this can merge with the decodeValue below...
     public mutating func decodeValue<M: ProtobufMessage>(key: String, message: inout M) throws {
+        guard let nameProviding = (M.self as? ProtoNameProviding.Type) else {
```
allevato: I don't think so, since unlike the visitor pattern that takes a message in its initializer, the message being decoded into here isn't known until the method is called.

Sources/SwiftProtobuf/FieldNameMap.swift
allevato: The compiler parses the `[ k: v, ... ]` literal and generates assembly code that creates a `ContiguousArray` of tuples. Since we can guarantee that the code generator generates them in sorted order, we *could* look at just preserving that array instead of building the number->name dictionary and then binary search it. But I think it's something we can revisit later.

Sources/SwiftProtobuf/ProtobufJSONEncoding.swift
tbkka: > .. because even Swift extensions can't override behaviors

Actually, they can.  A swift extension on a struct can override/replace a method defined in a protocol extension.  This is how Swift Protobuf allows end developers to override/replace the default hash and equality implementations, for instance.  (It defines hashValue and isEqual(to:) in a protocol whose default implementation just invokes the `_protoc_generated` version.)

Sources/SwiftProtobuf/ProtobufJSONTypes.swift
tbkka: There are several different pieces here:
  1. Standard JSON serialization of standard messages.
  2. Standard JSON serialization of WKTs
  3. 'Any JSON' serialization of standard messages  (which adds an `@type` field)
  4. 'Any JSON' serialization of WKTs (which adds an `@type` field and puts the custom JSON version into a `value` field)

I initially tried to build one JSON encoding flow that handled all four of these and it got to be a right mess.  I managed to tame it some by treating the last two as a fundamentally separate encoding/decoding flow (which share a lot of the back-end infrastructure with the first two).  That's how `serializeAnyJson()` came to exist.

Where it gets really confusing is the connection between the second item above (regular JSON encoding of Any fields) and the last two (the encoding of the Any field contents).

Reference/google/protobuf/any.pb.swift
tbkka: > ... description ... output would make it into emails/bug reports/etc

That's a good argument for using the text encoder as the default for debugDescription.

Sources/SwiftProtobuf/ProtobufJSONEncoding.swift
allevato: > A swift extension on a struct can override/replace a method defined in a protocol extension.

That's a different situation than the one I was referring to, which was that methods generated *directly. However, when I wrote that I was forgetting that `serializeJSON` was implemented in a protocol extension rather than being generated into the struct, so it should still be possible to generate the remaining WKTs and then introduce an extension that overrides `serializeJSON` to do the right thing, which is good.

I'm not crazy about the fact that it allows people to arbitrarily extend their own custom types with a different JSON representation (that's against spec), but we'd have to overhaul the design more to lock that down and I'm not sure it's worth it.

> This is how Swift Protobuf allows end developers to override/replace the default hash and equality implementations, for instance.

Which we shouldn't be allowing either, but that's a separate issue. Equality of messages is well-defined by protobuf (see #97), and users should not be allowed to provide a different implementation that could produce different results than what the spec defines.

PR #150, opened 2016-12-01
Update the protos copied from google/protobuf.

thomasvl:
- Add Makefile target to update the protos that are copied over. This hacks
  back in the custom swift prefixes until we look at upstreaming things.
- Regenerate the generated sources.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,14 +38,18 @@ AWK=awk
 # Installation directory
 BINDIR=/usr/local/bin
 
+# Install tool name
 INSTALL=install
 
-PROTOC_GEN_SWIFT=.build/debug/protoc-gen-swift
+# Where to find a google/protobufs checkout. This is only used for the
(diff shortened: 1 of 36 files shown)


Makefile
```diff
@@ -322,6 +326,22 @@ regenerate-test-protos: build
 		${GENERATE_SRCS} --tfiws_out=FileNaming=DropPath:Tests/SwiftProtobufTests $$t; \
 	done
 
+#
+# Helper to update the .proto files copied from the google/protobufs distro.
+#
+update-proto-files:
```
allevato: Would it be difficult to add an existence check for `${GOOGLE_PROTOBUFS_CHECKOUT}` and print a nice error message that says "Please clone google/protobuf as a sibling of swift-protobuf"?

(Or even clone it for them... but then you'd have to choose between http and ssh.)

allevato: Thanks!

PR #153, opened 2016-12-01
Focus on the official swift_prefix file option.

thomasvl:
- Mark the custom apple_swift_prefix as deprecated.
- Output a message telling folks to migrate.
- Update the one other test file missed to use the official option.
- Kill off option_apple_swift_prefix.proto, isn't used.


Protos/swift-options.proto
```diff
@@ -3,7 +3,9 @@ syntax="proto2";
```
thomasvl: Since the old version of descriptor.proto we had in this repo also had `swift_prefix`, we could just kill this file and the support for it since the .proto files will work with our old releases also.

thomasvl: Actually, https://github.com/search?l=Protocol+Buffer&q=apple_swift_prefix&ref=searchresults&type=Code&utf8=%E2%9C%93 seems to say no one but us is using the old option, so I'll update again to kill it outright.

thomasvl: Amended.

PR #155, opened 2016-12-06
Fix duplicate references in watchOS target

kitasuke:
Running build for watchOS target fails in master branch. These files below were referenced twice, so I just followed `update to recommended settings` on Xcode to fix.

```
Google_Protobuf_Struct.swift
Google_Protobuf_Timestamp_Extensions.swift
Google_Protobuf_Wrappers+Extensions.swift
source_context.pb.swift
timestamp.pb.swift
type.pb.swift
Varint.swift
wrappers.pb.swift
ZigZag.swift
```


thomasvl:
Thanks

PR #157, opened 2016-12-06
Swift 3.0.1

soffes:
Set Xcode to use Swift 3.0.1. Compiled and ran the tests after changing this and everything passed without any source changes 🎉


soffes:
Thank you!

PR #162, opened 2016-12-06
Bring over conformance test runner from apple/swift-protobuf-test-conformance

### LINKED ISSUE #160: Move conformance test runner into this project
Forking off from #125.

Having the conformance test runner in a separate project makes it difficult to iteratively develop and ensure that both the runner stays up to date and that changes we make to the runtime/codegen don't break conformance tests. We should add it as a top-level target in `swift-protobuf`'s package description, then delete the `swift-protobuf-test-conformance` project.

Although it will increase build times currently, keeping our code health up is (IMO) a higher priority right now.

Eventually, the accepted Swift Evolution proposal [SE-0146](https://github.com/apple/swift-evolution/blob/master/proposals/0146-package-manager-product-definitions.md) will be implemented that lets us define individual products, and clients can depend on just the runtime product to avoid building the entire project.

I'll prepare a PR that moves over the necessary files, then updates them to compile with the latest version of the runtime.

allevato:
Fixes #160. (Most of it, anyway, except for deleting the other repo.)


Documentation/CONFORMANCE_TESTS.md
```diff
@@ -0,0 +1,61 @@
...
+
+Google's protobuf project includes a conformance checking host program that generates a large number of test cases, feeds them to a test program that executes those test cases, then verifies the results.  This project provides the test program for Swift Protobuf that works with Google's conformance checking host to verify that Swift Protobuf is completely compatible with other implementations of Google's protobuf specification.
+
+## Requirements
+
+This test suite requires Swift 3.0, standard command-line tools such as make and awk, and a full source checkout of [Google's protobuf project](https://github.com/google/protobuf).
+
+## Building the Tests
+
+You should start by editing the `Makefile` and adjusting these two lines to match your system:
```
thomasvl: you can invoke make with: `make [target] PROTOC=[path] GOOGLE_PROTOBUFS_CHECKOUT=[path]` so no editing is needed.

Documentation/CONFORMANCE_TESTS.md
```diff
@@ -0,0 +1,61 @@
+# Swift Protobuf Conformance Tester
+
+**Welcome to Swift Protobuf!**
+
+Apple's Swift programming language is a perfect complement to Google's Protocol
+Buffer serialization technology.  They both emphasize high performance
+and programmer safety.
+
+Google's protobuf project includes a conformance checking host program that generates a large number of test cases, feeds them to a test program that executes those test cases, then verifies the results.  This project provides the test program for Swift Protobuf that works with Google's conformance checking host to verify that Swift Protobuf is completely compatible with other implementations of Google's protobuf specification.
```
thomasvl: Looks like this file is a mix of 80c wrapping unwrapped.  wrap it all so diffs are more useful?

Makefile
```diff
@@ -324,6 +337,10 @@ regenerate-test-protos: build
 		${GENERATE_SRCS} --tfiws_out=FileNaming=DropPath:Tests/SwiftProtobufTests $$t; \
 	done
 
+# Rebuild just the protos used by the conformance test runner.
+regenerate-conformance-protos: build
+	${GENERATE_SRCS} --tfiws_out=FileNaming=DropPath,Visibility=Public:Sources/Conformance ${CONFORMANCE_PROTOS}
```
thomasvl: Do they need to be public or are they built directly into the binary?

Makefile
```diff
@@ -346,6 +363,32 @@ update-proto-files:
...
+# Runs the conformance tests.
+test-conformance: $(SWIFT_CONFORMANCE_PLUGIN) $(CONFORMANCE_HOST) failure_list_swift.txt
+	( \
+		ABS_PBDIR=`cd ${GOOGLE_PROTOBUFS_CHECKOUT}; pwd`; \
+		$${ABS_PBDIR}/conformance/conformance-test-runner --failure_list failure_list_swift.txt $(SWIFT_CONFORMANCE_PLUGIN); \
+	)
+
+# The 'conformance-host' program is part of the protobuf project.
+# It generates test cases, feeds them to our plugin, and verifies the results:
+conformance-host: $(CONFORMANCE_HOST)
```
thomasvl: conformance-host should get added to .PHONY

Makefile
```diff
@@ -324,6 +337,10 @@ regenerate-test-protos: build
 		${GENERATE_SRCS} --tfiws_out=FileNaming=DropPath:Tests/SwiftProtobufTests $$t; \
 	done
 
+# Rebuild just the protos used by the conformance test runner.
+regenerate-conformance-protos: build
```
thomasvl: regenerate-conformance-protos shoudl get added to .PHONY

Sources/Conformance/main.swift
```diff
@@ -0,0 +1,127 @@
+// ProtobufRuntime/Soruces/Conformance/main.swift - Conformance test main
```
thomasvl: correct path comment?

Sources/Conformance/main.swift
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    switch request.requestedOutputFormat {
+    case .protobuf:
+        do {
+            response.protobufPayload = try testMessage.serializeProtobuf()
+        } catch let e {
+            response.serializeError = "Failed to serialize: \(e)"
+        }
+    case .json_:
```
thomasvl: what causes the trailing underscore?  Oh, collision with our json property?

Documentation/CONFORMANCE_TESTS.md
tbkka: Good point.  The best practice I learned from working with man files is:
        Each sentence starts a new line, wrap to 80c.
If you don't start a new line, then rewrapping tends to propagate diffs over long blocks.

Sources/Conformance/main.swift
tbkka: Enums have a generated 'json' property that exposes the JSON serialization of the value.

thomasvl: 🙁  Is that just so the message json support can call it?  Seems like we should leave as much of the enum namespace open for the enum instead.

Documentation/CONFORMANCE_TESTS.md
allevato: Done.

Documentation/CONFORMANCE_TESTS.md
allevato: I un-wrapped the first few lines to be consistent with the rest of the file.

Makefile
allevato: Internal works; removed public.

Makefile
allevato: Done.

Makefile
allevato: It's there :)  (Just out of order, fixed)

Sources/Conformance/main.swift
allevato: So close...

Makefile
```diff
@@ -346,6 +364,32 @@ update-proto-files:
...
+		$${ABS_PBDIR}/conformance/conformance-test-runner --failure_list failure_list_swift.txt $(SWIFT_CONFORMANCE_PLUGIN); \
+	)
+
+# The 'conformance-host' program is part of the protobuf project.
+# It generates test cases, feeds them to our plugin, and verifies the results:
+conformance-host: $(CONFORMANCE_HOST)
+
+$(CONFORMANCE_HOST):
+	( \
+		cd ${GOOGLE_PROTOBUFS_CHECKOUT}; \
```
tbkka: Just noticed this:  The macro here should be `GOOGLE_PROTOBUF_CHECKOUT`  (singular) since `protobuf` is the name of the project on github:  `github.com/google/protobuf/`

Issue #146, opened 2016-11-29
See about removing unneeded JSON support from Extensions


thomasvl:
Spun out of comments on https://github.com/apple/swift-protobuf/pull/145 -

The Extensions vars created include information for JSON support.  But the spec for [JSON](https://developers.google.com/protocol-buffers/docs/proto3#json) doesn't seem to include extensions.  So the extra support for JSON on extensions like should get removed.

_Note:_ a message in a proto3 syntax file can have a field defined in a proto2 syntax file, so JSON support is needed for proto2, it just seems like extensions aren't expected to survive.

tbkka:
@thomasvl:  Is this done?

I see no JSON specifics in the generated code (so at a minimum, I think `generator API` is no longer relevant).

I don't see any JSON support for Extensions in the library, either, but maybe I missed something?

thomasvl:
I honestly don't know.  I opened this purely on your comment on #145 -

> I implemented full JSON support for all of proto2 (including extensions, groups, etc) just to see what it would require. Turns out to be pretty straightforward, actually. We should probably rip that back out again to avoid confusion.

But I didn't go dig to see what it actually was.

tbkka:
Okay.  Looks like the last part of this was finished in d1dd91e5, so I'll close it.

PR #168, opened 2016-12-09
Significantly speed up JSON decoding

### LINKED ISSUE #154: JSON decoding is very slow
After updating the perf harness to measure JSON encoding and decoding, I found some troubling results. Run 100 times for a message with 100 `int32` fields:

```
"Encode JSON" took 0.025 sec
"Populate message fields" took 0.000 sec
"Encode binary" took 0.004 sec
"Decode JSON" took 12.800 sec
"Test equality" took 0.000 sec
"Decode binary" took 0.009 sec
----
Total execution time: 12.840 sec
```

A glance at the call trace shows that we're spending most of that time in `ProtobufJSONScanner.next()`, with much of *that* time being charged to `Character.init(_builtinExtendedGraphemeCluster:utf8CodeUnit:isASCII)`.

I spent the past couple hours digging into the scanner and came up with the following results:

* Initializing `Character` values seems to be *expensive*, especially in tight loops. There are probably penalties there from all the details of Unicode correctness that that type provides. By working with the string's `unicodeScalars` view instead, I was able to get rid of a significant amount of overhead, without sacrificing correctness.
* Building up strings character-by-character (or even scalar-by-scalar) is also expensive. Instead of using iterators and appending, I maintain indices that I advance and take slices, which can be converted back to strings as a contiguous unit much more cheaply.

The result: JSON decoding of 100 messages with 100 `int32` fields went down from 12.8 sec to 0.023 sec—a roughly **550x** improvement.

I finessed some of the edge cases that weren't being exercised by the perf harness (I didn't worry about Unicode escape sequences, surrogates, and such) so that I could focus on speeding up the common cases, but putting those edge cases back shouldn't penalize those common cases too much. I also focused solely on the scanner side of things; there may be gains within the parser itself that I haven't explored yet, as well.

I'll keep working on this and get a PR out in the next few days, hopefully.

allevato:
This achieves the ~550x runtime speedup for JSON decoding of common cases. There may be improvements to be made elsewhere in the decoder, but the changes to the scanner are where I believe the biggest gains can be made.

I've removed the custom number normalization code in favor of an approach closer to the C++ implementation, which simply converts numeric literals at scanning time to a `Double`, `Int64`, or `UInt64` based on its format. This is faster and the implementation is simpler than the previous approach. There is a minor loss of previous functionality here, in that we no longer support the full range of 64-bit integers as floating point literals. That's ok, however, for a couple reasons: (1) protobuf JSON only guarantees no precision loss for 64-bit integers if they are quoted, and (2) since Javascript/JSON represents all numbers as doubles and not all 64-bit ints can be represented that way, there is no guarantee that a sufficiently large number would have its least-significant digits preserved anyway. A fully-preserving floating-point literal could only be written by hand—not by actual Javascript code or other protobuf JSON encoding implementations—so the value of supporting this non-interoperable behavior in Swift was minimal.

I've also punted on the correct way to handle `-0` in integer contexts since the C++ implementation is inconsistent in this regard. We can update it later depending on how https://github.com/google/protobuf/issues/2474 is resolved.

Fixes #154.

While I was here, this change also removes proto2 extensions from JSON decoding, addressing part of #146. I haven't done the encoding side yet because that will require code generation changes, and small changes to the visitor API.


Tests/SwiftProtobufTests/Test_JSON.swift
```diff
@@ -273,14 +271,9 @@ class Test_JSON: XCTestCase, PBTestHelpers {
         assertJSONDecodeSucceeds("{\"singleInt64\": \"0\" }") {$0.singleInt64 == 0}
         // Protobuf JSON does accept exponential format
         assertJSONDecodeSucceeds("{\"singleInt64\":1e3}") {$0.singleInt64 == 1000}
-        assertJSONDecodeSucceeds("{\"singleInt64\":9.223372036854775807e18}") {$0.singleInt64 == Int64.max}
-        assertJSONDecodeSucceeds("{\"singleInt64\":9.223372036854775807e+18}") {$0.singleInt64 == Int64.max}
-        assertJSONDecodeSucceeds("{\"singleInt64\":-92.23372036854775808e17}") {$0.singleInt64 == Int64.min}
-        assertJSONDecodeSucceeds("{\"singleInt64\":9223372036854775807000e-3}") {$0.singleInt64 == Int64.max}
-        assertJSONDecodeSucceeds("{\"singleInt64\":9223372036854775807000000000e-9}") {$0.singleInt64 == Int64.max}
```
tbkka: These should still work if the numeric value is quoted, shouldn't they?

tbkka:
> There is a minor loss of previous functionality here, in that we no longer support the full range of 64-bit integers as floating point literals.

Have you looked at using Float80 for parsing these numbers instead?   That is a floating-point type that supports the full range of Int64.

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -14,245 +14,335 @@
...
-        }
+public class JSONScanner {
+
+  /// A stack of tokens that should be returned upon the next call to `next()`
+  /// before the input string is scanned again.
+  private var tokenPushback: [JSONToken]
+
+  /// The scanner uses the UnicodeScalarView of the string, which is
+  /// significantly faster than operating on Characters.
+  private let scalars: String.UnicodeScalarView
```
tbkka: I'm curious:  Is there additional memory overhead for this?   In particular, I'm not sure if `String.unicodeScalars` has to copy the original.

Tests/SwiftProtobufTests/Test_JSON.swift
allevato: Do you mean a quoted FP literal or normalized to integer form and quoted?

As they're written now, the 1st, 3rd, and 4th ones have too many significant digits to fit in a `double`, quoted or not, so they were no longer valid tests. The 2nd one still passes, but it isn't significantly different from another test that uses the same number shifted one decimal place over. Since the implementation has been converted to just use `Double(_: String)`, I didn't think it was worth it for us to essentially be testing Swift's stdlib's double-parsing, and the conformance tests cover this more thoroughly.

allevato:
> Have you looked at using Float80 for parsing these numbers instead? That is a floating-point type that supports the full range of Int64.

It would, but it would be inconsistent with the behavior of the other protobuf implementations and not provide any real benefit. JSON numbers are double-precision floating point; Javascript won't write a floating point number with more precision than that, and a conforming protobuf implementation will write 64-bit integers as quoted integers, not in scientific notation. So since the only way for such a FP literal to get into the parser is for someone to *write it by hand*, and that input would parse differently with other protobuf implementations, there's not really a reason to support it.

> I understand your argument for not supporting Int64 in non-quoted form; did you intend to also limit it in quoted form as well?

To make sure I know what you're asking here, do you mean treating quoted floating point literals with scientific notation differently in such a way that it preserves their precision beyond that supported by `Double`? If so, yes, I limited those intentionally, for the same reasons above.

tbkka:
> ...  Javascript won't write a floating point number with more precision ...

Not particularly relevant, but I'm really tired of having my Java, Swift, and Obj-C data interoperability broken just because Javascript (which I don't use) has gratuitous limitations.

tbkka:
Sorry:  Wrong button.

allevato:
> I see you've removed extension handling. What about groups? The Google specs don't seem to mention proto2 at all for JSON, so I've never been clear on whether groups should be supported or not.

Good question! I wasn't sure, so treating C++ as canonical, I tried it out and it looks like it drops groups entirely when JSON-encoding, and fails if it encounters something during JSON-decoding. It also looks like the conformance tests don't do anything with groups, either.

Just pushed a commit to have JSON decoding fail on groups like C++ does.

tbkka:
>  It also looks like the conformance tests don't do anything with groups, either.

It would be a great service to the protobuf world if someone expanded the conformance tests to cover proto2.  ;-)

Tests/SwiftProtobufTests/Test_JSON_Group.swift
```diff
@@ -1,45 +0,0 @@
...
-
-class Test_JSON_Group: XCTestCase, PBTestHelpers {
-    typealias MessageTestType = ProtobufUnittest_TestAllTypes
-
-    func testOptionalGroup() {
-        assertJSONEncode("{\"optionalgroup\":{\"a\":3}}") {(o: inout MessageTestType) in
-            var g = MessageTestType.OptionalGroup()
-            g.a = 3
-            o.optionalGroup = g
-        }
```
tbkka: We should probably keep a test here to verify that the decoder fails if it sees a group.  Among other things, that will document that we think this is actually the expected behavior:
```
assertJSONDecodeFails("{...}")
```

allevato: Yes, good point. I've added this file back and changed the tests to decode and verify failure. I'll add tests for encoding (verify groups are dropped) later, because I'd like to clean that up at the same time that I clean up extension encoding.

tbkka:
What specific test did you use for measuring the performance of this?

In working through the Protobuf Text, I've come up with some ideas for simplifying the JSON decoding logic, but I want to be careful to measure the performance impacts before I go too far down that path.

allevato:
> What specific test did you use for measuring the performance of this?

I used `perf-runner.sh 100 fixed32`; it's a nice baseline because it only deals with singular integer parsing and the metrics don't get bogged down by other things like array performance or UTF8 decoding of string values (although for JSON and text, there will naturally be some of that). So, it's a good starting point for squeezing out the most general performance improvements, and then we can dive deeper into specific types to see if there are other places where we can do better.

Issue #148, opened 2016-11-30
Replace debugDescription with text format


allevato:
Once we have text format implemented, the custom `debugDescription` implementation (that uses Swift field names instead of proto field names) feels redundant. If it's solely for debugging purposes, text format would suit that just as well, which would let us strip out a number of strings (all `swiftFieldName`s) entirely from our code.

The only other place where `swiftFieldName` is used I believe is for the custom mirror. Since we're already customizing the mirror representation, I don't think there would be any harm in using the text names for the fields there either instead of the Swift names. Users will see the proto names in the debugger instead of the Swift names, but I don't think that would be a major source of confusion. Certainly not worth having to either (1) keep storing the Swift names, or (2) reïmplement in the runtime the transformation done by the generator (which would have to include all the reserved word checking, too).

So, we can do this in two stages:
(1) Get rid of `swiftFieldName` and replace its uses with `protoFieldName`
(2) Once text format is implemented, update the `debugDescription` property to just do a text format serialization and delete the custom `ProtobufDebugDescriptionVisitor`.

tbkka:
In another thread, we discussed simply removing the custom mirror entirely.  I would rather do that than change the mirror to use proto names (I think it would actually be quite confusing to users to have the names they see in their source editor be different from those showing in the debugger).

Text serialization/deserialization is in the works, but still has a ways to go.

thomasvl:
@tbkka should we open an issue to drop mirror then and revisit it later?  Just so it isn't in the initial release locking us into it?

tbkka:
Done.  Although I think we're locked in either way:  If we don't have a custom mirror, then people will be relying on the behavior of the generic mirror.

thomasvl:
Done and all supporting bits gone.

PR #169, opened 2016-12-10
Protobuf text format support

tbkka:
I forked the initial text format support by @BradLarson and have significantly overhauled it:
* Many more tests, most copied directly from Google's C++ unit tests
* Much simplified compared to the original JSON support code
* Corrected support for number formats (octal, hex, decimal integers)
* Corrected support for bytes fields
* Corrected support for UTF-8 strings
* Support for a lot of legacy variations (e.g., semicolons between message fields, "<...>" message syntax, etc
* Support for maps
* Extensions
* Handling of WKTs, including Google's specialized Any format.


tbkka:
I'm proposing this as a replacement for Brad's PR #117

tbkka:
Working through extension and group support is raising a curious distinction.  There are three different "proto names" floating around here, and it's not entirely clear to me which one(s) should be emitted in the generated code:
* The field name that the user wrote in the proto file
* The name that protoc puts into FieldDescriptor.name
* The name used in the Protobuf Text format output

For example, for a group field:  the user will usually write a `CamelCase` name in the proto file, protoc converts this to lowercase `camelcase`, and Text format uses the `CamelCase` name.

For extensions, Text format uses a fully-qualified name in brackets, which is different from the other two.

Of course, for a group-valued extension, all three of these are different.

I'm inclined to use the Text format field name in the generated code, since I can't think of any other consumer of the proto field name.  That would include the brackets for extensions, etc.

thomasvl:
> I'm inclined to use the Text format field name in the generated code, since I can't think of any other consumer of the proto field name. That would include the brackets for extensions, etc.

The qualified name should be buildable if you are tracking the package/etc.  I haven't looked at your CL to see how you are doing things so that might not be easily possible.  The one advantage of building them would be if the file has a lot of extensions it could be less static data.

tbkka:
> The qualified name should be buildable if you are tracking the package/etc.

I'll take another look at this after I finish working through some WKT tests.

tbkka:
This seems pretty good now.  I spent the last couple of days working through some polish issues:
 * Performance.  Switching from `characters` to `unicodeScalars` gave an easy 20x performance gain on decoding.
 * Pretty.  Packed repeated fields now render with the `[...]` array notation.  This distinguishes packed from repeated fields in the output and makes the default output more compact (especially for proto3).
 * Cleanup.  I removed the last traces of `null` and backed out the placeholder options support (since I have no options yet).

There's still room for improvement, but I think this is solid enough to commit.  Feedback appreciated.

Sources/SwiftProtobuf/Errors.swift
```diff
@@ -52,6 +56,8 @@ public enum DecodingError: Error {
     case malformedProtobuf
     /// Field names were not compiled into the binary
     case missingFieldNames
+    /// The field identifier (name or number) was not recognized
+    case unknownField
     /// TODO: More here?
```
thomasvl: Slight aside - does it make sense to have different error enums for each of the different decoders so it's a little more clear what error can happen for a given operation?

Sources/SwiftProtobuf/TextDecoder.swift
```diff
@@ -0,0 +1,211 @@
...
+                    try scanner.skipRequired(token: .comma)
+                }
+                try S.setFromText(scanner: scanner, value: &value)
+            }
+        } else {
+            try S.setFromText(scanner: scanner, value: &value)
+        }
+    }
+
+    mutating func decodePackedField<S: FieldType>(fieldType: S.Type, value: inout [S.BaseType]) throws {
```
thomasvl: Needed?

Sources/SwiftProtobuf/TextDecoder.swift
```diff
@@ -0,0 +1,211 @@
...
+            if token == terminator {
+                return
+            }
+            switch token {
+            case .identifier("key"):
+                _ = try scanner.skipRequired(token: .colon)
+                try KeyType.setFromText(scanner: scanner, value: &keyField)
+            case .identifier("value"):
+                // Awkward:  If the value is message-typed, the colon is optional,
+                // otherwise, it's required.
```
thomasvl: Does this mean we parse even when it isn't correct?

Sources/SwiftProtobuf/TextEncoder.swift
```diff
@@ -0,0 +1,168 @@
...
+
+import Foundation
+
+public class TextEncoder {
+    var text: [String] = []
+    private var tabLevel = 0
+
+    init() {}
+
+    var result: String { return text.joined(separator: "") }
```
thomasvl: Doesn't collecting in an array and then joining mean we use a lot more memory?  Should we just build the string up and skip the join?

Sources/SwiftProtobuf/TextEncoder.swift
```diff
@@ -0,0 +1,168 @@
...
+
+    func append(text newText: String) {
```
thomasvl: Any of these helpers internal/private?

Sources/SwiftProtobuf/TextEncodingVisitor.swift
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    mutating public func withAbstractVisitor(clause: (inout Visitor) throws -> ()) throws {
+        var visitor: Visitor = self
+        try clause(&visitor)
+        encoder.text = (visitor as! TextEncodingVisitor).encoder.text
+    }
+
+    mutating public func visitUnknown(bytes: Data) {
+        // TODO: Does text encoding have any provision for representing proto2 unknown field?
```
thomasvl: Yes, they just get written out as [field number]: [value] -- https://github.com/google/protobuf/blob/master/src/google/protobuf/text_format.cc#L1936

SwiftProtobuf.xcodeproj/project.pbxproj
```diff
@@ -1826,7 +1918,7 @@
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SKIP_INSTALL = YES;
 				SWIFT_TREAT_WARNINGS_AS_ERRORS = YES;
-				SWIFT_VERSION = 3.0.1;
+				SWIFT_VERSION = "";
```
thomasvl: collision?

Tests/SwiftProtobufTests/Test_Text_proto2.swift
```diff
@@ -0,0 +1,37 @@
+// Test/Sources/TestSuite/Test_Test_proto3.swift - Exercise proto3 text format coding
```
thomasvl: Test_Test -> Test_Text

BradLarson:
Great work on this. To be honest, the support I added got me as far as I needed, and I've been using it in projects since October without a problem. It's held up well with the files that I've needed to parse, but those didn't include binary fields and some of the things that were incomplete in my implementation. I hadn't had a reason to enhance it, so other things took my time.

This seems to improve on all that, and has kept the structure in line with the rest of the project, so I've closed out my pull request in favor of this.

Sources/SwiftProtobuf/Errors.swift
tbkka: Good question:  I don't have a particularly strong opinion either way.

I've heard reasonable arguments on both sides.

Sources/SwiftProtobuf/TextDecoder.swift
tbkka: Yes:  this is how the generated code communicates the type that was declared in the proto file.  Since proto files can have `packed` fields (it's even the default in proto3 for many types), all decoders must implement this.

Sources/SwiftProtobuf/TextDecoder.swift
```diff
@@ -0,0 +1,211 @@
...
+                    try scanner.skipRequired(token: .comma)
+                }
+                try M.setFromText(scanner: scanner, value: &value)
+            }
+        } else {
+            try M.setFromText(scanner: scanner, value: &value)
+        }
+    }
+
+    mutating func decodeSingularGroupField<G: Message>(fieldType: G.Type, value: inout G?) throws {
```
tbkka: Again, this is called by the generated code and communicates the declared type of the field to the decoder.  For Text and JSON formats, there's no difference, but Protobuf Binary does distinguish groups and messages on the wire.

Sources/SwiftProtobuf/TextDecoder.swift
```diff
@@ -0,0 +1,211 @@
+// ProtobufRuntime/Sources/Protobuf/ProtobufTextDecoding.swift - Text format decoding
```
tbkka: I'm planning soon to do a semi-automated pass to fix all of these up across the project.

Sources/SwiftProtobuf/TextEncodingVisitor.swift
tbkka: This is optional in C++ so I've not yet implemented it here.

tbkka:
@BradLarson:  I'd greatly appreciate if you could test this against your project.  I've never used Protobuf Text in a real project, so have based this purely on reading the C++ implementation and unit tests and doing a few experiments.

Sources/SwiftProtobuf/TextDecoder.swift
tbkka: I think so, but I'm not entirely sure.  Given how organically Text format has evolved, it's not entirely clear what "correct" means.  (There are good reasons that no-one has tried to document this format.)  Regardless, I think this is pretty safe:  I'm pretty sure this does not introduce any ambiguity (there's no possibility of mis-parsing correct data), and the C++ implementation generally errs on the side of being overly accepting.

(I'm actually more worried that I've misunderstood somewhere and the ':' is really optional for all field types.  Fortunately, that would be a very easy fix.)

Sources/SwiftProtobuf/TextEncoder.swift
tbkka: Good question:  I'll experiment a little bit.  I'm not sure about memory use, but I have some performance profiling tools on hand, so I can quickly check the performance impact.

Sources/SwiftProtobuf/TextScanner.swift
```diff
@@ -0,0 +1,398 @@
...
+
+    private func parseUnsignedInteger() -> String? {
+        return nil
+    }
+
+    private func parseUnsignedNumber() throws -> String? {
+        var s = String()
+        while let c = scalarGenerator.next() {
+            switch c {
+            case "0", "1", "2", "3", "4", "5", "6", "7", "8", "9":
```
tbkka: No reason.  I should probably try to be more consistent.

Sources/SwiftProtobuf/Errors.swift
tbkka: Note:  Were we to separate them, it would probably be good to split this into separate enums:  `ProtobufDecodingErrors`, `JSONDecodingErrors`, etc.  That way, switch exhaustiveness can be leveraged to be sure you've covered every possible error.

tbkka: I've also wondered whether errors such as `unknownField` should be enhanced to `unknownField(String)` to carry the name that was actually seen.

Sources/SwiftProtobuf/TextDecoder.swift
```diff
@@ -0,0 +1,211 @@
...
+        if terminator == nil {
+            return
+        } else {
+            throw DecodingError.truncatedInput
+        }
+    }
+}
+
+struct TextFieldDecoder: FieldDecoder {
+    var rejectConflictingOneof: Bool {return true}
```
tbkka: I believe this is optional in the C++ implementation.  I think the C++ version has way too many options, so I've tried to implement this first with no options at all, trying to find the most generally useful approach.  I figure it's easy to introduce options later as real need arises.

Sources/SwiftProtobuf/TextEncodingVisitor.swift
thomasvl: If we do swap debugDescription to use text format (#148), we likely will want them so the info is complete.

tbkka: Ah.  Good point!

Sources/SwiftProtobuf/TextEncoder.swift
tbkka: Everything here is 'internal' (used only in the library) except for the `TextEncoder` class itself.  I've refactored slightly to eliminate this one redundant method.

Sources/SwiftProtobuf/TextEncodingVisitor.swift
tbkka: Added Issue #176 to capture this.

tbkka:
Added #177 to capture revisiting the generation of field and enum names for Protobuf Text format.

tbkka:
Performance note:  I copied the large example at the end of `Test_Text_proto3.swift` and built a simple C++ test around it.   On both encoding and decoding, the Swift implementation is about 8x slower than C++ right now for text format.  For anyone who would like to try reproducing this, the main part of my C++ test looks like this.  (Note that this was using "Release" build for swift and "-O2" for C++.)
```
std::string ref = "single_int32: 1\n .......";

double dtime(struct timeval tv) {
    return (double)tv.tv_sec + (double)tv.tv_usec / 1000000.0;
}

int main() {
    protobuf_unittest::TestAllTypes tester;
    struct timeval start, end;
    gettimeofday(&start, NULL);
    int count = 1000;
    for(int i = 0; i < count; ++i) {
        google::protobuf::TextFormat::ParseFromString(ref, &tester);
    }
    gettimeofday(&end, NULL);
    printf("\n\nElapsed Decoding: %f\n\n", dtime(end) - dtime(start));

    std::string str;
    google::protobuf::TextFormat::Printer printer;
    printer.SetUseUtf8StringEscaping(true);
    gettimeofday(&start, NULL);
    for(int i = 0; i < count; ++i) {
        printer.PrintToString(tester, &str);
    }
    gettimeofday(&end, NULL);
    printf("\n\nElapsed Encoding: %f\n\n", dtime(end) - dtime(start));
}
```

Sources/SwiftProtobuf/TextDecoder.swift
tbkka: I just finished updating the leading comments in all the files across the project, and did a separate pass over this branch to catch the files added here.

allevato:
> Performance note: I copied the large example at the end of Test_Text_proto3.swift and built a simple C++ test around it. On both encoding and decoding, the Swift implementation is about 8x slower than C++ right now for text format.

I see that you've changed the scanner to use `UnicodeScalar` instead of `Character`, which is good—you should be able to get some additional improvements by scanning with indices instead of building up the resulting tokens scalar-by-scalar. I'd recommend just taking the new `JSONScanner` and modifying it as needed (it should be trivial, since I built in enough primitives that it should be reüsable without too much effort).

That can be saved for a future PR, though, because it's probably not the only place where we'll want to improve perf.

allevato:
It looks like this is pretty far along now that I'm inclined to say merge it and we can open other more-targeted PRs for additional behavior needed or performance improvements. Especially since we want to use this to toss `debugDescription`, so with that and getting rid of the custom mirror, we can finally do away with the Swift names being generated.

PR #170, opened 2016-12-10
Add bundle version

soffes:
iTunes Connect will reject frameworks without a `CFBundleVersion`. It was already set to `CURRENT_PROJECT_VERSION` but that was empty. I put in the value of `1` for now.


thomasvl:
I guess we don't actually need to track the real version.

soffes:
@tbkka Thanks!

@thomasvl I usually use a build phase script to count the number of commits and use that as the version number. I do that for apps since you have to have a higher unique version than the last one you uploaded to iTunes Connect. For frameworks, I usually just use one, since no one ever looks at this. I usually set the short version number to the actual version (like `0.9.27`).

Happy to PR my script if you want though.

Issue #63, opened 2016-10-07
Look at compile times for generated code


thomasvl:
Since @allevato mentioned he'd seen some generated file crash the compiler and other take a while to compile…

Swift Weekly Issue 41 mentioned the [swift-dev thread](https://lists.swift.org/pipermail/swift-dev/Week-of-Mon-20161003/003099.html) on compile times.  It mentions some options (`-debug-time-function-bodies` & `-warn-long-function-bodies`), they might be useful to help see about improving the code generated.

thomasvl:
More info/links:

  * http://irace.me/swift-profiling
  * http://mjtsai.com/blog/2016/10/06/reducing-swift-compile-time/
  * http://mjtsai.com/blog/2016/04/29/speeding-up-slow-swift-build-times/

Some comments about code changes that helped compile times in there and/or things linked off them.

tbkka:
Ah, yes.  The `-debug-time-function-bodies` option and I are old friends.  ;-)  There's definitely still room for improvement, though.

thomasvl:
@allevato I think you sorta did this recently.  I believe the answer was there wasn't anything really worth changing, but it might make sense to capture some of what you did find here.

I'm also not sure it is worth adding any of those compiler flags into the local builds, since there didn't see to be much we could do for things.

allevato:
The biggest takeaways I found were that type checking was slowest for things involving global operators with a lot of overloads, and the ones we use most frequently in generated code are `~=` (indirectly, for matching `case`s) and `!=` (mostly in `isEqualTo`, but also to a lesser extent in the `has*` properties). So, large enums showed type checking slowness in their initializers (because we have a `case` for every raw value) and messages with large numbers of fields suffered in a handful of places.

There are horrible ways that I would *not* recommend to speed this up (basically create our own `notEqualTo` function to avoid the large global overload space for `!=`), but the improvements we're talking about would be on the order of milliseconds and it's not worth making our generated code horrible to get such a small gain.

[...]

PR #174, opened 2016-12-13
Issue97 equality

thomasvl:
This "fixes" #97, but I'm not tagging it as such because I'm going to try to do a follow up more complete tests around equality and hash before calling the issue fixed.


Sources/PluginLibrary/plugin.pb.swift
```diff
@@ -126,7 +126,7 @@ public struct Google_Protobuf_Compiler_CodeGeneratorRequest: SwiftProtobuf.Messa
 
   public func _protoc_generated_isEqualTo(other: Google_Protobuf_Compiler_CodeGeneratorRequest) -> Bool {
     if fileToGenerate != other.fileToGenerate {return false}
-    if (parameter != other.parameter) {return false}
+    if ((_parameter != nil || other._parameter != nil) && (_parameter == nil || other._parameter == nil || _parameter! != other._parameter!)) {return false}
```
allevato: I might be missing something... but do we need all the nil checks? Wouldn't `_parameter != other._parameter` be sufficient?

* If both are nil/unset, they're equal
* If both are set, it'll compare the values
* If one is nil and the other isn't, it doesn't matter what the value is, they're not equal because we ignore default values

thomasvl: Yea, this is just what some of the existing code was doing, so I was just following that hoping there was reason for it, I'm fine changing it to something better.

tbkka: Good point.  I keep forgetting that there's a pre-defined == for `Optional<T:Equatable>` in the standard library.

Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -302,14 +302,12 @@ struct MessageFieldGenerator {
...
-        } else if let def = swiftProto2DefaultValue {
+        } else {
             var name = name
             if !usesHeapStorage {
                 name = "_" + name
             }
-            return "(((\(name) != nil && \(name)! != \(def)) || (other.\(name) != nil && other.\(name)! != \(def))) && (\(name) == nil || other.\(name) == nil || \(name)! != other.\(name)!))"
-        } else {
-            return "(\(name) != other.\(name))"
+            return "((\(name) != nil || other.\(name) != nil) && (\(name) == nil || other.\(name) == nil || \(name)! != other.\(name)!))"
```
tbkka: Tony's right:  The inequality test here should simplify to just
```
return "(\(name) != other.\(name))"
```
Ditto above for the optional group/message case.

thomasvl: Actually, shouldn't message/group and repeated fields all use the underscore private also to avoid looking at the default value?

thomasvl:
New set of commits up.

thomasvl:
Seeing the new code, I also wonder about generating:

```swift
if _a != other._a ||
  _b != other._b ||
  _c != other._c {
  return false;
}
return true
```

or just returning the result directly:

```swift
return (_a != other._a ||
  _b != other._b ||
  _c != other._c)
```

thomasvl:
@allevato were you looking for some numbers with this cl?  I don't think we're currently set up to track compile times.  I believe we have some issues open to look at flags for doing that, but don't currently have it.

allevato:
> Seeing the new code, I also wonder about generating:

I'd be very worried about the compiler's ability to handle expressions as long as we would generate for some messages. I don't know if conditionals would work differently than arithmetic, but I've done expressions with something like 8-10 terms and it's barfed saying "the expression was too complex to evaluate".

> were you looking for some numbers with this cl?

No, just an off-hand comment about curiosity. It's easy enough to sync back to before this change and time it myself.

tbkka:
Actually, you should be able to just emit:
```swift
return (
  _a == other._a &&
  _b == other._b &&
  _c == other._c
)
```

It would be interesting to see how that fares.

thomasvl:
In the interest of pushing forward, how about sticking with the current code (close to the old) and when someone works on #63 this might be something that gets revisited?

allevato:
Sounds good to me. I'd like to analyze the effect of having expressions with potentially tens to a hundred terms on compile times before going down that path, but it shouldn't block this.

PR #178, opened 2016-12-15
Swift and protobuf language utilities, following @dpassage

tbkka:
This implements some basic language utilities along the lines of @dpassage's work on improving enum case stemming.

I've put them into `PluginLibrary` since I think they are generally useful, not just for protoc-gen-swift.

Includes a _very_ basic test.


tbkka:
@dpassage:  Here's an alternate formulation of the basic Swift and protobuf identifier recognition that should be more portable.

Let me know if you think this is a good approach:  If so, I can merge this and you can use it for your enum fixes.

dpassage:
Yes, this looks good, and thanks for your help! Also appreciate the idea of putting it into `PluginLibrary` rather than adding a unit test target to `protoc-gen-swift` - that hadn't occurred to me.

One thought - your implementation of `isValidSwiftIdentifier` accepts `_`, and it appears that you can't actually use than in a declaration - at the very least, Xcode 8.2 calls the following an error:

```swift
enum Bar {
    case _
}
```

For our purposes, it should probably be invalid.

tbkka:
Good catch!  I've updated the tests and implementation to reject `_`.

dpassage:
Nice! +1 would merge again

Sources/PluginLibrary/ProtoLanguage.swift
```diff
@@ -0,0 +1,42 @@
...
+/// Utility functions for dealing with proto language syntax
+///
+// -----------------------------------------------------------------------------
+
+import Swift
+import Foundation
+
+public func isValidProtobufIdentifier(_ s: String) -> Bool {
+    var i = s.characters.makeIterator()
+    if let first = i.next() {
```
thomasvl: Maybe a comment with a link to where this was taken from?

Sources/PluginLibrary/SwiftLanguage.swift
```diff
@@ -0,0 +1,109 @@
...
+///
+/// Utility functions for dealing with Swift language issues
+///
+// -----------------------------------------------------------------------------
+
+import Swift
+import Foundation
+
+fileprivate func isSwiftIdentifierHeadCharacter(_ c: UnicodeScalar) -> Bool {
+    switch c.value {
```
thomasvl: Same, maybe a comment link in these to say where this info came from?

Sources/PluginLibrary/ProtoLanguage.swift
```diff
@@ -0,0 +1,42 @@
...
+// -----------------------------------------------------------------------------
+///
+/// Utility functions for dealing with proto language syntax
+///
+// -----------------------------------------------------------------------------
+
+import Swift
+import Foundation
+
+public func isValidProtobufIdentifier(_ s: String) -> Bool {
```
thomasvl: Is this used?

btw-I'm mixed on some of these validations.  For them to get to the plugin, protoc has to have parsed and approved them already.  So a lot of our validation doesn't really have to handle everything Swift can do as much as just handle what is valid proto but not valid Swift.

tbkka: I share your misgivings about the valid protobuf test.  I may drop that.

For the Swift identifier test, I went with something more ambitious here primarily to make sure we're very clear about what this support code does.  I expect that test to get used a lot so I don't want there to be any confusion about it.

Sources/PluginLibrary/ProtoLanguage.swift
tbkka: Done.

Sources/PluginLibrary/SwiftLanguage.swift
tbkka: Done.

PR #182, opened 2016-12-19
Workaround to build on Xcode 8.2

kitasuke:
Running build fails on Xcode 8.2. I got the error message below, even though the Build Settings already set `NO` to Use Legacy Swift Language Version.

```
“Use Legacy Swift Language Version” (SWIFT_VERSION) is required to be configured correctly for targets which use Swift. Use the [Edit > Convert > To Current Swift Syntax…] menu to choose a Swift version or use the Build Settings editor to configure the build setting directly.
```

Here might be temporally workaround for this issue. I don't know why this works, but setting 3.0 to SWIFT_VERSION in project file solved this issue. Do you have any idea or better solution for this?


kitasuke:
Seems like it's fixed on Xcode 8.2.1.

PR #183, opened 2016-12-19
Better better case enum stemming

### LINKED ISSUE #138: Enum case stemming can result in invalid Swift

dpassage:
This fixes #138 with an assist from @tbkka on the cross-platform validation functions.

Thanks again for all your work on this!

### DIFF
--- a/Protos/unittest_swift_enum.proto
+++ b/Protos/unittest_swift_enum.proto
@@ -41,4 +41,8 @@ message SwiftEnumTest {
         ENUM_TEST_2_FIRST_VALUE = 1;
         SECOND_VALUE = 2;
     }
+    enum EnumTestNoStem {
+        ENUM_TEST_NO_STEM_1 = 1;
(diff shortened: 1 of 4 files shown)


tbkka:
Looks good!

dpassage:
Hooray, thanks!

PR #185, opened 2016-12-20
Add text format support, C++, and visualizations to perf harness.

allevato:
The perf harness now encodes and decodes text format as part of its benchmarks. It also creates a C++ harness and runs that alongside the Swift one so we can compare stats. (Objective-C coming later.)

I also wanted to be able to visualize the results better, so I created a little HTML dashboard. As the harness runs, it dumps formatted data to a partial Javascript file, which I hand off to plotly.js to create some nice box charts. I also show some median timings for each test and comparative stats, so we can see how we line up. The background color ranges are somewhat arbitrary: green is 3x or less, yellow is 10x or less, red is higher. We can tweak it if we wish later.

The results are concatenated to the same results.js file in the _results directory, so multiple runs can be compared easily in the same document—just refresh.

Also, added an "all" field type to perf_runner.sh that lets me run the harness multiple times for all the types supported, instead of doing it one by one.

Click screenshot to see example:

example


allevato:
Since this is purely performance harness related, I'm going to assume it's uncontroversial and merge it.

tbkka:
This looks great!

PR #188, opened 2016-12-21
Visitor API cleanups

### LINKED ISSUE #171: Remove Mirror.
Following the conversation in #148:  We should remove the custom Mirror implementation.

There are a couple of reasons for this:
* Mirror is the last remaining consumer of the Swift field names; removing it will provide minor code savings.
* The goal of Mirror is to hide the distinction between messages generated with in-line fields and messages generated with copy-on-write backing storage.  But we also want to refactor the boilerplate into library protocol extensions; such extensions would necessarily expose that distinction.

It has also been suggested that we keep the custom Mirror but change it to use proto names; I think this would be very confusing to users, since Mirror is used by debugging tools and this would lead to a situation where debugging tools were printing the wrong field names.

### LINKED ISSUE #181: Visitor and related API improvements
* Should the `Visitor` protocol be `class`-constrained? This would let us get rid of the awkward `withAbstractVisitor` dance that's required because they would no longer have to be passed as `inout`, and polymorphism on the protocol type will just work cleanly when visitors recurse on nested messages.

* A small enhancement, but the concrete visitors currently do everything as a side effect of initialization: for example, calling `ProtobufEncodingVisitor(message: foo, pointer: bar)` causes `foo` to be serialized into `bar`, and the actual visitor afterwards is unneeded (hence the need for `_ = ProtobufEncodingVisitor(...)` at the call site). We should separate these concerns by having the call sites initialize visitors and pass them to `traverse` as separate operations. (This also becomes much easier if we get rid of `withAbstractVisitor` above.)

* We have metatypes like `ProtobufInt32`, `ProtobufFloat`, etc., which conform to `FieldType` (among others) and contain static vars/funcs to provide visitors and decoders with various information (wire format, corresponding Swift type) and encoding/decoding logic. For messages, however, the message type itself conforms to `FieldType`. This adds a large number of static methods that are implementation details to the public API of messages. What if we create a `ProtobufMessageType<M: Message>` metatype instead for these? Separating out the low-level encoding/decoding will clear up some noise from the public API of actual messages. (The same might be true for enums, as well.)

allevato:
Fixes #171.
Partially fixes #181 (the first two bullets).

* Make `Visitor` protocol class-constrained.
* Simplify `protoFieldNumber` argument label to just `fieldNumber`. (We should do the same across the code base.)
* Remove side effects from visitor initializers, except `JSONEncodingVisitor` (needs more work to cleanly handle special cases).
* Reformat some files.
* Delete `MirrorVisitor` and remove `CustomReflectable` conformance from `Message`.
* Remove some explicit tab width settings that have collected in the PBX file.


allevato:
@tbkka Anything else you want to see on this?

tbkka:
Sorry for the long delay; I was on vacation over the holidays.

Summary:  This looks great.  Let's merge it.

Random observations:
 * THANK YOU for documenting the Visitor protocol.   That was greatly needed.
 * Removing the initializer side effects is  nice code cleanup.

It recently occurred to me that there's an interesting trade-off between using a reference type for things like this vs. using an inout struct type:  The former reduces casting woes, but the latter makes the lifetime more explicit.  I think in this case, it turns out to be a win, but I'm starting to prefer inout structs in general.  Something to think about...

PR #190, opened 2016-12-22
Tbkka JSON decoder simplification

tbkka:
This combines the decoder simplification work I recently did for Text decoding with @allevato's work on JSON performance.

It drops 250 lines of code, eliminates quite a few structs and should generally make the code easier to follow.


tbkka:
Specifically, this builds on @allevato's work in #168.

tbkka:
Compared to master, this is:
  * About 60k smaller dylib (after stripping)
  * About 1k smaller harness (after stripping)
  * Almost identical performance (within a few percent)

Sources/SwiftProtobuf/JSONDecoder.swift
```diff
@@ -10,148 +10,517 @@
...
-/// pushback and convenience functions for iterating over complex
-/// structures.
+/// Note: Only `init(json:)` and `nextToken()` are public (they're used
+/// by the test harness and @testable breaks release builds at this writing).
+/// The rest is `private` where possible.
 ///
-public struct JSONDecoder {
-    private var scanner: JSONScanner
-    public var complete: Bool {return scanner.complete}
+public class JSONDecoder {
```
allevato: From a general parsing-theory point of view, I actually prefer having the tokenizer/scanner be a separate type from the parser/decoder. It just looks now like this file is as long as the Decoder/Scanner together were previously, and I'd be surprised if there were any performance advantages by merging them—were there?

Sources/SwiftProtobuf/ProtoNameProviding.swift
```diff
@@ -36,6 +36,14 @@ public protocol ProtoNameProviding {
...
   /// extension sets for names as well.
   func _protobuf_fieldNumber(forProtoName name: String) -> Int?
+
+  /// Returns the field number for the field with the given json name.
+  ///
+  /// The default implementation looks up the field in the static name map,
+  /// which is sufficient for proto3. For proto2 extensions, making this an
+  /// instance method allows generated messages to override it and ask their
+  /// extension sets for names as well.
+    func _protobuf_fieldNumber(forJSONName name: String) -> Int?
```
allevato: Nit: Indentation is off here.

Sources/SwiftProtobuf/ProtoNameProviding.swift
```diff
@@ -36,6 +36,14 @@ public protocol ProtoNameProviding {
   /// instance method allows generated messages to override it and ask their
   /// extension sets for names as well.
   func _protobuf_fieldNumber(forProtoName name: String) -> Int?
+
+  /// Returns the field number for the field with the given json name.
+  ///
+  /// The default implementation looks up the field in the static name map,
+  /// which is sufficient for proto3. For proto2 extensions, making this an
```
allevato: The comment about proto2 extensions doesn't apply for JSON.

tbkka: Good point.

This implies that there should not be any reason to override this lookup.

tbkka: Fixed comment.

Sources/SwiftProtobuf/ProtoNameProviding.swift
tbkka: Fixed indentation.

Sources/SwiftProtobuf/JSONDecoder.swift
tbkka: Combining the types eliminates some of the creation of FieldDecoder objects along the way.  That didn't provide any performance gains (the FieldDecoder structs were mostly just wrappers around a scanner reference), but did clean up some of the code a bit.

The file is a little bit shorter due to the removal of the extra `FieldDecoder` types.

Sources/SwiftProtobuf/ProtoNameProviding.swift
allevato: > This implies that there should not be any reason to override this lookup.

Indeed! That means you can actually remove this method entirely and just have its callers access the static `_protobuf_fieldNames` mapping directly. That would prevent the method from being overridden (say, by an extension). It's unlikely someone would do that, but it would be safer to forbid it.

PR #192, opened 2016-12-22
Perf tweaks

tbkka:
@allevato:  What do you think of these two changes?

### DIFF
--- a/Performance/perf_runner.sh
+++ b/Performance/perf_runner.sh
@@ -106,10 +106,14 @@ function run_harness_and_concatenate_results() {
     "$language": {
 EOF
 
+  echo "Running $language test harness alone..."
+  sleep 3
+  DYLD_LIBRARY_PATH="$script_dir/_generated" "$harness" "$partial_results"
+  sleep 3
+
   echo "Running $language test harness in Instruments..."
   instruments -t "$script_dir/Protobuf" -D "$results_trace" \
-      "$harness" -e DYLD_LIBRARY_PATH "$script_dir/_generated" \
-      "$partial_results"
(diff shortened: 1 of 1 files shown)


Performance/perf_runner.sh
```diff
@@ -34,7 +34,7 @@ readonly script_dir="$(dirname $0)"
 
 # Change this if your checkout of github.com/google/protobuf is in a different
 # location.
-readonly GOOGLE_PROTOBUF_CHECKOUT="$script_dir/../../protobuf"
+readonly GOOGLE_PROTOBUF_CHECKOUT=${GOOGLE_PROTOBUF_CHECKOUT:-"/Users/tkientzle/projects/protobuf"}
```
allevato: Well, this one only applies to your system ;)

Performance/perf_runner.sh
```diff
@@ -106,10 +106,14 @@ function run_harness_and_concatenate_results() {
     "$language": {
 EOF
 
+  echo "Running $language test harness alone..."
+  sleep 3
+  DYLD_LIBRARY_PATH="$script_dir/_generated" "$harness" "$partial_results"
```
allevato: +1, this is probably the right thing to do. It likely won't affect the comparative nature of the timings, but it'll give us truer absolute numbers.

Performance/perf_runner.sh
tbkka: Oops.  Fixed the default here.

tbkka:
I didn't really look at the Instruments overhead.  What I noticed rather was that the variability was really high.  Before this, for example, I might see most samples were 0.1s but one would be ~15s.  After this, I still see some outliers but nothing of quite that magnitude any more.

PR #195, opened 2017-01-05
Performance: Use UnicodeScalarView for text decoding

tbkka:
This applies the same techniques to the text decoder
that @allevato developed in #168 for the JSON decoder.

This seems to improve text decoder performance by an additional
40-60% compared to the previous code.  We're now about 4x-5x
slower than C++ instead of 7x-8x slower.

### DIFF
--- a/Sources/SwiftProtobuf/TextDecoder.swift
+++ b/Sources/SwiftProtobuf/TextDecoder.swift
@@ -26,7 +26,7 @@ public struct TextDecoder {
     public var complete: Bool {return scanner.complete}
 
     internal init(text: String, extensions: ExtensionSet? = nil) {
-        scanner = TextScanner(text: text, tokens: [], extensions: extensions)
+        scanner = TextScanner(text: text, extensions: extensions)
     }
 
     internal init(scanner: TextScanner) {
--- a/Sources/SwiftProtobuf/TextScanner.swift
+++ b/Sources/SwiftProtobuf/TextScanner.swift
@@ -15,94 +15,22 @@
 import Foundation
 import Swift
 
-private func parseIdentifier(prefix: String, scalarGenerator: inout String.UnicodeScalarView.Generator) -> String? {
-    var result = prefix
-    var previousScalarGenerator = scalarGenerator
-    while let c = scalarGenerator.next() {
-        switch c {
-        case "a"..."z", "A"..."Z", "0"..."9", "_":
-            result.append(String(c))
-            previousScalarGenerator = scalarGenerator
-        default:
-            scalarGenerator = previousScalarGenerator
-            return result
-        }
-    }
-    return result
-}
-
-/// Parse the rest of an [extension_field_name] in the input, assuming the
-/// initial "[" character has already been read (and is in the prefix)
-/// This is also used in Any for the typeURL, so we include "/", "."
-private func parseExtensionIdentifier(prefix: String, scalarGenerator: inout String.UnicodeScalarView.Generator) -> String? {
-    var result = prefix
-    if let c = scalarGenerator.next() {
-        switch c {
-        case "a"..."z", "A"..."Z":
-            result.append(String(c))
-        default:
-            return nil
-        }
-    } else {
-        return nil
-    }
-    while let c = scalarGenerator.next() {
-        switch c {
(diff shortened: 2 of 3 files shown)


tbkka:
Observation:  An interesting experiment would be to try reworking the text decoder to use the UTF-8 form instead of the UnicodeScalars form.  It's not a trivial translation and I have no idea what the perf impact would be (probably minor, but hard to tell for sure without trying).  Fortunately, all the special characters are ASCII, so you can avoid most of the usual headache of tracking code point boundaries in UTF-8.

allevato:
I considered whether UTF-8 would be better when I did the JSON decoder work, but I think it would end up performing strictly worse. The work I did on that (and the discovery that `Character` performs surprisingly poorly) inspired me to do some deeper analysis and blog about it (shameless self promotion): https://medium.com/@tonyallevato/strings-characters-and-performance-in-swift-a-deep-dive-b7b5bde58d53

To summarize, the `_StringCore` class that provides the backing storage for `String` can store text in either 7-bit ASCII (treating the buffer as bytes) or UTF-16 code units (treating the buffer as 16-bit ints). `UnicodeScalarView` is very fast, because:

* If the backing store contains ASCII, they can be cheaply widened to their corresponding 32-bit `UnicodeScalar` values. If the backing store contains UTF-16, converting to `UnicodeScalar` also involves simple integer widening in the vast majority of cases (non-surrogate code units are 95% of the code space); for surrogates, it only has to look one code unit ahead and do simple bit-math to convert it.
* As such, `UnicodeScalarView.Index` is very simple, and the traversal operations like `index(after:)` are fast—as is taking slices.

For `UTF8View`, there are some optimizations that let it take faster code paths when the underlying buffer is known to be ASCII, but in order to keep the implementation mostly unified, it does some memory shuffling to make it work. Essentially it decodes at most 8 bytes worth of UTF-8 from the underlying buffer at a time (using a `UInt64` as a fast "chunk" to avoid repeated allocations): if the buffer is ASCII, it does a simple `memcpy` into the chunk, but if it contains UTF-16, it has to transcode it. So even in the ASCII case, the logic is much more complex.

Then, another problem is that we don't know what the underlying `_StringCore`'s representation will be—any string we hand it *could* be ASCII or *could* be UTF-16, and if it's the latter, `UTF8View` will perform much more poorly compared to `UnicodeScalarView`.

And there's another catch: in addition to ASCII and UTF-16 data, `_StringCore` also supports using a Cocoa buffer for the `NSString` bridging case. AFAIK, those buffers are always UTF-16 because `NSString` stores its data with the Objective-C `unichar` type. So if someone passes our decoder a string that was obtained from a Foundation API on an Apple platform, we *definitely* lose the fast path for UTF-8.

thomasvl:
Re: ObjC NSStrings - there was a change a few years ago to let them be UTF8 also, but I think that was mainly for constant strings.

In the JSON case, I'd expect most folks to actually send UTF8 over the wire, so there likely wouldn't be a transform needed unless it happens when making a String out of it.

Which actually brings up and interesting question, does it make sense to have a JSON parsing api that takes a Data?  That way something like gRPC could take the raw bytes off the wire, make a Data, and just feed that for parsing instead of taking the hit to convert it to a String first?  In the past I've relied on NSString's initWithBytesNoCopy:length:encoding:freeWhenDone: to do stuff like this without extra copies, but maybe we can wrap it up nicely?

allevato:
Having a `Data` parsing path is an interesting idea. That could end up with performance savings in places we aren't currently measuring—outside the parsing code itself—because if file/stream APIs hand users `Data` to begin with, they wouldn't have to do potentially costly `Data` <-> `String` conversions.

If we did that, we wouldn't want to have two completely different code paths. IMO, `UTF8View` is still going to yield strictly poorer performance than `UnicodeScalarView` just by looking at the implementations of both types. So then the question becomes, can we create an efficient `StringOrData` abstraction (with a similar abstraction for its indices) that we could wrap either one in, hand that off to the decoder, and have the decoder do the right thing in both situations, and would the performance improvements from working with just `Data` be worth it? Maybe—it would be an interesting problem to solve.

thomasvl:
The downside of having `Data` interfaces is we can no longer assume valid UTF8, so validation becomes our problem.

PR #196, opened 2017-01-05
Performance harness fixes...

tbkka:
A few fixes/improvements for the perf harness:
 * Make it work when you invoke it as `Performance/perf_runner.sh`  In particular, the C++ protoc invocation is very sensitive to the current directory.
 * Write separate trace files for each language
 * Open the Swift trace file at the end of the run (before, it opened the last one, which was always C++)
 * In the summary table, decorate the fastest result in green, show multipliers for the slower language (even when C++ is slower ;-)

### DIFF
--- a/Performance/js/harness-visualization.js
+++ b/Performance/js/harness-visualization.js
@@ -102,25 +102,16 @@
 
   // Populates a multiplier cell with the ratio between the given two values
   // and sets its background color depending on the magnitude.
-  function populateMultiplierCell(cell, language, swiftValue, otherValue) {
-    if (language != 'Swift') {
-      var multiplier = swiftValue / otherValue;
-
-      if (multiplier < 1) {
-        cell.text('(<1x)');
-      } else {
-        cell.text('(' + multiplier.toFixed(0) + 'x)');
-      }
-
-      if (multiplier < 3) {
-        cssClass = 'bg-success';
+    function decorateMultiplierCell(cell, multiplier) {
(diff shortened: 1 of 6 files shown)


tbkka:
To clarify, here's the output I'm now seeing for one of my test runs.  Note that Swift is actually faster than C++ here for the JSON decode and equality tests:
screen shot 2017-01-05 at 3 42 35 pm

Performance/js/harness-visualization.js
```diff
@@ -102,25 +102,16 @@
...
-        cell.text('(' + multiplier.toFixed(0) + 'x)');
-      }
-
-      if (multiplier < 3) {
-        cssClass = 'bg-success';
+    function decorateMultiplierCell(cell, multiplier) {
+      if (multiplier == 1) {
+          cell.addClass('bg-success');
+      } else if (multiplier < 3) {
+//        cssClass = 'bg-success';
```
thomasvl: Drop this case?

tbkka: I've adjusted the comments to clarify that this case is deliberately doing nothing.  (This is responsible for the white cells above, which have multipliers >1 but <3.

I've also updated the comments and fixed the indentation for this function.

tbkka:
I'd be very interested to see other languages added to this, partly just out of curiosity, but also to put the Swift numbers into better perspective.  🤔

tbkka:
I don't think there's anything controversial here, so I'll merge it.

PR #199, opened 2017-01-07
Binary decode performance improvements

tbkka:
This improves the performance of the binary decoder by more than double in most cases I've tested.  I'm now routinely seeing 5x-6x compared to C++ (where I was routinely seeing 8x-15x before).

It refactors down to a single field decoder type, reduces the FieldType hooks for protobuf, eliminates a lot of unnecessary object creations, improves decoding of strings and fixed-size numeric values.

From my profiling, the low-level parsing of fields is actually pretty competitive with C++.  The higher-level parsing structure still has a lot of extra object and data shuffles that I'll continue to work on.

### DIFF
--- a/Reference/conformance/conformance.pb.swift
+++ b/Reference/conformance/conformance.pb.swift
@@ -265,7 +265,7 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message, SwiftProtobuf.Prot
       self = .None
     }
 
-    public mutating func decodeField(setter: inout SwiftProtobuf.FieldDecoder, protoFieldNumber: Int) throws {
+    public mutating func decodeField<T: SwiftProtobuf.FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws {
       if self != .None && setter.rejectConflictingOneof {
         throw SwiftProtobuf.DecodingError.duplicatedOneOf
       }
@@ -330,7 +330,7 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message, SwiftProtobuf.Prot
 
   init() {}
 
-  public mutating func _protoc_generated_decodeField(setter: inout SwiftProtobuf.FieldDecoder, protoFieldNumber: Int) throws {
+  public mutating func _protoc_generated_decodeField<T: SwiftProtobuf.FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws {
     switch protoFieldNumber {
     case 1, 2: try payload.decodeField(setter: &setter, protoFieldNumber: protoFieldNumber)
     case 3: try setter.decodeSingularField(fieldType: Conformance_WireFormat.self, value: &requestedOutputFormat)
@@ -397,7 +397,7 @@ struct Conformance_ConformanceResponse: SwiftProtobuf.Message, SwiftProtobuf.Pro
       self = .None
     }
 
-    public mutating func decodeField(setter: inout SwiftProtobuf.FieldDecoder, protoFieldNumber: Int) throws {
+    public mutating func decodeField<T: SwiftProtobuf.FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws {
       if self != .None && setter.rejectConflictingOneof {
         throw SwiftProtobuf.DecodingError.duplicatedOneOf
       }
@@ -556,7 +556,7 @@ struct Conformance_ConformanceResponse: SwiftProtobuf.Message, SwiftProtobuf.Pro
 
   init() {}
 
-  public mutating func _protoc_generated_decodeField(setter: inout SwiftProtobuf.FieldDecoder, protoFieldNumber: Int) throws {
+  public mutating func _protoc_generated_decodeField<T: SwiftProtobuf.FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws {
     switch protoFieldNumber {
     case 1, 6, 2, 3, 4, 5: try result.decodeField(setter: &setter, protoFieldNumber: protoFieldNumber)
     default: break
@@ -817,7 +817,7 @@ struct Conformance_TestAllTypes: SwiftProtobuf.Message, SwiftProtobuf.Proto3Mess
 
     init() {}
 
-    func decodeField(setter: inout SwiftProtobuf.FieldDecoder, protoFieldNumber: Int) throws {
+    func decodeField<T: SwiftProtobuf.FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws {
       switch protoFieldNumber {
       case 1: try setter.decodeSingularField(fieldType: SwiftProtobuf.ProtobufInt32.self, value: &_optionalInt32)
       case 2: try setter.decodeSingularField(fieldType: SwiftProtobuf.ProtobufInt64.self, value: &_optionalInt64)
@@ -1539,7 +1539,7 @@ struct Conformance_TestAllTypes: SwiftProtobuf.Message, SwiftProtobuf.Proto3Mess
       self = .None
     }
 
-    public mutating func decodeField(setter: inout SwiftProtobuf.FieldDecoder, protoFieldNumber: Int) throws {
+    public mutating func decodeField<T: SwiftProtobuf.FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws {
       if self != .None && setter.rejectConflictingOneof {
         throw SwiftProtobuf.DecodingError.duplicatedOneOf
       }
@@ -1741,7 +1741,7 @@ struct Conformance_TestAllTypes: SwiftProtobuf.Message, SwiftProtobuf.Proto3Mess
 
       init() {}
 
-      func decodeField(setter: inout SwiftProtobuf.FieldDecoder, protoFieldNumber: Int) throws {
+      func decodeField<T: SwiftProtobuf.FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws {
         switch protoFieldNumber {
         case 1: try setter.decodeSingularField(fieldType: SwiftProtobuf.ProtobufInt32.self, value: &_a)
         case 2: try setter.decodeSingularMessageField(fieldType: Conformance_TestAllTypes.self, value: &_corecursive)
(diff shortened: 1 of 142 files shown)


tbkka:
More progress:  Changing the generated decodeField() to be generic in the type of the FieldDecoder yields another big boost.  Without this, the decoders have to construct a lot of `FieldDecoder` protocol witnesses; the generic form avoids the additional object creation.  This does modestly increase the size of the generated code (about 1.7k for 100 fields in my tests), but I think the speedup (around 20%) is worth it.

tbkka:
To summarize the perf improvements for the binary decoder over this branch, I measured a bunch of cases on my laptop:
  "100 fixed64" went from 5ms (master) to 1.1ms (this branch)
  "100 int64" went from 2.4ms (master) to 1.2ms (this branch)
  "100 string" went from 13ms (master) to 12ms (this branch)
  "100 repeated int64" went from 18ms (master) to 9.5ms (this branch)

tbkka:
I incorporated some suggestions which gave us another 2x speedup for singular fields and an additional big boost for strings:

"100 fixed64" is now 0.55ms (2.6x of C++)
"100 int64" is now  0.7ms (1.6x of C++)
"100 string" is now 3ms (1.6x of C++)
"100 repeated int64" is about the same (around 2.6x of C++)

tbkka:
Any concerns before I merge this?

Sources/SwiftProtobuf/JSONTypeAdditions.swift
```diff
@@ -665,8 +665,8 @@ public extension Message {
         while true {
             let key = try decoder.nextKey()
             if let protoFieldNumber = fieldNames.fieldNumber(forJSONName: key) {
-                var fieldDecoder: FieldDecoder = decoder
-                try decodeField(setter: &fieldDecoder, protoFieldNumber: protoFieldNumber)
+                var t = decoder
```
thomasvl: t?  any reason for cryptic naming in this case compared to before?

Sources/SwiftProtobuf/ProtobufDecoder.swift
```diff
@@ -21,564 +21,205 @@
...
+        available = count
+        fieldStartP = p
+        self.extensions = extensions
     }
 
-    mutating func decodeSingularField<S: FieldType>(fieldType: S.Type, value: inout S.BaseType?) throws {
-        try S.setFromProtobufFixed8(fixed8: fixed8, value: &value)
-        consumed = true
+    public mutating func reset() {
+        consumed = false
```
thomasvl: Should fieldNumber get reset?

thomasvl: What about unknownData, etc?  Some of those also get lazy created/appended, should reset ensure they are cleared out also?

SwiftProtobuf.xcodeproj/project.pbxproj
```diff
@@ -1439,8 +1447,8 @@
 				F44F938E1DAEA76700BC5B85 /* JSONTypeAdditions.swift in Sources */,
 				AAF2EE011DFB1450007B510F /* JSONIntegerConverting.swift in Sources */,
 				F44F93801DAEA76700BC5B85 /* Google_Protobuf_Wrappers+Extensions.swift in Sources */,
-				F44F938C1DAEA76700BC5B85 /* JSONDecoder.swift in Sources */,
```
thomasvl: Just checking, this wasn't a merge issue was it?

Sources/SwiftProtobuf/ProtobufDecoder.swift
thomasvl: Is it possible to craft a test where reset() comes into play?  To ensure all state is properly clears between uses of the decoder?

SwiftProtobuf.xcodeproj/project.pbxproj
tbkka: Urk.  It certainly was.  Thanks for catching it!

Sources/SwiftProtobuf/JSONTypeAdditions.swift
tbkka: Improved the name.

Sources/SwiftProtobuf/ProtobufDecoder.swift
tbkka: I just removed reset() entirely as it's no longer needed.  (The switch to passing decoders via generics removed the need to access the decoder through a protocol witness.)

thomasvl:
LGTM.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -43,18 +43,20 @@ public protocol Message: CustomDebugStringConvertible {
...
   ///
   /// This is the core method used by the deserialization machinery.
   ///
   /// Note that this is not specific to protobuf encoding; formats that use
   /// textual identifiers translate those to protoFieldNumbers and then invoke
   /// this to decode the field value.
-  mutating func decodeField(setter: inout FieldDecoder,
-                            protoFieldNumber: Int) throws
+ mutating func decodeField<T: FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws
+//  mutating func decodeField(setter: inout FieldDecoder, protoFieldNumber: Int) throws
```
allevato: Delete commented out line?

Sources/SwiftProtobuf/Message.swift
```diff
@@ -197,7 +202,8 @@ public protocol _MessageImplementationBase: Message, Hashable, MapValueType, Fie
   // The compiler actually generates the following methods. Default
   // implementations below redirect the standard names. This allows developers
   // to override the standard names to customize the behavior.
-  mutating func _protoc_generated_decodeField(setter: inout FieldDecoder,
+  mutating func _protoc_generated_decodeField<T: FieldDecoder>(setter: inout T,
+//  mutating func _protoc_generated_decodeField(setter: inout FieldDecoder,
```
allevato: Delete commented out line?

Sources/SwiftProtobuf/Message.swift
```diff
@@ -215,8 +221,8 @@ public extension _MessageImplementationBase {
     try _protoc_generated_traverse(visitor: visitor)
   }
 
-  mutating func decodeField(setter: inout FieldDecoder,
-                            protoFieldNumber: Int) throws {
+  mutating func decodeField<T: FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws {
+//  mutating func decodeField(setter: inout FieldDecoder, protoFieldNumber: Int) throws {
```
allevato: Delete commented out line?

Sources/SwiftProtobuf/FieldTypes.swift
```diff
@@ -47,24 +47,11 @@ public protocol FieldType {
...
-    /// Set the variable from a block of bytes
-    static func setFromProtobufBuffer(buffer: UnsafeBufferPointer<UInt8>, value: inout BaseType?) throws
-    // Special variant used when decoding packed enum fields.
-    // The extra `unknown` argument will be updated with a re-encoded
-    // representation of any unknown enum values.
-    static func setFromProtobufBuffer(buffer: UnsafeBufferPointer<UInt8>, value: inout [BaseType], unknown: inout Data) throws
+    /// Set the variable from a protobuf decoder
+    static func setFromProtobuf(decoder: inout ProtobufDecoder, value: inout BaseType?) throws -> Bool
+    static func setFromProtobuf(decoder: inout ProtobufDecoder, value: inout BaseType) throws -> Bool
+    /// Set the variable from a protobuf decoder
```
allevato: Super-nit: This comment isn't necessary

Reference/conformance/conformance.pb.swift
```diff
@@ -265,7 +265,7 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message, SwiftProtobuf.Prot
       self = .None
     }
 
-    public mutating func decodeField(setter: inout SwiftProtobuf.FieldDecoder, protoFieldNumber: Int) throws {
+    public mutating func decodeField<T: SwiftProtobuf.FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws {
```
allevato: At first I thought you might have added an associated type or `Self` requirement to `FieldDecoder` that would prompt this change, but I don't see that. Is it a performance boost because it causes the compiler to specialize implementations instead of using dynamic dispatch? If so, is the size impact on the runtime and generated code minimal?

Sources/SwiftProtobuf/Message.swift
tbkka: Deleted.

Sources/SwiftProtobuf/Message.swift
tbkka: Deleted.

Sources/SwiftProtobuf/Message.swift
tbkka: Deleted.

Sources/SwiftProtobuf/FieldTypes.swift
tbkka: Deleted.

Reference/conformance/conformance.pb.swift
tbkka: It's a performance boost because it allows you to avoid creating objects elsewhere.

With the old pattern, you had to create a `FieldDecoder` protocol witness to pass into `decodeField`, which led to patterns like this:
```
      // In decoder
      var fieldDecoder: FieldDecoder = self
      message.decodeField(setter: &fieldDecoder, ...)
```
(or worse, creating a new object for the field decoder).  This both copies the decoder object, and also creates an additional protocol witness.  With the generic form, the decoder can just implement the `FieldDecoder` protocol and be passed directly in without copying and without creating a witness.

By my measurement, this adds only a few bytes per field to the generated compiled code.  (17 bytes per field in the one measurement I clearly remember)  I've not attempted to look carefully at the runtime library size, but I don't expect much impact there either.

If you want to dig more carefully into the implications, try checking out 5ce85d4 (just before I made the switch to the generic form) and 	c0f1a33 (just after).

tbkka: P.S.  I suspect a similar change on the encoder side might have similar benefits.

PR #200, opened 2017-01-09
Conformance test corrections

tbkka:
This fixes a number of conformance test issues:

 * Three tests of 'Any' transcoding were failing because the type being tested had not been registered with `Google_Protobuf_Any`.  Fix: add the necessary proto and register it before running the conformance cases

 * Four "known failures" were in tests that no longer exist.  Fix: remove them from the known failures list


Sources/Conformance/main.swift
```diff
@@ -120,6 +120,8 @@ func singleTest() throws -> Bool {
 }
 
 Google_Protobuf_Any.register(messageType: Conformance_TestAllTypes.self)
+Google_Protobuf_Any.register(messageType: ProtobufTestMessages_Proto3_TestAllTypes.self)
```
thomasvl: I haven't looked - What's the threading on this?  i.e. - Since it is a global registry what happen if some some subsystem is registering while another us using an Any?

For [ObjC I avoid the registry](https://github.com/google/protobuf/blob/master/objectivec/GPBWellKnownTypes.h#L216-L230), seems like we might be able to template the function to take a Swift protobuf class instead and confirm the Any is that type before extracting instead of relying on a registry?

tbkka: The registration per se only needs to be done once at program start; after that, the registry itself is read-only.  We could serialize access, but that seems overkill right now.

Note that we are not simply unpacking the Any field in this case.   Rather, we're decoding a protobuf binary message (which may potentially be quite complex with Any fields buried at any point), then re-encoding as JSON.  As a side-effect of transcoding, Any fields buried within the message must be unpacked and repacked.  This in turn requires access to the type.  Unlike Obj-C, Swift has no mechanism for dynamically locating a type by name at run-time.

To avoid the registry, we could either:
 * Pass in a collection of types to the serializer (similar to how we handle extensions during decoding)
 * Require the user to manually inspect every Any field and unpack/rewrite it (which converts an Any field holding a protobuf to an Any field holding a message -- the latter can then be serialized to any supported format)

thomasvl: Startup - The catch is someone could be using code that uses protos internally, so they don't know what it has to have registered.  Forcing those modules to add a _Startup_ method that has to be called also sorta sucks.

I guess I should go look at what you have on Any (I haven't yet).  The other languages all don't do any implicit decoding of an Any.  Instead when the outer message is decoded, the Any stays with it's content as a blob and there are helpers to extract/decode the Any to specific types.  An application doesn't anyways want the Any decoded, and even when it does, the code is only going to know how to deal with a specific number set of them.

tbkka: That assumes you are using descriptors to translate JSON to protobuf, which in turn relies on a registry of descriptors.

tbkka: Also, to be clear:  The registry is *only* needed if you are transcoding messages between different encodings (e.g., protobuf-to-JSON or vice-versa).  It's not needed in any other case.  (In particular, I do not need any registry to decode JSON and recode back to JSON, unlike the C++ implementation which relies on the descriptor registry in this case.)

tbkka:
I'm going to merge this since it fixes obvious problems.  (The question about Any transcode support is a separate issue.)

PR #201, opened 2017-01-09
Fix test suite build on Linux

tbkka:
This corrects two failures when building the test suite on Linux:

1. The script that collects the list of tests was not ignoring commented-out tests.

2. On Linux, using the `Data` type requires you to first import Foundation

### DIFF
--- a/CollectTests.awk
+++ b/CollectTests.awk
@@ -69,7 +69,7 @@ BEGIN {
 }
 
 
-/ *func *test.*/ {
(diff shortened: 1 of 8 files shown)


thomasvl:
Is there any way to make macOS also require the Foundation import so we don't break things?

tbkka:
I plan to file a bug with the Swift team about that; I'll let you know if I find out anything useful.

PR #203, opened 2017-01-09
Fix tests on Linux

tbkka:
Data on Linux seems to have problems with `hashValue` when there is no content.  Work around this for now...


Sources/SwiftProtobuf/HashVisitor.swift
```diff
@@ -37,7 +37,9 @@ final class HashVisitor: Visitor {
   init() {}
 
   func visitUnknown(bytes: Data) {
-    mix(bytes.hashValue)
+    if bytes.count > 0 { // Workaround for Linux Foundation bug
```
thomasvl: Does it make sense to put a linux test around this as a flag for when it could come out?

PR #206, opened 2017-01-12
Update Message's "with" builder.

cburrows:
This changes with() so that it follows the standard library pattern for
higher order functions. In particular, the function type argument is
declared with throws, which with() rethrows, and it is not given an
external parameter name.

This enables easy error handling using with() and closures that throw.


tbkka:
This is very nice.  Thanks a lot!

PR #210, opened 2017-01-19
Text decode perf

tbkka:
This comprises a sizable speedup of the text decoder, mostly by propagating the patterns from the recent binary coding performance work.  In particular, this moves to a fully schema-driven parse (which eliminates the need for TextToken), pushes the proto3 singular distinction down to the TextDecoder, and eliminates a lot of intermediate objects.

For '100 int64', this is only about 50% slower than the C++ implementation (look at TextScanner.nextFieldNumber() for an idea about how we might close the remaining gap in this case).  In particular, the int parser now directly converts the UTF8 code points into an integer, eliminating the intermediate string.

'100 string' is about 2.5x slower than C++ due to the time required to build a String from a part of a String.UTF8View.  There may be ways to improve this.

I've also added some tests to better exercise `nan` and boolean constants and improved the error handling and reporting in a few tests.

Note that I've also adjusted the code generation slightly:  Extension field names are now stored in the generated code without the surrounding `[...]` characters.


tbkka:
This seems pretty serviceable now.  I have some ideas for a future round of perf work, but this is a definite improvement.

I'm in the process of taking these same ideas through the JSON decoder, and it seems to be working equally well there.

Sources/SwiftProtobuf/TextScanner.swift
```diff
@@ -15,55 +15,227 @@
 import Foundation
 import Swift
 
+private func fromHexDigit(_ c: UInt8) -> UInt8? {
+  if c >= 48 && c <= 57 {
```
allevato: I may have glossed over this in the similar JSON PR, unfortunately :(  Swift has `UInt8(ascii:)` that takes a `UnicodeScalar` and produces the equivalent 8-bit int if it's a valid ASCII character. It's also able to optimize the whole thing down to just the single integer load instruction. I just tested it and `UInt8(ascii: "0")` becomes simple `mov al, 0x30` as expected.

Using that initializer instead of the raw ASCII would make this code a lot more readable/maintainable.

Reference/google/protobuf/unittest_optimize_for.pb.swift
```diff
@@ -179,9 +179,9 @@ struct ProtobufUnittest_TestOptimizedForSize: SwiftProtobuf.Message, SwiftProtob
...
+    static let testExtension = SwiftProtobuf.MessageExtension<OptionalExtensionField<SwiftProtobuf.ProtobufInt32>, ProtobufUnittest_TestOptimizedForSize>(protoFieldNumber: 1234, fieldNames: .same(proto: "protobuf_unittest.TestOptimizedForSize.test_extension", swift: "ProtobufUnittest_TestOptimizedForSize_testExtension"), defaultValue: 0)
 
-    static let testExtension2 = SwiftProtobuf.MessageExtension<OptionalMessageExtensionField<ProtobufUnittest_TestRequiredOptimizedForSize>, ProtobufUnittest_TestOptimizedForSize>(protoFieldNumber: 1235, fieldNames: .same(proto: "[protobuf_unittest.TestOptimizedForSize.test_extension2]", swift: "ProtobufUnittest_TestOptimizedForSize_testExtension2"), defaultValue: ProtobufUnittest_TestRequiredOptimizedForSize())
+    static let testExtension2 = SwiftProtobuf.MessageExtension<OptionalMessageExtensionField<ProtobufUnittest_TestRequiredOptimizedForSize>, ProtobufUnittest_TestOptimizedForSize>(protoFieldNumber: 1235, fieldNames: .same(proto: "protobuf_unittest.TestOptimizedForSize.test_extension2", swift: "ProtobufUnittest_TestOptimizedForSize_testExtension2"), defaultValue: ProtobufUnittest_TestRequiredOptimizedForSize())
```
thomasvl: aside: diffs on these would be nicer if we made them multiple lines.

Sources/SwiftProtobuf/TextScanner.swift
```diff
@@ -15,55 +15,227 @@
...
+        while index != utf8.endIndex {
+            let u = utf8[index]
+            switch u {
+            case 32, 9, 10, 13: // space, tab, NL, CR
+                index = utf8.index(after: index)
                 lastIndex = index
-            case "#":
-                while index != scalars.endIndex {
+            case 35: // #
+                while index != utf8.endIndex {
```
thomasvl: Do you want to advance once before the loop?  Otherwise, you check fetch/check the '#' again, no?

thomasvl: Or use a repeat...while?

Sources/SwiftProtobuf/TextScanner.swift
```diff
@@ -75,206 +247,525 @@ public class TextScanner {
...
-                index = scalars.index(after: index)
+                let s = String(utf8[start..<index])
+                index = utf8.index(after: index)
                 return s
             }
-            index = scalars.index(after: index)
-            if c == "\\" {
-                if index == scalars.endIndex {
+            index = utf8.index(after: index)
+            if c == 92 { //  \
```
thomasvl: nit: maybe // '\' for these sorta comments, to make it clear the character you are talking about?

SwiftProtobuf.xcodeproj/project.pbxproj
```diff
@@ -1447,7 +1439,6 @@
 				F44F938E1DAEA76700BC5B85 /* JSONTypeAdditions.swift in Sources */,
 				AAF2EE011DFB1450007B510F /* JSONIntegerConverting.swift in Sources */,
 				F44F93801DAEA76700BC5B85 /* Google_Protobuf_Wrappers+Extensions.swift in Sources */,
-				F44F938C1DAEA76700BC5B85 /* JSONDecoder.swift in Sources */,
```
thomasvl: Stray delete again?

Sources/SwiftProtobuf/TextScanner.swift
tbkka: Thanks for the nudge:  I've been chafing about how illegible all these numbers are.

I just went through and added symbolic constants for all of them.

Sources/SwiftProtobuf/TextScanner.swift
tbkka: Swapped out the numbers with symbolic constants throughout.

Sources/SwiftProtobuf/TextScanner.swift
tbkka: I just added an extra increment here to avoid the duplicate check.  While here, I also removed the unnecessary "lastIndex" tracking.

Reference/google/protobuf/unittest_optimize_for.pb.swift
tbkka: I went ahead and adjusted the generator to break these across lines.  Hopefully that will improve future diffs.

SwiftProtobuf.xcodeproj/project.pbxproj
tbkka: I think I inadvertently deleted/readded that file at some point, and this is Xcode GCing the stray reference.

Reversing this out properly would be complicated, so I think I'll leave it.

tbkka:
Thanks!  Now I need to reconcile this with my WIP JSON perf.  Another PR coming soon...

PR #211, opened 2017-01-20
Make perf_runner work with Swift 3.1

tbkka:
In particular, the 'product' support for SwiftPM is in flux,
and some people seem confused by the dynamic library
in the package manifest.


allevato:
FYI, I just noticed that these changes introduced some fragility in the perf harness build; I'm preparing a PR to fix it.

PR #217, opened 2017-02-03
Add missing triple-back-ticks to GENERATED_CODE.md

salutis:


tbkka:
Good catch!  Thank you!

Issue #111, opened 2016-10-19
Investigate if we can validate the protoc version being used


thomasvl:
Opening this to capture it and think on it, but we may just be stuck without a real solution.  ☹️ 

PR #95 started using the protoc provided JSON names instead of trying to create them within the generator (plugin).  This was needed to ensure an explicit json_name is honored by Swift.  The catch is protoc sets these strings in all cases.  The logic being it could allow a single implementation to transform the names so all languages are consistent.

PR #105 was done because my clean build showed things as not up to date.  What @allevato and I finally realized is this was a side effect of my version of google/protobuf being newer, so it had commits that improved (fixed?) more of the json support in the C++ library.

So the net result of PR #95 is swift-protobuf depends on developers using a 3.1 (maybe even later) release of protoc to get valid JSON support from the generated Swift classes.

We might be able to suggest the proto interface between protoc and plugins get expanded to relay the version so a plugin can check it, but the question becomes what version is passed, and how does google/protobuf know when to revise it?  As we discovered with this case, any change to the JSON name generation could justify a change to that version interface if it is being exposed to allow plugins to ensure conformance.

tbkka:
protoc 3.0 shipped with known-broken JSON support.   So we know that using the protoc-generated JSON names with protoc 3.0 will not work correctly.  But there's nothing we can do about it.

Yes, protoc should advertise its version when it writes out the plugin data.

thomasvl:
fyi - https://github.com/google/protobuf/pull/2437

tbkka:
Huzzah!  Among other things, I'd like to include the protoc and plugin versions in the template we put at the top of each generated output file.  I suspect that will help when trying to diagnose issues reported by users.  So that boilerplate might look like:
```
/*
 * DO NOT EDIT.
 *
 * Generated by protoc 3.1.2 and protoc-gen-swift 0.9.36
 * Source: unittest_swift_all_required_types.proto
 */
```

thomasvl:
While I think of it…

Having a line like:

> ` * Generated by protoc 3.1.2 and protoc-gen-swift 0.9.36`

Would mean every release of google/protobuf and every release of this project would cause generated sources and reference files to be stale and need to regenerate.  Or we look at filtering these lines just for the files within this repo.

tbkka:
It probably makes sense to strip those lines from the sources in this repo to avoid the self-churn.

thomasvl:
[...]

Issue #219, opened 2017-02-06
Add version stamp to generated source


thomasvl:
Add a version stamp to the generated source, so when the time comes when we do have the change things we can make it an explicit compile error instead of a cascade of errors.

thomasvl:
Fixed by #234

PR #225, opened 2017-02-07
Includes protoc and plugin version in generated sources.

thomasvl:
Hopefully will help trouble shooting in the future as it leaves a
breadcrumb about generation.

Filter out the versions in the checked in files so as not to
cause a lot of churn within this project.

Fixes https://github.com/apple/swift-protobuf/issues/111


thomasvl:
Having now implemented this, I'm of mixed feelings.

- Part of me feels like this is sorta like putting in date/time stamps; and something that makes it harder to have hermetic builds.  https://github.com/apple/swift-protobuf/issues/219 will add a something to cause mismatches to have compile errors, so I'm not completely sure how useful this info ends up being.
- Instead of the post processing sed, it could also be done via a undocumented compiler option (or env variable check?) that we set during generation.
- Currently I'm filtering all the generated source that is checked in, the real problem cases would seem to be the Reference directory, do we only want to post process there?

Thoughts?

tbkka:
Good points.

Off the top of my head, here are a couple of possible alternatives:
* We could mitigate the churn problem by trimming the version info (just major.minor, leaving off the fine details).
* We could emit the detailed version info to stderr as build log information instead of putting it in the generated sources.  That would expose the information to people diagnosing build issues.  (Of course, some build systems interpret any output to stderr as evidence of a build failure.)

thomasvl:
My fear would be when there is a problem, those point version matter; which likely would mean using the short version won't get us much.  When the short version does really change, odds are the work for #219 will call it out instead.

I'm a big fan a silent builds unless there is a problem.  If builds are chatty folks are more likely to miss real warnings/errors.

allevato:
IMO, the version of `protoc` and `protoc-gen-swift` used to generate the protos seems less important than the compatibility version introduced in #234, which will encode a more meaningful (at least to the runtime, not necessarily to humans) version number in the source. So including the other version numbers seems like it would just make noisier diffs without a lot of value.

The argument could be made that the `protoc` version matters in situations where old compilers were broken (i.e., JSON names), but I'm not sure that happens frequently enough to worry about it here?

thomasvl:
Even in the old protoc case, we're printing out a warning during generation.  Once could argue that warning might not be seen by folks, but we still don't have anything like `#warning`, so the best we could do is a generated comment if we want something to live in the file for these case.

Anyway, I'm on the fence.

thomasvl:
@tbkka @allevato any verdict here?

I guess we can submit this as is, and also revisit it if the info proves to be a problem for folks using the plugin.

allevato:
Sorry, thought I replied to this yesterday.

I don't see a problem with the way it's being done now. Every generated Swift file will contain the compatibility version now, which is probably more important than the specific protoc or protoc-gen-swift version.

thomasvl:
I'll close this out without landing it can close off #111 since we do now issue warnings from the plugin.  If we find a need for it, we can always re-investigate something like this in the future.

PR #228, opened 2017-02-07
Switch debugDescription to TextFormat.

thomasvl:
- Remove DebugDescriptionVisitor.swift
- Remap debugDescription to return the TextFormat.

Re https://github.com/apple/swift-protobuf/issues/148


thomasvl:
I think @allevato is going to open up another issue, looking at this we're realizing a bunch of methods tagged as throwing really shouldn't be able to, so while they call `traverse()`, they really should internally error handle themselves.

thomasvl:
I'll do a follow up cl after this lands to remove the `swiftName` from `FieldNameMap.Names`

PR #232, opened 2017-02-07
Switch debugDescription to TextFormat

thomasvl:
Take 2, with the tests update this time…


Tests/SwiftProtobufTests/Test_AllTypes_Proto3.swift
```diff
@@ -777,13 +777,17 @@ class Test_AllTypes_Proto3: XCTestCase, PBTestHelpers {
         assertDecodeFails([253, 1, 77])
         assertDecodeFails([254, 1, 78])
         assertDecodeFails([255, 1, 79])
-        assertDebugDescription("Proto3TestAllTypes(repeatedInt32:[1])") {(o: inout MessageTestType) in
+        // WHY ISN'T THIS LIKE PROTO2?
+        //   "Proto3TestAllTypes:\nrepeated_int32: 1\n"
+        assertDebugDescription("Proto3TestAllTypes:\nrepeated_int32: [1]\n") {(o: inout MessageTestType) in
```
thomasvl: @tbkka why is this using [] here?

Tests/SwiftProtobufTests/Test_AllTypes_Proto3.swift
```diff
@@ -777,13 +777,17 @@ class Test_AllTypes_Proto3: XCTestCase, PBTestHelpers {
...
             o.repeatedInt32 = [1]
         }
-        assertDebugDescription("Proto3TestAllTypes()") {(o: inout MessageTestType) in
+        assertDebugDescription("Proto3TestAllTypes:\n") {(o: inout MessageTestType) in
             o.repeatedInt32 = []
         }
-        assertDebugDescription("Proto3TestAllTypes(repeatedInt32:[1,2])") {(o: inout MessageTestType) in
+        // WHY ISN"T THIS LIKE PROTO2?
+        //  "Proto3TestAllTypes:\nrepeated_int32: 1\nrepeated_int32: 2\n"
+        assertDebugDescription("Proto3TestAllTypes:\nrepeated_int32: [1, 2]\n") {(o: inout MessageTestType) in
```
thomasvl: Same

Tests/SwiftProtobufTests/Test_AllTypes_Proto3.swift
```diff
@@ -1175,7 +1179,9 @@ class Test_AllTypes_Proto3: XCTestCase, PBTestHelpers {
         assertEncode([154, 3, 2, 2, 3]) {(o: inout MessageTestType) in
             o.repeatedNestedEnum = [.bar, .baz]
         }
-        assertDebugDescription("Proto3TestAllTypes(repeatedNestedEnum:[.bar,.baz])") {(o: inout MessageTestType) in
+        // WHY ISN'T THIS LIKE PROTO2?
+        //  "Proto3TestAllTypes:\nrepeated_nested_enum: BAR\nrepeated_nested_enum: BAZ\n"
+        assertDebugDescription("Proto3TestAllTypes:\nrepeated_nested_enum: [BAR, BAZ]\n") {(o: inout MessageTestType) in
```
thomasvl: And this one.

Tests/SwiftProtobufTests/Test_AllTypes_Proto3.swift
thomasvl: Ah, packed vs. non packed.  Other languages always use long form, and have a print option to ask for the short form.  I wonder if we should special case one item and skip the [] in that case?

tbkka: I'm open to changes here.  My thinking was that for a debug dump (primary use for Text format), it is useful to visually distinguish packed vs. non-packed.  For non-debug uses, "packed" is a specific request to minimize the data size.

Put together, I think these two considerations argue for always using the `[...]` form for packed and the long form for non-packed.

thomasvl: Two catches:
1. A proto3 syntax files has no control over if things are packed or not.
1. packed just controls how well behaved encoders write out the bytes, they have to be able to read both forms.

So I'm not so sure the end developers would understand the distinction.  But they also aren't likely to be comparing dumps and notice the difference.  😄   Since I was updating tests, it stuck out as I had to make otherwise identical changes in the files.

tbkka: > A proto3 syntax files has no control over if things are packed or not.

That would argue for not using `[...]` in the 1-item case .

> packed just controls how well behaved encoders write out the bytes, they have to be able to read both forms.

The text decoder does always support both forms, so we're already okay there.  (In fact, my work branch is dropping the packed/repeated distinction entirely in the generated decode logic, since all decoders always accept either form.)

thomasvl: Ok, land this as is and do the tweaks you mentioned in your cl (and fix up these tests to match the changes)?

tbkka: On further thought:
> A proto3 syntax files has no control over if things are packed or not.

But even in that case, I think the `[...]` has some value as a debug aid for visually distinguishing repeated fields.

thomasvl: Ok, sounds like this is all follow up for the TextFormat changes in general, and not specific to this CL, so I'll go ahead and land this so I can finish cleaning out the info the plugin is capturing that was just to support this.

Issue #35, opened 2016-10-03
Generated message members should be reördered to best expose what users are interested in


allevato:
Folks who use generated protos are likely going to want to open those up to see what properties are available (specifically, how the fields in their messages map to properties in Swift).

The code generator should put the stuff of highest user interest at the top—properties, which have simple enough implementations that they don't really get in the way, and then nested types—and put all the generated implementation details further down, out of the user's way.

Another interesting idea would be to move the generated implementation details into an extension deeper in the file, bubbling as much of the "interface" of the generated messages to the top as possible. Some things can't be done this way (specifically, stored properties cannot be put in extensions), but we could do a fairly good job of segregating what users care about from the things only the runtime needs to know.

tbkka:
Great point:  Readability counts, even for generated code.

thomasvl:
Ok, formatting question, give:

```protobuf
message Msg {
  int32 field1 = 1;
  message SubMsg {
    string name = 1;
  }
  SubMsg field2 = 2;
  enum SubEnum {
    NONE = 0;
    FOO = 1;
  }
  SubEnum field3 = 3;
}
```

How do we want the generated code to look?

_A_
```swift
struct Msg {
  enum SubEnum {
    case NONE
    case FOO
  }
  struct SubMsg {
    var name: String
  }
  var field1: Int32
  var field2: SubMsg
  var field3: SubEnum
}
```

_B_
```swift
struct Msg {
  var field1: Int32
  var field2: SubMsg
  var field3: SubEnum
  enum SubEnum {
    case NONE
    case FOO
  }
  struct SubMsg {
    var name: String
  }
}
```

_C_
```swift
struct Msg {
  var field1: Int32
  var field2: SubMsg
  var field3: SubEnum
}

extension Msg {
  enum SubEnum {
    case NONE
    case FOO
  }
  struct SubMsg {
    var name: String
  }
}
```

Or something else?

@tbkka @allevato

allevato:
My thoughts (with no obvious winner):

* My preferred personal style in general is to put nested types at the top of their parent's scope, before any properties/initializers/methods. However, for the proto case where the code isn't being written by the user, I can see the value of pushing them to the bottom so that the first thing the user sees after the struct header is the meat of *that* type.

* The extension approach has the benefit of letting us push deeply nested things to the left, because we can do this:

```swift
extension Msg.SubMsg.SubSubMsg {
  struct SubSubSubMsg {
  }
  enum SubSubSubEnum {
[...]

PR #234, opened 2017-02-07
Add version check "trick" to ensure compatibility between generated code and runtime.

allevato:
I haven't regenerated the protos yet to use this—I wanted to run the scheme and naming by everyone before filling up the PR with all of that. Once it's signed off, I'll regenerate and add those.


Sources/protoc-gen-swift/FileGenerator.swift
```diff
@@ -370,5 +370,21 @@ class FileGenerator {
             p.outdent()
             p.print("]\n")
         }
+
+        generateVersionCheck(printer: &p)
```
thomasvl: Should we getting it at the top of the file so it is the first error instead of the last?

Sources/protoc-gen-swift/Version.swift
```diff
@@ -17,6 +17,16 @@ struct Version {
     static let minor = 9
     static let revision = 28
     static let versionString = "\(major).\(minor).\(revision)"
+
+    // The "compatibility version" of the runtime library. This should be less
+    // than or equal to (major * 10000 + minor * 100 + revision). We guarantee
```
thomasvl: Should we not even hit at a tie between the two numbers?  After all, we don't want to have to change this, it is something we'll have to do if we make breaking changes to the generated|runtime api interfaces.

Sources/protoc-gen-swift/FileGenerator.swift
allevato: That's a good reason for doing it, but it's also counter to #35—putting stuff the user is most interested in (in the typical case) up top.

The protocol could be collapsed into a one-liner, but I wouldn't want to lose the block of comments that actually explains what it's doing.

Sources/protoc-gen-swift/Version.swift
allevato: That might help mitigate confusion about whether the version means "protos are **only** compatible with this version" rather than "protos are compatible with this version and up, until the next time we change it."

We could just use an integer starting at 1 that we increment every time we have to make a breaking change. WDYT?

Sources/protoc-gen-swift/FileGenerator.swift
thomasvl: I took #35 to be without the structs – if you click on a type and jump to the definition, the first thing within it should be the real methods users will care about.  If we smash this block up at the top of with the generated comment, they sorta go together.

I'm not sure if swiftc does this, but at one point clang was giving up after too many errors, so my fear would be this error wouldn't get printed because of all the other things that this really is meant to cover.

Sources/protoc-gen-swift/Version.swift
thomasvl: Works for me.

allevato:
I've force-pushed the changes you suggested. I also tweaked the names of the version check tokens to try to make the errors as legible as possible. Each mismatched version token produces four errors, but the final note should help users diagnose the problem. If we changed the version token in the runtime to `ProtobufAPIVersion_2`, they would see this if you tried to compile old generated protos against it:

```
note: possibly intended match '_GeneratedWithProtocGenSwiftVersion.Version' (aka '_GeneratedWithProtocGenSwiftVersion._1') does not conform to 'ProtobufAPIVersion_2'
```

The name of the generated struct, `_GeneratedWithProtocGenSwiftVersion`, should be obscure enough that it hopefully never collides with a top-level symbol in generated protos.

Sources/protoc-gen-swift/FileGenerator.swift
```diff
@@ -371,4 +373,17 @@ class FileGenerator {
...
     }
+
+    private func generateVersionCheck(printer p: inout CodePrinter) {
+        let v = Version.compatibilityVersion
+        p.print("// If the compiler emits an error on this type, it is because this file\n")
+        p.print("// was generated by a version of the `protoc` Swift plug-in that is\n")
+        p.print("// incompatible with the version of SwiftProtobuf to which you are linking.\n")
+        p.print("// Please ensure that your are building against the same version of the API\n")
+        p.print("// that was used to generate this file.\n")
+        p.print("fileprivate struct _GeneratedWithProtocGenSwiftVersion: SwiftProtobuf.ProtobufAPIVersionCheck {\n")
```
thomasvl: I guess none of the WKTs are currently generated, would the "SwiftProtobuf." need to be dropped if we're generating something within the library itself?

tbkka: Actually, _all_ of the WKTs are currently generated except for Any and Struct/Value/ListValue.

We do not need to drop the `SwiftProtobuf.` module name for code inside the library.

allevato: Yeah, it's ok to use the module name within the same module. (In fact, it's required to access a global symbol within a scope where a more local name is shadowing it. That's not the case here, but it's harmless.)

PR #235, opened 2017-02-08
Print times for 1 operation, use tabular output, skip C++ Instruments

tbkka:
This changes the timing printout to divide by the run_count, effectively
printing the time for a single operation.  This also required changing the
units to microseconds instead of milliseconds.

This makes it feasible to change the run_count and get comparable results,
which is useful as things get faster and you want to bump up the run_count
to get better precision.  I've gone ahead and bumped the run_count to 1000
while I'm here.

I've also reworked the display during the run to use a tabular format
and reordered the overall flow to first run C++ alone, then Swift alone,
and finally run Swift under Instruments.  This allows you to look at and
think about the values while waiting on Instruments.

I've also shuffled the shell script a bit to omit C++ under Instruments;
it's easy to put that back if you really want to.

### DIFF
--- a/Performance/Harness.cc
+++ b/Performance/Harness.cc
@@ -39,26 +39,26 @@ Harness::Harness(std::ostream* results_stream) :
     repeated_count(10) {}
 
 void Harness::write_to_log(const string& name,
-                           const vector<milliseconds_d>& timings) const {
+                           const vector<microseconds_d>& timings) const {
   if (results_stream == nullptr) {
     return;
   }
 
   (*results_stream) << "\"" << name << "\": [";
   for (const auto& duration : timings) {
-    auto millis = duration_cast<milliseconds_d>(duration);
-    (*results_stream) << millis.count() << ", ";
+    auto micros = duration_cast<microseconds_d>(duration);
+    (*results_stream) << micros.count() / run_count << ", ";
   }
   (*results_stream) << "]," << endl;
 }
 
 Harness::Statistics Harness::compute_statistics(
     const vector<steady_clock::duration>& timings) const {
-  milliseconds_d::rep sum = 0;
-  milliseconds_d::rep sqsum = 0;
+  microseconds_d::rep sum = 0;
+  microseconds_d::rep sqsum = 0;
 
   for (const auto& duration : timings) {
-    auto millis = duration_cast<milliseconds_d>(duration);
+    auto millis = duration_cast<microseconds_d>(duration);
     auto count = millis.count();
     sum += count;
     sqsum += count * count;
--- a/Performance/Harness.cc
+++ b/Performance/Harness.cc
@@ -58,8 +58,8 @@ Harness::Statistics Harness::compute_statistics(
   microseconds_d::rep sqsum = 0;
 
   for (const auto& duration : timings) {
-    auto millis = duration_cast<microseconds_d>(duration);
-    auto count = millis.count();
+    auto micros = duration_cast<microseconds_d>(duration);
+    auto count = micros.count();
     sum += count;
     sqsum += count * count;
   }
--- a/Performance/Harness.h
+++ b/Performance/Harness.h
@@ -46,7 +46,9 @@ public:
   void run();
 
 private:
-  /** Include fractional parts in milliseconds. */
+  /** Microseconds unit for representing times. */
+  typedef std::chrono::duration<double, std::micro> microseconds_d;
+  /** Milliseconds unit for representing times. */
   typedef std::chrono::duration<double, std::milli> milliseconds_d;
 
   /**
@@ -62,6 +64,7 @@ private:
   std::ostream* results_stream;
 
   /** The number of times to loop the body of the run() method. */
+  /* Increase this to get better precision. */
   int run_count;
 
   /** The number of times to measure the function passed to measure(). */
@@ -70,8 +73,11 @@ private:
   /** The number of times to add values to repeated fields. */
   int repeated_count;
 
+  /** Ordered list of task names */
+  std::vector<std::string> subtask_names;
+
   /** The times taken by subtasks during each measured attempt. */
-  std::map<std::string, std::vector<milliseconds_d>> subtask_timings;
+  std::map<std::string, std::vector<microseconds_d>> subtask_timings;
 
   /** Times for the subtasks in the current attempt. */
   std::map<std::string, std::chrono::steady_clock::duration> current_subtasks;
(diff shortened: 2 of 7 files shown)


tbkka:
To better explain, here's what I now see in the Terminal when I run `perf_runner.sh 100 int64`:
```
Generating test proto with 100 fields of type int64...
Generating C++ harness source...
Building C++ test harness...

real	0m4.735s
user	0m3.616s
sys	0m0.823s

Running C++ test harness alone...
Running each check 100 times, times in µs
   Populate fields   Decode binary     Decode JSON       Decode text       
            Encode binary     Encode JSON       Encode text       Equality          
  1    0.088    1.465    2.935   74.252  220.326   15.755   36.354    6.387
  2    0.186    1.871    3.037   84.573  254.732   20.008   44.580   19.271
  3    0.227    2.126    3.740   89.114  278.118   21.380   53.968    8.798
  4    0.329    2.066    3.271   80.189  313.558   18.314   42.509    7.529
  5    0.188    2.084    3.209   84.750  319.369   19.449   44.940    8.163
  6    0.276    2.243    3.740   90.546  276.544   24.308   49.184    8.985
  7    0.176    2.031    4.277  119.104  254.927   20.588   45.454    8.065
  8    0.179    1.854    2.792   78.609  275.404   17.099   40.427    8.453
  9    0.216    2.376    3.703   88.336  270.600   23.609   49.019    9.192
 10    0.094    1.170    2.161   69.876  258.500   14.874   36.970    6.229
Relative stddev = 8.3%
C++ harness size before stripping: 93236 bytes
C++ harness size after stripping:  78684 bytes

Generating Swift harness source...
Building Swift test harness...

real	0m4.148s
user	0m3.608s
sys	0m0.391s

Swift dylib size before stripping: 2452628 bytes
Swift dylib size after stripping:  1733460 bytes

Running Swift test harness alone...
Running each check 100 times, times in µs
   Populate fields   Decode binary     Decode JSON       Decode text       
            Encode binary     Encode JSON       Encode text       Equality          
  1    1.413    9.768    3.729  127.768   22.372   60.101   65.941    0.340
  2   30.923   25.119    3.826  135.316   25.809   62.451   69.580    0.382
  3    1.932    9.437    4.328  141.954   31.898   73.284   73.265    0.551
  4    3.490   12.770    5.476  230.172   38.812   76.019   82.230    0.716
  5    2.286   10.470    4.948  219.118   33.392   69.118   74.826    0.768
  6    3.009   11.477    5.139  165.793   37.222   73.311   79.420    1.310
  7    1.737    8.749    4.057  129.859   25.780   63.378  112.988    0.421
  8    2.227   10.369    4.480  142.947   29.240   69.811   74.301    0.650
  9    2.838   11.501    5.921  150.654   33.741  120.012   76.078    0.678
 10    3.480   11.830    5.286  161.589   38.108   78.054   82.538    0.820
Relative stddev = 11.9%

Swift harness size before stripping: 273072 bytes
Swift harness size after stripping:  158392 bytes

Running Swift test harness in Instruments...
```

If you change the `run_count` in the Harness driver, you'll see (about) the same values, just a lower relative stddev.

Performance/Harness.swift
```diff
@@ -14,6 +14,14 @@
 
 import Foundation
 
+private func padded(_ input: String, to width: Int) -> String {
+    var s = input
```
allevato: Instead of a loop, what about reducing this to something like...
```
return input + String(repeating: " ", count: max(0, width - input.characters.count))
```

(Not actually verified, maybe typos.)

Performance/Harness.swift
```diff
@@ -96,10 +129,11 @@ class Harness {
     _ name: String,
     block: () throws -> Result
   ) rethrows -> Result {
+    taskNames.append(name)
     let start = Date()
     let result = try block()
     let end = Date()
-    let diff = end.timeIntervalSince(start) * 1000
+    let diff = end.timeIntervalSince(start) / Double(runCount) * Double(1000000.0)
```
allevato: Nit: The literal 1000000.0 doesn't need to be explicitly made into `Double`

Performance/Harness.h
```diff
@@ -112,42 +118,62 @@ void Harness::measure(const Function& func) {
...
   // Do each measurement multiple times and collect the means and standard
   // deviation to account for noise.
   for (int attempt = 1; attempt <= measurement_count; attempt++) {
-    printf("Attempt %d, %d runs\n", attempt, run_count);
     current_subtasks.clear();
 
     auto start = steady_clock::now();
-    func();
+    for (auto i = 0; i < run_count; i++) {
+        subtask_names.clear();
```
allevato: Though it may be negligible, it's worth noting that this will add the vector's `clear()` overhead into the timings we collect. (Likewise for the Swift one.)

Performance/Harness.cc
```diff
@@ -39,26 +39,26 @@ Harness::Harness(std::ostream* results_stream) :
...
 Harness::Statistics Harness::compute_statistics(
     const vector<steady_clock::duration>& timings) const {
-  milliseconds_d::rep sum = 0;
-  milliseconds_d::rep sqsum = 0;
+  microseconds_d::rep sum = 0;
+  microseconds_d::rep sqsum = 0;
 
   for (const auto& duration : timings) {
-    auto millis = duration_cast<milliseconds_d>(duration);
+    auto millis = duration_cast<microseconds_d>(duration);
```
tbkka: Good catch.  Fixed.

Performance/Harness.h
```diff
@@ -46,7 +46,9 @@ class Harness {
   void run();
 
 private:
-  /** Include fractional parts in milliseconds. */
+  /** Microseconds unit for representing times. */
+  typedef std::chrono::duration<double, std::micro> microseconds_d;
+  /** Milliseconds unit for representing times. */
   typedef std::chrono::duration<double, std::milli> milliseconds_d;
```
tbkka: Good catch!  It was still being used for the overall time summary until (at the last minute) I realized that I didn't care about the overall time, just the relative stddev.

Performance/Harness.h
tbkka: Yeah, that's hard to avoid without reordering the overall test flow.

Performance/Harness.swift
tbkka: That's much cleaner.  Done.

Performance/Harness.swift
tbkka: Removed the unnecessary cast.

Performance/Harness.swift
```diff
@@ -14,6 +14,14 @@
 
 import Foundation
 
+private func padded(_ input: String, to width: Int) -> String {
```
tbkka: Oops.  Fixed.

Performance/Harness.h
```diff
@@ -62,6 +64,7 @@ class Harness {
   std::ostream* results_stream;
 
   /** The number of times to loop the body of the run() method. */
+  /* Increase this to get better precision. */
```
tbkka: Done.

PR #236, opened 2017-02-09
Docs update for `Message` and `Visitor`

dpassage:
A first pass at improving the documentation for the `Message` and `Visitor` protocols.

I've tried to make sure things render appropriately in both jazzy and in Xcode's Quick Help, but not everything comes out perfect in both places.

One annoyance with jazzy in particular: The style guide wants each line no wider than 80 columns, so some of the longer methods on `Visitor` take multiple lines. Jazzy doesn't render that well, only including the first line of the declaration.  Xcode's Quick Help does just fine, so I'll call that a Jazzy issue.

I know the API's still a moving target, but hopefully this improves things. Both style correctness feedback welcome!

This PR increases documentation coverage from 48% to 58%.


Sources/SwiftProtobuf/Google_Protobuf_Any.swift
```diff
@@ -41,6 +41,16 @@ fileprivate func typeName(fromMessage message: Message) -> String {
 }
 
 public extension Message {
+
+  /// Assumes that the data provided when creating the `Google_Protobuf_Any`
+  /// instance is encoded in the default manner, whether's in in binary
+  /// protobuf or JSON formats.
+  ///
+  /// Can be overridden to change how the data should be interpreted. See
```
tbkka: Actually, this should never be overridden.  It's purely a convenience wrapper for `Google_Protobuf_Any.unpackTo(target:)`.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -18,64 +18,100 @@
...
+  /// Name of the message from the original .proto file
   var protoMessageName: String { get }
+
+  /// Name of the protobuf package from the original .proto file
   var protoPackageName: String { get }
+
+  /// Prefix used for this message's type when encoded as an `Any`
   var anyTypePrefix: String { get }
+
+  /// Fully qualifed name used for this message's type when encoed as an `Any`
```
tbkka: Typo: "encoed" should be "encoded"

Sources/SwiftProtobuf/Message.swift
```diff
@@ -18,64 +18,100 @@
...
   ///
   /// This is the core method used by the deserialization machinery.
   ///
   /// Note that this is not specific to protobuf encoding; formats that use
   /// textual identifiers translate those to protoFieldNumbers and then invoke
   /// this to decode the field value.
- mutating func decodeField<T: FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws
+  ///
+  /// - Parameters:
+  ///   - setter: A `FieldDecoder` whose type matches the field
```
tbkka: The "type matches" comment here is actually not important.  I would say that:

> The message will call the method corresponding to the type of this field.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -18,64 +18,100 @@
...
   /// Decode a field identified by a field number (as given in the .proto file).
-  /// The Message will call the FieldDecoder method corresponding
+  /// The `Message` will call the `FieldDecoder` method corresponding
   /// to the declared type of the field.
   ///
   /// This is the core method used by the deserialization machinery.
   ///
   /// Note that this is not specific to protobuf encoding; formats that use
   /// textual identifiers translate those to protoFieldNumbers and then invoke
   /// this to decode the field value.
```
tbkka: I don't know if this helps, but you could include a simple example of what a typical `decodeField()` implementation looks like:
```
public mutating func decodeField<T: FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws {
    switch protoFieldNumber {
    case 1: setter.decodeSingularInt32Field(value: &_myInt32Field)
    case 2: setter.decodeRepeatedStringField(value: &_someListOfStrings)
    default: break
   }
}
```

tbkka: I can envision some people wanting to override this in order to avoid decoding certain fields.  To do so, the easiest way would be to leverage the `_protoc_generated_decodeField()` created by the code generator:
```
public mutating func decodeField<T: FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws {
   switch protoFieldNumber {
   case 1: break // Don't decode this field at all
   default: try _protoc_generated_decodeField(setter: &setter, protoFieldNumber: protoFieldNumber)
   }
}
```

Sources/SwiftProtobuf/Message.swift
```diff
@@ -89,21 +125,28 @@ public protocol Message: CustomDebugStringConvertible {
   // google.protobuf.Any support
   //
 
-  // Decode from an `Any` (which might itself have been decoded from JSON,
-  // protobuf, or another `Any`).
+  /// Decode from an instance of `Any` (which might itself have
+  /// been decoded from JSON, protobuf, or another `Any`).
+  ///
+  /// - Parameter any: item to decode
+  /// - Throws: an instance of `DecodingError` if the message cannot be parsed
```
tbkka: TODO: example

Sources/SwiftProtobuf/Message.swift
```diff
@@ -89,21 +125,28 @@ public protocol Message: CustomDebugStringConvertible {
...
   init(any: Google_Protobuf_Any) throws
 
   /// Serialize as an `Any` object in JSON format.
   ///
   /// For generated message types, this generates the same JSON object as
   /// `serializeJSON()` except it adds an additional `@type` field.
```
tbkka: More explanation:  This is used by the JSON serialization machinery to allow certain types to override how they are represented within `Any` containers.  You should not normally call or override this method.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -18,64 +18,100 @@
...
+  /// `traverse` methods.  The details get a little involved due to
   /// the need to allow particular messages to override particular
   /// behaviors for specific encodings, but the general idea is quite simple.
   func traverse(visitor: Visitor) throws
 
-  //
-  // Protobuf Binary decoding
-  //
+  /// Attempts to decode raw bytes into this message.
+  ///
```
tbkka: You should not normally call or override this method.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -89,21 +125,28 @@ public protocol Message: CustomDebugStringConvertible {
...
   //
   // JSON encoding/decoding support
   //
 
-  /// Overridden by well-known-types with custom JSON requirements.
+  /// Serialize using the standard protobuf JSON representation.
+  ///
+  /// - Throws: an instance of `EncodingError` on failure
   func serializeJSON() throws -> String
   init(decoder: inout JSONDecoder) throws
```
tbkka: We should separately document the `init(decoder: inout JSONDecoder)` here.  It is an internal helper that is overridden by messages that have specialized JSON encodings.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -112,17 +155,33 @@ public protocol Message: CustomDebugStringConvertible {
   // They are implemented in the protocol, not in the generated structs,
   // so can be overridden in user code by defining custom extensions to
   // the generated struct.
+
+  /// Hash value generated from this message's contents, for conformance with
+  /// the `Hashable` protocol
   var hashValue: Int { get }
+
+  /// Textual representation of this message's contents suitable for debugging,
+  /// for conformance with the `CustomDebutStringConvertible` protocol.
```
tbkka: The output format is identical to `serializeText()`.

tbkka:
Note:  Reviewing this, I see a lot of things should ordinarily not be called nor overridden.  I'm trying to remove some of those, but any that remain should be clearly documented and the names should be prefixed with `_` (or in the case of initializers, the first parameter name should be prefixed).

Sources/SwiftProtobuf/Message.swift
thomasvl: Actually it is:

> \[class name\]:
> \[serializeText()\]

Namely because if they are in a collection, it wouldn't be clear what they actual options were when dumping the collections description.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -89,21 +125,28 @@ public protocol Message: CustomDebugStringConvertible {
...
+  ///
+  /// - Throws: an instance of `EncodingError` on failure
   func serializeAnyJSON() throws -> String
 
   //
   // JSON encoding/decoding support
   //
 
-  /// Overridden by well-known-types with custom JSON requirements.
+  /// Serialize using the standard protobuf JSON representation.
```
allevato: For functions that return a value, please express the documentation in terms of the value that it returns rather than the action that it performs. For example, "Returns the JSON-serialized representation of this message."

(I have another issues open to improve the name of this method itself and others.)

Sources/SwiftProtobuf/Message.swift
```diff
@@ -112,17 +155,33 @@ public protocol Message: CustomDebugStringConvertible {
...
+  /// If you override this, make sure you maintain the property that values
+  /// which are `==` to each other have identical `hashValues`, providing a
+  /// custom implementation of `==` if necessary.
   var hashValue: Int {
     let visitor = HashVisitor()
     try? traverse(visitor: visitor)
     return visitor.hashValue
   }
 
+  /// Uses a `Visitor` to recursively generate a description. May be overridden
```
allevato: The fact that a `Visitor` is used here is an unimportant detail—it doesn't need to be included (and the implementation of this function is changing soon anyway with a PR from @thomasvl).

Sources/SwiftProtobuf/Visitor.swift
```diff
@@ -90,9 +71,11 @@ public protocol Visitor: class {
   func visitUnknown(bytes: Data)
 }
 
-/// Forwarding default implementations of some visitor methods, for convenience.
 extension Visitor {
 
+  /// Forwards to `visitRepeatedField`. Implementors who need to handle
```
allevato: Does Jazzy actually pick up documentation for protocol extension methods? If so, that's great—and appears to go beyond what Xcode does. 👍

dpassage:
Thanks for the feedback everyone! I'll have a revision up probably tomorrow.

Sources/SwiftProtobuf/Message.swift
dpassage: Is this a method that end users should be overriding? I was under the impression that it's exclusively generated code.

Sources/SwiftProtobuf/Message.swift
dpassage: Example of usage or implementation?

Sources/SwiftProtobuf/Message.swift
```diff
@@ -18,64 +18,100 @@
...
+///     func insertIntoSet<M: Message & Hashable>(message: M) {
+///         mySet.insert(message)
+///     }
+///
+/// The actual functionality is implemented either in the generated code or in
+/// default implementations of the below methods and properties. Some of them,
+/// including `hashValue` and `debugDescription`, are designed to let you
+/// override the functionality in custom extensions to the generated code.
 public protocol Message: CustomDebugStringConvertible {
+  /// Builds an instance of the message with all fields initialized to
```
dpassage: Check.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -18,64 +18,100 @@
...
+  /// their default values.
   init()
 
   // Metadata
   // Basic facts about this class and the proto message it was generated from
   // Used by various encoders and decoders
+  /// Name of this message's concrete Swift type
```
dpassage: Check

Sources/SwiftProtobuf/Message.swift
```diff
@@ -18,64 +18,100 @@
...
   var swiftClassName: String { get }
+
+  /// Name of the message from the original .proto file
```
dpassage: check

Sources/SwiftProtobuf/Message.swift
dpassage: Check

Sources/SwiftProtobuf/Visitor.swift
dpassage: Yup! And it renders it nicely:

image

dpassage:
OK, I think I got most of the comments. Thanks again, tech writing isn't my day job. :)

tbkka:
@dpassage:  Now that we've been through (hopefully the last major) churn on these APIs, would you be able to take another pass at it?

dpassage:
Yes, I'll do that; should be later this week...

dpassage:
API layout changed enough between then and now that I've created a new PR, #361. Hopefully all the points raised on this one are addressed over there.

PR #237, opened 2017-02-09
Remove extra Swift naming info

thomasvl:
I believe this takes care of everything left in the runtime/generated sources around Swift names that is no longer needed.  Let me know if I missed another, otherwise, after landing this I'll closed out #148.


tbkka:
Nice work!

Issue #12, opened 2016-09-28
Revisit public method/property/initializer signatures


allevato:
Several of the method/property/initializer signatures feel a bit un-Swifty and unclear about their purpose or their inputs/outputs.

For example, `init(protobuf:)` takes a `Data` argument but that's not clear from the argument label; likewise `serializeProtobuf()` returns `Data`. Off the top of my head, names that would better fit with [Swift's API naming guidelines](https://swift.org/documentation/api-design-guidelines/) would be `init(data:)` and `serializedData()`, as examples.

Nailing down best practices for API naming on the swift-evolution mailing list was a huge effort, and we should audit the user-facing APIs in SwiftProtobuf and ensure that they follow the same patterns for consistency.

allevato:
Added note: It's also important that we do this very soon, to determine what changes we need to make before it gets too late to smoothly break things. (At least for methods/properties/initializers, we can keep the old ones around with a deprecation period).

tbkka:
init(data:) and serializedData() are ambiguous, since they don't specify the format of the data.  It is possible, for instance, that we would someday provide support for NSKeyedArchiver serialization, which also produces and consumes a data object.

init(protobufData:) would satisfy both concerns.  I could live with that.

A similar problem would exist if you tried renaming init(json:) to init(string:).  That doesn't allow for the string being protobuf text vs. JSON.

thomasvl:
I don't think anyone would suggest `init(string:)` because of the ambiguity within the protobuf space.  Playing devil's advocate, there are a few ways to express an object tree in JSON, but hopefully we don't need it to be `init(protobufJSON:)` since the protobuf context is implied.

A fair mount of the  [protocol buffer docs](https://developers.google.com/protocol-buffers/) (and other language apis), seem to just talk about _serialize_ being the action of creating a binary blob.  So within the protobuf space, it seems like it is likely defined.  Maybe `init(serialized:)`?

allevato:
I would expect the string methods to be differentiated by something like `init(textString:)` and `init(jsonString:)`. But as we all know, naming things is the hardest part of software design. :)

[...]

PR #243, opened 2017-02-09
Performance: Decoder architecture and API changes

tbkka:
This is admittedly a pretty big change, even without the Test and Reference protos.

The end result is about a 20% speedup in the binary decoder and a 3x speedup in the JSON decoder.  The binary decoder is closer to the performance of Google's C++ implementation, and the JSON decoder is now faster than the C++ text decoder and comparable in speed to Foundation's JSONSerialization toolkit.

This rests on several big changes, which includes API shifts with repercussions across a lot of code:
* `FieldDecoder` is now simply `Decoder`
* Messages now have a `decodeMessage()` which simply holds the obvious while loop:  calling `Decoder.nextFieldNumber()` to iterate over all the field numbers and passing that directly to `decodeField()`.  This allows decoders to call `decodeMessage()` once instead of calling `decodeField()` for every field, this significantly reduces the copy-on-write overhead for larger messages.
* Note:  I thought I would be able to fold `decodeField()`entirely into `decodeMessage()` but `decodeField()` is still used for Protobuf group parsing and deferred JSON parsing for Any.
* `Decoder` now has a separate method for every combination of singular/repeated and primitive field type.  This reduces the call depth for binary decoding and eliminates some generic overhead.
* As a result of the previous, `FieldType` is ignorant of specific decoders now:  `setFromProtobuf`, `setFromJSON`, and `setFromText` are gone since the type-specific decoding has moved directly into the decoder proper.
* `FieldType` now has a single generic decoder reflection that's used by Map and Extensions.
* `Message` and `Enum` no longer conform to `FieldType`, which reduces a lot of namespace pollution in the generated code.
* While I was adapting JSON decoder to the new `Decoder` interface, I went ahead and converted it to work against the binary data directly (rather than write the whole new interface against the old design and then rewrite it again).
* Note: Text is still using the old String-based design.  Converting it to the same design as the JSON decoder should yield comparable performance gains.

I've taken the liberty of adjusting some field and method names to (hopefully) be clearer than before.


tbkka:
Another thing:  I went ahead and split out the decoder error codes.

Sources/Conformance/conformance.pb.swift
```diff
@@ -180,18 +180,18 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message, SwiftProtobuf.Prot
       self = .None
     }
 
-    public mutating func decodeField<T: SwiftProtobuf.FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws {
-      if self != .None && setter.rejectConflictingOneof {
-        throw SwiftProtobuf.DecodingError.duplicatedOneOf
+    public mutating func decodeField<T: SwiftProtobuf.Decoder>(decoder: inout T, fieldNumber: Int) throws {
+      if self != .None {
+        try decoder.handleConflictingOneOf()
```
tbkka: Since there's no longer a single `DecodingError`, I wanted to provide a way for individual decoders to throw appropriate errors.  So I changed the boolean property `rejectConflictingOneof` to a method that is a no-op for decoders that don't want to reject conflicts here.

Sources/SwiftProtobuf/JSONTypeAdditions.swift
```diff
@@ -420,40 +216,25 @@ public extension Message {
...
+            }
+            if !decoder.scanner.complete {
+                throw JSONDecodingError.trailingGarbage
+            }
         }
     }
 
-    init(decoder: inout JSONDecoder) throws {
-        self.init()
+    public mutating func decodeJSON(from decoder: inout JSONDecoder) throws {
```
tbkka: I've managed to bring the three decoders into a similar structure:
* initializers present the user-facing API (here that's `init(json:String)` and `init(jsonUTF8:Data)`
* There's a `decode<Format>(from: inout <Format>Decoder)` on `Message` that provides the override hook for specific messages to override specific formats.  The contracts aren't entirely well-documented yet.

Sources/SwiftProtobuf/Map.swift
```diff
@@ -15,10 +15,26 @@
...
 {
-    typealias Key = KeyType.BaseType
-    typealias Value = ValueType.BaseType
+    public typealias Key = KeyType.BaseType
+    public typealias Value = ValueType.BaseType
+    public typealias BaseType = Dictionary<Key, Value>
+}
+
+public struct ProtobufMessageMap<KeyType: MapKeyType, ValueType: Message>
+    where KeyType.BaseType: Hashable
```
tbkka: Since `Message` and `Enum` are no longer `FieldType`, we can't use the generic `ProtobufMap` for them.  So I ended up triplicating most of the map support.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -104,7 +107,8 @@ public protocol Message: CustomDebugStringConvertible {
 
   /// Overridden by well-known-types with custom JSON requirements.
   func serializeJSON() throws -> String
-  init(decoder: inout JSONDecoder) throws
+
+  mutating func decodeJSON(from: inout JSONDecoder) throws
```
tbkka: Note the parallel structure: `decodeJSON(from:)`, `decodeText(from:)`, and `decodeProtobuf(from:count:extensions:)`.

Hmmm...  I should review the `decodeProtobuf()` form and see if it can accept a `ProtobufDecoder` instead.

Sources/SwiftProtobuf/TextDecoder.swift
```diff
@@ -21,67 +21,449 @@ import Swift
...
-public struct TextDecoder: FieldDecoder {
-    private var scanner: TextScanner
-    public var complete: Bool {return scanner.complete}
-    public var rejectConflictingOneof: Bool {return true}
+public struct TextDecoder: Decoder {
+    internal var scanner: TextScanner
+    private var fieldCount = 0
+    private var terminator: UInt8?
+    private var fieldNameMap: FieldNameMap?
+    private var messageType: Message.Type?
```
tbkka: Text and JSON decoders now have common overall structure:  the "scanner" holds overall state (position within the stream) and the "decoder" holds the per-message state (field name map, field count, message type, etc).

Sources/SwiftProtobuf/TextDecoder.swift
```diff
@@ -21,67 +21,449 @@ import Swift
...
 
-    internal mutating func decodeFullObject<M: Message>(message: inout M, terminator: UInt8?) throws {
-        guard let nameProviding = (M.self as? ProtoNameProviding.Type) else {
-            throw DecodingError.missingFieldNames
+
+    public mutating func handleConflictingOneOf() throws {
+        throw TextDecodingError.conflictingOneOf
+    }
+
+    public mutating func nextFieldNumber() throws -> Int? {
```
tbkka: `nextFieldNumber()` is called from `decodeMessage()` to iterate over the field numbers.  This requires the decoder to store the field name map and other message-level parsing state.

Sources/SwiftProtobuf/TextDecoder.swift
```diff
@@ -21,67 +21,449 @@ import Swift
...
+        } else if terminator == nil {
+            return nil
+        } else {
+            throw TextDecodingError.truncated
+        }
+
+    }
+
+    public mutating func decodeSingularFloatField(value: inout Float) throws {
+        try scanner.skipRequiredColon()
```
tbkka: For JSON, the colon is skipped when we read the key.  But Text format is less consistent about when a colon is required -- message-valued fields don't require them, other types do -- so the colon needs to be swallowed here when we parse the value.

Sources/SwiftProtobuf/TextScanner.swift
```diff
@@ -264,14 +259,16 @@ private func decodeString(_ s: String) -> String? {
 ///
 /// TextScanner has no public members.
 ///
-public class TextScanner {
+public struct TextScanner {
```
tbkka: I made the TextScanner a struct mainly for consistency.  It seems to make very little performance difference whether you (a) copy the scanner into a sub-decoder and copy back or (b) refcount a class reference.  This is especially true for JSON, where the scanner state is literally just a pointer and a count.

Sources/SwiftProtobuf/duration.pb.swift
```diff
@@ -117,10 +117,16 @@ public struct Google_Protobuf_Duration: SwiftProtobuf.Message, SwiftProtobuf.Pro
 
   public init() {}
 
-  public mutating func _protoc_generated_decodeField<T: SwiftProtobuf.FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws {
-    switch protoFieldNumber {
-    case 1: try setter.decodeSingularField(fieldType: SwiftProtobuf.ProtobufInt64.self, value: &seconds)
-    case 2: try setter.decodeSingularField(fieldType: SwiftProtobuf.ProtobufInt32.self, value: &nanos)
+  public mutating func _protoc_generated_decodeMessage<D: SwiftProtobuf.Decoder>(decoder: inout D) throws {
```
tbkka: I plan to look into moving this boilerplate into a library extension.  It will be interesting to see what performance and/or code size impact that has.

Sources/SwiftProtobuf/Enum.swift
```diff
@@ -17,7 +17,7 @@
 
 import Swift
 
-public protocol Enum: RawRepresentable, Hashable, CustomDebugStringConvertible, FieldType, MapValueType {
+public protocol Enum: RawRepresentable, Hashable, CustomDebugStringConvertible {
     init()
     init?(jsonName: String)
     init?(protoName: String)
```
thomasvl: Can/Should these last two move into an internal extension?  i.e. - does the inner plumbing on the library only need to call them and do we see a need to expose them to developers?

Sources/SwiftProtobuf/Google_Protobuf_Any.swift
```diff
@@ -18,6 +18,33 @@
 import Foundation
 import Swift
 
+/// Any objects can be parsed from Protobuf Binary, Protobuf Text, or JSON.
```
thomasvl: s/Protobuf Text/TextFormat/ ?

Sources/SwiftProtobuf/FieldNameMap.swift
```diff
@@ -8,6 +8,82 @@
 //
 // -----------------------------------------------------------------------------
 
+/// Proto field names are always ASCII, so we can improve the
+/// performance of some of our parsing by storing them as
+/// arrays of UInt8 to avoid the overhead of creating and
+/// comparing Unicode Strings.
```
allevato: This feels a bit rough and tricky to maintain going forward—I think we can improve this quite a bit.

There are two situations where we need to create instances of these field names: in the generated code (where it's a literal string), and while parsing JSON/text input (where we have a UBP<UInt8>). In both cases, we know they should be ASCII.

So, we could design our own small `ASCIIString` type as follows:
* It internally holds an `UnsafeBufferPointer<UInt8>` to the string contents
* It conforms to `ExpressibleByStringLiteral` using `StaticString` as its `StringLiteralType`. This handles the case for the names in generated code; since you can access the `UBP<UInt8>` of a `StaticString` directly, and since static strings are stored in the data segment, we don't have to worry about retention.
* It has another `init` that takes a `UBP<UInt8>` directly, which handles the case of names scanned during parsing. As long as we ensure that an instance created this way doesn't escape its calling context, we're good.
* It conforms to `Hashable` and `Equatable` for the dictionary key use case. (`Equatable` can just be implemented in terms of `UBP<UInt8>.elementsEqual` instead of hand-rolling a loop.)

A side effect of this is that `ASCIIString` will have to be public since it will become the `FieldNameMaps.Names` payload type instead of `String`.

This should give you the same performance benefits but will be easier to work with in the future. (Indeed, the performance might be a bit better because you're eliminating all the `String` usage, including the ones that get created when the dictionary is first initialized. So there should be some heap memory usage improvements here too.)

Sources/SwiftProtobuf/Decoder.swift
```diff
@@ -0,0 +1,132 @@
...
+    // but someone might want to track OneOf conflicts in the future.
+    mutating func handleConflictingOneOf() throws
+
+    // Get the next field number.  For JSON and Text, the decoder
+    // translates name to number at this point, based on information
+    // it obtained from the message when it was initialized.
+    mutating func nextFieldNumber() throws -> Int?
+
+    // Primitive field decoders
+    mutating func decodeSingularFloatField(value: inout Float) throws
```
allevato: This also feels like it's leading up to the elimination of the `FieldType` types entirely...

tbkka: I don't think the `FieldType` types can go away entirely unless we commit to exporting a separate type for every possible Map and Extension.  But I do think all of the `FooTypeAddition.swift` files will disappear by the time this is done.

tbkka: Having the singular forms return might work well.  I had kept them this way for consistency (with both the repeated forms and with the singular message form), but the duplication is getting annoying.

One oddity, though:  JSON parsing of `null` resets the field.  That's the one place today where the optional and non-optional forms here differ:  The optional form resets the field to nil, the non-optional form resets it to zero/false/etc.  Any thoughts?

Sources/SwiftProtobuf/FieldNameMap.swift
tbkka: This is a great idea.  I'll give it a try.

Sources/SwiftProtobuf/Decoder.swift
allevato: Urgh, you're right, the JSON ones do differ.

I wonder what the performance difference would be then if you made these methods always return an optional; then proto2 generated methods would just assign it directly, while proto3 would assign `decodeSingular*Field(...) ?? <the default>`. Unwrapping optionals *should* be fast...

So unless you want to try that now, maybe measuring just that change is better done in a future PR.

Sources/SwiftProtobuf/ExtensibleMessage.swift
```diff
@@ -32,21 +32,21 @@ public protocol ExtensibleMessageStorage: class {
 
 public extension ExtensibleMessageStorage {
     public func setExtensionValue<F: AnyExtensionField>(ext: MessageExtension<F, ExtendedMessage>, value: F.ValueType) {
-        extensionFieldValues[ext.protoFieldNumber] = ext.set(value: value)
+        extensionFieldValues[ext.fieldNumber] = ext.set(value: value)
```
tbkka: Note: This implements one of the items proposed in #12

Sources/SwiftProtobuf/Enum.swift
tbkka: It can't be private to the library:  This is the equivalent of `ProtoNameProviding` on `Message` -- it's how the generated code exposes literal names of `Enum` cases.

It does feel like we should be able to somehow reuse `FieldNameMap` here to provide name/number conversions.

allevato: > It does feel like we should be able to somehow reuse FieldNameMap here to provide name/number conversions.

This is a great idea. 👍  There may be some tricks we have to play with unrecognized cases, but I think we can make it work.

Sources/SwiftProtobuf/FieldNameMap.swift
tbkka: > A side effect of this is that ASCIIString will have to be public since it will become the FieldNameMaps.Names payload type instead of String.

Does it?  Would it be a problem if we made the payload type be StaticString and kept the custom type private?

tbkka: I just pushed a commit with this change.  A few interesting points:

 * I did not have to make AsciiFieldName public:  It gets initialized directly from the string literals, so we just need to say we want StaticString there and the type inference lines it all up for us.

* I did have to change the code generation to provide a way for TextDecoder to notice when it's encoding extensions -- the former code relied on the extension to be able to create `FieldNameMap.Names` objects at run-time, which we can't do with `StaticStrings`.  So I added a (long overdue) path for encoders to actually know when they are encoding extensions.

tbkka: I went ahead and flipped the TextScanner over to build an `Array<UInt8>` instead of a `String` for the name lookup.  That was an easy >30% speedup there.  ;-)

allevato: Good point—I was thinking we needed to preserve the untyped-literalness of the string all the way through to our custom type, but that's not needed for `StaticString` (we aren't at risk of allocating anything we don't expect to).

Sources/SwiftProtobuf/FieldNameMap.swift
```diff
@@ -24,62 +24,30 @@
 private let i_2166136261 = Int(bitPattern: 2166136261)
 private let i_16777619 = Int(16777619)
 
-private enum AsciiFieldName: Hashable {
-    case array([UInt8])
-    case mem(UnsafeBufferPointer<UInt8>)
+private struct AsciiFieldName: Hashable {
```
allevato: We could just call this `ASCIIName` based on your comment elsewhere that it could be used for enum names too.

Sources/SwiftProtobuf/FieldNameMap.swift
```diff
@@ -94,38 +62,47 @@ public struct FieldNameMap: ExpressibleByDictionaryLiteral {
...
-    public var protoName: String {
+    public var protoStaticStringName: StaticString {
       switch self {
       case .same(proto: let name): return name
       case .unique(proto: let name, json: _): return name
       }
     }
 
+    /// Returns the proto (and text format) name in the bundle.
+    public var protoName: String {
```
allevato: Do we need the `String` variants of these properties/methods now?

Sources/SwiftProtobuf/ProtoNameProviding.swift
```diff
@@ -42,8 +34,12 @@ extension ProtoNameProviding {
   public func _protobuf_fieldNames(for number: Int) -> FieldNameMap.Names? {
     return Self._protobuf_fieldNames.fieldNames(for: number)
   }
+}
 
-  public func _protobuf_fieldNumber(forProtoName name: String) -> Int? {
-    return Self._protobuf_fieldNames.fieldNumber(forProtoName: name)
-  }
+
+extension ProtoNameProviding where Self: ExtensibleMessage {
```
allevato: What's going on here? :)

Sources/SwiftProtobuf/Decoder.swift
tbkka: Filed Issue #246 to capture this.

Sources/SwiftProtobuf/Enum.swift
tbkka: Captured in Issue #247.

Sources/Conformance/conformance.pb.swift
```diff
@@ -245,18 +245,24 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message, SwiftProtobuf.Prot
 
   init() {}
 
-  public mutating func _protoc_generated_decodeField<T: SwiftProtobuf.FieldDecoder>(setter: inout T, protoFieldNumber: Int) throws {
-    switch protoFieldNumber {
-    case 1, 2: try payload.decodeField(setter: &setter, protoFieldNumber: protoFieldNumber)
-    case 3: try setter.decodeSingularField(fieldType: Conformance_WireFormat.self, value: &requestedOutputFormat)
+  public mutating func _protoc_generated_decodeMessage<D: SwiftProtobuf.Decoder>(decoder: inout D) throws {
```
allevato: Something I've been thinking about lately—instead of having so much duplicated generation code for messages with storage and messages without (and having these methods on both the message and the storage), we should just generate everything in the message and have it reach into the storage as needed to access the field properties. This would mean methods like this, which are currently generated differently in each case, could instead be implemented in a protocol extension.

tbkka: > ... we should just generate everything in the message and have it reach into the storage as needed to access the field properties...

This would require care to ensure that "reaching into the storage" did not pay too much for the copy-on-write and refcount overhead.  In particular, you don't want to go through the accessors for each separate field.

Sources/SwiftProtobuf/FieldNameMap.swift
tbkka: None of our code uses this, so we could drop them.

If we did open up `FieldNameMap` for users, then I would prefer to have `String`-based access rather than exposing the `UBP` ones.

Sources/SwiftProtobuf/ProtoNameProviding.swift
tbkka: Ooops.  That was an experiment to see if I could move some of the overrides out of the generated code.   I'll remove it.

tbkka: Removed.

Sources/Conformance/conformance.pb.swift
allevato: Hmm. Fields that are arrays or maps can be stored in the message itself unconditionally, since they're already CoW types. For simple scalars, I would hope that there wouldn't be any additional retain/release overhead to reach into the class and pull the values out, but based on some of the generated assembly I saw when I was submitting a `Character` change PR to Swift stdlib, you might be right that they don't get optimized away very well.

We should profile it and see what we end up with, but it doesn't need to be done by 1.0. At the very least, it would be nice to figure out how to unify the code in the generator that writes these out so that there aren't two copies of it that need to be kept up-to-date.

thomasvl: +1 to simplifying `storage`.  If we move the repeated/map fields out, that alone would be a nice help.  I've got some ideas for how to otherwise simplify some of the generator, but I'm hesitant to start in on it with the other stuff going on likely making changes in those also.

tbkka: I do want to revisit this post-1.0.

The current design puts everything on the struct or everything in the CoW storage class; it should be feasible to use both strategies in the same message for different fields.

Associated with this is a performance exploration -- intuitively, it seems that very large structs would be a performance problem so that we should limit the total number of fields directly on the struct and push others into the heap storage.  Protobuf has a general notion that field numbers <16 should be faster than others, so this may be one of the considerations.

Post-1.0, I want to look at overhauling the generator code to introduce a modeling layer that represents the logical structure.  Then the generator logic looks like:  build the model from the descriptor, then walk the model to emit the code.  This is essentially the same idea as the C++ distinction between 'DescriptorProto' and 'Descriptor' -- the latter is what I'm calling the "model".  If done carefully, this model would also simplify separately building the JSON/Text support and simplify tools such as gRPC.

Among other things, I would hope that introducing this modeling layer would make it easier to experiment with field storage organization, by providing a clean place to handle per-field storage decisions.

Sources/SwiftProtobuf/Decoder.swift
tbkka: As I recorded in Issue #246, this definitely hurts performance (with no offsetting improvement in code size).

Sources/SwiftProtobuf/FieldNameMap.swift
tbkka: I shortened the name to `AsciiName` as you suggested.

allevato: Nit: Capitalize `ASCII` in `ASCIIName`; from the Swift API design guidelines:

> Acronyms and initialisms that commonly appear as all upper case in American English should be uniformly up- or down-cased according to case conventions:
>
> ```swift
> var utf8Bytes: [UTF8.CodeUnit]
> var isRepresentableAsASCII = true
> var userSMTPServer: SecureSMTPServer
> ```

tbkka:
Apart from regenerating the protos, I think this is ready to commit now.  Any concerns?

PR #244, opened 2017-02-10
Remove hack for required field serialization.

thomasvl:
Remove the hack that always writes out the default value, instead conditionally
write them out.  This will be needed to eventually have full and partial
serialization methods.

Regenerate with hack removed.


thomasvl:
fyi - tests all still pass because a while ago we disabled all the required field tests since they were testing behaviors that never were to spec.

Issue #133, opened 2016-11-03
OneOf enum None case name


soffes:
Should this be renamed to `none` (from `None`) for consistency with the rest of the generated cases and Swift 3's convention of lowercase enum cases?

Also, should another name be considered to avoid confusion with `Optional.none`?

tbkka:
It was spelled in uppercase to ensure that it could not conflict with a generated case.  Otherwise, we would need special checks in case someone had a oneof field called 'none'.

We could use some other convention, of course.   I have no opposition to changing the name if someone could suggest something better.

soffes:
Ah. That makes sense. Could you just omit the `None` case and make the enum optional instead? I think that feels better. I'll think about it some more.

By the way, totally love that `oneof` get generated this way! So amazing! :heart:

allevato:
Vacation drive-by comment: Generating `oneof` cases as optionals is an interesting idea. It does make pattern matching a little trickier, because you'd have to do this:

```
switch msg.someOneOf {
case .foo(let value)?:  // or case .some(.foo(let value))
  ...
case .none:  // or case nil; in either case, note the lack of question mark
  ...
}
```

(I've encountered a few experienced Swift programmers who weren't aware of the `case .foo?` shorthand. I was once one of them. So one could argue that the patterns needed to use the field would be slightly as discoverable.)

But setting that parenthetical aside, maybe that's not a bad thing because it enforces the fact that either one of the fields has "some" value or "none" of them do.

I pushed for non-optionals everywhere earlier in our generated properties because we needed to handle default values correctly, but unlike non-`oneof` fields, there isn't a default value here to worry about. The case being "none" means that none of its fields are set, and the default values of those individual fields are still accessible using those fields' properties instead of the case, if need be.

I'll have to sleep on it to see if I can come up with any situations where this would break expectations. But it sounds promising.

thomasvl:
I guess one upside is also clearing the `oneof` becomes setting it to `nil` (instead of setting it to `.None`).

I wonder how likely folks would be to nest the whole switch/case under a `if let`, because they don't know the `?` part?

tbkka:
Added labels "API change" (since this would affect the API seen by users) and "generator API" (since it affects the API between the generated code and the library).

thomasvl:
@allevato - Assigning to you given the last major comment seemed to be you'd figure out if there could be issues with it.

[...]

PR #250, opened 2017-02-13
Add isInitialized to Message

thomasvl:
I still need to look at pulling over tests for the new method before this is ready to submit, but I wanted to get this up for initial feedback.

### DIFF
--- a/Sources/SwiftProtobuf/Internal.swift
+++ b/Sources/SwiftProtobuf/Internal.swift
@@ -1,0 +1,42 @@
+// Sources/SwiftProtobuf/Message.swift - Message support
+//
+// Copyright (c) 2014 - 2017 Apple Inc. and the project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/master/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+///
+/// Internal helpers on Messages for the library. These are public
+/// just so the generated code can call them, but shouldn't be called
+/// by developers directly.
+///
+// -----------------------------------------------------------------------------
+
+
+import Foundation
+import Swift
+
+public struct Internal {
+  private init() {}
+
+  public static func areAllInitialized(_ listOfMessages: [Message]) -> Bool {
+    for msg in listOfMessages {
+      if !msg.isInitialized {
+        return false
+      }
+    }
+    return true
+  }
+
+  public static func areAllInitialized<K: Hashable>(_ mapToMessages: [K: Message]) -> Bool {
+    for (_, msg) in mapToMessages {
+      if !msg.isInitialized {
+        return false
+      }
+    }
+    return true
+  }
+
+}
--- a/Sources/SwiftProtobuf/Message.swift
+++ b/Sources/SwiftProtobuf/Message.swift
@@ -41,6 +41,10 @@ public protocol Message: CustomDebugStringConvertible {
   var anyTypePrefix: String { get }
   var anyTypeURL: String { get }
 
+  /// Check if all required fields (if any) have values set on this message
+  /// on any messages withing this message.
+  var isInitialized: Bool { get }
+
   //
   // General serialization/deserialization machinery
   //
@@ -116,6 +120,12 @@ public protocol Message: CustomDebugStringConvertible {
 }
 
 public extension Message {
+
+  var isInitialized: Bool {
+    // The generated code will include a specialization as needed.
+    return true;
+  }
+
   var hashValue: Int {
     let visitor = HashVisitor()
     try? traverse(visitor: visitor)
--- a/Sources/protoc-gen-swift/MessageFieldGenerator.swift
+++ b/Sources/protoc-gen-swift/MessageFieldGenerator.swift
@@ -287,6 +287,30 @@ struct MessageFieldGenerator {
         (descriptor.options.hasPacked ? descriptor.options.packed : isProto3)}
     var isRepeated: Bool {return descriptor.isRepeated}
 
+    // True/False if the field will be hold a Message. If the field is a map,
+    // the value type for the map is checked.
+    var fieldHoldsMessage: Bool {
+        switch descriptor.label {
+        case .required, .optional:
+            let type = descriptor.type
+            return type == .message || type == .group
+        case .repeated:
+            let type = descriptor.type
+            if type == .group { return true }
+            if type == .message {
+                let m = context.getMessageForPath(path: descriptor.typeName)!
+                if m.options.mapEntry {
+                    let valueField = m.field[1]
+                    return valueField.type == .message
+                } else {
+                    return true
+                }
+            } else {
+                return false
+            }
+        }
+    }
+
     var name: String {return descriptor.name}
 
     var swiftBaseType: String {return descriptor.getSwiftBaseType(context: context)}
--- a/Sources/protoc-gen-swift/MessageGenerator.swift
+++ b/Sources/protoc-gen-swift/MessageGenerator.swift
@@ -94,17 +94,21 @@ func hasMessageField(descriptor: Google_Protobuf_DescriptorProto, context: Conte
 
 class StorageClassGenerator {
     private let fields: [MessageFieldGenerator]
+    private let oneofs: [OneofGenerator]
     private let descriptor: Google_Protobuf_DescriptorProto
     private let messageSwiftName: String
     private let isProto3: Bool
     private let isExtensible: Bool
+    private let context: Context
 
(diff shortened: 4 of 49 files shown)


Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -739,3 +944,40 @@ class MessageGenerator {
...
+      if (f.isMessage || f.isGroup) &&
+        HasRequiredFieldsInner(context.getMessageForPath(path: f.typeName)!) {
+        return true
+      }
+    }
+
+    return false
+  }
+
+  let msgDesc = context.getMessageForPath(path: fieldDesc.typeName)!
```
thomasvl: Our use of getMessageForPath suddenly worries me.  [descriptor.proto's](https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.proto#L174-L178) seems to say that string we're using might not be full paths, and could just be the leaf from the parent, but the way we use it it seems like it always has to be full.

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -718,6 +813,116 @@ class MessageGenerator {
...
+          if f.descriptor.isMessage && HasRequiredFields(f.descriptor, context: context) {
+            p.print("case .\(f.swiftName)(let v):\n")
+            p.indent()
+            p.print("if !v.isInitialized {return false}\n")
+            p.outdent()
+          } else {
+            needsDefault = true
+          }
+        }
+        p.print("case .None:\n")
```
thomasvl: When #133 is done, we can drop the `.None` cases from there, shortening them a hair.

Tests/SwiftProtobufTests/unittest.pb.swift
```diff
@@ -6947,6 +6991,18 @@ struct ProtobufUnittest_TestRequiredOneof: SwiftProtobuf.Message, SwiftProtobuf.
 
     init() {}
 
+    var isInitialized: Bool {
```
thomasvl: Does it make sense to try to optimize this in the generator, and if only one case of the oneof is a message that needs to be checked, instead generate an if to test that one instead of a switch?

thomasvl:
Ok, tests added, and I managed to find a bug. I've left the bug fix and regeneration as a set of commits in the chain so the diffs are more inspectable, but I'll reorder/squash them into the first three commits before we merge this so the history on Master is a little more clean/easier to follow.

Sources/protoc-gen-swift/MessageGenerator.swift
tbkka: As far as I can tell, protoc does actually guarantee that these strings are always full paths.  This does not seem to be documented anywhere, though.

thomasvl: Upstreaming the issue (https://github.com/google/protobuf/issues/2711) to hopefully get the definition to say they are always complete.

Sources/SwiftProtobuf/Internal.swift
```diff
@@ -0,0 +1,42 @@
...
+// -----------------------------------------------------------------------------
+
+
+import Foundation
+import Swift
+
+public struct Internal {
+  private init() {}
+
+  public static func areAllInitialized<M: Message>(_ listOfMessages: [M]) -> Bool {
```
allevato: I wonder how different the performance/code size would be if we made this non-generic and the argument type `[Message]` instead (and below).

In this case, since we don't need the generics, we should probably drop them.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -41,6 +41,10 @@ public protocol Message: CustomDebugStringConvertible {
   var anyTypePrefix: String { get }
   var anyTypeURL: String { get }
 
+  // Check if all required fields (if any) have values set on this message
```
allevato: Use three-slashes so this gets picked up as a doc-comment.

Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -52,6 +52,28 @@ extension Google_Protobuf_FieldDescriptorProto {
         let m = context.getMessageForPath(path: typeName)!
         return m.options.mapEntry
     }
+    // Returns true/false if the field will be hold a Message. map<> is the
+    func getFieldHoldsMessage(context: Context) -> Bool {
+        switch label {
+        case .required:
```
allevato:
```swift
case .required, .optional:
  return type == .message || type == .group
```

Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -52,6 +52,28 @@ extension Google_Protobuf_FieldDescriptorProto {
...
+            fallthrough
+        case .optional:
+            return type == .message || type == .group
+        case .repeated:
+            if type == .group { return true }
```
allevato: The indentation in this block looks like a combination of 2 and 4 spaces, making it a little hard to follow.

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -718,6 +813,116 @@ class MessageGenerator {
         p.print("}\n")
     }
 
+    func generateIsInitialized(printer p: inout CodePrinter) {
+
```
allevato: Delete extra blank line

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -739,3 +944,40 @@ class MessageGenerator {
         }
     }
 }
+
+// The logic for this check comes from google/protobuf; the C++ and Java generators specificly.
+//
+// This is a helper for generating isInitialized methods.
+fileprivate func HasRequiredFields(_ fieldDesc: Google_Protobuf_FieldDescriptorProto, context: Context) -> Bool {
```
allevato: Nit: lowercase `hasRequiredFields`

Sources/SwiftProtobuf/Internal.swift
```diff
@@ -0,0 +1,42 @@
...
+/// just so the generated code can call them, but shouldn't be called
+/// by developers directly.
+///
+// -----------------------------------------------------------------------------
+
+
+import Foundation
+import Swift
+
+public struct Internal {
```
thomasvl: @allevato @tbkka Does it make sense to move any other stuff here?  Especially some of the helper types used in parsing/generating?  I assume they don't have to physically move, just get updated to nest them in an extension of this.

allevato: There will be some limits to what we can move—e.g., generic types can't be nested yet, so those would have to stay at the top level. But there's probably a good amount of public-as-implementation-detail stuff that we could put in there and it would be a good idea to do that.

Sources/SwiftProtobuf/Message.swift
thomasvl: Done.

Sources/SwiftProtobuf/Internal.swift
thomasvl: Done

Sources/protoc-gen-swift/MessageFieldGenerator.swift
thomasvl: Done.

Sources/protoc-gen-swift/MessageFieldGenerator.swift
thomasvl: I think I got it all.

Sources/protoc-gen-swift/MessageGenerator.swift
thomasvl: Done.

Sources/protoc-gen-swift/MessageGenerator.swift
thomasvl: This sucked.  As `swiftc` then got grumpy about the local with the same name, so I went ahead and renamed it changing the calling a little.

Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -52,6 +52,26 @@ extension Google_Protobuf_FieldDescriptorProto {
         let m = context.getMessageForPath(path: typeName)!
         return m.options.mapEntry
     }
+    // Returns true/false if the field will be hold a Message. map<> is the
+    func getFieldHoldsMessage(context: Context) -> Bool {
```
allevato: Is there a reason this is a separate function, instead of putting the body in the property below that calls it? That would get rid of the awkward name.

thomasvl: Just following the general flow between the two, moved.

thomasvl:
Reordered and folded into 4 logic commits for landing.

PR #252, opened 2017-02-13
Allow GOOGLE_PROTOBUF_CHECKOUT to be set in the environment.

tbkka:


Makefile
```diff
@@ -45,7 +45,7 @@ INSTALL=install
 # checkout. Invoke make with GOOGLE_PROTOBUF_CHECKOUT=[PATH_TO_CHECKOUT] to
 # override this value, i.e. -
 #   make [TARGET] GOOGLE_PROTOBUF_CHECKOUT=[PATH_TO_CHECKOUT]
-GOOGLE_PROTOBUF_CHECKOUT=../protobuf
+GOOGLE_PROTOBUF_CHECKOUT?=../protobuf
```
thomasvl: We don't actually document using the environment for this, the info right above only mentions passing it via `make`, if we want to support the env for these, shouldn't we do it for others?  (`PROTOC`, etc.)

tbkka: The GNU Make manual makes a good point that you should be cautious about accepting settings from the environment, so I'm not in a hurry to do this globally.

Issue #207, opened 2017-01-13
Fix up timestamp helpers to match spec


thomasvl:
Quick scan of the code seems to say the helpers added in timestamp.proto aren't correct.  The key part being: https://github.com/google/protobuf/blob/master/src/google/protobuf/timestamp.proto#L103-L106  nanos is always positive.

https://github.com/google/protobuf/blob/master/src/google/protobuf/util/time_util.cc#L68-L72 is some of the logic to steal.

thomasvl:
I believe this was fixed with #253

PR #253, opened 2017-02-13
Update Timestamp for correct spec.

cburrows:
Timestamp and Duration have differing specs for what constitute valid
values. A Timestamp always has nanos in [0, 999999999), whereas a
Duration generally has nanos with signs that match seconds. The behavior
for Timestamp was clearly copied from Duration and this fixes that. The
tests were testing the wrong thing.

I'd like to add more tests.

https://github.com/apple/swift-protobuf/issues/207


thomasvl:
Re more tests, I think we can actually upstream something to the conformance tests.  Send in a json strings that is negative, and then ensure the proto that comes back is correct.  That has the advantage of getting all the languages covered also.  I'd likely do both timestamp and duration to ensure both behaviors are covered.

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
```diff
@@ -191,14 +195,12 @@ private func parseTimestamp(s: String) throws -> (Int64, Int32) {
...
-        || (seconds > 0 && nanos < 0)
-        || (seconds < -62135596800)
-        || (seconds == -62135596800 && nanos < 0)
-        || (seconds >= 253402300800)) {
-            return nil
+    guard isWithinValidTimestampRange(seconds: seconds, nanos: nanos) else {
+        return nil
     }
 
+    let (seconds, nanos) = normalizeForTimestamp(seconds: seconds, nanos: nanos)
```
thomasvl: isWithinValidTimestampRange also calls normalizeForTimestamp, should we just normalize first, and then validate and make validate skip that step?

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
```diff
@@ -270,49 +272,58 @@ public extension Google_Protobuf_Timestamp {
...
-    if n >= 1000000000 || n <= -1000000000 {
-        s += Int64(n) / 1000000000
-        n = n % 1000000000
+private func normalizeForTimestamp(seconds: Int64, nanos: Int32) -> (seconds: Int64, nanos: Int32) {
+    // The Timestamp spec says that nanos must be in the range [0, 999999999), as in actual
+    // modular arithmetic. Integer % and / do not reflect this, so mod() and div() are defined
+    // for this purpose. Factored out, they make normalization very straightforward.
+    func mod<T : SignedInteger>(_ a: T, _ b: T) -> T {
+        let b = abs(b)
+        return a % b >= 0 ? a % b : a % b + b
```
thomasvl: does this end up doing the a%b twice?  i.e. should we use a local for it?

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
```diff
@@ -270,49 +272,58 @@ public extension Google_Protobuf_Timestamp {
...
-    var s = seconds
-    var n = nanos
-    if n >= 1000000000 || n <= -1000000000 {
-        s += Int64(n) / 1000000000
-        n = n % 1000000000
+private func normalizeForTimestamp(seconds: Int64, nanos: Int32) -> (seconds: Int64, nanos: Int32) {
+    // The Timestamp spec says that nanos must be in the range [0, 999999999), as in actual
+    // modular arithmetic. Integer % and / do not reflect this, so mod() and div() are defined
+    // for this purpose. Factored out, they make normalization very straightforward.
+    func mod<T : SignedInteger>(_ a: T, _ b: T) -> T {
```
thomasvl: Do both of these really need to be generic?  Seems like the types are consistent in the usage.

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
```diff
@@ -15,6 +15,10 @@
 
 import Swift
 
+let minTimestampSeconds: Int64 = -62135596800  // 0001-01-01T00:00:00Z
```
thomasvl: Private on all three?

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
```diff
@@ -224,17 +207,30 @@ private func formatTimestamp(seconds: Int64, nanos: Int32) -> String? {
     // The following is crude, but it works.
     // TODO: If String(format:) works, that might be even better
```
thomasvl: Is this still true?  https://github.com/apple/swift-protobuf/blob/5702ce39fa52d950479024b21f6236bfa938566e/Performance/Harness.swift#L103 is using it.

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
```diff
@@ -224,17 +207,30 @@ private func formatTimestamp(seconds: Int64, nanos: Int32) -> String? {
...
     // (it was broken on Linux a while back...)
-    let result = (FormatInt(n: Int32(year), digits: 4)
+
+    func formatInt(n: Int32, digits: Int) -> String {
```
thomasvl: If we really need this, this seem like it could be sorta slow.

thomasvl:
fyi - I'd suggest doing reindents in a sep PR so looking at the combined diff on github is doable.  I ended up going per commit to try and figure out the changes.

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
cburrows: Done

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
cburrows: I am happy to defer to your opinion on style.

However, I did confirm that common subexpression elimination works correctly in this case and % is computed at most once for any invocation.

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
cburrows: I wrote them in their full generality but then found that I didn't have an obvious "math utils" collection to put them in. I'll tone it down.

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
cburrows: (when swiftc -O)

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
cburrows: Done

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
cburrows: I don't know ... I would you object to making that change in another PR?

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
cburrows: Yeah I will definitely make another change to get rid of the custom string formatting as soon as I'm in a place to test it well enough.

cburrows:
I removed the re-indent.

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
thomasvl: That's fine, lets make sure we track it.

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
```diff
@@ -270,49 +266,53 @@ public extension Google_Protobuf_Timestamp {
...
+        assert(b > 0)
+        return a % b >= 0 ? a % b : a % b + b
     }
-    if s > 0 && n < 0 {
-        n += 1000000000
-        s -= 1
-    } else if s < 0 && n > 0 {
-        n -= 1000000000
-        s += 1
+    func div(_ a: Int32, _ b: Int32) -> Int32 {
```
thomasvl: If you make this Int64 it might remove the cast below.

cburrows: Yeah it does, but then I need to put the cast into the return expression for div, which is ... not right.

I'm still getting used to Swift's refusal to allow for implicit conversions for integer types when they ought to be safe. Or, in this case, I would like to see an operator + with signature (Int64, Int32) -> Int64. It makes me wonder if I'm thinking about something wrong.

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
```diff
@@ -270,49 +266,53 @@ public extension Google_Protobuf_Timestamp {
...
-private func normalizedTimestamp(seconds: Int64, nanos: Int32) -> Google_Protobuf_Timestamp {
-    var s = seconds
-    var n = nanos
-    if n >= 1000000000 || n <= -1000000000 {
-        s += Int64(n) / 1000000000
-        n = n % 1000000000
+private func normalizeForTimestamp(seconds: Int64, nanos: Int32) -> (seconds: Int64, nanos: Int32) {
+    // The Timestamp spec says that nanos must be in the range [0, 999999999), as in actual
+    // modular arithmetic. Integer % and / do not reflect this, so mod() and div() are defined
+    // for this purpose. Factored out, they make normalization very straightforward.
```
thomasvl: Why doesn't simple logic like the C++ has work?  https://github.com/google/protobuf/blob/master/src/google/protobuf/util/time_util.cc#L62-L72

cburrows: Well, right, that works, but I already have a follow-up change to factor out some general purpose math and time functions (including div and mod), at which point this code becomes:

s = s + div(n,x)
n = mod(n,x)

which seems simpler to me. In the sense that it expresses the actual definition of the thing being computed. Do you agree? Then I get to reuse those functions to clear up other inscrutable arithmetic code around here (look at the computation of secondsSinceStartOfDay for instance). I can pull those changes into this PR if you like.

thomasvl:
Are you doing that followup on this PR, or did you want to land this and then do the follow up?

cburrows:
@thomasvl I ported the followup into a commit here--happy to do on this PR if it looks OK to you.

thomasvl:
Might as well finish it out here.  I'll let @allevato comment on that part, since he's the better voice on Swift style for this stuff.

Sources/SwiftProtobuf/TimeUtils.swift
```diff
@@ -0,0 +1,53 @@
...
+
+let minutesPerDay: Int32 = 1440
+let minutesPerHour: Int32 = 60
+let secondsPerDay: Int32 = 86400
+let secondsPerHour: Int32 = 3600
+let secondsPerMinute: Int32 = 60
+let nanosPerSecond: Int32 = 1000000000
+
+func timeOfDayFromSecondsSince1970(seconds: Int64) -> (hh: Int32, mm: Int32, ss: Int32) {
+    let secondsSinceMidnight = Int32(mod(seconds, 86400))
```
allevato: Can you use the rest of the constants you defined above throughout here?

Sources/SwiftProtobuf/TimeUtils.swift
```diff
@@ -0,0 +1,53 @@
...
+    let ss = mod(secondsSinceMidnight, 60)
+    let mm = mod(div(secondsSinceMidnight, 60), 60)
+    let hh = div(secondsSinceMidnight, 3600)
+    
+    return (hh: hh, mm: mm, ss: ss)
+}
+
+func julianDayNumberFromSecondsSince1970(seconds: Int64) -> Int64 {
+    return div(seconds + 210866803200, 86400)
```
allevato: Can you give a better name to, or at least document, 210866803200?

Sources/SwiftProtobuf/Google_Protobuf_Timestamp_Extensions.swift
allevato: Is that code committed? I still see the old version here.

Sources/SwiftProtobuf/TimeUtils.swift
cburrows: Done

Sources/SwiftProtobuf/TimeUtils.swift
cburrows: Done

cburrows:
@allevato In response to "I still see the old version here" -- you're talking about the String formatting? I guess just let me fold in the String.init(format:_:) code that I have if we want to change this for this PR. BRB.

cburrows:
PTAL

cburrows:
Hold off a moment--I'm testing.

cburrows:
Ok, validated on Linux (for the string formatting concerns). Ready to merge. @allevato

PR #263, opened 2017-02-15
Make oneof fields optional and remove the .None case.

### LINKED ISSUE #133: OneOf enum None case name

allevato:
This removes the `.None` case from generated `oneof` enums and goes a bit further by making the property itself optional. This provides nice cohesion with Swift optionals that isn't possible with other fields—since a `oneof` itself doesn't have a default value (its individual fields do), we can take advantage of Swift's own `Optional` type here to convey the difference between a `oneof` that has a value and one for which none of its fields is present.

This allows users to clear the `oneof` by setting it to `nil` (or `.none`). It also necessitates a change in the patterns used in `case` clauses. The former `case .foo(let bar)` now must be written `case .foo(let bar)?` (note the terminal question mark) if one wishes to not manually unwrap it first.

Implementation-wise, the most significant part of this change is the removal of `ExpressibleByNilLiteral` from `OneofEnum` and changing the decode method into an initializer. The latter was required because the decoder logic previously assumed that a (possibly `.None`) value was always present, which is no longer the case when the value is optional and `nil` when unset.

Fixes #133.

### DIFF
--- a/Reference/conformance/conformance.pb.swift
+++ b/Reference/conformance/conformance.pb.swift
@@ -147,44 +147,34 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message, SwiftProtobuf.Prot
   ]
 
 
-  enum OneOf_Payload: ExpressibleByNilLiteral, SwiftProtobuf.OneofEnum {
+  enum OneOf_Payload: SwiftProtobuf.OneofEnum {
     case protobufPayload(Data)
     case jsonPayload(String)
-    case None
 
     static func ==(lhs: Conformance_ConformanceRequest.OneOf_Payload, rhs: Conformance_ConformanceRequest.OneOf_Payload) -> Bool {
       switch (lhs, rhs) {
       case (.protobufPayload(let l), .protobufPayload(let r)): return l == r
       case (.jsonPayload(let l), .jsonPayload(let r)): return l == r
-      case (.None, .None): return true
       default: return false
       }
     }
 
-    init(nilLiteral: ()) {
-      self = .None
-    }
-
-    init() {
-      self = .None
-    }
-
-    mutating func decodeField<T: SwiftProtobuf.Decoder>(decoder: inout T, fieldNumber: Int) throws {
-      if self != .None {
-        try decoder.handleConflictingOneOf()
-      }
+    init?<T: SwiftProtobuf.Decoder>(byDecodingFrom decoder: inout T, fieldNumber: Int) throws {
       switch fieldNumber {
       case 1:
         var value = Data()
         try decoder.decodeSingularBytesField(value: &value)
         self = .protobufPayload(value)
+        return
       case 2:
         var value = String()
         try decoder.decodeSingularStringField(value: &value)
         self = .jsonPayload(value)
+        return
       default:
-        self = .None
+        break
       }
+      return nil
     }
 
     func traverse(visitor: SwiftProtobuf.Visitor, start: Int, end: Int) throws {
@@ -197,15 +187,13 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message, SwiftProtobuf.Prot
         if start <= 2 && 2 < end {
           try visitor.visitSingularField(fieldType: SwiftProtobuf.ProtobufString.self, value: v, fieldNumber: 2)
         }
-      case .None:
-        break
       }
     }
   }
 
   var protobufPayload: Data {
     get {
-      if case .protobufPayload(let v) = payload {
+      if case .protobufPayload(let v)? = payload {
         return v
       }
       return Data()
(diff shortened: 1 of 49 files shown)


Sources/Conformance/conformance.pb.swift
```diff
@@ -147,44 +147,34 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message, SwiftProtobuf.Prot
...
       case 2:
         var value = String()
         try decoder.decodeSingularStringField(value: &value)
         self = .jsonPayload(value)
+        return
       default:
-        self = .None
+        break
       }
+      return nil
```
thomasvl: This reads odd to me because of the "return" vs "return nil".  What is we change the default to just be "return nil", and remove the "return" from the valid cases instead?  Then the one failing case ends up exiting early with a return and everything else just falls through without special attention?

Reference/conformance/conformance.pb.swift
```diff
@@ -147,44 +147,34 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message, SwiftProtobuf.Prot
...
-    init(nilLiteral: ()) {
-      self = .None
-    }
-
-    init() {
-      self = .None
-    }
-
-    mutating func decodeField<T: SwiftProtobuf.Decoder>(decoder: inout T, fieldNumber: Int) throws {
-      if self != .None {
```
thomasvl: Too bad this test has to go at every calling place instead, did you happen to run numbers to see how much the binary grows?  Speaking of, any changes in performance numbers since it is now an optional?

Sources/SwiftProtobuf/Google_Protobuf_Struct.swift
```diff
@@ -304,16 +307,16 @@ public struct Google_Protobuf_Value: Message, Proto3Message, _MessageImplementat
     }
 
     public func _protoc_generated_traverse(visitor: Visitor) throws {
-        try kind.traverse(visitor: visitor, start:1, end: 7)
+        try kind?.traverse(visitor: visitor, start:1, end: 7)
     }
 
     // Storage ivars
-    private var kind = Google_Protobuf_Value.OneOf_Kind()
+    private var kind: Google_Protobuf_Value.OneOf_Kind?
```
allevato: No, optionals are automatically initialized to nil. (We can probably shave off some codegen elsewhere by doing that.)

Sources/SwiftProtobuf/Google_Protobuf_Struct.swift
```diff
@@ -407,69 +410,67 @@ public struct Google_Protobuf_Value: Message, Proto3Message, _MessageImplementat
...
         case boolValue(Bool)
         case structValue(Google_Protobuf_Struct)
         case listValue(Google_Protobuf_ListValue)
-        case None
-
-        public init(nilLiteral: ()) {
-            self = .None
-        }
-
-        public init() {
```
allevato: No, it doesn't make sense for a `oneof` case property to be initialized without *some* field set now that they're optional.

Sources/Conformance/conformance.pb.swift
allevato: The default case doesn't work because there are conditionals inside the other cases, so they would also have to be updated to return nil; it's easier to just have early return in the success cases and a single `return nil` for the failure case.

Reference/conformance/conformance.pb.swift
allevato: `oneof`s are only decoded in one place, so "every calling place" is just moving it from the `oneof`'s decode method to the message's decode method. So, there should be no noticeable code size difference.

Performance-wise, our harness doesn't exercise `oneof`s right now. :(  Given the rarity of `oneof`s compared to other field types, any performance impact is probably negligible, and a reasonable price to pay for the right API.

Sources/SwiftProtobuf/Google_Protobuf_Struct.swift
```diff
@@ -248,7 +248,10 @@ public struct Google_Protobuf_Value: Message, Proto3Message, _MessageImplementat
     mutating public func _protoc_generated_decodeField<T: Decoder>(decoder: inout T, fieldNumber: Int) throws {
         switch fieldNumber {
         case 1, 2, 3, 4, 5, 6:
-            try kind.decodeField(decoder: &decoder, fieldNumber: fieldNumber)
+            if kind != nil {
```
allevato: Each decoder overrides it to handle it as needed. JSON throws an exception; binary decoding does nothing. We should verify that text format throwing is the right behavior.

Sources/Conformance/conformance.pb.swift
allevato: Ah, I was looking at a different one (proto2, which had optionals and if-statements). Using a default branch in the cases where conditionals aren't present is tricky because they can be mixed (message vs. scalar fields, for instance).

Sources/SwiftProtobuf/Google_Protobuf_Struct.swift
tbkka: If I remember correctly, the C++ text decoder optionally rejects conflicting one of fields.  If we need to support such an option someday, `handleConflictingOneOf()` can conditionally throw.

Issue #222, opened 2017-02-06
Look into failing conformance tests


thomasvl:
PR #220 ended up needing some conformance tests to be tagged as failing, likely want all those passing for 1.0 Release.

tbkka:
More detail:  Six tests were having issues:

`Required.ProtobufInput.ValidDataRepeated.DOUBLE.JsonOutput`:  The conformance tester is complaining that it cannot parse the following, but I don't understand why (it looks correct to me):
```
"{\"repeatedDouble\":[0.1,1.79769313486232e+308,2.2250738585072e-308]}"
```

`Required.ProtobufInput.RepeatedScalarSelectsLast.DOUBLE.JsonOutput`:  The conformance tester is complaining that 2.2250738585072e-308 (in our JSON formatted response) is not equal to the expected value 2.2250738585072014e-308.  I suspect that the conformance tester should accept this as a pass.

`Required.JsonInput.AnyNested.JsonOutput` and `Required.JsonInput.AnyNested.ProtobufOutput`:  I have fixes for these in PR #243.

`Recommended.JsonInput.BoolFieldDoubleQuotedTrue` and `Recommended.JsonInput.BoolFieldDoubleQuotedFalse`:  Our JSON parser currently accepts `"true"` and `"false"` (quoted) as synonyms for `true` and `false` in all cases.  This could be fixed by adding a flag so the boolean parser could know whether it was currently parsing a map key or not.

thomasvl:
re - 2.2250738585072e-308 vs 2.2250738585072014e-308 - They aren't actually the same.  Are we losing precision some place?

tbkka:
PR #243 now includes fixes for the two `Recommended` failures as well.

thomasvl:
More failures:`Required.JsonInput.DurationMaxValue.JsonOutput` and `Required.JsonInput.DurationMinValue.JsonOutput`.  Recording them in https://github.com/apple/swift-protobuf/pull/266.

@cburrows - fyi, incase this came out of the recent changes in there.

allevato:
AFAIK they were broken before @cburrows' change was merged.

I remember having all conformance tests pass after I did my original JSON decoder overhaul a while back, so I wonder if some of Tim's changes since then caused floating point literals to be parsed or encoded differently.

tbkka:
I fixed the Duration breakage in PR #273.

cburrows:
Thanks @tbkka, sorry about that. I commented there.

[...]

PR #273, opened 2017-02-16
Issue #222: Duration range truncation

tbkka:
`String(format:)` treats `%d` as a 32-bit integer, truncating
the 64-bit seconds value used in `Google_Protobuf_Duration`.
Change to `%ld` to preserve the 64-bit value.


cburrows:
Should these be %lld?

Issue #272, opened 2017-02-16
Revisit helper method naming for collisions


thomasvl:
Message fields in a proto2 file get `has*` and `clear*` methods generated also.  These are coming out of the namespace for fields, so there could be another field that generates the same name.  If that happens we'll generate code that doesn't compile.

The only options that jump to mind:
- Switch to generating `has_*` and `clear_*` instead.
- Beef up the generator to detect this, and do _something_, my guess is we'd have to rename the property starting with `has`/`clear` to add a `_p` on the end so the has/clear names are available for the other property.

thomasvl:
While objc did the `has_*` style to avoid naming conflicts, it does sorta stick out from an API pov.  So unless someone really objects, I'm closing this and saying we're going with our current choices since they are more _swifty_ even though they could cause some conflicts.

Issue #271, opened 2017-02-16
Revisit edge cases for reserved word handling and sanitizing method.


thomasvl:
Opening after realized and sending out #270.

Things we need to double check:
1. Any place we add a prefix/suffix to something that gets sanitized (like #270)
1. Anything that is a static in one of the Message could collide with a field that generates the same name.
1. Exactly how we sanitize can also come into play, we need to ensure it doesn't generate another name that collide with something that could also be generated.

thomasvl:
Number 3 is a problem we just add "Message", "Oneof", "Enum" to things.

thomasvl:
I believe all the other issues around naming are closed with current decisions/code.  So this issue becomes the last thing.  My goal is to expand the tests #270 did by writing a script to generate all the Swift reserved words as as well as the ones form the library in ever possible place (message name, enum name, enum value, field name, etc.). That will gives us something to ensure we've got all the cases properly handed and code always compiles if any of those names happen to get used.  Also by making it script driven it should be a little easier to maintain/update with the language.

thomasvl:
All the cases I can found are covered.  Closing this out as the reserved word handing stuff can happen post milestone.

PR #274, opened 2017-02-16
Naming cleanup and tests

thomasvl:
Working on parts of #271 and #272


Tests/SwiftProtobufTests/Test_Naming.swift
```diff
@@ -184,3 +184,165 @@ class Test_FieldNamingInitials: XCTestCase {
     msg.clearTheURL()
   }
 }
+
+class Test_ExtensionNamingInitials_MessageScoped: XCTestCase {
+  func testLowers() {
+    var msg = SwiftUnittest_Names_ExtensionNamingInitials()
+
+    msg.SwiftUnittest_Names_Lowers_http = 1
```
thomasvl: So these are the interesting cases at the moment.  If you compare this to the pain fields (above).  There the "http" comes first, so lower make sense, but here because it is always scoped with the message that defined it, should we just go to uppercase for the "http" at the end?  Notice whatis looks like on the `has` and `clear` below also.

thomasvl: @allevato @tbkka any thoughts here?

tbkka: I like the way you have it here, personally, with simple combining of the scope name and field name.

thomasvl: Ok, I'll go ahead and land this.  We can always revisit it we change our mind (before 1.0 is done), it just caused me some pause as I wrote the more complete tests.

PR #278, opened 2017-02-17
Move and tweak naming of ExtensibleMessageStorage.

thomasvl:
For the things that are `public` because the generated code has to be able to use them, but really are internal details to the library/runtime, does it make sense to move/name them to help call this out?  (If Swift let me stick this inside the `struct Internal` I would, but that isn't allowed.)

### DIFF
--- a/Reference/google/protobuf/unittest.pb.swift
+++ b/Reference/google/protobuf/unittest.pb.swift
@@ -4694,7 +4694,7 @@ struct ProtobufUnittest_TestFieldOrderings: SwiftProtobuf.Message, SwiftProtobuf
     200: .unique(proto: "optional_nested_message", json: "optionalNestedMessage"),
   ]
 
-  private class _StorageClass: SwiftProtobuf.ExtensibleMessageStorage {
+  private class _StorageClass: SwiftProtobuf._ExtensibleMessageStorage {
     typealias ExtendedMessage = ProtobufUnittest_TestFieldOrderings
     var extensionFieldValues = SwiftProtobuf.ExtensionFieldValueSet()
     var unknownFields = SwiftProtobuf.UnknownStorage()
@@ -8341,7 +8341,7 @@ struct ProtobufUnittest_TestParsingMerge: SwiftProtobuf.Message, SwiftProtobuf.P
     20: .unique(proto: "RepeatedGroup", json: "repeatedgroup"),
   ]
 
-  private class _StorageClass: SwiftProtobuf.ExtensibleMessageStorage {
+  private class _StorageClass: SwiftProtobuf._ExtensibleMessageStorage {
     typealias ExtendedMessage = ProtobufUnittest_TestParsingMerge
     var extensionFieldValues = SwiftProtobuf.ExtensionFieldValueSet()
     var unknownFields = SwiftProtobuf.UnknownStorage()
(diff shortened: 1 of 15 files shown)


thomasvl:
Expanding on my initial comment, I'm wonders a few things here:
1. Do we want to do rename these sort things to include `Internal_` to hopefully hit that while they are `public` visibility, we might change them in the future, they aren't said to be a stable API.
2. Do we want to move them into files that also hopefully indicate they are internal, gain to help indicate this.

Given we are also working on better generated docs, I'm a little concerned what will show up there, so directly indicating they aren't really public api seems like it might be a good thing.

allevato:
Given limitations of Swift with regard to nesting certain types right now, it would probably be better to uglify the names with an `Internal_` prefix across the board; that way we don't have a weird convention of "some things are nested, some aren't, because *reasons*".

I'm less crazy about changing the file names; people shouldn't be working with our source directly, but instead getting it from SwiftPM or Cocoapads or just importing the project into an Xcode workspace.

SwiftPM won't let us create an "Internal" subdirectory right now because of its rigid directory structure expectations, but it's something we might be able to do in the future. In our Xcode project, we could at least create an "Internal" group under "Sources" and move some things there.

thomasvl:
@allevato So yes on the rename? Yes or no on the move in this case?  What else would be on your list for doing this?

@tbkka how about you?

allevato:
@thomasvl Yes to this rename, no to moving it into `Internal`.

In general, maybe we should only have `Internal` contain functions, not other types, for now. Once we can *uniformly* nest types, we could consider dropping `Internal_` prefixes and making them extensions into `Internal` instead (which would be an easy source change, just replacing `Internal_` with `Internal.` across the board).

allevato:
Oh, and to answer the last question: Other types on the list would be all the concrete decoders/encoders, for one. Users don't need to access those directly.

tbkka:
The Swift standard library does have a long-standing convention of using an underscore `_` prefix on "private" methods and on the first argument name for "private" initializers.

We've also used `_protobuf` as a prefix for "private" things elsewhere.

Reference/google/protobuf/unittest.pb.swift
```diff
@@ -4647,7 +4647,7 @@ struct ProtobufUnittest_TestFieldOrderings: SwiftProtobuf.Message, SwiftProtobuf
     200: .unique(proto: "optional_nested_message", json: "optionalNestedMessage"),
   ]
 
-  private class _StorageClass: SwiftProtobuf.ExtensibleMessageStorage {
+  private class _StorageClass: SwiftProtobuf.Internal_ExtMsgStorage {
```
tbkka: I'm not a fan of cryptic abbreviations.

Sources/SwiftProtobuf/Internal.swift
```diff
@@ -40,3 +40,32 @@ public struct Internal {
   }
 
 }
+
+// SwiftProtobuf Internal: Common support for storage classes to handle extension fields
+public protocol Internal_ExtMsgStorage: class {
+  associatedtype ExtendedMessage: Message
+  var extensionFieldValues: ExtensionFieldValueSet {get set}
+  func setExtensionValue<F: AnyExtensionField>(ext: MessageExtension<F, ExtendedMessage>, value: F.ValueType)
+  func getExtensionValue<F: AnyExtensionField>(ext: MessageExtension<F, ExtendedMessage>) -> F.ValueType
```
tbkka: Side note:  I don't think `setExtensionValue` and `getExtensionValue` need to be here.  The default implementations below should never, ever be overridden.

tbkka:
Apart from the unnecessary abbreviation, this is a good idea.  Let's do it.

Sources/SwiftProtobuf/Internal.swift
thomasvl: Dropped

thomasvl:
Ok, new commit up inplace of the previous one, this just adds the `_` leaving the naming/location as is and drops some things from the protocol that don't need to be there.

thomasvl:
@tbkka ping - this ok with you?  If so I'll land and start doing this for the naming issues I took.

PR #280, opened 2017-02-17
Don't declare Message conformance, Proto[23]Message includes that.

thomasvl:
No need to say all generated messages conform to SwiftProtobuf.Message as
SwiftProtobuf.Proto[23]Message both extend it, so it is already expressed.

### DIFF
--- a/Reference/conformance/conformance.pb.swift
+++ b/Reference/conformance/conformance.pb.swift
@@ -135,7 +135,7 @@ enum Conformance_WireFormat: SwiftProtobuf.Enum {
 ///     1. parse this proto (which should always succeed)
 ///     2. parse the protobuf or JSON payload in "payload" (which may fail)
 ///     3. if the parse succeeded, serialize the message in the requested format.
-struct Conformance_ConformanceRequest: SwiftProtobuf.Message, SwiftProtobuf.Proto3Message, SwiftProtobuf._MessageImplementationBase, SwiftProtobuf.ProtoNameProviding {
+struct Conformance_ConformanceRequest: SwiftProtobuf.Proto3Message, SwiftProtobuf._MessageImplementationBase, SwiftProtobuf.ProtoNameProviding {
   static let protoMessageName: String = "ConformanceRequest"
   static let protoPackageName: String = "conformance"
   static let _protobuf_fieldNames: FieldNameMap = [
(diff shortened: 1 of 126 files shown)


tbkka:
I disagree.

A lot of folks seem to get started by first running their protos through the generator and then looking at the generated code.  Even though this is technically duplicative, I think it's important for documentation purposes.

allevato:
> A lot of folks seem to get started by first running their protos through the generator and then looking at the generated code. Even though this is technically duplicative, I think it's important for documentation purposes.

Taking this to its logical conclusion, doesn't that imply that any time someone conforms to a protocol, they should list all of its ancestor protocols as well?

Is it really asking too much to expect users to follow an inheritance chain once, the first time they look at some code?

thomasvl:
When I realized this the other day, I went back to double check that Proto[23]Message were Messages, since listing it would seem to imply it was needed for some reason. To someone that knows the language, expressing something the language/type system already does is an indicator the author(s) doesn't realize what the code they've written already expresses.

allevato:
> When I realized this the other day, I went back to double check that Proto[23]Message were Messages

+1, I've had to do the same thing. This feels like it's sacrificing legibility for most users of the language in order to avoid a one-time learning opportunity on behalf of users with little Swift experience.

tbkka:
My original thinking was for `Proto[23]` to be mixin/marker protocols.  One way to do that would be the following:

```
 ///
-public protocol Proto2Message: Message {
+public protocol Proto2Message {
   var unknownFields: UnknownStorage { get set }

 ///
-public protocol Proto3Message: Message {
+public protocol Proto3Message {
 }

 /// Proto2 messages preserve unknown fields
-public extension Proto2Message {
+public extension Proto2Message where Self: Message {
     public mutating func decodeBinary(from bytes: UnsafePointer<UInt8>, count: Int, extensions: ExtensionSet?) throws {
         var decoder = BinaryDecoder(forReadingFrom: bytes, count: count, extensions: extensions)

 // Proto3 messages ignore unknown fields
-public extension Proto3Message {
+public extension Proto3Message where Self: Message {
     public mutating func decodeBinary(from bytes: UnsafePointer<UInt8>, count: Int, extensions: ExtensionSet?) throws {
```

I'm pretty sure that Thomas' change would encourage a lot of people to create variables of type `Proto2Message` when I would rather see them use `Message`.  But maybe that's not so bad.

thomasvl:
> I'm pretty sure that Thomas' change would encourage a lot of people to create variables of type `Proto2Message` when I would rather see them use `Message`. But maybe that's not so bad.

Yes, with protocols we can split like you suggest, but I'm not sure there is value, it feels like it ends up causing more questions: _Why does't `Proto2Message` extend `Message`?  What would it mean to have a `Proto3Message` that is **not** a `Message`?_

We're also only taking about the cases where a developer is writing code to act generically on `Message`s.  The vast majority of their code will be using concrete leaf classes because they want to get/set the fields.  The few cases I'd think they will want generic code, they likely will quickly realize they need `Message` if they really are dealing with both types of `Messages`.  Sorta like if they are writing generic code they will realize where they need to use `ExtensibleMessage`.

Anyway, not overly critical, we can let this sit while other things shake out.

tbkka:
After thinking about it some more, I agree.  (I was overthinking this.)  Let's merge Thomas' change and move on to other things...

thomasvl:
Ok, merging.

PR #289, opened 2017-02-21
Remove traverse method from _StorageClass.

allevato:
This is a proof-of-concept for removing the traverse/decode/isEqual/etc. methods from `_StorageClass` and have the methods in the message itself just reach into the storage as needed.

@tbkka had some concerns that this would result in excess retain/releases for each field access that would slow down the operations. That's true by default—the workaround is to use `withExtendedLifetime`, which essentially wraps a block in a retain/release for an object so that the individual accesses inside the block don't need to be guarded.

Here are the perf harness results to show encoding is unaffected (the stats for operations that don't use `traverse` have been removed):

### Current HEAD
```
   Populate fields   Decode binary     Decode JSON       Decode text       
            Encode binary     Encode JSON       Encode text       Equality          
  1             8.436           116.082           83.152                 
  2             8.011           115.860           94.119                 
  3             7.382           113.005           82.662                 
  4             7.735           112.838           83.348                 
  5             7.371           111.767           81.422                 
  6             6.860           109.529           86.798                 
  7             7.039           112.621           84.683                 
  8             7.449           122.730           84.335                 
  9             7.446           112.201           80.305                 
 10             7.488           127.730           87.780                 
Relative stddev = 3.3%
```

### This PR
```
   Populate fields   Decode binary     Decode JSON       Decode text       
            Encode binary     Encode JSON       Encode text       Equality          
  1             8.010           114.422           82.248                 
  2             7.054           114.266           83.320                 
  3             6.391           103.471           75.926                 
  4             6.318           102.302           74.459                 
  5             6.430           103.060           75.744                 
  6             6.244           100.742           74.443                 
  7             6.838           105.088           77.048                 
  8             7.539           121.632           88.102                 
  9             6.841           108.022           79.561                 
 10             6.230           101.811           74.250                 
Relative stddev = 6.1%
```

In fact, the stats in this PR are better, but that might just be statistical noise.

If everything looks good here, I can update the rest of the methods the same way. This has *huge* benefits: the code generator won't have duplicated code paths anymore, and we now have the ability to make smarter decisions about which fields go into storage vs. which go into the message. (For example, arrays for repeated fields and maps are already copy-on-write, so they don't need to be added to the storage class).

### DIFF
--- a/Reference/google/protobuf/any_test.pb.swift
+++ b/Reference/google/protobuf/any_test.pb.swift
@@ -80,18 +80,6 @@ struct ProtobufUnittest_TestAny: SwiftProtobuf.Message, SwiftProtobuf.Proto3Mess
       }
     }
 
-    func traverse(visitor: SwiftProtobuf.Visitor) throws {
-      if _int32Value != 0 {
-        try visitor.visitSingularField(fieldType: SwiftProtobuf.ProtobufInt32.self, value: _int32Value, fieldNumber: 1)
-      }
-      if let v = _anyValue {
-        try visitor.visitSingularMessageField(value: v, fieldNumber: 2)
-      }
-      if !_repeatedAnyValue.isEmpty {
-        try visitor.visitRepeatedMessageField(value: _repeatedAnyValue, fieldNumber: 3)
-      }
-    }
-
     func isEqualTo(other: _StorageClass) -> Bool {
       if _int32Value != other._int32Value {return false}
       if _anyValue != other._anyValue {return false}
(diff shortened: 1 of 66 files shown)


tbkka:
Nice work!  Does this show any improvement in compiled code size?

allevato:
> Does this show any improvement in compiled code size?

Only slightly: the harness size pre-stripping the old way was 277,688 bytes; afterwards it was 277,576. The stripped sizes are the same.

That's fairly expected; the only thing this gets rid of in the compiled code is the extra stack frame preparation for the `traverse` method in `_StorageClass` and the symbols associated with it; the real improvement is in the reduction of generator complexity.

allevato:
I'll merge this in, and then make the same change for the other methods in a separate PR.

PR #291, opened 2017-02-22
Performance overhaul for encoding API

tbkka:
This comprises a number of changes to how serialization is handled:
    
1. All of the encoders are now structs that are passed as `inout` parameters.
    
2. `traverse` methods are now generic in the visitor type instead of taking a protocol type.  This eliminates the creation of protocol witnesses.

3. `traverse` now calls type-specific methods directly, e.g., `visitSingularStringField()`.  This significantly shortens the call depth and eliminates generic reflection.  This is a big overall speedup for binary coding.

4. Text and JSON now construct UTF8 byte sequences directly,  instead of appending to strings.

5. Text and JSON now use `StaticString` field names.

6. Many small micro-optimizations throughout.

Here are a few numbers measured on my laptop:
```    
"100 int64":      C++  Master  This PR
Encode Binary   1.667   8.343    3.772
Encode JSON    78.684 168.201   28.859
Encode Text    17.912 104.672   26.875
    
"100 string":     C++  Master  This PR
Encode Binary   5.350  58.713   13.625
Encode JSON    77.743 239.404   29.504
Encode Text    18.358 174.878   29.545
    
"100 double":     C++  Master  This PR
Encode Binary   0.921   8.099    3.748
Encode JSON    95.823 128.070   24.358
Encode Text    38.967 100.073   25.291
```
    
(This last is cheating a little; our JSON/Text encoders now use a fast path if the double value happens to be an integer, and our test suite only uses integral values for exercising numeric fields.)

Bonus: Most of the encoder-specific `FieldType` methods are gone, replaced by a standard set of abstract reflectors used for maps and extensions.
    
Additional Bonus: All five encoding visitors and all but one of the low-level encoders are now `internal` to the library.  The only exception is `JSONEncoder` which is public but has no public members.


Sources/Conformance/conformance.pb.swift
```diff
@@ -175,15 +175,15 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message, SwiftProtobuf.Prot
       return nil
     }
 
-    fileprivate func traverse(visitor: SwiftProtobuf.Visitor, start: Int, end: Int) throws {
+    fileprivate func traverse<V: SwiftProtobuf.Visitor>(visitor: inout V, start: Int, end: Int) throws {
```
tbkka: Notable change:  `traverse` is generic in the visitor type and accepts the visitor as an `inout` parameter.

Sources/Conformance/conformance.pb.swift
```diff
@@ -175,15 +175,15 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message, SwiftProtobuf.Prot
...
       switch self {
       case .protobufPayload(let v):
         if start <= 1 && 1 < end {
-          try visitor.visitSingularField(fieldType: SwiftProtobuf.ProtobufBytes.self, value: v, fieldNumber: 1)
+          try visitor.visitSingularBytesField(value: v, fieldNumber: 1)
```
tbkka: This seems to be the cause of a small but definite reduction in generated code size.  I'm not sure if it's the simpler call sequence or not referring to the `FieldType` that does it.

Sources/SwiftProtobuf/BinaryEncoder.swift
```diff
@@ -82,63 +80,41 @@ public struct BinaryEncoder {
...
+    mutating func putFixedUInt64(value: UInt64) {
+        var v = value.littleEndian
         let n = MemoryLayout<UInt64>.size
-        withUnsafePointer(to: &v) { v -> () in
-            v.withMemoryRebound(to: UInt8.self, capacity: n) { p -> () in
-                let buff = UnsafeBufferPointer<UInt8>(start: p, count: n)
-                append(contentsOf: buff)
-            }
-        }
+        memcpy(pointer, &v, n)
```
tbkka: I experimented a bit with different ways to put a fixed-size number into the serialized output.  Using `memcpy` like this is the best I've found so far.

Sources/SwiftProtobuf/BinaryEncoder.swift
```diff
@@ -82,63 +80,41 @@ public struct BinaryEncoder {
...
-            // TODO: There has got to be a better way to do this...
-            stringWithNul.withUnsafeBufferPointer { bp -> () in
-                bp.baseAddress?.withMemoryRebound(to: UInt8.self, capacity: stringLength) { p -> () in
-                    let stringWithoutNul = UnsafeBufferPointer<UInt8>(start: p, count: stringLength)
-                    append(contentsOf: stringWithoutNul)
-                }
-            }
+        let count = value.utf8.count
+        putVarInt(value: count)
+        for b in value.utf8 {
```
tbkka: Using the `utf8` iterator like this seems to be pretty quick -- I think in this case the compiler can inline it pretty efficiently.

Sources/SwiftProtobuf/BinaryEncodingSizeVisitor.swift
```diff
@@ -18,54 +18,195 @@ import Foundation
 /// Visitor that calculates the binary-encoded size of a message so that a
 /// properly sized `Data` or `UInt8` array can be pre-allocated before
 /// serialization.
-final class BinaryEncodingSizeVisitor: Visitor {
+internal struct BinaryEncodingSizeVisitor: Visitor {
```
tbkka: All the visitor types are `internal struct` now.

Sources/SwiftProtobuf/BinaryEncodingSizeVisitor.swift
```diff
@@ -18,54 +18,195 @@ import Foundation
...
+    serializedSize += tagSize + MemoryLayout<Float>.size
   }
 
-  func visitRepeatedField<S: FieldType>(fieldType: S.Type,
-                                        value: [S.BaseType],
-                                        fieldNumber: Int) throws {
-    let tagSize = FieldTag(fieldNumber: fieldNumber,
-                           wireFormat: S.protobufWireFormat).encodedSize
-    serializedSize += value.count * tagSize
+  public mutating func visitSingularDoubleField(value: Double, fieldNumber: Int) throws {
```
tbkka: This is all quite repetitive but at least it's pretty easy to read and understand -- the verbosity here is in exchange for dropping the various helper methods off of the `FieldType`s.

Sources/SwiftProtobuf/BinaryTypeAdditions.swift
```diff
@@ -15,235 +15,6 @@
 
 import Foundation
```
tbkka: Huzzah!  All gone!

Sources/SwiftProtobuf/Decoder.swift
```diff
@@ -124,7 +124,7 @@ public protocol Decoder {
     // (ProtobufMessageMap)
     mutating func decodeMapField<KeyType: MapKeyType, ValueType: MapValueType>(fieldType: ProtobufMap<KeyType, ValueType>.Type, value: inout ProtobufMap<KeyType, ValueType>.BaseType) throws
     mutating func decodeMapField<KeyType: MapKeyType, ValueType: Enum>(fieldType: ProtobufEnumMap<KeyType, ValueType>.Type, value: inout ProtobufEnumMap<KeyType, ValueType>.BaseType) throws where ValueType.RawValue == Int
-    mutating func decodeMapField<KeyType: MapKeyType, ValueType: Message>(fieldType: ProtobufMessageMap<KeyType, ValueType>.Type, value: inout ProtobufMessageMap<KeyType, ValueType>.BaseType) throws
+    mutating func decodeMapField<KeyType: MapKeyType, ValueType: Message & Hashable>(fieldType: ProtobufMessageMap<KeyType, ValueType>.Type, value: inout ProtobufMessageMap<KeyType, ValueType>.BaseType) throws
```
tbkka: To fix an issue with the Hash visitor, I had to make the map value types be `& Hashable` throughout.

Sources/SwiftProtobuf/ExtensionFields.swift
```diff
@@ -146,12 +143,9 @@ public struct RepeatedExtensionField<T: FieldType>: ExtensionField {
...
   }
 
-  public func traverse(visitor: Visitor) throws {
+  public func traverse<V: Visitor>(visitor: inout V) throws {
     if value.count > 0 {
-      try visitor.visitRepeatedField(
-        fieldType: T.self,
-        value: value,
-        fieldNumber: protobufExtension.fieldNumber)
+      try T.visitRepeated(value: value, fieldNumber: protobufExtension.fieldNumber, with: &visitor)
```
tbkka: `visitSingular`, `visitRepeated`, and `visitPacked` here are the new general-purpose reflectors that just dispatch to the appropriate `visitor.visit<modifier><type>Field()` method.

Sources/SwiftProtobuf/Google_Protobuf_Wrappers+Extensions.swift
```diff
@@ -30,33 +30,6 @@ protocol ProtobufWrapper {
...
-  // NOTE: The `init(_ value: WrappedType.BaseType)` initializer repeated below
-  // should theoretically be able to go here and be declared public, but this
-  // causes linker errors in release builds (see issue #70). If this is indeed a
-  // bug and should be allowed, we should move the initializer back into this
-  // extension once it's fixed, to reduce a small amount of code duplication/
-  // bloat.
-
-  func serializeWrapperJSON() throws -> String {
-    var encoder = JSONEncoder()
-    try WrappedType.serializeJSONValue(encoder: &encoder, value: value)
```
tbkka: I've not yet completely eliminated `serializeJSONValue`, but this is chipping away at it.

Sources/SwiftProtobuf/ProtoNameResolvers.swift
```diff
@@ -12,22 +12,20 @@
 /// This type contains helper functions to resolve field names based on their
 /// numbers during encoding.
 ///
-/// This type is public since it might be useful to clients who want to write
-/// advanced custom visitors.
-public enum ProtoNameResolvers {
+internal enum ProtoNameResolvers {
```
tbkka: I made this `internal` because changing it to `StaticString` makes it a lot more specialized; I'm not sure if it's really something we want to expose in this form.

Sources/SwiftProtobuf/ProtoNameResolvers.swift
```diff
@@ -12,22 +12,20 @@
...
   /// the given message or in any set extensions.
   ///
   /// If the name cannot be resolved (because the field number is not defined
   /// on the message or any of its extensions, or names were not compiled into
   /// the binary), then the resolver returns nil.
-  public static func protoFieldNameResolver(
+  static func protoFieldNameResolver(
     for message: Message
-  ) -> (Int) -> String? {
+  ) -> (Int) -> StaticString? {
```
tbkka: Changing to `StaticString` here is a noticeable performance bump since the JSON and Text encoders can now just copy the UTF8 bytes across.

Sources/SwiftProtobuf/TextFormatEncoder.swift
```diff
@@ -14,165 +14,261 @@
...
+private let asciiLowerN = UInt8(ascii: "n")
+private let asciiLowerR = UInt8(ascii: "r")
+private let asciiLowerT = UInt8(ascii: "t")
+private let asciiLowerV = UInt8(ascii: "v")
+
 /// TextFormatEncoder has no public members.
-public final class TextFormatEncoder {
-    var text: String = ""
+internal struct TextFormatEncoder {
+    private var data = [UInt8]()
```
tbkka: I also tried `Data()` here but `[UInt8]()` was faster.

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
```diff
@@ -14,209 +14,409 @@
...
-      encoder.startField(name: "value")
-      try ValueType.serializeTextValue(encoder: encoder, value: v)
-      encoder.endField()
-      encoder.endObject()
-      encoder.endField()
-    }
+      try _visitMap(map: value, fieldNumber: fieldNumber) {
+          (visitor: inout TextFormatEncodingVisitor, key, value) throws -> () in
+          try KeyType.visitSingular(value: key, fieldNumber: 1, with: &visitor)
+          try ValueType.visitSingular(value: value, fieldNumber: 2, with: &visitor)
```
tbkka: Note the maps here are using the general-purpose reflectors on FieldType, which is what allows us to lose the specialized `serializeTextValue` methods we used to have.

Tests/SwiftProtobufTests/Test_JSON.swift
```diff
@@ -653,21 +666,206 @@ class Test_JSON: XCTestCase, PBTestHelpers {
 class Test_JSONPacked: XCTestCase, PBTestHelpers {
     typealias MessageTestType = Proto3TestPackedTypes
 
+    func testPackedFloat() {
```
tbkka: I filled in a lot of missing test methods for packed fields in JSON.

Tests/SwiftProtobufTests/Test_Struct.swift
```diff
@@ -284,7 +284,7 @@ class Test_JSON_Value: XCTestCase, PBTestHelpers {
 
         let structValue = try Google_Protobuf_Value(jsonString: "{\"a\":1.0}")
         let d = structValue.debugDescription
-        XCTAssertEqual(d, "SwiftProtobuf.Google_Protobuf_Value:\nstruct_value {\n  fields {\n    key: \"a\"\n    value: {\n      number_value: 1\n    }\n  }\n}\n")
+        XCTAssertEqual(d, "SwiftProtobuf.Google_Protobuf_Value:\nstruct_value {\n  fields {\n    key: \"a\"\n    value {\n      number_value: 1\n    }\n  }\n}\n")
```
tbkka: An actual bug fix:  `value` here is a message field, so shouldn't have a colon.  (Google's C++ implementation accepts a colon here but prefers to not write one.)

Sources/Conformance/conformance.pb.swift
thomasvl: Was there a change in binary size with this?

Sources/SwiftProtobuf/BinaryEncoder.swift
```diff
@@ -82,63 +80,41 @@ public struct BinaryEncoder {
...
-                let buff = UnsafeBufferPointer<UInt8>(start: p, count: n)
-                append(contentsOf: buff)
-            }
-        }
+        memcpy(pointer, &v, n)
+        pointer = pointer.advanced(by: n)
     }
 
     mutating func putFloatValue(value: Float) {
         var v = value
```
thomasvl: This feels odd, doesn't float/double have little/big endian differences?  So would this be right on a big endian machine?

Sources/SwiftProtobuf/BinaryEncodingSizeVisitor.swift
thomasvl: Nice!

Sources/SwiftProtobuf/JSONEncoder.swift
```diff
@@ -53,68 +129,116 @@ public struct JSONEncoder {
...
+                let s = String(value)
                 append(text: s)
             }
         }
     }
-    mutating func putInt64(value: Int64, quote: Bool) {
-        // Always quote integers with abs value > 2^53
-        if quote || value > 0x1FFFFFFFFFFFFF || value < -0x1FFFFFFFFFFFFF {
-            append(text: "\"" + String(value) + "\"")
+    private mutating func appendUInt(value: UInt64) {
```
thomasvl: Is this really faster then using the builtin conversion to a string?

Sources/SwiftProtobuf/JSONEncoder.swift
```diff
@@ -53,68 +129,116 @@ public struct JSONEncoder {
...
-    mutating func putStringValue(value: String) {
-        let hexDigits = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "A", "B", "C", "D", "E", "F"];
-        append(text: "\"")
+
+    internal mutating func putBoolValue(value: Bool) {
+        if isMapKey {
+            data.append(asciiDoubleQuote)
+        }
+        if value {
+            data.append(contentsOf: [asciiLowerT, asciiLowerR, asciiLowerU, asciiLowerE])
```
thomasvl: Is this really better than using literal "true"?

Sources/SwiftProtobuf/ProtoNameResolvers.swift
thomasvl: We can always expose something later if there is reason, but once it is exposed, we're stuck until we want to break api, so I'd try to expose as little as possible.

Sources/SwiftProtobuf/TextFormatEncoder.swift
```diff
@@ -14,165 +14,261 @@
...
-        text.append("\n")
+    mutating func endField() {
+        data.append(asciiNewline)
     }
 
-    func startObject() {
+    mutating func startObject() {
         tabLevel += 1
-        text.append("{\n")
+        data.append(asciiOpenCurlyBracket)
```
thomasvl: For a bunch of these, would we be better with constants have the two characters so it is a single append call?

Sources/SwiftProtobuf/TextFormatEncoder.swift
```diff
@@ -14,165 +14,261 @@
...
 
-    func endField() {
-        text.append("\n")
+    mutating func endField() {
+        data.append(asciiNewline)
     }
 
-    func startObject() {
+    mutating func startObject() {
         tabLevel += 1
```
thomasvl: So if you keep a var currentIndent = "" instead of a tabLevel, you can simply add two spaces when increasing and truncate two when decreasing.  Then the indent is always a single append instead of a loop with two appends per cycle.

Tests/SwiftProtobufTests/Test_JSON.swift
```diff
@@ -653,21 +666,206 @@ class Test_JSON: XCTestCase, PBTestHelpers {
...
+        assertJSONEncode("{\"packedFloat\":[1]}") {(o: inout MessageTestType) in
```
thomasvl: Didn't float and double have different code paths for non integer values?  Do you want coverage for those also?

Sources/SwiftProtobuf/TextFormatEncoder.swift
allevato: `Data` has some interesting locking behavior over the `NSData` that it wraps internally, I found, by digging into the implementation. Doing byte-by-byte operations on it is slower than it needs to be :(

Sources/SwiftProtobuf/JSONEncoder.swift
tbkka: Yes, lots faster.

Sources/Conformance/conformance.pb.swift
tbkka: The PR as a whole cuts binary size by about 4k for 100 fields (regardless of field type).  I didn't measure the impact for each separate change in this case, but I believe when I did a similar rearchitecture on the decoder side, that the generic here had no impact.

Sources/SwiftProtobuf/TextFormatEncoder.swift
tbkka: Not easy to do:  These are numeric constants, not strings, and the append calls get inlined so I think this is actually faster.

Sources/SwiftProtobuf/TextFormatEncoder.swift
tbkka: Done.

Sources/SwiftProtobuf/JSONEncoder.swift
tbkka: > ... I get kind of an "assembly language in Swift syntax" feeling ...

Agreed.  My primary goals for this PR are to:
1. get the API to a place where very good performance is possible.   The easiest way to prove that is to actually demonstrate such performance.
2. isolate the ugly bits.

If we can achieve the first goal, it means that we should be able to commit to the generated code API for a long time.  If we can achieve the second, we'll be able to iterate on better ways to express these things.  I also expect that such games will be less necessary over time as the Swift library implementation continues to improve.

Right now, this PR still has a flaw in exposing the `JSONEncoder` and the associated `FieldType` bits.  I think I see a way to fix that but it might take me a few days to work through the details...

Tests/SwiftProtobufTests/Test_JSON.swift
tbkka: Yes, they do.  Yes, I do.  Done.

Sources/SwiftProtobuf/TextFormatEncoder.swift
```diff
@@ -86,17 +85,18 @@ internal struct TextFormatEncoder {
...
         data.append(asciiOpenCurlyBracket)
         data.append(asciiNewline)
     }
 
     mutating func endObject() {
-        tabLevel -= 1
-        for _ in 0..<tabLevel {
-            data.append(asciiSpace)
-            data.append(asciiSpace)
+        for _ in 1...tabSize {
```
thomasvl: No clue if it is any better: removeSubrange(_:) or removeLast()?

Sources/SwiftProtobuf/BinaryEncoder.swift
tbkka: Unfortunately, I think typed `UnsafePointer` is subject to alignment problems (it optimizes to a load/store that may have alignment assumptions).  More experimentation is definitely in order here.

Sources/SwiftProtobuf/BinaryEncoder.swift
tbkka: I had thought that all modern processors used the same endianness for floating point, but I dug a little more -- it appears I was wrong about that.

So I expanded this a bit:  It now reads into a correspondingly-sized `UInt`, then swaps to native endianness, then copies to a `Float` or `Double`.

Fixed.

Sources/SwiftProtobuf/JSONEncoder.swift
tbkka: Some interesting numbers from `100 bool`:  The array form here does a JSON encode in about 18us.  Using a literal constant `String` bumps that up to 23us.  But a `StaticString` is about the same as the array form, so I've switched to that.  That also provides a way to use literal constants for other multi-character items as Thomas suggested elsewhere.

Sources/SwiftProtobuf/TextFormatEncoder.swift
tbkka: I've changed a bunch of them to use literal `StaticString`:  It doesn't seem to impact performance much, but it's certainly more legible.

thomasvl:
@tbkka I think something went wrong:

```
/Users/thomasvl/src/github/swift-protobuf/Sources/SwiftProtobuf/JSONEncoder.swift:133:24: error: cannot invoke initializer for type 'Int64' with an argument list of type '(exactly: Double)'
            if let v = Int64(exactly: value) {
                       ^
/Users/thomasvl/src/github/swift-protobuf/Sources/SwiftProtobuf/JSONEncoder.swift:133:24: note: overloads for 'Int64' exist with these partially matching parameter lists: (exactly: UInt8), (exactly: Int8), (exactly: UInt16), (exactly: Int16), (exactly: UInt32), (exactly: Int32), (exactly: UInt64), (exactly: Int64), (exactly: UInt), (exactly: Int)
            if let v = Int64(exactly: value) {
                       ^
/Users/thomasvl/src/github/swift-protobuf/Sources/SwiftProtobuf/TextFormatEncoder.swift:140:24: error: cannot invoke initializer for type 'Int64' with an argument list of type '(exactly: Double)'
            if let v = Int64(exactly: value) {
                       ^
/Users/thomasvl/src/github/swift-protobuf/Sources/SwiftProtobuf/TextFormatEncoder.swift:140:24: note: overloads for 'Int64' exist with these partially matching parameter lists: (exactly: UInt8), (exactly: Int8), (exactly: UInt16), (exactly: Int16), (exactly: UInt32), (exactly: Int32), (exactly: UInt64), (exactly: Int64), (exactly: UInt), (exactly: Int)
            if let v = Int64(exactly: value) {
                       ^
error: exit(1): /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift-build-tool -f /Users/thomasvl/src/github/swift-protobuf/.build/debug.yaml
make: *** [build] Error 1
```

That's with Xcode 8.2.1

tbkka:
Ugh.  Thanks for catching and fixing this.

PR #296, opened 2017-02-22
Unify the naming of the +Extensions files.

thomasvl:
Google_Protobuf_Wrappers+Extensions.swift was the only one using '+', so swap
to '_' to match the other three.


thomasvl:
I'm assuming this is the one we want, if I should flip this and rename the other three, just say so.  😄

thomasvl:
Just realized we also have _Message+UInt8ArrayHelpers.swift_ over in the _Tests_ directory.  So what's the verdict?  '\_' or '+' for these sorta things.  Before this PR, we have 3 files with '\_' and 2 with '+'.

tbkka:
Probably '+' since that would mimic the Obj-C pattern.

thomasvl:
New commit up using '+' in all cases.

PR #301, opened 2017-02-22
Use our init(safely:) for float/double -> Int64 conversions.

thomasvl:
init(exactly:) for double is only in Swift 3.1, so use our helper
so we support building in the current GM Swift.


thomasvl:
@tbkka @allevato do we want to do `#if swift(>=3.0)` around this?

PR #303, opened 2017-02-22
Add partial support to binary serialization support.

thomasvl:
Also add tests to confirm partial and full serialization work as
expected.

### DIFF
--- a/Sources/SwiftProtobuf/BinaryTypeAdditions.swift
+++ b/Sources/SwiftProtobuf/BinaryTypeAdditions.swift
@@ -19,7 +19,18 @@ import Foundation
 /// Messages
 ///
 public extension Message {
-    func serializedData() throws -> Data {
+    /// Serializes the message to the Protocol Buffer binary serialization format.
+    ///
+    /// - Parameters:
+    ///   - partial: The binary serialization format requires all `required` fields
+    ///     be present; when `partial` is `false`, `EncodingError.missingRequiredFields`
+    ///     is throw if any were missing. When `partial` is `true`, then partial
+    ///     messages are allowed, and `Message.isRequired` is not checked.
+    /// - Throws: An instance of `EncodingError` on failure .
+    func serializedData(partial: Bool = false) throws -> Data {
+        if !partial && !isInitialized {
+            throw BinaryEncodingError.missingRequiredFields
+        }
         let requiredSize = try serializedDataSize()
         var data = Data(count: requiredSize)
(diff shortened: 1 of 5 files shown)


Sources/SwiftProtobuf/BinaryTypeAdditions.swift
```diff
@@ -19,7 +19,18 @@ import Foundation
...
-    func serializedData() throws -> Data {
+    /// Serializes the message to the Protocol Buffer binary serialization format.
+    ///
+    /// - Parameters:
+    ///   - partial: The binary serialization format requires all `required` fields
+    ///     be present; when `partial` is `false`, `EncodingError.missingRequiredFields`
+    ///     is throw if any were missing. When `partial` is `true`, then partial
+    ///     messages are allowed, and `Message.isRequired` is not checked.
+    /// - Throws: An instance of `EncodingError` on failure .
+    func serializedData(partial: Bool = false) throws -> Data {
```
tbkka: Should this be different for proto2 vs. proto3?  (Since proto3 doesn't have `required` and never will.)

allevato: A proto3 message could still have a proto2 message nested inside it, so `partial` needs to be supported there as well.

Tests/SwiftProtobufTests/Test_Required.swift
```diff
@@ -167,7 +167,7 @@ class Test_Required: XCTestCase, PBTestHelpers {
     }
```
tbkka: Just noticed this:  We should have a real text format serializer method in addition to the `debugDescription` hook (because people actually do use text format as an interchange format).  That would remove the need to play games just above to massage the expected text into the `debugDescription` format.

Sources/SwiftProtobuf/BinaryTypeAdditions.swift
tbkka: Ah.  Good point.

Tests/SwiftProtobufTests/Test_Required.swift
thomasvl: We do (https://github.com/apple/swift-protobuf/blob/master/Sources/SwiftProtobuf/TextFormatTypeAdditions.swift#L22), it still throws, which have an issue to clean up.  So for the moment I believe the `debugDescription` is a hair easier to deal with.  Once it does't throw, we're probably good to use it in testing asserts.

PR #308, opened 2017-02-23
Allow `visitUnknown()` to throw errors

tbkka:
This corrects an earlier oversight:  The expanded `Visitor` protocol in PR #291 allowed visitors to throw errors for every method except `visitUnknown`.

Correcting this requires `unknownFields.traverse` to be able to throw, which in turn affects the generated code.


Sources/SwiftProtobuf/UnknownStorage.swift
```diff
@@ -25,11 +25,15 @@ public struct UnknownStorage: Equatable {
 
   public init() {}
 
+  public init(data: Data) {
```
thomasvl: Needed?  I was going to say now that append is internal, the only thing that can get into the unknowns should come from within the library, so shouldn't that ensure the data is always valid and there be no need to throw?

tbkka:
> ... now that append is internal ... shouldn't that ensure the data is always valid and there be no need to throw?

The point of this PR is simply consistency.  Either all visitor methods should be allowed to throw or only a small select set should.  Right now, all but one are allowed to throw, and this PR fixes that.

I believe that for our current visitors, encoding failures can only happen within messages (for example, Duration, Timestamp, and FieldMask can all fail to encode in certain circumstances), so it might suffice to only allow certain visitor methods to throw:  message visitors and visitors that may recursively contain messages (which includes groups, extensions, oneofs, and message-valued maps).  But it would take a little while to work through and verify that.

tbkka:
That wasn't a new initializer; it was the old one from before I merged your recent changes.  I merged master and now it's gone.

PR #309, opened 2017-02-23
Remove remaining methods from _StorageClass.

allevato:
`_StorageClass` now only contains stored properties and the `copy` method. The rest of the logic is in the message struct, which dots into storage as needed (using `withExtendedLifetime` to avoid spurious retain/releases).

An interesting thing to note: Decoding groups isn't exercised by our harness, and I imagine it has been slightly slower because `BinaryDecoder` calls `decodeField` directly, which would dip into storage each time. This PR doesn't fix that, because that method can't be wrapped with `withExtendedLifetime` (the whole loop in `decodeMessage` has to be wrapped); it would be worth looking into more deeply later (one way would be to introduce a `decodeGroup` on `Message`, but it sucks having yet another method to deal with), but it seems like a trickier change, and groups are probably infrequent enough that we can punt on it for now.

Also did some cleanup in the message generator in general.


tbkka:
I haven't worried about group performance since groups have been deprecated in proto2 for a long time and they've been entirely dropped from proto3.

Reference/google/protobuf/any_test.pb.swift
```diff
@@ -123,35 +108,42 @@ struct ProtobufUnittest_TestAny: SwiftProtobuf.Proto3Message, SwiftProtobuf._Mes
...
+    try withExtendedLifetime(_storage) { (_storage: _StorageClass) in
+      while let fieldNumber = try decoder.nextFieldNumber() {
+        try decodeField(decoder: &decoder, fieldNumber: fieldNumber)
+      }
+    }
   }
 
   mutating func _protoc_generated_decodeField<D: SwiftProtobuf.Decoder>(decoder: inout D, fieldNumber: Int) throws {
-    try _uniqueStorage().decodeField(decoder: &decoder, fieldNumber: fieldNumber)
+    switch fieldNumber {
```
tbkka: Does the `withExtendedLifetime` in `decodeMessage` cover the `_storage` dereferences here when `decodeField` is called from `decodeMessage`?

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -13,855 +13,550 @@
...
       }
+      p.outdent()
+      p.print("}\n")
+    }
 
-      if isExtensible {
-        ensureFunctionStarted()
-        p.print("if !extensionFieldValues.isInitialized {return false}\n")
+    // ivars
+    if storage != nil {
```
tbkka: You could move this block up and combine it with the case that generates the storage class.  That way, the ivars (in the else clause below) would be in about the same place in the file regardless.

tbkka:
This is very nicely done.

Reference/google/protobuf/any_test.pb.swift
allevato: Yes; there's a comment in the generator describing this (fortunate!) behavior.

Sources/protoc-gen-swift/MessageGenerator.swift
allevato: Done.

Sources/SwiftProtobuf/timestamp.pb.swift
```diff
@@ -119,6 +118,7 @@ public struct Google_Protobuf_Timestamp: SwiftProtobuf.Proto3Message, SwiftProto
   ///   inclusive.
   public var nanos: Int32 = 0
 
+
```
thomasvl: Looks like we picked up a double blank some place.

allevato: Too slow! 😞  (I'll merge in a quick fix.)

Sources/Conformance/test_messages_proto3.pb.swift
```diff
@@ -1673,480 +1150,987 @@ struct ProtobufTestMessages_Proto3_TestAllTypes: SwiftProtobuf.Proto3Message, Sw
...
   }
 
   func _protoc_generated_isEqualTo(other: ProtobufTestMessages_Proto3_TestAllTypes) -> Bool {
-    return _storage === other._storage || _storage.isEqualTo(other: other._storage)
-  }
-
-  private mutating func _uniqueStorage() -> _StorageClass {
-    if !isKnownUniquelyReferenced(&_storage) {
-      _storage = _storage.copy()
+    return withExtendedLifetime((_storage, other._storage)) { (_storage, other_storage) in
```
thomasvl: What about the _storage === other._storage case to short circuit?

allevato: It's no longer possible to short-circuit on storage identity because unknown fields and extensions aren't stored there. So, knowing that two messages have the same storage instance doesn't tell you anything about those, and knowing that two messages have different storage instances doesn't let you assume that the contents aren't different because they could have diverged and then been set back to match.

allevato: Actually, there's still a possible improvement: if we know the storage identities are the same, we don't have to check the individual fields in them. So while we can't short circuit the *whole thing*, we could at least wrap the checks of the `_storage` fields in a conditional that only runs if `_storage !== other_storage`.

The logic will get a little tricker when we start pulling subsets of fields out of storage, though, because we'll have to keep those separate instead of interleaving them.

thomasvl: Yup, but reordering will be a good thing to localize memory access.  I've got some ideas for changes to the generators to help with this, but I was thinking those might happen post 1.0 milestone.

Reference/google/protobuf/any_test.pb.swift
tbkka: @allevato:  I just noticed this:  How are you ensuring proper value semantics here?  You've dropped the `_uniqueStorage()` copy-on-write support for these mutating methods, so it looks like two copies of the same object now risk mutating the shared backing store.

That's clearly not a problem in the case where decoding is occurring during initialization, but now that we have `merge` methods, decoding can occur at any time.

allevato: Good point... that might have gotten left on the floor and wasn't caught because the CoW tests we have didn't have to deal with merging at the time.

I'll write some tests and fix it up. It's going to suck to have to call `_uniqueStorage()` on every `decodeField` call when we know that we really only need to call it around `decodeMessage` in most cases, but groups calling into `decodeField` directly make that not possible. But I guess again, if groups are rare, the performance penalty won't hit most people.

tbkka: We might be able to improve this some by breaking up `decodeField` into a static func that takes the storage reference it's given.  That way, decodeMessage still only pays CoW overhead once, not once per field:
```
mutating func _protoc_generated_decodeMessage(...) {
   _ensureUniqueStorage()
   withExtendedLifetime ... {
      while let fieldNumber = nextFieldNumber {
          MyType._protobuf_generated_static_decodeField(_storage, ...)
      }
   }
}

mutating func _protobuf_generated_decodeField(...) {
    _ensureUniqueStorage()
    withExtendedLifetime {
         MyType._protobuf_generated_static_decodeField(storage: _storage, etc...)
    }
}

private static func _protobug_generated_static_decodeField(storage: _StorageClass, etc...) {
     ... do decodeField work without CoW here...
}
```

The extra hit for groups is unfortunate but I haven't yet come up with a clean way to avoid it.

tbkka: Note:  The idea above would also eliminate the need for `withExtendedLifetime` in these methods, I think?

allevato: If we were to go that route, I think we may as well just write a `decodeGroup` method instead that does the same thing as `decodeMessage` but with a different terminating condition (look for the right end-tag). The amount of increased codegen would be about the same and that would make a little more sense to the reader.

I think we would still need `withExtendedLifetime`; I'm not sure the change you suggested would affect the way reference counts are added/subbed by the parameter passing.

PR #310, opened 2017-02-23
Remove extra blank line; fix equality short circuiting.

allevato:
It turns out that removing the storage identity check in `isEqualTo` was too agressive; while we can't use it to *completely* short circuit the test like we used to (because unknowns and extensions don't live in storage), we can at least guard the individual storage property checks with it so that we don't waste time comparing them if `_storage === other_storage`.


allevato:
There's some additional tweaking we could do here (no need to retain the object if they're the same, so we could move the check outside `withExtendedLifetime`), but that will require changing the lifetime helper a bit. I'll do that in a future PR; it's not urgent.

PR #311, opened 2017-02-23
Use WireFormat enum directly instead of just the rawValue.

tbkka:
rawValue seemed to be more performant at one point, but more
detailed perf testing shows no significant difference.

### DIFF
--- a/Sources/SwiftProtobuf/BinaryDecoder.swift
+++ b/Sources/SwiftProtobuf/BinaryDecoder.swift
@@ -32,7 +32,7 @@ struct BinaryDecoder: Decoder {
     // Whether or not the field value  has actually been parsed
     private var consumed = true
     // Wire format for last-examined field
-    private var fieldWireFormat: UInt8 = WireFormat.varint.rawValue
+    private var fieldWireFormat = WireFormat.varint
     // Field number for last-parsed field tag
     private var fieldNumber: Int = 0
     // Collection of extension fields for this decode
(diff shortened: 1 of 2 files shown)


Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -86,7 +86,11 @@ struct BinaryDecoder: Decoder {
         fieldEndP = nil
         let start = p
         let c0 = start[0]
-        fieldWireFormat = c0 & 7
+        if let wireFormat = WireFormat(rawValue: c0 & 7) {
```
allevato: We have `FieldTag` to isolate this bit twiddling away from the decoders themselves, but we don't have a method specifically for this case since you're working with only the least-significant byte of the tag (which I guess isn't a problem in itself, we can ignore the higher ones for this particular operation, it's just another reason this extremely low-level rewrite bums me out).

It would be worth checking to see if the compiler can optimize `if let wireFormat = FieldTag(rawValue: Int32(c0))?.wireFormat` enough that it boils down to the same operations you have here, at least.

tbkka:
> Does WireFormat need to be public or is it internal to the library now?

Thanks for pointing that out.  It can now be internal.

Sources/SwiftProtobuf/BinaryDecoder.swift
tbkka: I tried using `FieldTag` here to parse out the field number and wire format, but it measured decidedly slower, even after repeated test runs.  I've not dug into the assembly to figure out what's really going on, so it's possible I was just unlucky when I ran the tests.  Certainly worth further investigation.

PR #312, opened 2017-02-23
Underscore prefix the ProtobufMap types.

thomasvl:
They should be considered library details, so give them an underscore
prefix so developer won't depend on them.


tbkka:
These are no more or less internal than the Extension generics or the bare FieldTypes they both utilize.  Were you planning to mark all of those as well?  (I see reasonable arguments both ways, but we should be consistent one way or the other.)

thomasvl:
Yes, I thought that was the intent.  Tag everything public because it is needed by the generated code with the leading underscore if it isn't something we want to developers using a truly _public_ api.

thomasvl:
That's also why I'm chipping away at these in individual CLs, to hopefully finish sorting out what is public but internal vs what is public and really public.  Since we'll be stuck with those decisions for a while.

PR #315, opened 2017-02-23
Text encoder now writes out binary unknown fields

tbkka:
I believe this matches the output of the C++ text formatter.

The text encoder now uses the binary decoder to parse out the unknown data and prints the results.

I ended up refactoring the Text encoder to simplify things a bit.  For example, writing a message field now just brackets the contents with `startMessageField`/`endMessageField` and no longer needs to `startObject`/`endObject` to adjust the indentation.


Tests/SwiftProtobufTests/Test_Text_Unknown.swift
```diff
@@ -0,0 +1,87 @@
+// Tests/SwiftProtobufTests/Test_Text_Unknown.swift - Exercise unknown field text format coding
```
thomasvl: New file, so add it to the Xcode project?

tbkka: Thanks for the reminder!

Fixed.

PR #317, opened 2017-02-24
Make JSONEncoder internal, remove `serializeJSONValues` from `FieldType`

tbkka:
This hides the last piece of internal encoding machinery inside the
library where it belongs.  The only public pieces of the encoding
side now are the various encoding methods on `Message`, the
`Visitor` protocol itself, and the general-purpose reflectors on
`FieldType`.  All of the actual JSON, Binary, Text, and hash visitors
and their format-specific supporting machinery are completely internal.

Because of JSON's distinctive map coding, sealing the JSON encoding
machinery involved creating a separate visitor just for JSON maps.
This visitor accepts alternate keys and values and writes them into
the map structure (unlike the regular JSON encoding visitor which
writes complete key/value fields).

While here, comment the JSON encoder more, simplify some of the JSON
coding idioms, and improve the performance of the JSON bytes encoder.


Sources/SwiftProtobuf/JSONEncoder.swift
```diff
@@ -25,42 +25,38 @@ private let asciiSeven = UInt8(ascii: "7")
...
+    digits.append(contentsOf: asciiUpperA...asciiUpperZ)
+    digits.append(contentsOf: asciiLowerA...asciiLowerZ)
+    digits.append(contentsOf: asciiZero...asciiNine)
+    digits.append(asciiPlus)
+    digits.append(asciiForwardSlash)
+    return digits
+}()
 
-public struct JSONEncoder {
+internal struct JSONEncoder {
```
thomasvl: \o/

Sources/SwiftProtobuf/JSONEncoder.swift
```diff
@@ -155,56 +172,75 @@ public struct JSONEncoder {
...
-        if isMapKey {
-            data.append(asciiDoubleQuote)
-        }
     }
+
+    /// Append a string value escaping special characters as needed.
     internal mutating func putStringValue(value: String) {
-        let hexDigits = [asciiZero, asciiOne, asciiTwo, asciiThree, asciiFour, asciiFive, asciiSix, asciiSeven,
-                         asciiEight, asciiNine, asciiUpperA, asciiUpperB, asciiUpperC, asciiUpperD, asciiUpperE, asciiUpperF];
+        let hexDigits = [asciiZero, asciiOne, asciiTwo, asciiThree, asciiFour,
```
thomasvl: Is it worth doing the same thing you did with base64Digits?

tbkka: Good point.  Done.

Issue #177, opened 2016-12-15
Revisit Text format field and enum naming


tbkka:
The Protobuf Text format encoder/decoder currently relies on some awkward sources of name information:
* Field names are expected to be stored by the generator in full format (in particular, extensions get stored as "[package.scope.field]").  We should be able to economize here.
* Enum names are generated by stripping quotes off the JSON serialization.  I need to double-check whether enum names are always the same between JSON and Protobuf Text.  If so, it would make more sense to have the JSON serialization add the quotes than to have the Protobuf Text serializer remove them.

thomasvl:
Where does this one stand, is it done or getting finished out with your other PRs?

tbkka:
Yes, I'll be cleaning this up along with my other decoder work...

thomasvl:
Ok, tagged with the milestone so we track them all.

tbkka:
This was actually finished a while back.

thomasvl:
@tbkka did you want to do anything about your first bullet?  I did the second (enums), but I didn't look at the first.

tbkka:
Yes, I had hoped to look at that.  Thanks for the reminder; re-opening.

tbkka:
PRs #383 and #384 made some progress here.  The field name for an extension is now a simple StaticString.  If I can push that to be a String, then it can be computed using the same approach as `protoMessageName`.

tbkka:
PR #400 explored sharing the package name, but that turned out to actually make the generated and compiled code larger.

Closing this.

PR #320, opened 2017-02-24
Rework enums to use the same name mapping as message fields.

### LINKED ISSUE #247: Use `FieldNameMap` on `Enum` for textual names
As discussed in PR #243:

https://github.com/apple/swift-protobuf/pull/243/files/e397b41fa359a4aa7f6fec465b4a2d357b661d65#diff-7ed1330a33d576e5f9e1354cfb77e795L23

we should be able to reuse `FieldNameMap` with `Enum` to provide number ⇔ name mapping.  A few notes:
* Maybe rename `FieldNameMap` to just `NameMap`?
* Can we use the exact same `ProtoNameProviding` protocol here?

Among other things, this would work well with a future generator option to omit the textual names.

allevato:
This cuts a decent amount of codegen and makes the access to enum names internal only. Users shouldn't need access to those any more than they need the field names. (They *can* still get at them because the name map is public, of course, but it's not convenient and not meant to be.)

Fixes #247.


Sources/SwiftProtobuf/Enum.swift
```diff
@@ -15,21 +15,48 @@
...
+  public var hashValue: Int {
+    return rawValue
+  }
+
+  /// Internal convenience property representing the name of the enum value (or
+  /// `nil` if it is an `UNRECOGNIZED` value or doesn't provide names).
+  ///
+  /// Since the text format and JSON names are always identical, we don't need
+  /// to distinguish them.
+  internal var name: StaticString? {
```
thomasvl: So what does this mean as far as an enum having a case of "name"?  i.e. - should we prefix these?

allevato: Cases won't collide with instance properties. (For all intents and purposes, a case is a static method that returns an instance of the enum.)

Sources/SwiftProtobuf/NameMap.swift
```diff
@@ -0,0 +1,155 @@
+// Sources/SwiftProtobuf/fieldNameMap.swift - Bidirectional field number/name mapping
```
thomasvl: Update comment for new name.

Looks like git didn't think this came from the other file.  :(

allevato: Oops, find-and-replace gone awry too.

There are significant enough changes within the file to defeat git's crappy rename detection :(

Sources/SwiftProtobuf/type.pb.swift
```diff
@@ -71,39 +76,11 @@ public enum Google_Protobuf_Syntax: SwiftProtobuf.Enum {
...
-      }
-    }
-  }
-
-  public var _protobuf_jsonName: String? {
-    get {
-      switch self {
-      case .proto2: return "SYNTAX_PROTO2"
-      case .proto3: return "SYNTAX_PROTO3"
-      case .UNRECOGNIZED: return nil
```
thomasvl: Did we have a test case that covered this one?

tbkka:
Looks like this also fixes #177?

Sources/protoc-gen-swift/EnumCaseGenerator.swift
```diff
@@ -0,0 +1,56 @@
+// Sources/protoc-gen-swift/EnumGenerator.swift - Enum case logic
```
thomasvl: Update

allevato: And this is why it's not a great idea to include the name/path of a source file in the file...

allevato:
> Looks like this also fixes #177?

I believe it does (in part; @thomasvl already did some of the work, like removing the quotes from the JSON names, IIRC).

Sources/SwiftProtobuf/type.pb.swift
allevato: Looks like there are some in `Test_Enum`.

Reference/conformance/conformance.pb.swift
```diff
@@ -88,43 +94,12 @@ enum Conformance_WireFormat: SwiftProtobuf.Enum {
...
-  init?(jsonName: String) {
-    switch jsonName {
-    case "UNSPECIFIED": self = .unspecified
-    case "PROTOBUF": self = .protobuf
-    case "JSON": self = .json
-    default: return nil
-    }
-  }
-
-  init?(protoName: String) {
```
NachoSoto: @allevato how come this is gone now? I was relying on this method myself. Any chance we can bring it back? Thanks!

allevato: Can you file an issue describing your use case in detail?

Before the 1.0 milestone we're trying to lock down as much API as we can; it's much easier to add things later than it is to remove or change things once users depend on them. Just as we don't expose the field names as strings publicly, we want to discourage people depending on conversions between the raw enum names and the actual Swift values from outside the standard text format and JSON encoding/decoding processes.

bordunosp: condemn. I need string value of enum type

PR #321, opened 2017-02-24
Bump run_count to 1000, add autoreleasepool, support `bool` fields

tbkka:
Three minor improvements that I've been using in my local branch:

* Bump the run_count from 100 to 1000.  SwiftProtobuf is faster now,
  so we need more runs to get useful detail from Instruments.  ;-)
  The higher run_count also brings down the relative stddev, which
  makes it easier to see differences of 10%-20% (with run_count 100,
  you can easily see differences of 2x, but not much less).

* Add an autoreleasepool around the big loop for the Swift version.
  In particular, Data tends to leak autoreleased objects.

* Add support for `bool` fields.  Swift `bool` cannot be initialized
  with a number, so we need a special case to initialize them.

### DIFF
--- a/Performance/Harness.swift
+++ b/Performance/Harness.swift
@@ -29,7 +29,7 @@ class Harness {
 
   /// The number of times to loop the body of the run() method.
   /// Increase this to get better precision...
-  var runCount = 100
+  var runCount = 1000
 
   /// The number of times to call append() for repeated fields.
   let repeatedCount: Int32 = 10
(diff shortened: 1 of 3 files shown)


Performance/perf_runner_swift.sh
```diff
@@ -28,6 +28,9 @@ function print_swift_set_field() {
       echo "          message.field$num.append(Data(repeating:$((num)), count: 20))"
       echo "        }"
       ;;
+    repeated\ bool)
+      echo "        message.field$num = [true, false, true, false, true, false, true, false]"
```
allevato: I would complain that this ignores the value of `repeatedCount` if I actually made smart use of it anywhere. :)

Performance/Harness.swift
```diff
@@ -125,13 +125,15 @@ class Harness {
...
     block: () throws -> Result
   ) rethrows -> Result {
-    taskNames.append(name)
-    let start = Date()
-    let result = try block()
-    let end = Date()
-    let diff = end.timeIntervalSince(start) / Double(runCount) * 1000000.0
-    currentSubtasks[name] = (currentSubtasks[name] ?? 0) + diff
-    return result
+      return try autoreleasepool { () -> Result in
```
allevato: I wonder how much overhead this adds to the overall timings. It probably doesn't matter though, since the only place where we use the overall timings is computing the stddev.

Issue #259, opened 2017-02-15
Avoid naming conflicts


thomasvl:
This somewhat overlaps with #40 (validate reserved words), #12 (revisit public names), and #109 (struct for metadata).

In addition to ensuring the library uses the naming we want and don't steal too many names from the Message field, enum name, etc. namespace; there's another catch.  In cases like proto2 syntax message fields, the get/set for the variable uses the expected field name, but we use a private variable for the storage that gets a leading underscore.  That likely means atleast on message all properties that are `_[a-zA-Z0-9]+` are also usable by the generated code (exact details being specific to the generator logic).  So the generated details where we have simply put an underscore on the front are likely not safe; they all should probably have `_protobuf` or `_protoc` on the front, or we should use a trailing `_` marker instead.

thomasvl:
Some of our functions might also be stealing from the field/enum namespace.  If the function takes no named arguments (`func foo(_ blah: String)`), doesn't that collide with a field of `foo` trying to be added to the message?

thomasvl:
@tbkka what was the original intent for when to use `_protoc_` vs. `_protobuf_` for naming things?  We have a mix, but I'm not sure we've been consistent in the naming.

Especially as we look to move more _private but public_ things to have a leading underscore, I'm thinking some of these actually are exposed to be hooked (as we do in a few WKTs), so they already violate the leading underscore rule.

tbkka:
I used `_protoc_generated_` for functions whose implementation was written by the code generator.  These are proxied by non-prefixed versions in protocols so that people can override them.  (For example, `traverse` is by default just a facade for `_protoc_generated_traverse`.)  In particular, people have specifically asked that it be easy to override `debugDescription`, equality, and hash.

I recently started using `_protobuf_` for _private but public_ implementation details but I'd also be happy to just use a plain underscore in this case.

One odd trivia:  field accessors need to not be prefixed in the generated struct, but I had prefixed them with `_`  in the storage class as that was the easiest way to avoid any naming conflicts there.

thomasvl:
Ok, converging from the other PRs and comments here:

* `_*` for all type (protocol, struct, class) names that is `public` because the generated code need to reference it but we don't want to be a _public_ api folks should rely on.
* `_protobuf_*` for any property or function that is `public` because the generated code need to call it.
* `_protobuf_generated_*` for anything that should be generated.

[...]

PR #324, opened 2017-02-24
Remove more "public" tagging of things.

thomasvl:
This is about the last of the easy pickoffs of visibility.  There is a lot exposed from decoders and field types.  @tbkka do you want to make a double check on both the types and methods within the types to see if anything can be flipped internal? Once that's done, I think I can do a last round of underscore prefixing on the things that have to be public but are internal details.


tbkka:
I'm looking at the decoders and field types now.  I think the field types are going to stay pretty much how they are currently, but I have some ideas for hiding more of the decoding machinery.

thomasvl:
Ok, I'll land this.

My current proposal for my final push on some of the renaming is in https://github.com/apple/swift-protobuf/issues/259.  Once that's agreed I'll also hit those.

PR #327, opened 2017-02-25
Fix copy-on-write behavior for merged messages.

allevato:


allevato:
I realized over the weekend that this is actually wrong—`decodeMessage` extends the lifetime of the old storage but doesn't copy it, and then `decodeField` does the copy, so it won't work the way we expect. Working on the fix.

allevato:
Ok, pushed updates; I've fixed the problem by calling `_uniqueStorage` before extending the lifetime and taking it out of `decodeField`. As such, there should be no performance hit, since `decodeField` is left untouched.

This is safe because the only time `decodeField` is called directly is when decoding groups or when decoding the JSON fields of an `Any`. In both of those cases, it's being done on a newly created message, so we don't have to worry about CoW behavior here.

Other attempts to fix it more cleanly got tangled up in other messy decoder logic, so this will have to do for now.

Reference/google/protobuf/any_test.pb.swift
```diff
@@ -107,6 +107,7 @@ struct ProtobufUnittest_TestAny: SwiftProtobuf.Proto3Message, SwiftProtobuf._Mes
   init() {}
 
   mutating func _protoc_generated_decodeMessage<D: SwiftProtobuf.Decoder>(decoder: inout D) throws {
+    _ = _uniqueStorage()
```
tbkka: FYI:  I've been toying with the idea of changing `_uniqueStorage()` to not return the reference.  It's not 1.0 critical since it's both defined and used purely within generated code (so it's not a library API issue).

Reference/google/protobuf/any_test.pb.swift
```diff
@@ -107,6 +107,7 @@ struct ProtobufUnittest_TestAny: SwiftProtobuf.Proto3Message, SwiftProtobuf._Mes
...
     try withExtendedLifetime(_storage) { (_storage: _StorageClass) in
```
tbkka: After thinking about it over the weekend, I'm not entirely clear on why the `withExtendedLifetime` helps here; I've not come up with any scenario where `_storage` could get released while this method is running.

Reference/google/protobuf/any_test.pb.swift
```diff
@@ -107,6 +107,7 @@ struct ProtobufUnittest_TestAny: SwiftProtobuf.Proto3Message, SwiftProtobuf._Mes
...
       while let fieldNumber = try decoder.nextFieldNumber() {
         try decodeField(decoder: &decoder, fieldNumber: fieldNumber)
```
tbkka: The `decodeField` method documentation should clearly specify that it does not guarantee CoW semantics and is dangerous to use except during initialization.

Reference/google/protobuf/any_test.pb.swift
allevato: That's exactly what I've been thinking, which is what makes it so puzzling that the compiler would generate a retain/release for every access instead of using the ownership information that it should already have.

Reference/google/protobuf/any_test.pb.swift
allevato: Done.

PR #330, opened 2017-02-27
Rename ExtensionFieldValueSet fieldNames to _protobuf_fieldNames.

thomasvl:
Working on prefixing all internal functions with _protobuf to get them out of
the message/enum namespaces and mark them as not really for use by developers.


thomasvl:
@tbkka please confirm, I believe this should be an internal detail like the matching concept is on Messages.

PR #331, opened 2017-02-27
Remove ExtensionFieldValueSet being a Sequence.

thomasvl:


thomasvl:
@tbkka I'm happy to drop this, just figured I'd send it out as part of checking if it really needed to be a sequence.

thomasvl:
The whole `SubSequence` part of `Sequence` is what made me think we likely didn't really want to promise `Sequence`.

PR #332, opened 2017-02-27
Prefix some part of MessageExtensions with _protobuf_.

thomasvl:
Working on marking things that aren't part of the public api
with _protobuf.

### DIFF
--- a/Sources/SwiftProtobuf/MessageExtension.swift
+++ b/Sources/SwiftProtobuf/MessageExtension.swift
@@ -38,7 +38,7 @@ public class MessageExtension<FieldType: ExtensionField, MessageType: Message>: 
         self.messageType = MessageType.self
         self.defaultValue = defaultValue
     }
-    public func set(value: FieldType.ValueType) -> AnyExtensionField {
+    public func _protobuf_set(value: FieldType.ValueType) -> AnyExtensionField {
         var f = FieldType(protobufExtension: self)
         f.value = value
         return f
--- a/Sources/SwiftProtobuf/MessageExtension.swift
+++ b/Sources/SwiftProtobuf/MessageExtension.swift
@@ -32,8 +32,8 @@ public class MessageExtension<FieldType: ExtensionField, MessageType: Message>: 
     public var _protobuf_fieldNames: _NameMap.Names
     public let messageType: Message.Type
     public let defaultValue: FieldType.ValueType
-    public init(fieldNumber: Int, fieldNames: _NameMap.Names, defaultValue: FieldType.ValueType) {
-        self.fieldNumber = fieldNumber
+    public init(_protobuf_fieldNumber: Int, fieldNames: _NameMap.Names, defaultValue: FieldType.ValueType) {
+        self.fieldNumber = _protobuf_fieldNumber
         self._protobuf_fieldNames = fieldNames
         self.messageType = MessageType.self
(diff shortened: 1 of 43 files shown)


Sources/SwiftProtobuf/MessageExtension.swift
```diff
@@ -19,22 +19,22 @@
 /// pieces.
 public protocol MessageExtensionBase {
     var fieldNumber: Int { get }
```
thomasvl: I wasn't completely sure here, I'm guessing we want to expose the `fieldNumber` and `messageType`?

Sources/SwiftProtobuf/MessageExtension.swift
```diff
@@ -19,22 +19,22 @@
...
 
 /// A "Message Extension" relates a particular extension field to
 /// a particular message.  The generic constraints allow
 /// compile-time compatibility checks.
 public class MessageExtension<FieldType: ExtensionField, MessageType: Message>: MessageExtensionBase {
     public let fieldNumber: Int
-    public var fieldNames: _NameMap.Names
+    public var _protobuf_fieldNames: _NameMap.Names
     public let messageType: Message.Type
     public let defaultValue: FieldType.ValueType
```
thomasvl: Should `defaultValue` also get prefixed?

Sources/SwiftProtobuf/MessageExtension.swift
```diff
@@ -19,22 +19,22 @@
...
     public init(fieldNumber: Int, fieldNames: _NameMap.Names, defaultValue: FieldType.ValueType) {
```
thomasvl: I think this is one of the few initializers we have that really is just for the generated code, do we want to do anything with the naming of this?

Sources/SwiftProtobuf/MessageExtension.swift
tbkka: That sounds right.

Sources/SwiftProtobuf/MessageExtension.swift
allevato: The Swift stdlib prefixes the first argument label with `_` to indicate "public but private". You could prepend it with that, or with `_protobuf_` if you wanted to make it even uglier/scarier.

Sources/SwiftProtobuf/MessageExtension.swift
tbkka: I don't think so:  `defaultValue` comes straight from the proto file so I think it's something we can commit to.

Sources/SwiftProtobuf/MessageExtension.swift
tbkka: The convention I've seen elsewhere is to prefix the first argument label:
```
  public init(_protobuf_fieldNumber: Int, fieldNames: ...)
```

thomasvl: Do we do `_protobuf_` on the first arg (`fieldNumber`) or should we put it on `fieldNames` to match the properties in general?

thomasvl: New commit prefixes the first arg to init.

thomasvl:
Added the prefix to the `set()` method also since it also seems like it should be internal.

PR #333, opened 2017-02-27
Ensure map decoding signals fields being consumed

thomasvl:
This isn't meant to currently be a PR, I'm opening it for discussion because I'm a little confused.

### DIFF
--- a/Sources/SwiftProtobuf/BinaryDecoder.swift
+++ b/Sources/SwiftProtobuf/BinaryDecoder.swift
@@ -895,6 +895,7 @@ internal struct BinaryDecoder: Decoder {
 
         if let k = k, let v = v {
             value[k] = v
+            consumed = true
         } else {
             throw BinaryDecodingError.malformedProtobuf
         }
@@ -926,6 +927,7 @@ internal struct BinaryDecoder: Decoder {
 
         if let k = k, let v = v {
             value[k] = v
+            consumed = true
         } else {
             throw BinaryDecodingError.malformedProtobuf
         }
@@ -957,6 +959,7 @@ internal struct BinaryDecoder: Decoder {
 
         if let k = k, let v = v {
             value[k] = v
+            consumed = true
         } else {
             throw BinaryDecodingError.malformedProtobuf
         }
--- a/Tests/SwiftProtobufTests/Test_Map.swift
+++ b/Tests/SwiftProtobufTests/Test_Map.swift
@@ -68,7 +68,13 @@ class Test_Map: XCTestCase, PBTestHelpers {
         assertDecodeSucceeds([10, 4, 8, 1, 16, 2]) {
             $0.mapInt32Int32 == [1: 2]
         }
-        assertDecodeFails([11, 4, 8, 1, 16, 2]) // Bad wire type
+        // TODO: This current doens't fail -
+        // 1. The comment imples it should be a bad wire type, but that doesn't
+        //    appear to be true, it is a field 1 startGroup.
+        // 2. The current known field support seems not to handle startGroups
+        //    correctly in that they don't seem to push everything in until the
+        //    endGroup.
+//        assertDecodeFails([11, 4, 8, 1, 16, 2]) // Bad wire type
     }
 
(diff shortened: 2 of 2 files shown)


Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -895,6 +895,7 @@ internal struct BinaryDecoder: Decoder {
 
         if let k = k, let v = v {
             value[k] = v
+            consumed = true
```
thomasvl: @tbkka When looking at moving unknowns into the base Message, and always haven't them, so Maps tests started to failing, looking a little it seems like maps never say they consumed their input for the might have been ended up in both the unknowns and the real map field.  Does this fix seem correct?

Tests/SwiftProtobufTests/Test_Map.swift
```diff
@@ -68,6 +68,7 @@ class Test_Map: XCTestCase, PBTestHelpers {
         assertDecodeSucceeds([10, 4, 8, 1, 16, 2]) {
             $0.mapInt32Int32 == [1: 2]
         }
+        // TODO: Why should this fail?  It appears to actually be valid.
```
thomasvl: After making the consumed changed, this test starts failing.  The comment says it is an invalid bad wire type, but I think it actually is a field 1 startGroup.  Where it gets interesting is I don't see our unknown handling property dealing with this.  Our generated `messageDecode` is:

```swift
  mutating func _protoc_generated_decodeMessage<D: SwiftProtobuf.Decoder>(decoder: inout D) throws {
    try withExtendedLifetime(_storage) { (_storage: _StorageClass) in
      while let fieldNumber = try decoder.nextFieldNumber() {
        try decodeField(decoder: &decoder, fieldNumber: fieldNumber)
      }
    }
  }
```

But how do groups factor into this.  Isn't this going to say it is field 1, but never really capture that it it was a startGroup and there *has* to be an endGroup marker and all sub data could into this messages unknowns.

Sources/SwiftProtobuf/BinaryDecoder.swift
tbkka: I believe your proposed fix here is correct.

I'm a little dismayed that our tests did not catch this earlier; I should go back and figure out what tests we're missing.

As a minor matter of style, I would put this statement outside of this conditional, and might even pull it up to just after `getFieldBodyBytes` above.  When you put it here, it makes `consumed` look like it's flagging decoding success when it's really intended to just mark that the field has been inspected.

thomasvl: So in addition to the test I tagged in the second commit in this PR putting it any other place than here causes some other things to fail.  I need to go back to the C++ code to figure out what happens if a map is missing the key or value on the wire.  As we're currently written, I don't think we'll parse it, but when I did maps for C++ I almost thing the C++ would end up using the default value.  The really crappy part is for a proto2 syntax file, the map can be to enums, so the value could be non zero making it harder to hardcode.

tbkka: The default initializer for an enum should DTRT for that case.

thomasvl: Right now the code makes both Key and Value optional, so unless we decode them we don't assign into the map.  I'll keep digging for the C++ side.

thomasvl: Ok, found some old email to the protobuf-team confirming the default value part.  I'll look at fixing that up in another PR.

Tests/SwiftProtobufTests/Test_Map.swift
thomasvl: @tbkka I'll update this PR to disable this test for now, but do you recall the intent here and how were groups supposed to be handled with unknowns, I'm I reading this correctly as not handing correctly?

thomasvl: Updated the second commit to disable this assert and opened #334 to make sure we have this right.

Issue #335, opened 2017-02-27
Map parsing handle missing key/values


thomasvl:
From talking to the protobuf-team at Google, if a Map entry is missing the key or value, the default should be used.  The catch here is a proto2 syntax enum can have a non zero value as the default, so that has to pick up the default correctly.

Spun out of #333

PR #336, opened 2017-02-27
Map entry decoding uses the default values if key/value are missing.

thomasvl:
- Add tests to cover missing key/value.

Fixes https://github.com/apple/swift-protobuf/issues/335

### DIFF
--- a/Sources/SwiftProtobuf/BinaryDecoder.swift
+++ b/Sources/SwiftProtobuf/BinaryDecoder.swift
@@ -892,13 +892,13 @@ internal struct BinaryDecoder: Decoder {
         if !subdecoder.complete {
             throw BinaryDecodingError.trailingGarbage
         }
-
-        if let k = k, let v = v {
-            value[k] = v
-            consumed = true
-        } else {
-            throw BinaryDecodingError.malformedProtobuf
-        }
+        // A map<> definition can't provide a default value for the keys/values,
+        // so it is safe to use the proto3 default to get the right
+        // integer/string/bytes. The one catch is a proto2 enum (which can be the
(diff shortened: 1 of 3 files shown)


Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -892,13 +892,13 @@ internal struct BinaryDecoder: Decoder {
...
-            consumed = true
-        } else {
-            throw BinaryDecodingError.malformedProtobuf
-        }
+        // A map<> definition can't provide a default value for the keys/values,
+        // so it is safe to use the proto3 default to get the right
+        // integer/string/bytes. The one catch is a proto2 enum (which can be the
+        // value) can have a non zero value, but that case is the next
+        // custom decodeMapField<>() method and handles it.
+        value[k ?? KeyType.proto3DefaultValue] = v ?? ValueType.proto3DefaultValue
```
thomasvl: The other option is to make k and v non optional, but that ripples through all the MapKey/MapValue types as there is no decodeSingle that takes a non option, can easily be added, but I wasn't sure if it was worth it.

tbkka: Hmmm....  It seems you could use the same idiom you used below:  `ValueType()` for the default value and even `KeyType()` for the default key.

In fact, I think proto3DefaultValue can maybe be eliminated since the proto3 default value is also the Swift default initialized value?

thomasvl: `ValueType()`/`KeyType()` weren't working when it needed to be an int/float/etc.

PR #337, opened 2017-02-28
Make json and text decoders internal

tbkka:
I managed to make the `JSONDecoder` and `TextFormatDecoder` internal to the library by having them perform dynamic type checks on each message instead of having format-specific helpers in the `Message` protocol.  This does add a small run-time penalty, but will make it much easier to evolve the decoders in the future.


Sources/SwiftProtobuf/Message.swift
```diff
@@ -90,16 +94,12 @@ public protocol Message: CustomDebugStringConvertible {
...
-  //
   // google.protobuf.Any support
   //
 
-  // Decode from an `Any` (which might itself have been decoded from JSON,
-  // protobuf, or another `Any`).
+  /// Decode this object from an `Any` (which might itself have been
+  /// decoded from JSON, protobuf, or another `Any`).  This is the
+  /// analog of `unpack` in the Google C++ or Java API.
   init(any: Google_Protobuf_Any) throws
```
thomasvl: Does any of this free up these to move out of Message.swift and into a JSON additions like was done for the other formats?

tbkka: Good idea!  This initializer is now in Google_Protobuf_Any as a general Message extension (but never needs to be overridden).  The encoding is now hidden within Google_Protobuf_Any.

Question:  Should this initializer be `init(unpacking:)` instead of `init(any:)`?  I noticed in updating this that Value also has an `init(any:)` that takes a `Swift.Any` and it seems that might be confusing.

thomasvl: @allevato you know naming better than me…

My guess would be something like `init(unpackingAny:)`.

thomasvl: I guess maybe `init(message:)` should also get names for the reverse what ever is picked?

allevato: I'm down with `init(unpackingAny:)`. 👍 

I don't think we need to worry about confusion with Swift's built-in `Any` then, since the verb and argument type will make it clear.

tbkka: `init(unpackingAny:)` it is, then!

I also filled in a bunch of comments around the Any methods.

PR #340, opened 2017-02-28
Expose UnknownStorage.data as a r/o Data.

thomasvl:
Gives developers a way to tell something is there, even if we
don't really have help for parsing it.

### DIFF
--- a/Sources/SwiftProtobuf/UnknownStorage.swift
+++ b/Sources/SwiftProtobuf/UnknownStorage.swift
@@ -17,7 +17,7 @@
 import Foundation
 
 public struct UnknownStorage: Equatable {
-  internal var data = Data()
+  public private(set) var data = Data()
 
   public static func ==(lhs: UnknownStorage, rhs: UnknownStorage) -> Bool {
     return lhs.data == rhs.data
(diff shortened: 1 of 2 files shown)


Sources/SwiftProtobuf/UnknownStorage.swift
```diff
@@ -17,7 +17,7 @@
 import Foundation
 
 public struct UnknownStorage: Equatable {
-  internal var data = Data()
+  public private(set) var data = Data()
```
allevato: `internal(set)`? (Sorry, didn't realize it was originally internal, not private, when we chatted offline.)

thomasvl: I'm gonna go with private for the moment because we don't actually want to force set the data, it should be always coming in via `append()`

thomasvl:
Update tests where I was looking at unknowns to be much simpler by using this.

tbkka:
This is a good change:  `Data` is the right type to expose here, it makes perfect sense to share a "raw" version of this with our clients, and we do want it to be read-only.

allevato:
> `Data` is the right type to expose here

For now, maybe—but we eventually want a way for users to look at the field numbers and wire-format values individually like the other languages allow. But with the implementation we have now, exposing the data is better than exposing nothing, since building that other API would take longer and I don't think we need it for 1.0.

tbkka:
> ...we eventually want a way for users to look at the field numbers and wire-format values individually like the other languages allow

Absolutely.  Constructing such a view does carry a CPU and memory cost, of course, so I would prefer to make that explicit and not construct such a view until the user actually asks for it.  The `Data` form is efficient and compact for cases where users don't ever need to directly access individual unknown fields.

allevato:
Yep, building the smart view lazily for the user should be trivial.

Issue #342, opened 2017-02-28
Unknown fields include fields with wrong wire type


thomasvl:
Not really covered in the protobuf docs, but you can find it in the code for other languages - If a field number on the wire is the wrong wire type, it should get sent into unknown fields instead of failing the decode.

The C++ is likely easiest to spot this as the decode is done via tag (field number + wire type) -

- Decoding a field number checks the tag - https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.pb.cc#L1006-L1008
- Ends up in the `handle_unusual` case, where `SkipField()` pushes it into unused - https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.pb.cc#L1020-L1028

tbkka:
> ...but you can find it in the code for other languages...

Hmmm...  Based on some comments on the protobuf mailing list, I thought that the Go version also considered mismatched wire type to be a fatal error, though I've not looked carefully.

(Personally, I prefer the error in this case, since I've been bitten a few times by overly-lenient parsers; but I will defer to the group consensus on this.)

thomasvl:
I haven't looked at the go code, but I think the issue will be we need to be able to parse anything parsable on other languages.

Issue #334, opened 2017-02-27
Validate Group parsing


thomasvl:
Spun out of #333.

If the `nextFieldNumber` loop sees a field that is a startGroup, it doesn't seem to ensure everything to the endGroup goes into the unknowns for the message.

The unknown handling needs to deal with groups, if the field number wasn't known, it needs to ensure all the subfields make it into the messages unknowns.  The catch is this can be recursive, there could be another startGroup so it needs to read until both endGroups.

We might want to double check that the normal group parsing ensure the endGroup is found and that the startGroup wire type was seen.  (on quick look, I think the startGroup check is there, just not sure about the endGroup vs. data happening to end)

thomasvl:
Fixed up a few things, last thing on my list is unknownFields within the group are completely dropped at the moment.

tbkka:
There is a `skip` method used for grabbing the body unknown fields that should know how to skip an entire group.  It sounds like it was never fully tested...

thomasvl:
The problem is nothing puts the unknowns on to the group.  I think I've got things sorted out now, going to do some more tests and a perf test to check the impact on things, and hopefully get the PR out early tomorrow.

PR #343, opened 2017-02-28
Map decoding behaviors

thomasvl:
The individual commits should make it easier to see the sub things being addressed.


Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -917,6 +917,12 @@ internal struct BinaryDecoder: Decoder {
                 try KeyType.decodeSingular(value: &k, from: &subdecoder)
             case 2: // Value is a message type
```
tbkka: Please update the comment here (or remove it)

Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -917,6 +917,12 @@ internal struct BinaryDecoder: Decoder {
...
                 try subdecoder.decodeSingularEnumField(value: &v)
+                if v == nil {
```
tbkka: This was simpler than I thought it would be.

I don't suppose anything similar could ever apply to the key here?  I can't come up with any case that would.

thomasvl: Not yet, but when dealing with #342 there could be; but that's a future PR.  I'm almost down doing tests to go with #334, so that will be a PR (likely tomorrow given the time here).

Sources/SwiftProtobuf/BinaryDecoder.swift
thomasvl: Done.

PR #344, opened 2017-02-28
Use more-accurate JSON float and double output when appropriate.

tbkka:
This addresses Issue #222 by fixing the last failing conformance test.

The full logic used for writing a float or double to JSON is now:

* If the number is nan or infinite, write the appropriate string
* If the number is an exact integer, write as an integer (for performance)
* Try using the default string conversion
* Reparse the result just above:
   * If the reparsed result is exactly the same, use it as-is
   * If not, reformat with `DBL_DIG + 2` digits (or `FLT_DIG + 2` for floats)

This is slightly different than the logic used by the C++ implementation:  we short-circuit the integer format for performance reasons, and in our floating-point trial conversion, we just use the default conversion instead of explicitly specifying `DBL_DIG` or `FLT_DIG`.

### DIFF
--- a/Sources/SwiftProtobuf/JSONEncoder.swift
+++ b/Sources/SwiftProtobuf/JSONEncoder.swift
@@ -131,8 +131,33 @@ internal struct JSONEncoder {
     }
 
     /// Append a float value to the output.
+    /// This handles Nan and infinite values by
+    /// writing well-known string values.
     internal mutating func putFloatValue(value: Float) {
-        putDoubleValue(value: Double(value))
+        if value.isNaN {
+            append(staticText: "\"NaN\"")
+        } else if !value.isFinite {
+            if value < 0 {
+                append(staticText: "\"-Infinity\"")
+            } else {
+                append(staticText: "\"Infinity\"")
+            }
+        } else {
+            if let v = Int64(safely: Double(value)) {
+                appendInt(value: v)
+            } else {
+                let s = String(value)
+                let reparsed = Float(s)
+                // If exact match, then default precision is sufficient:
+                if value == reparsed {
+                    append(text: s)
+                } else {
+                    let precision = FLT_DIG + 2
+                    let s = String(format: "%.*g", precision, Double(value))
+                    append(text: s)
+                }
+            }
+        }
     }
 
     /// Append a double value to the output.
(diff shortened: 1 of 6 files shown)


Sources/SwiftProtobuf/JSONEncoder.swift
```diff
@@ -148,13 +173,19 @@ internal struct JSONEncoder {
                 append(staticText: "\"Infinity\"")
             }
         } else {
-            // TODO: Be smarter here about choosing significant digits
-            // See: protoc source has C++ code for this with interesting ideas
             if let v = Int64(safely: value) {
```
thomasvl: If we skip this and just do the lower logic, what gets printed for whole number? Seemed like the C++ code just used the format string no matter what.

Sources/SwiftProtobuf/JSONEncoder.swift
```diff
@@ -148,13 +173,19 @@ internal struct JSONEncoder {
...
             } else {
                 let s = String(value)
-                append(text: s)
+                let reparsed = Double(s)
+                // If exact match, then default precision is sufficient:
+                if value == reparsed {
+                    append(text: s)
+                } else {
+                    let precision = DBL_DIG + 2
+                    let s = String(format: "%.*g", precision, value)
```
thomasvl: Do we need any logic like this in TextFormat and/or the plugin when capturing a custom default for a double (field or extension field)?  (I though the API C++ format/parse code was common to all these places)

Sources/SwiftProtobuf/JSONEncoder.swift
tbkka: With the special case for integers, we get `0` and `1`; without it, we get `0.0` and `1.0`.  Not significant, though the tests would need minor adjustments.

But:  With the special case for integers, `100 double` takes 19µs; without it, it takes 134µs.  For people who happen to be storing integer values in double fields, that's a big difference.

Sources/SwiftProtobuf/JSONEncoder.swift
tbkka: Good point.

I added a bunch of tests to JSON and Text to verify round-tripping of floating-point numbers through those formats.  These tests are able to reproduce the conformance failures for Double (but not for Float for reasons I don't yet understand).

I've updated Text with the same logic as JSON.

I checked the plugin code as well:  The descriptor that protoc sends to the plugin contains a textual form of the default value.  Apart from translating "inf" and "nan", we just pass the literal text straight through to the Swift source code.  So the numeric literal that gets written to the Swift source code is actually formatted by the C++ library code.

Sources/SwiftProtobuf/TextFormatEncoder.swift
```diff
@@ -145,13 +172,19 @@ internal struct TextFormatEncoder {
                 append(staticText: "inf")
             }
         } else {
-            // TODO: Be smarter here about choosing significant digits
-            // See: protoc source has C++ code for this with interesting ideas
             if let v = Int64(safely: value) {
                 appendInt(value: v)
```
thomasvl: Since int64 in JSON use strings, I'm still a little worried that large values here could have trouble depending on what language the receiver is written in.  But I guess we deal with that when the time comes.

Sources/SwiftProtobuf/JSONEncoder.swift
thomasvl: Re float - maybe because you are testing on macOS and in 64 bit so float is a larger data type?  Try the iOS 32bit build where float would be smaller?

PR #345, opened 2017-03-01
Help the compiler pick the right `div()` function

tbkka:
The C standard library exposes (via the `Darwin` module)
another function called `div()` with different return types.
By specifying the return type here, we can ensure the
compiler consistently picks our `div()` function from
MathUtils.swift instead of the Darwin one.


cburrows:
@tbkka ugh this is really unfortunate. Maybe we should rename ours?

tbkka:
Renaming might make this simpler/clearer.

PR #347, opened 2017-03-01
Drop Proto2Message/Proto3Message distinction

thomasvl:


Sources/SwiftProtobuf/BinaryTypeAdditions.swift
```diff
@@ -99,11 +99,9 @@ public extension Message {
             throw BinaryDecodingError.missingRequiredFields
         }
     }
-}
 
-/// Proto2 messages preserve unknown fields
-public extension Proto2Message {
-    public mutating func _protobuf_mergeSerializedBytes(from bytes: UnsafePointer<UInt8>, count: Int, extensions: ExtensionSet?) throws {
+    /// SwiftProtobuf Internal: Common support for decoding.
```
allevato: Technically don't need the first part of this comment anymore, since outsiders won't see it.

thomasvl: I haven't looked at the jazzy docs, does it strip the "internal" stuff?

dpassage: Yes; by default, jazzy only renders docs for symbols which are `public`.

PR #350, opened 2017-03-01
Fold contiguous oneof field number sequences into a range.

allevato:
When decoding, we now fold oneofs with more than two field numbers into ranges: e.g., `case 1, 6, 2, 3, 4, 5` into `case 1...6` (note that out-of-order fields in the original .proto are handled correctly).

Two fields are still written as `case 1, 2`, for no reason other than it saves a character compared to `case 1...2`. The number of comparisons would be the same anyway.


Tests/SwiftProtobufTests/unittest_swift_fieldorder.pb.swift
```diff
@@ -344,7 +344,7 @@ struct Swift_Protobuf_TestFieldOrderings: SwiftProtobuf.Proto2Message, SwiftProt
     case 11: try decoder.decodeSingularStringField(value: &_storage._myString)
     case 1: try decoder.decodeSingularInt64Field(value: &_storage._myInt)
```
thomasvl: Clearly we don't sort the fields in general for these other cases…

PR #353, opened 2017-03-02
Complete group decoding support.

thomasvl:


thomasvl:
There is a small additional check without the loop for field parsing but the cost according to the `perf_runner.sh` seems to be within the range of noise.

The other approach would seem to be to make a GroupBinaryDecoder that just wraps the BinaryDecoder forwarding everything to the wrapped one with the extra checks adding to `nextFieldNumber()`.  That would mean the cost of looking for the `.groupEnd` only happens when parsing fields within a group, but would come at some more code cost since we'd be adding another `Decoder`.

Thoughts?

Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -143,6 +146,11 @@ internal struct BinaryDecoder: Decoder {
         }
         if fieldNumber != 0 {
             consumed = false
+
+            // Reached the end of the current group.
+            if fieldWireFormat == .endGroup, groupFieldNumber == fieldNumber {
```
allevato: Use `&&` instead of comma when the second clause doesn't depend on a binding from the first clause.

thomasvl: Done.  Left over from when i had a let groupFieldNumber = groupFieldNumber.

Issue #265, opened 2017-02-16
Confirm design were Any knowledge leaks all the way into Message


thomasvl:
If you look at the C++, thinks like `type.googleapis.com`, and the type urls don't bleed all the way into the base message class, instead the Any implementation is the only thing that needs to know about these defaults: https://github.com/google/protobuf/blob/39f9b43219bc5718b659ed72a2130a7b2ce66108/src/google/protobuf/any.h#L53-L62.  When filling in an Any the `type.googleapis.com` can be the default or another call can be used to provide a custom value.

Likewise, our `Message` protocol has `serializeAnyJSON`, which shows the knowledge.  That should likely live in over in Any, or if really needed, Any can add it as an extension.

thomasvl:
Took a quick scan of the Java Any support, they seem to have a type registry that comes into play so the things can be looked up and decoded during the JSON parse.  The helpers for `pack()` operations also take the type url with the default.  So this approach seems to keep everything out of the base message and in the helpers instead.

It's likely also worth pointing out, for a field that is an Any, C++ (and it seems Java) don't actually use the generated proto class.  They instead use a stand in with the extra methods and they hook it into serialization/parsings for the different formats as needed.

thomasvl:
All moved out.

PR #356, opened 2017-03-03
Move anyTypePrefix and anyTypeURL off message.

thomasvl:
- Remove the static properties from Message.
- Take a typePrefix to the Any initializer that defaults.
- If a typeURL came into an Any, honor it in all the transforms instead
  of potentially overriding it with values from the Message type.

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Any.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Any.swift
@@ -535,15 +535,18 @@ public struct Google_Protobuf_Any: Message, _MessageImplementationBase, _ProtoNa
     // into an object, then reserializing back to JSON.
     internal func encodedJSONString() throws -> String {
         if let message = _message {
-            // We were initialized from a message object, means typeURL
-            // also was initialized.
+            // We were initialized from a message object.
+
+            // We should have been initialized with a typeURL, but
+            // ensure it wasn't cleared.
+            let url = typeURL ?? buildTypeURL(forMessage: message, typePrefix: defaultTypePrefix)
             if let m = message as? _CustomJSONCodable {
                 // Serialize a Well-known type to JSON:
                 let value = try m.encodedJSONString()
-                return "{\"@type\":\"\(typeURL!)\",\"value\":\(value)}"
+                return "{\"@type\":\"\(url)\",\"value\":\(value)}"
             } else {
                 // Serialize a regular message to JSON:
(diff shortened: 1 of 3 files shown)


thomasvl:
With @tbkka change to move the rest of they Any knowledge out of message, I think this would be the last parts for #265

Sources/SwiftProtobuf/Google_Protobuf_Any.swift
```diff
@@ -545,12 +556,12 @@ public struct Google_Protobuf_Any: Message, _MessageImplementationBase, _ProtoNa
                 if let messageType = Google_Protobuf_Any.wellKnownTypes[messageTypeName] {
                     let m = try messageType.init(unpackingAny: self)
                     let value = try m.jsonString()
-                    return "{\"@type\":\"\(type(of: m).anyTypeURL)\",\"value\":\(value)}"
+                    return "{\"@type\":\"\(typeURL)\",\"value\":\(value)}"
```
thomasvl: This and the one right below are sorta interesting to me, we have a typeURL already set, but we weren't using it in the json generated.

allevato: This looks reasonable to me, but I don't know enough about `Any` to make a good judgment, probably.

Sources/SwiftProtobuf/Google_Protobuf_Any.swift
```diff
@@ -506,11 +516,11 @@ public struct Google_Protobuf_Any: Message, _MessageImplementationBase, _ProtoNa
     /// Traversal-based JSON encoding of a standard message type
     /// This mimics the standard JSON message encoding logic, but adds
     /// the additional `@type` field.
-    private func _serializeAnyJSON(for message: Message) throws -> String {
+    private func _serializeAnyJSON(for message: Message, typeURL: String) throws -> String {
```
tbkka: Looks like we can make this static.

Sources/SwiftProtobuf/Google_Protobuf_Any.swift
tbkka: This looks good to me.

Sources/SwiftProtobuf/Google_Protobuf_Any.swift
thomasvl: Moved it up to be a `fileprivate`.

PR #357, opened 2017-03-04
Improve enum alias handling.

### LINKED ISSUE #354: Handle enum aliases correctly
We currently generate unique cases for every enum value, including aliases. This makes using them difficult, because Swift's exhaustiveness requirement requires the user to list all the aliases for a particular case. It also means they could do something weird, like have different code paths for aliases of the same case.

We can handle aliases by observing that Swift enum cases (without associated values) are essentially static properties of the enum type, and we can alias them by creating our own statics:

```protobuf
enum Foo {
  option allow_alias = true;
  BAR = 0;
  SOME_OLDER_BAR = 0;
}
```
becomes
```swift
enum Foo {
  case bar
  static let someOlderBar = bar
}
```

One side effect of this is that changing a case to an alias is a source-breaking change—the alias can't be used in a `case` clause. This seems to be unavoidable, but it's probably still the right thing to do compared to the alternative.

I also just noticed that I overlooked this when I converted enums to use `NameMap` ([see example](https://github.com/apple/swift-protobuf/blob/master/Tests/SwiftProtobufTests/unittest_proto3.pb.swift#L110-L117)). I'm surprised none of the tests caught this; I can't dig into it until tomorrow but it was either luck, or all the tests that would be affected rely on round-tripping, which would mask this problem.

The name issue will require a slight addition to the `NameMap` logic to support multiple names for a particular number (so that we can parse them all) but have the "primary" one that gets used when going from number to name.

allevato:
Generate `static let` properties for enum aliases instead of separate cases, to avoid requiring original values *and* aliases to be present in exhaustive switches.

Updated `NameMap` to support multiple names for a particular number, and added tests to verify that the primary (first encountered) name is used during encoding text format/JSON and that all names (primary and aliases) are accepted when decoding.

Fixes #354.

### DIFF
--- a/Sources/SwiftProtobuf/NameMap.swift
+++ b/Sources/SwiftProtobuf/NameMap.swift
@@ -66,10 +66,10 @@ public struct _NameMap: ExpressibleByDictionaryLiteral {
     /// other.
     case unique(proto: StaticString, json: StaticString)
 
-    /// Used for enum cases only to represent a value's primary name (the first
-    /// defined case) and its aliases. The JSON and text format names for enums
-    /// are always the same.
-    case aliased(primary: StaticString, aliases: [StaticString])
+    /// Used for enum cases only to represent a value's primary proto name (the
+    /// first defined case) and its aliases. The JSON and text format names for
+    /// enums are always the same.
+    case aliased(proto: StaticString, aliases: [StaticString])
 
     // TODO: Add a case for JSON names that are computable from the proto name
     // using the same algorithm implemented by protoc; for example,
@@ -81,7 +81,7 @@ public struct _NameMap: ExpressibleByDictionaryLiteral {
       switch self {
       case .same(proto: let name): return name
       case .unique(proto: let name, json: _): return name
-      case .aliased(primary: let name, aliases: _): return name
+      case .aliased(proto: let name, aliases: _): return name
       }
     }
 
@@ -90,7 +90,7 @@ public struct _NameMap: ExpressibleByDictionaryLiteral {
       switch self {
       case .same(proto: let name): return name
       case .unique(proto: _, json: let name): return name
-      case .aliased(primary: let name, aliases: _): return name
+      case .aliased(proto: let name, aliases: _): return name
       }
     }
 
@@ -100,7 +100,7 @@ public struct _NameMap: ExpressibleByDictionaryLiteral {
       switch self {
       case .same(proto: let name): return [name]
       case .unique(proto: let name, json: _): return [name]
-      case .aliased(primary: let name, aliases: let aliases):
+      case .aliased(proto: let name, aliases: let aliases):
         var names = [name]
         names.append(contentsOf: aliases)
         return names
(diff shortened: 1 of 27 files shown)


Sources/SwiftProtobuf/NameMap.swift
```diff
@@ -66,6 +66,11 @@ public struct _NameMap: ExpressibleByDictionaryLiteral {
     /// other.
     case unique(proto: StaticString, json: StaticString)
 
+    /// Used for enum cases only to represent a value's primary name (the first
+    /// defined case) and its aliases. The JSON and text format names for enums
+    /// are always the same.
+    case aliased(primary: StaticString, aliases: [StaticString])
```
thomasvl: Should this be aliased(proto:aliases:) to follow the naming of the other?  i.e. - it is a refinement of same where there are aliases?

Sources/protoc-gen-swift/EnumGenerator.swift
```diff
@@ -113,8 +128,8 @@ class EnumGenerator {
     } else {
       p.print("\(visibility)static let _protobuf_nameMap: SwiftProtobuf._NameMap = [\n")
       p.indent()
-      for c in enumCases {
-        p.print("\(c.number): .same(proto: \"\(c.protoName)\"),\n")
+      for c in enumCases where !c.isAlias {
```
thomasvl: For this and the two loops below, does it make sense to use firstCases and sort based on the numbers?

Tests/SwiftProtobufTests/Test_EnumWithAliases.swift
```diff
@@ -0,0 +1,62 @@
+// Tests/SwiftProtobufTests/Test_Enum.swift - Exercise generated enums
```
thomasvl: Update

Tests/SwiftProtobufTests/Test_EnumWithAliases.swift
```diff
@@ -0,0 +1,62 @@
...
+//
+// Copyright (c) 2014 - 2016 Apple Inc. and the project authors
```
thomasvl: 2017

Sources/SwiftProtobuf/NameMap.swift
allevato: Done.

Sources/protoc-gen-swift/EnumGenerator.swift
allevato: Done.

Tests/SwiftProtobufTests/Test_EnumWithAliases.swift
allevato: Done.

Tests/SwiftProtobufTests/Test_EnumWithAliases.swift
allevato: Done.

PR #360, opened 2017-03-06
Add init/getter to Duration for TimeInterval

cburrows:
In Swift, TimeIntervals are durations in seconds, and this adds some
code and tests to work with them from Google_Protobuf_Duration.

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Duration+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Duration+Extensions.swift
@@ -134,6 +134,10 @@ extension Google_Protobuf_Duration: _CustomJSONCodable {
 extension Google_Protobuf_Duration: ExpressibleByFloatLiteral {
     public typealias FloatLiteralType = Double
 
+    /// Returns a `Google_Protobuf_Duration` initialized to have the same
+    /// meaning as the given literal, when interpreted as a duration in seconds.
+    /// The result will be rounded to the nearest nano if the input has
+    /// precision beyond 1e-9.
     public init(floatLiteral value: Double) {
         let sd = trunc(value)
         let nd = round((value - sd) * TimeInterval(nanosPerSecond))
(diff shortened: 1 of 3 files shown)


Sources/SwiftProtobuf/Google_Protobuf_Duration+Extensions.swift
```diff
@@ -130,18 +132,41 @@ extension Google_Protobuf_Duration: _CustomJSONCodable {
     }
 }
 
-
 extension Google_Protobuf_Duration: ExpressibleByFloatLiteral {
     public typealias FloatLiteralType = Double
 
     public init(floatLiteral value: Double) {
+        // TODO: this rounds the nanos part towards zero, not to the nearest
```
cburrows: I actually changed this to match init(timeInterval:) but reverted that in favor of a discussion. Right now my position is that these two things should match, and that the implementation in init(timeInterval:) is correct. If you wind up doing some innocuous looking Double arithmetic and the answer is just epsilon less than 2, I think you want (2, 0) not (1, 999999999).

thomasvl: So if the developer types a TimeInterval of 1.9999999999 will that stay or get rounded to 2.0 also?

cburrows: So you've typed ten 9's. That is an order of magnitude closer to 2.0 than it is to 1.999999999 (with nine 9's), so I am advocating that it should result in (2, 0).

cburrows: And also right now if they ask for the literal conversion they get (1, 999999999) but if they use the TimeInterval init they get (2, 0)--the combination of which is at least very hard to explain.

thomasvl: Ah, now I follow.  Yea, and we likely need to call it out in the docs out the ranges impact the rounding.

Sources/SwiftProtobuf/Google_Protobuf_Duration+Extensions.swift
```diff
@@ -130,15 +132,32 @@ extension Google_Protobuf_Duration: _CustomJSONCodable {
     }
 }
 
-
 extension Google_Protobuf_Duration: ExpressibleByFloatLiteral {
     public typealias FloatLiteralType = Double
 
     public init(floatLiteral value: Double) {
```
thomasvl: Lets go ahead and document this then while you are here.

cburrows: Done.

cburrows:
@thomasvl Rebased too. PTAL?

PR #361, opened 2017-03-07
Update the `Message` documentation.

dpassage:
Updates #236 given the new API layout. `Message` is now fully documented.

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Any.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Any.swift
@@ -83,6 +83,9 @@ public extension Message {
   ///
   /// See `Google_Protobuf_Any.unpackTo()` for more discussion.
   ///
+  /// - Parameter unpackingAny: the message to decode.
+  /// - Throws: an instance of `AnyUnpackError`, `JSONDecodingError`, or
+  ///   `BinaryDecodingError` on failure.
   public init(unpackingAny: Google_Protobuf_Any) throws {
     self.init()
     try unpackingAny.unpackTo(target: &self)
--- a/Sources/SwiftProtobuf/JSONTypeAdditions.swift
+++ b/Sources/SwiftProtobuf/JSONTypeAdditions.swift
@@ -15,9 +15,9 @@
 
 import Foundation
 
-///
-/// Messages
-///
+//
+// Messages
+//
 public extension Message {
     func jsonString() throws -> String {
         var visitor = JSONEncodingVisitor(message: self)
@@ -27,11 +27,21 @@ public extension Message {
         return visitor.stringResult
     }
 
+    /// Creates an instance of the message by deserializing the given
+    /// JSON-format `String`.
+    ///
+    /// - Throws: an instance of `JSONDecodingError` if the JSON cannot be
+    ///   decoded.
     public init(jsonString: String) throws {
         let data = jsonString.data(using: String.Encoding.utf8)!
         try self.init(jsonUTF8Data: data)
     }
 
+    /// Creates an instance of the message by deserializing the given
+    /// `Data` as UTF-8 encoded JSON.
+    ///
+    /// - Throws: an instance of `JSONDecodingError` if the JSON cannot be
+    ///   decoded.
     public init(jsonUTF8Data: Data) throws {
         self.init()
         try jsonUTF8Data.withUnsafeBytes { (bytes:UnsafePointer<UInt8>) in
--- a/Sources/SwiftProtobuf/Message.swift
+++ b/Sources/SwiftProtobuf/Message.swift
@@ -71,7 +71,9 @@ public protocol Message: CustomDebugStringConvertible {
   /// The `Message` will call the Decoder method corresponding
   /// to the declared type of the field.
   ///
-  /// This is the core method used by the deserialization machinery.
+  /// This is the core method used by the deserialization machinery. It is
+  /// `public` to enable users to implement their own encoding formats; it
+  /// should not be called otherwise.
   ///
   /// Note that this is not specific to protobuf encoding; formats that use
   /// textual identifiers translate those to fieldNumbers and then invoke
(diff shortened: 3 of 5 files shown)


Sources/SwiftProtobuf/Google_Protobuf_Any.swift
```diff
@@ -83,6 +83,9 @@ public extension Message {
   ///
   /// See `Google_Protobuf_Any.unpackTo()` for more discussion.
   ///
+  /// - Parameter unpackingAny: the message to decode.
+  /// - Throws: an instance of `AnyUnpackError`, `JSONDecodingError`, or
+  ///   `BinaryDecodingError` on failure.
   public init(unpackingAny: Google_Protobuf_Any) throws {
```
thomasvl: fyi - https://github.com/apple/swift-protobuf/pull/356 could end up changing this.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -67,6 +81,12 @@ public protocol Message: CustomDebugStringConvertible {
   /// semantics for messages with heap storage; it should only be called on
   /// newly-created messages or messages where the storage has been ensured
   /// unique.
+  ///
+  /// - Parameters:
+  ///   - decoder: a `Decoder`; the `Message` will call the method
+  ///     corresponding to the type of this field.
+  ///   - protoFieldNumber: the number of the field to decode.
+  /// - Throws: an error on failure or type mismatch.
   mutating func decodeField<D: Decoder>(decoder: inout D, fieldNumber: Int) throws
```
thomasvl: I believe this is meant to be internal to the library, and I thought @allevato was hoping to maybe even remove it.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -76,18 +96,19 @@ public protocol Message: CustomDebugStringConvertible {
   /// then invokes `decodeField` above.
   mutating func decodeMessage<D: Decoder>(decoder: inout D) throws
```
thomasvl: Same, internal to the library?

Sources/SwiftProtobuf/Message.swift
```diff
@@ -96,15 +117,22 @@ public protocol Message: CustomDebugStringConvertible {
   // JSON encoding/decoding support
   //
 
-  /// Returns a JSON-coded representation of this object as a String.
+  /// Returns a JSON-encoded representation of this object as a `String`.
   func jsonString() throws -> String
```
thomasvl: @tbkka didn't this move out with the other json support?

Sources/SwiftProtobuf/Message.swift
```diff
@@ -142,7 +184,12 @@ public extension Message {
     return result
   }
 
+  /// The literal `type.googleapis.com`; may be overridden if your
+  /// message's type should be encoded differently.
   static var anyTypePrefix: String { return "type.googleapis.com" }
```
thomasvl: These two may also be going away with that PR.

Sources/SwiftProtobuf/Message.swift
thomasvl: I guess not, should we move the overrides of this into _CustomJSONCodable and use that same pattern?

Sources/SwiftProtobuf/Message.swift
allevato: Correct—this method is mostly only internal to the library, and someone should only call it directly if they were implementing their own `Decoder`. The documentation should reflect that.

Sources/SwiftProtobuf/Message.swift
thomasvl: Scratch that, I think this was the one we wanted to keep public, but decodeField was the one we were sorta hopping to kill/making internal/etc.

allevato: Yes, same as above.

Sources/SwiftProtobuf/Message.swift
dpassage: check, adding text to that effect.

Sources/SwiftProtobuf/Google_Protobuf_Any.swift
dpassage: Should we leave its documentation alone, then?

Sources/SwiftProtobuf/Message.swift
dpassage: Should I remove the docs, then?

Sources/SwiftProtobuf/Google_Protobuf_Any.swift
thomasvl: You can update it, and try to merge up/update my PR after it lands.

Sources/SwiftProtobuf/Message.swift
thomasvl: fyi - #362 will move this out.

Sources/SwiftProtobuf/Message.swift
thomasvl: Eh, I guess keep them so @tbkka can agree with my PR.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -54,10 +68,12 @@ public protocol Message: CustomDebugStringConvertible {
   //
 
   /// Decode a field identified by a field number (as given in the .proto file).
-  /// The Message will call the Decoder method corresponding
+  /// The `Message` will call the Decoder method corresponding
   /// to the declared type of the field.
   ///
-  /// This is the core method used by the deserialization machinery.
+  /// This is the core method used by the deserialization machinery. It is
```
thomasvl: I think you want this text also on decodeMessage

Sources/SwiftProtobuf/Message.swift
tbkka: #362 moved `func jsonString()` out of the Message protocol, but it is still part of the public API and still needs to be documented in `JSONTypeAdditions.swift`.

Sources/SwiftProtobuf/JSONTypeAdditions.swift
```diff
@@ -15,9 +15,9 @@
...
 import Foundation
 
-///
-/// Messages
-///
+//
+// Messages
+//
 public extension Message {
     func jsonString() throws -> String {
```
tbkka: This should be public and should be documented.

dpassage:
Thanks, @tbkka !

PR #363, opened 2017-03-08
Linux fixes

tbkka:
This fixes two failures seen with Swift 3.0.2 Release on Ubuntu 16.04.

One appears to be a Swift bug:  attempting to convert an empty string to UTF8 consistently fails with some versions of Swift.

The other provides basic definitions of two C library constants that we use when converting floats and doubles to strings.

### DIFF
--- a/Sources/SwiftProtobuf/MathUtils.swift
+++ b/Sources/SwiftProtobuf/MathUtils.swift
@@ -12,6 +12,12 @@
 ///
 // -----------------------------------------------------------------------------
 
+#if os(Linux)
+// Linux doesn't seem to define these by default.
+internal let FLT_DIG=6
+internal let DBL_DIG=15
+#endif
+
 /// Remainder in standard modular arithmetic (modulo). This coincides with (%)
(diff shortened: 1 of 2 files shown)


Sources/SwiftProtobuf/MathUtils.swift
```diff
@@ -12,6 +12,12 @@
 ///
 // -----------------------------------------------------------------------------
 
+#if os(Linux)
+// Linux doesn't seem to define these by default.
```
thomasvl: Should an issue get opened and linked here for followup?

allevato: I think this would be fixed by importing `Glibc` on Linux?

tbkka: Importing `Glibc` does not fix this.

tbkka: Added a comment linking to my bug report at
  https://bugs.swift.org/browse/SR-4198

PR #366, opened 2017-03-08
Get rid of hand-written Struct, Value, and ListValue.

allevato:
There was some trickiness to get the null special case to work for `Value`, but I think hooking into `CustomJSONCodable` to do it is cleaner than having `JSONDecoder` coupled directly to `Value`.

Another change was to get rid of the standalone JSON decodings of `null` into a message—for example, `Google_Protobuf_Int32Value(jsonString: "null")`. This request doesn't make much sense—aside from the special case for `Google_Protobuf_Value`, `null` is only used to indicate the *absence* of a nested message. So, the only place where `null` should be decoded as a message, like the wrappers, is when it's a value in another dictionary; the tests have been updated to reflect that.

The last change was to fix the behavior of an empty `Google_Protobuf_Value`—since a oneof can be unset, an empty `Value` is not the same as one with kind `nullValue`. The C++ implementation treats an empty `Value` as an error if you try to encode it, so we do the same here.

Part of #13.


Tests/SwiftProtobufTests/Test_Struct.swift
```diff
@@ -288,18 +281,6 @@ class Test_JSON_Value: XCTestCase, PBTestHelpers {
     }
 
     func testValue_list() throws {
-        XCTAssertEqual(Google_Protobuf_ListValue(any: [1]),
```
tbkka: Just to make sure I understand:  It looks like you dropped the initializers that take `Swift.Any`, so you've also dropped these corresponding tests?

Tests/SwiftProtobufTests/Test_Struct.swift
```diff
@@ -171,18 +161,21 @@ class Test_JSON_ListValue: XCTestCase, PBTestHelpers {
...
-        XCTAssertNotEqual(empty, Google_Protobuf_Value(numberValue: 1))
+        do {
+            _ = try empty.jsonString()
+            XCTFail("Encoding should have thrown .missingValue, but it succeeded")
+        } catch let e as JSONEncodingError {
+            XCTAssertEqual(e, JSONEncodingError.missingValue)
+        } catch {
+            XCTFail("Encoding should have thrown .missingValue, but instead it threw: \(error)")
+        }
     }
```
tbkka: Should we test other behaviors of a default `Google_Protobuf_Value()` object?

For example, I presume it should successfully binary encode (to a zero-length value), should successfully text encode (to an empty "{}"), and should compare not-equal to `nil`.

Tests/SwiftProtobufTests/Test_Struct.swift
```diff
@@ -49,7 +49,6 @@ class Test_Struct: XCTestCase, PBTestHelpers {
 
     func test_JSON() {
         assertJSONDecodeSucceeds("{}") {$0.fields == [:]}
-        assertJSONDecodeSucceeds("null") {$0.fields == [:]}
```
tbkka: Should we change this to
```
assertJSONDecodeFails("null")
```
instead of deleting it?

Tests/SwiftProtobufTests/Test_Wrappers.swift
```diff
@@ -58,11 +60,14 @@ class Test_Wrappers: XCTestCase {
     func testFloatValue() throws {
         var m = Google_Protobuf_FloatValue()
         XCTAssertEqual("0", try m.jsonString())
-        XCTAssertEqual(m, try Google_Protobuf_FloatValue(jsonString:"null"))
```
tbkka: If decoding a bare `"null"` should fail, then we should test that here.

Tests/SwiftProtobufTests/Test_Struct.swift
allevato: Done.

Tests/SwiftProtobufTests/Test_Struct.swift
allevato: Done.

Tests/SwiftProtobufTests/Test_Wrappers.swift
allevato: Done for each wrapper.

Tests/SwiftProtobufTests/Test_Struct.swift
allevato: Ah yes, I forgot to mention this. Once I moved the convenience initializers to extensions, I started getting an odd ambiguity error for the `Any` ones. I ended up thinking about them some more and decided that they were probably not worth the potential problems they add—the fact that they reduce type safety and silently ignore values that don't match any of the allowed `Value` kinds seemed like it could cause problems that users would have difficulty tracking down.

allevato: Ah yes, I forgot to mention this. Once I moved the convenience initializers to extensions, I started getting an odd ambiguity error for the `Any` ones. I ended up thinking about them some more and decided that they were probably not worth the potential problems they add—the fact that they reduce type safety and silently ignore values that don't match any of the allowed `Value` kinds seemed like it could cause problems that users would have difficulty tracking down.

PR #367, opened 2017-03-09
Rename metadata properties

tbkka:
This simply adds `_protobuf_` to the old `protoPackageName` and `protoMessageName` properties to avoid name conflicts.


thomasvl:
Since package name and message name are static, doesn't that move them off the field namespace?  i.e. - a instant property with the same name would actually work?  If we do want to prefix them, should they just be `protobuf_` so avoid the namespace problems, but no loading underscore since they are actually meant to be public and not just an impl detail of the library?

tbkka:
Good point.  Closing.

PR #368, opened 2017-03-09
Ensure internal state gets reset.

thomasvl:
During json decode, once the object is started, ensure the internal state is
completely reset even if the decode doesn't complete.


thomasvl:
I didn't see anything immediately having issues here, but just noticed the flow when looking at other parts, so I figured it was probably better to play it safe to ensure if the call flow happened nothing bad happened.

tbkka:
Good catch.  Decoding the empty JSON object `{}` should always result in an empty `Any`.

PR #370, opened 2017-03-10
Ensure generated JSON is always valid.

thomasvl:
Since typeURL can be set by the developer, we need to ensure it
doesn't contain something that could make invalid JSON, so always
use a JSONEncoder so the string can be properly escaped.


tbkka:
Good point.   Should we include test for this case?

thomasvl:
Yup, added new commit covering both paths.

PR #372, opened 2017-03-10
Linux string correctness

tbkka:
This works around a couple of bugs in the String type on Linux:
* `String.data()` crashes for empty strings -- workaround is to test for an empty string first
* `NSString(bytes:length:encoding:)` stops at zero bytes on Linux -- workaround is to explicitly look for a zero byte and use a much slower method when we see one.  (Note that even with the additional check for a zero byte, it's still much faster to use `NSString(bytes:length:encoding:)` when it does work.)

The first bug above seems to be fixed in newer snapshots, but I believe it still appears in the most recent official release build.

The second bug has been reported: https://bugs.swift.org/browse/SR-4216

### DIFF
--- a/Sources/SwiftProtobuf/StringUtils.swift
+++ b/Sources/SwiftProtobuf/StringUtils.swift
@@ -17,7 +17,40 @@ import Foundation
 // Convert a pointer/count that refers to a block of UTF8 bytes
 // in memory into a String.
 // Returns nil if UTF8 is invalid.
+
+// This is painfully slow but seems to work correctly on every platform.
+#if os(Linux)
+fileprivate func slowUtf8ToString(bytes: UnsafePointer<UInt8>, count: Int) -> String? {
+    var s = ""
+    let buffer = UnsafeBufferPointer<UInt8>(start: bytes, count: count)
+    var bytes = buffer.makeIterator()
+    var utf8Decoder = UTF8()
+    while true {
+        switch utf8Decoder.decode(&bytes) {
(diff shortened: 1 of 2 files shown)


tbkka:
This fixes most (but not all) of the failures reported in #260.

Sources/SwiftProtobuf/StringUtils.swift
```diff
@@ -17,7 +17,37 @@ import Foundation
 // Convert a pointer/count that refers to a block of UTF8 bytes
 // in memory into a String.
 // Returns nil if UTF8 is invalid.
+
+// This is painfully slow but seems to work correctly on every platform.
```
thomasvl: Update comment since it is only used for linux?

Sources/SwiftProtobuf/StringUtils.swift
```diff
@@ -17,7 +17,37 @@ import Foundation
...
+    }
+}
+#endif
+
 internal func utf8ToString(bytes: UnsafePointer<UInt8>, count: Int) -> String? {
+    if count == 0 {
+        return ""
+    }
+#if os(Linux)
+    // As of March, 2017, the NSString(bytes:length:encoding:)
```
thomasvl: Bug link to track getting it fixed?

tbkka: Done.

Sources/SwiftProtobuf/StringUtils.swift
tbkka: Done.

PR #374, opened 2017-03-10
Change how Any caches when created from JSON.

thomasvl:
- Store all the content JSON in a Data so it is more compact compared to
  a [String:String].
- When decoding use JSONDecoder.decodeFullObject(), this sends things
  through the plumbing to ensure the Message has uniqueStorage, otherwise
  unpackTo could have been used into something using shared storage.

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Any.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Any.swift
@@ -236,7 +236,7 @@ public struct Google_Protobuf_Any: Message, _MessageImplementationBase, _ProtoNa
                 } catch {
                     return nil
                 }
-            } else if let _ = _jsonFields, let typeURL = typeURL {
+            } else if let _ = _contentJSON, let typeURL = typeURL {
                 // Transcode JSON-to-protobuf by decoding/recoding:
                 // Well-known types are always available:
                 let encodedTypeName = typeName(fromURL: typeURL)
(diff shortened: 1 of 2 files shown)


Sources/SwiftProtobuf/Google_Protobuf_Any.swift
```diff
@@ -491,36 +494,52 @@ public struct Google_Protobuf_Any: Message, _MessageImplementationBase, _ProtoNa
...
-                    if let fieldNumber = fieldNames.number(forJSONName: k) {
-                        let raw = v.data(using: String.Encoding.utf8)!
-                        try raw.withUnsafeBytes { (bytes: UnsafePointer<UInt8>) in
-                            var decoder = JSONDecoder(utf8Pointer: bytes, count: raw.count)
-                            try target.decodeField(decoder: &decoder, fieldNumber: fieldNumber)
-                            if !decoder.scanner.complete {
-                                throw JSONDecodingError.trailingGarbage
-                            }
-                        }
+                let asciiOpenCurlyBracket = UInt8(ascii: "{")
```
tbkka: Interesting.  I would have added the curly brackets when creating `_contentJSON`, but I guess that makes the WKT case a lot more complex?

thomasvl: Yea, that's the tradeoff I wasn't sure of.  We could use a slice of the data to drop those during serialization also.  My thought was it probably comes down to what patterns do we think are common and what should we optimize for.

thomasvl:
Yup, last use of `decodeField`, so I have some work in progress to drop that.  I'll clean it up and get it out next.

PR #378, opened 2017-03-12
Remove decodeField methods and directly generate decodeMessage.

thomasvl:


Sources/SwiftProtobuf/Message.swift
```diff
@@ -60,30 +60,6 @@ public protocol Message: CustomDebugStringConvertible {
...
-  ///
-  /// - Parameters:
-  ///   - decoder: a `Decoder`; the `Message` will call the method
-  ///     corresponding to the type of this field.
-  ///   - protoFieldNumber: the number of the field to decode.
-  /// - Throws: an error on failure or type mismatch.
-  mutating func decodeField<D: Decoder>(decoder: inout D, fieldNumber: Int) throws
-
   /// Decode all of the fields from the given decoder.
   ///
```
tbkka: Please update this comment (it still refers to `decodeField`)

tbkka:
I'm curious:  were you able to measure any performance difference from this?

Sources/SwiftProtobuf/Message.swift
thomasvl: New commit up with updated docs.

thomasvl:
> I'm curious: were you able to measure any performance difference from this?

I'll try to do it this morning before pushing.

Looks like it might be time to update the perf harness to support a "mix" of fields on a message.  And also update the to support skipping C++/Instruments, and display the results of previous run vs. new run.  Since changes like this are more about just how it may or maybe have changed the ObjC speeds and not so much in comparison to the C++.

thomasvl:
On my laptop, perfwise, the difference is in the noise.  Some of the fields seemed faster some slower.  So I'm gonna blame that all on noise in the system.  But the binaries sizes were smaller were slightly smaller in all cases, which makes sense.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -62,12 +62,25 @@ public protocol Message: CustomDebugStringConvertible {
 
   /// Decode all of the fields from the given decoder.
   ///
-  /// This is generally a simple loop that repeatedly gets the next
-  /// field number from `decoder.nextFieldNumber()` and
-  /// then invokes `decodeField` above.
+  /// This is a simple loop that repeatedly gets the next field number
+  /// from `decoder.nextFieldNumber()` and then usees the number returned
```
allevato: s/usees/uses/

Sources/SwiftProtobuf/Message.swift
```diff
@@ -62,12 +62,25 @@ public protocol Message: CustomDebugStringConvertible {
...
+  /// and the type information from the original .proto file to  decide
```
allevato: delete extra space between "to  decide"

Sources/SwiftProtobuf/Message.swift
```diff
@@ -62,12 +62,25 @@ public protocol Message: CustomDebugStringConvertible {
...
+  /// what type of data should be decoded for that field.  The corresponding
+  /// method on the decoder is then called to get the field value.
   ///
-  /// If you're not implementing a custom encoding format, you probably
-  /// shouldn't call this.
+  /// This is the core method used by the deserialization machinery. It is
+  /// `public` to enable users to implement their own encoding formats; it
```
allevato: s/implement their own encoding formats/implement their own encoding formats by conforming to `Decoder`/

Sources/SwiftProtobuf/Message.swift
```diff
@@ -62,12 +62,25 @@ public protocol Message: CustomDebugStringConvertible {
...
+  /// should not be called otherwise.
+  ///
+  /// Note that this is not specific to protobuf encoding; formats that use
```
allevato: s/protobuf encoding/binary encodng/

Sources/SwiftProtobuf/Message.swift
```diff
@@ -62,12 +62,25 @@ public protocol Message: CustomDebugStringConvertible {
...
+  /// textual identifiers translate those to fieldNumbers and also go through
```
allevato: s/fieldNumbers/field numbers/

Sources/SwiftProtobuf/Message.swift
```diff
@@ -62,12 +62,25 @@ public protocol Message: CustomDebugStringConvertible {
...
+  /// this to decode messages.
+  ///
+  /// - Parameters:
+  ///   - decoder: a `Decoder`; the `Message` will call the method
+  ///     corresponding to the type of this field.
+  /// - Throws: an error on failure or type mismatch.  The error types
```
allevato: s/The error types will come from the `Decoder`./The type of error thrown depends on which decoder is used./

Sources/SwiftProtobuf/Message.swift
thomasvl: Done

Sources/SwiftProtobuf/Message.swift
thomasvl: Done

Sources/SwiftProtobuf/Message.swift
thomasvl: Done

Sources/SwiftProtobuf/Message.swift
thomasvl: Done

Issue #379, opened 2017-03-13
Does the Google_Protobuf_Any registry work with modules/frameworks?


thomasvl:
Any project that needs to transform their Any usage between JSON/BinarySerialization will have to registry the mapping. So, given…

- Project A uses some protos as an impl detail, so it registers them.
- Project B uses some protos as an impl detail, so it also registers them.
- App XYZ comes along, and uses Projects A and B.

What happens if there was overlap between the protos in A & B?  Won't Swift modules successfully let both A & B have their own private copies of the message?  Yet how will the registry end up working?  The last one registered will _win_, meaning one library will never successfully be able to do the JSON->Binary conversion because the `Message.Type` registered will be wrong for them. App XYZ is the one that's going to run into this problem, not Project A or project B.

thomasvl:
Tagging this with API change and Milestone 1 since I fear this is an API change to address it.

tbkka:
> Any project that needs to transform their Any usage between JSON/BinarySerialization will have to registry the mapping.

There are two areas to worry about:

* Transcoding.  If you are transcoding, then the registered types are used only to decode/recode the data when you reserialize.  This is fundamentally no different than what C++ does -- it uses the registered descriptors to translate between JSON and binary encodings.  If there is a conflict (different protos with the same name), then it will not behave predictably.

* Packing/unpacking:  This doesn't use the registered types; it uses the type provided by the user at the point of packing/unpacking, so is unaffected by this.  (This should work correctly even if the Any is holding a deserialized object, as the unpack code serializes and deserializes if the provided type is not identical.)

thomasvl:
Transcoding - The C++ is a little different in that they have descriptors that get _globally_ registered via module initializers (load time); and ensure there are no collisions around the message names they use.  So they can't really get into the case where they have >1 message class claiming a given type.  Even if it was the same class, just linked into different shared libs, things blow up just loading the two libraries into the same process space.  Swift's scoping of classes to modules seems to mean the binary can have the class twice because the end up being unique; pushing the problem to our registry.

We could also make registration assert/fail if the a given type is already registered.  Recrafts the problem into something where an App might have to go to their dependencies that share a proto and force them to come up with a way for them to share a module that provides it.

tbkka:
> Swift's scoping of classes to modules seems to mean the binary can have the class twice because the end up being unique; pushing the problem ...

[...]

Issue #376, opened 2017-03-10
Google_Protobuf_Any registry isn't threadsafe


thomasvl:
Access to `knownTypes` needs to be guarded in some way, since `register()` mutates it, if anyone happens to do any action on an Any that causes it to get checked, they could crash.

Asserting all registration should happen early in an app life cycles doesn't really work as someone might load sub modules that used protos on demand, so the app developer doesn't even know the registration is happening.

R/W lock could work, but cheapest might be to just use a dispatch queue and funnel the access via dispatch sync.

PR #380, opened 2017-03-13
Move the registry into an extension with helpers for access.

thomasvl:
This was meant as a first step towards #376 (and to just split things out to be a little easier to find/follow), but in doing it I realized #379, so I guess the factoring out could be useful, but the whole registry might need to be revisited.  ☹️

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
@@ -1,0 +1,89 @@
+// Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift - Registry for JSON support
+//
+// Copyright (c) 2014 - 2017 Apple Inc. and the project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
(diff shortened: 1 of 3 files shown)


Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
```diff
@@ -0,0 +1,100 @@
...
+    ///   Binary format, the Any object will need to complete the
+    ///   deferred deserialization of the JSON object before it
+    ///   can re-encode.
+    ///
+    /// Note that well-known types are pre-registered for you and
+    /// you do not need to register them from your code.
+    ///
+    /// Also note that this is not needed if you only decode and encode
+    /// to and from the same format.
+    @discardableResult static public func register(messageType: Message.Type) {
```
allevato: This doesn't need `@discardableResult` since this function has no result.

thomasvl: Oops, was playing with making it return a bool to cover collisions when I wondered about the problem now in 379.

PR #384, opened 2017-03-15
Simplify extension naming

tbkka:
Another round of improvements to the name lookup handling.

Big achievement:  The details of how name lookup is done for extension fields is now an internal detail shared with the Text format encoder and decoder (which is the only client).  In particular, the old generated `_protobuf_fieldNames(for:)` method is completely gone.

The Extensions themselves now have a simple property `fieldName: StaticString` that's completely untangled from the `NameMap` machinery.  Along the way, this required exposing the `extensionFieldValues` property, which allowed moving more extension boilerplate into the library (off of the generated code), and we lost the last of the name resolver utilities.


Sources/PluginLibrary/descriptor.pb.swift
```diff
@@ -2208,29 +2208,7 @@ public struct Google_Protobuf_FileOptions: SwiftProtobuf.Message, SwiftProtobuf.
...
-    return ext.defaultValue
-  }
-
-  public func hasExtensionValue<F: SwiftProtobuf.ExtensionField>(ext: SwiftProtobuf.MessageExtension<F, Google_Protobuf_FileOptions>) -> Bool {
-    return _extensionFieldValues[ext.fieldNumber] is F
-  }
-  public func _protobuf_names(for number: Int) -> _NameMap.Names? {
-    return Google_Protobuf_FileOptions._protobuf_nameMap.names(for: number) ?? _extensionFieldValues._protobuf_fieldNames(for: number)
-  }
+  public var _extensionFieldValues = SwiftProtobuf.ExtensionFieldValueSet()
```
thomasvl: Did you want `_protobuf_` as the prefix like other things did?

Sources/SwiftProtobuf/ProtoNameResolvers.swift
```diff
@@ -1,33 +0,0 @@
-// Sources/SwiftProtobuf/ProtoNameResolvers.swift - Resolve proto field names
```
thomasvl: Remove it from the Xcode project?

tbkka: Ouch.  Good catch!  Fixed.

Sources/PluginLibrary/descriptor.pb.swift
tbkka: I've added the prefix and regenerated the protos.

PR #386, opened 2017-03-16
Compute json field names

tbkka:
More work on field naming.

This introduces a new NameDescription enum case called `.standard(proto: String)`.  It indicates that the JSON name should be computed from the proto name using "the standard JSON name conversion."  This significantly reduces the number of literal strings in the generated code.  We still use `.same()` when the strings are in fact identical to avoid unnecessary string computation at runtime.

This also refactors the `_NameMap` API somewhat.  The `_NameMap.NameDescription` enum are just dumb descriptions that are processed when the `_NameMap` is initialized into `_NameMap.Names` bundles.  Each `Names` bundle has a `proto` and `json` component that exposes both a `String` and an `UnsafeBufferPointer<UInt8>`.  The latter references stable storage of the UTF8 components.  (If the original name is a StaticString, the UTF8 references that directly; otherwise, the UTF8 is cached in a malloc()-ed block of memory for stability.)

This has no impact on Text performance since the proto names are still coming from StaticString objects.  This doesn't degrade JSON encoding or decoding performance because the cached UTF8 bytes are still available for direct use in both encoding and decoding lookup.

For now, I've sealed a lot of the APIs, mostly just to audit what's really being used outside of the library today:  Only the `_NameMap` and `_NameMap.NameDescription` types are public.  We'll need to selectively open up the `_NameMap` API further to allow third parties to actually use the string information.


Sources/SwiftProtobuf/NameMap.swift
```diff
@@ -8,147 +8,217 @@
 //
 // -----------------------------------------------------------------------------
 
-// Proto names are always ASCII and are always literals, so we can improve the
-// performance of some of our parsing by storing them as pointers to the
-// underlying bytes of a StaticString to avoid the overhead of creating and
-// comparing Unicode Strings.
+import Darwin
```
thomasvl: Needed?

Sources/SwiftProtobuf/NameMap.swift
```diff
@@ -8,147 +8,217 @@
...
-
-/// A simple byte wrapper than can hold (and hash/compare) either a
-/// pointer/count to bytes in memory. We don't worry about retention because the
-/// data is assumed to either (1) come from literal strings in the data segment
-/// or (2) not escape the lifetime of the `String` object from which the data
-/// was taken. This is inherently unsafe; use caution when using this type.
-private struct ASCIIName: Hashable {
+/// This must be exactly the same as the corresponding code in the
+/// protoc-gen-swift code generator.  Changing it will break
+/// compatibility of the library with older generated code.
```
thomasvl: I wonder if this should be a function from the library just to be shared instead of duplicated?

thomasvl:
Was there any change in binary sizes with this?

Sources/SwiftProtobuf/NameMap.swift
tbkka: No longer needed:  I was using malloc/free to manage the memory for caching UTF8 bytes, but I changed it to use UnsafeMutableRawPointer instead.

Sources/SwiftProtobuf/NameMap.swift
tbkka: I'm ambivalent about that.

I see two arguments _against_ sharing a single implementation of `toJsonFieldName`:

1. It's behavior is part of the contract between the library and the generated code; changing it will break compatibility with older library or generated code.  If having one shared instance might tempt someone to assume it's safe to change it, then it might be better to not share it.

2. I don't see a reason for this to be in the library API.

Do you think it should be shared?

tbkka:
> Was there any change in binary sizes with this?

For example, `type.pb.swift.o` went from 427276 (master) to 421112 (this branch).  I presume at least some of that is because 6 field names switched from `.unique(proto:json:)` to `.standard(proto:)`.

Sources/SwiftProtobuf/NameMap.swift
```diff
@@ -43,21 +41,24 @@ private func toJsonFieldName(_ s: String) -> String {
 /// string data.  Track the buffers and release all of those buffers
 /// in case we ever get deallocated.
 fileprivate class InternPool {
-  private var interned = Set<UnsafePointer<UInt8>>()
+  private var interned = [UnsafeBufferPointer<UInt8>]()
 
   func intern(utf8: String.UTF8View) -> UnsafeBufferPointer<UInt8> {
     let count = utf8.count
-    let raw = malloc(count)!
+    let raw = UnsafeMutableRawPointer.allocate(bytes: count, alignedTo: 1)
```
thomasvl: Could some of this be collapsed with https://developer.apple.com/reference/swift/unsafemutablepointer/2295026-allocate?

```swift
let buff = UnsafeMutablePointer<UInt8>.allocate(capacity: count)
_ = buff.initialize(from: utf8)
```

?

tbkka: `UnsafeMutablePointer.initialize()` won't work due to:

> warning: 'initialize(from:)' is deprecated: it will be removed in Swift 4.0.  Please use 'UnsafeMutableBufferPointer.initialize(from:)' instead

but I did manage to simplify it a bit.

Sources/SwiftProtobuf/NameMap.swift
```diff
@@ -8,183 +8,266 @@
...
-  var hashValue: Int {
-    var h = i_2166136261
-    for byte in buffer {
-      h = (h ^ Int(byte)) &* i_16777619
-    }
-    return h
+  func intern(utf8: String.UTF8View) -> UnsafeBufferPointer<UInt8> {
+    let bytePointer = UnsafeMutablePointer<UInt8>.allocate(capacity: utf8.count)
+    let mutable = UnsafeMutableBufferPointer<UInt8>(start: bytePointer, count: utf8.count)
+    _ = mutable.initialize(from: utf8)
```
thomasvl: Oops, it was this PR:

```
Sources/SwiftProtobuf/NameMap.swift:49:9: error: value of type 'UnsafeMutableBufferPointer<UInt8>' has no member 'initialize'
    _ = mutable.initialize(from: utf8)
        ^~~~~~~ ~~~~~~~~~~
```

allevato: Fixed by #394.

PR #387, opened 2017-03-16
Generate Any by using a hand written storage class.

thomasvl:
The long commit chain is to hopefully make it easier to follow the flow of the code moving around instead of a few final comments that gut things in hard to follow ways.

I think there are a few other cleanups that could be done, but I wanted to get the migration out first before changing other parts.


Reference/google/protobuf/any.pb.swift
```diff
@@ -147,38 +155,52 @@ struct Google_Protobuf_Any: SwiftProtobuf.Message, SwiftProtobuf._MessageImpleme
...
+        try visitor.visitSingularBytesField(value: _storage._value, fieldNumber: 2)
+      }
     }
     try unknownFields.traverse(visitor: &visitor)
   }
 
   func _protobuf_generated_isEqualTo(other: Google_Protobuf_Any) -> Bool {
-    if typeURL != other.typeURL {return false}
-    if value != other.value {return false}
+    if _storage !== other._storage {
```
allevato: <s>Why not just `if _storage !== other._storage && !_storage.isEqualTo(other: other._storage) {return false}`?</s> Never mind, now I see the different code paths that the generator can take that produced this.

Shouldn't the other generated protos with storage change slightly as well because of the generator change?

Sources/protoc-gen-swift/MessageStorageClassGenerator.swift
```diff
@@ -107,3 +125,20 @@ class MessageStorageClassGenerator {
     p.print("}\n")
   }
 }
+
+/// Cusotm generator for storage of an google.protobuf.Any.
```
allevato: s/Cusotm/Custom/

Reference/google/protobuf/any.pb.swift
```diff
@@ -125,6 +125,14 @@ struct Google_Protobuf_Any: SwiftProtobuf.Message, SwiftProtobuf._MessageImpleme
     1: .unique(proto: "type_url", json: "typeUrl"),
     2: .same(proto: "value"),
   ]
+  internal var _storage = AnyMessageStorage()
+
+  internal mutating func _uniqueStorage() -> AnyMessageStorage {
```
tbkka: If you added a `typealias _Storage = AnyMessageStorage`, could you use the standard boilerplate for `_uniqueStorage` here instead of customizing it with the new return type?

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -0,0 +1,447 @@
...
+  // state instead to better enforce this; but that also means we could
+  // never got a model were we might also be able to cache the things
+  // we have.
+  var _valueData: Data? = Data()
+  var _message: Message?
+  var _contentJSON: Data?  // Any json parsed from with the @type removed.
+
+  init() {}
+
+  func copy() -> AnyMessageStorage {
```
tbkka: Bit of an aside:  Is there any benefit to using copy initializers instead of a copy method like this?  The method form needs to initialize every field before copying everything over; would a copy initializer avoid that?

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -0,0 +1,447 @@
...
+    if let message = _message {
+      protobuf = try message.serializedData(partial: true)
+    } else if let value = _valueData {
+      protobuf = value
+    }
+    if let protobuf = protobuf {
+      // Decode protobuf from the stored bytes
+      if protobuf.count > 0 {
+        try protobuf.withUnsafeBytes { (p: UnsafePointer<UInt8>) in
+          try target._protobuf_mergeSerializedBytes(from: p, count: protobuf.count, extensions: nil)
```
tbkka: Another aside:  Does C++ have a way to provide extensions to `unpack` in case the Any contains a proto2 that might want to use extensions?

Reference/google/protobuf/any.pb.swift
thomasvl: Nope.  The override is only used on the Any's storage.  The reality is I didn't really have to do this, I could have let the default _p_g_isEqualTo get generated and just hooked in an extension (since we make equality replaceable like that), but it seemed odd to generate a method that is dead code and we always replace.

Sources/protoc-gen-swift/MessageStorageClassGenerator.swift
thomasvl: Fixed (in new commit to keep things followable).

Sources/SwiftProtobuf/Google_Protobuf_Any+AnyUnpackError.swift
```diff
@@ -0,0 +1,42 @@
+// Sources/SwiftProtobuf/Google_Protobuf_Any+AnyUnpackError.swift - Any Unpacking Errors
```
tbkka: Shouldn't this file just be called `AnyUnpackError.swift`?

Sources/SwiftProtobuf/any.pb.swift
```diff
@@ -0,0 +1,207 @@
...
+        case 2: try decoder.decodeSingularBytesField(value: &_storage._value)
+        default: break
+        }
+      }
+    }
+  }
+
+  public func traverse<V: SwiftProtobuf.Visitor>(visitor: inout V) throws {
+    try withExtendedLifetime(_storage) { (_storage: AnyMessageStorage) in
+      try _storage.preTraverse()
```
tbkka: This is very clever.  I like the economy here of having the generator insert just a single error-checking function into the default template.

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -67,6 +68,9 @@ class MessageGenerator {
       swiftRelativeName = sanitizeMessageTypeName(file.swiftPrefix + descriptor.name)
       swiftFullName = swiftRelativeName
     }
+    self.isAnyMessage = (isProto3 &&
+                         descriptor.name == "Any" &&
```
tbkka: Interesting; I would have used `file.package == ".google.protobuf"` as a condition here.  I suppose the filename check does the same thing, though.

Sources/SwiftProtobuf/Google_Protobuf_Any+AnyUnpackError.swift
thomasvl: Done.  (I talked myself out of that originally to try and hint the connection)

Sources/SwiftProtobuf/AnyMessageStorage.swift
thomasvl: C++ has that global registry of all descriptors/extensions, so they never have to pass a registry.  ☹️

Sources/SwiftProtobuf/AnyMessageStorage.swift
thomasvl: Captured in #388

Sources/protoc-gen-swift/MessageGenerator.swift
thomasvl: The name and file check came from the C++ code, but adding package since I have no problem being overly paranoid (the leading `.` isn't there though).

Reference/google/protobuf/any.pb.swift
thomasvl: Nice, done.

Reference/google/protobuf/any.pb.swift
thomasvl: Oh, also hooking it this way lets the unknowns get compared without me having to think about it again.  That's one thing I'll bet folks will forget in their hooks (and extensions).

PR #390, opened 2017-03-17
Performance: Text decoder infrastructure

tbkka:
This switches the Text decoder over to use the same overall design as the JSON decoder, for a significant boost in performance.

For `int64` fields, this now runs as much as 2x faster than Google's C++ implementation.  There are still specific performance issues with particular field types, but I think this shows that the core decoder design -- including the field name lookup and schema-driven decoding -- is fast enough.

### DIFF
--- a/Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
+++ b/Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
@@ -59,6 +59,24 @@ internal struct TextFormatEncodingVisitor: Visitor {
     self.encoder = encoder
   }
 
+  private mutating func startField(lookingUp fieldNumber: Int) throws {
+      // This compact form of startField(lookingUp:) slows the entire
+      // TextFormat encoding process by over 2x.
+      //let buff = try protoFieldName(for: fieldNumber)
+      //startField(name: buff)
+
+      // This manually unrolled version is *much* faster.
+      if let protoName = nameMap?.names(for: fieldNumber)?.proto {
+          encoder.startField(name: protoName.utf8Buffer, inExtension: inExtension)
+      } else if let protoName = nameResolver[fieldNumber] {
+          encoder.startField(name: protoName, inExtension: inExtension)
+      } else if let extensionName = extensions?[fieldNumber]?.protobufExtension.fieldName {
+          encoder.startField(name: extensionName, inExtension: inExtension)
+      } else {
+          throw TextFormatEncodingError.missingFieldNames
+      }
+  }
+
   private func protoFieldName(for number: Int) throws -> UnsafeBufferPointer<UInt8> {
     if let protoName = nameMap?.names(for: number)?.proto {
         return protoName.utf8Buffer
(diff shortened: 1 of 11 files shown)


Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
```diff
@@ -59,6 +59,24 @@ internal struct TextFormatEncodingVisitor: Visitor {
...
+
+      // This manually unrolled version is *much* faster.
+      if let protoName = nameMap?.names(for: fieldNumber)?.proto {
+          encoder.startField(name: protoName.utf8Buffer, inExtension: inExtension)
+      } else if let protoName = nameResolver[fieldNumber] {
+          encoder.startField(name: protoName, inExtension: inExtension)
+      } else if let extensionName = extensions?[fieldNumber]?.protobufExtension.fieldName {
+          encoder.startField(name: extensionName, inExtension: inExtension)
+      } else {
+          throw TextFormatEncodingError.missingFieldNames
```
thomasvl: Since this is encoding, shouldn't we be able to fall back to just doing the field number here?  Or is this a throw because it never should happen?  (i.e. - unknowns are a different path?)

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
```diff
@@ -59,6 +59,24 @@ internal struct TextFormatEncodingVisitor: Visitor {
     self.encoder = encoder
   }
 
+  private mutating func startField(lookingUp fieldNumber: Int) throws {
+      // This compact form of startField(lookingUp:) slows the entire
```
allevato: Delete commented out code?

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
allevato: +1. If it should never happen, make it an assert instead of a throw.

tbkka: 💡!

Even apart from unknown fields, I just realized that text serialization can always succeed even without field names since we can always fall back to the number form.  I'll refactor this a bit to make that work.

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
tbkka: Removed.  (It was there to explain why there were duplicated paths, but I just refactored to eliminate the duplication in a different way.)

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
tbkka: > If it should never happen...

Actually, someday it will be able to happen -- someday, we'll support generating code without name maps.  At that time, it will be possible to have valid decoded messages without field names.

thomasvl:
Looks like this is dependent on the prerelease Swift:

```
Sources/SwiftProtobuf/NameMap.swift:49:9: error: value of type 'UnsafeMutableBufferPointer<UInt8>' has no member 'initialize'
    _ = mutable.initialize(from: utf8)
        ^~~~~~~ ~~~~~~~~~~
```

thomasvl:
Oops, wrong PR.

PR #391, opened 2017-03-18
Major speedup of floating-point handling in JSON/Text

tbkka:
Previously, Float and Double handling for JSON/Text were handled with
the `Double(String)`, `Float(String)`, `String(Float)`, and
`String(Double)` initializers.  This was very slow, with a lot
of time spent in constructing String temporaries and translating those
temporaries to and from UTF8.

To decode a float, for instance, we would convert the UTF8 into
a String, then the Swift Standard Library would convert the string
back into UTF8 in order to use the C library `strtof` function
internally.

This code cuts out the middleman: We pass the UTF8 bytes directly to
the C library functions `strtod` and `vsnprintf` ourselves, avoiding
the intermediate String objects entirely.

For decode, this is a 10x improvement, putting float/double
parsing much more nearly on par with integer handling.

For encode, this is a big (4x) improvement, but we still have
a long way to go.  Invoking `vsnprintf` from Swift requires using
Swift's expensive `CVarArg` facility; profiling shows about 50% of
our total time being spent just to marshal varargs.

Actual results: After modifying the `perf_runner` test to write out
numbers with fractional parts, I now see the following results for
encoding/decoding 100 float fields to/from Text format (JSON results
are similar):

* Before: 600 µs Text format encoding, 260 µs decoding
* After: 140 µs Text format encoding, 22 µs decoding
* C++: 45 µs Text format encoding, 36 µs decoding

For comparison, encoding 100 int64 fields using the SwiftProtobuf text
encoder takes about 12µs on the same hardware.


allevato:
Hoping this doesn't sound too critical—

Making things faster is a good goal, but it feels like we're going even further down the path of "assembly language with Swift syntax" that I mentioned before with these changes. Are we sacrificing readability and maintainability of the runtime library by doing this? Are we going to be able to look at this code in 6 months and understand it easily?

When I made the original JSON performance boost, it was still very much idiomatic Swift—it used the high-level string view APIs, indexes, and so forth—but as time goes on, more and more of the runtime library looks less and less like Swift. The readability and design of the code has taken a distant back seat to squeezing out extra microseconds, and I wonder if we need to rebalance these priorities.

For binary encoding, I'm willing to concede that dropping into unsafe-land to squeeze out performance gains is probably something we need to do, since the whole point of the encoding is to be small and fast. But for text and JSON, if we've reached the point where we're mostly passing around raw UTF-8 pointers and calling into C functions, what's the point of writing it in Swift at all? We could get even better performance by implementing the whole thing in C and wrapping it in a Swift interface. _(Obviously I don't want to do this.)_

Another fear I have is that we're setting the efficiency bar impossibly high, and when the time comes that we want to revisit parts of the runtime design and improve them, we'll hesitate because they might result in a slight runtime increase in the raw numbers, even if those numbers would still be "great" by an objective standard.

We're already matching or beating C++ in text/JSON performance metrics, so is it worth going farther beyond that when we end up paying for it in code readability?

Sources/SwiftProtobuf/DoubleFormatter.swift
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    func doubleToUtf8(_ d: Double) -> UnsafeBufferPointer<UInt8> {
+        return _doubleToUtf8(d, digits: DBL_DIG + 2)
+    }
+
+    private func _doubleToUtf8(_ d: Double, digits: Int32) -> UnsafeBufferPointer<UInt8> {
+        // Split work area into an area to hold the (null-terminated) format string
+        // and an area to receive the formatted text:
+        let fmt = work.prefix(8)
```
thomasvl: Why 8?

tbkka:
> Hoping this doesn't sound too critical...

Although motivated by performance, this particular PR is also a step towards developing UTF-8-oriented support code that I think will make the JSON and Text logic easier to understand.  Specifically, I've created a `DoubleFormatter` that hides details of converting between UTF-8 and Float/Double.  Isolating those details made it feasible to explore using C library functions that happen to be much better suited to UTF-8-oriented operations.  My description above explains this in holistic terms:  How the compiled form now works after everything gets optimized down.  But I think the source is actually clearer than before.

Swift's standard String support is heavily biased towards UTF-16 and has a lot of compromises in its UTF-8 handling right now.  JSON and Text formatting and parsing are inherently UTF-8 based, and it's not yet entirely clear to me how best to express UTF-8 operations in Swift.  Today, we're using a mix of techniques: JSON/Text encoding relies heavily on `Array<UInt8>`, while decoding is using simple pointers a lot.  Two easy improvements I hope to work on soon:  migrating a lot of the simple pointers to buffer pointers, and developing tailored types to wrap key concepts (`NameMap.Name` and `DoubleFormatter` are two steps in that direction).  I expect that future Swift evolution will provide even better tools.

> We could get even better performance by implementing the whole thing in C ... (Obviously I don't want to do this.)

Nor do I.  But I'm not averse to selectively using small bits of C where C APIs happen to be a better fit for the task at hand.  Parsing floats and doubles from UTF-8 is one of those cases; `strtod` is almost exactly the right API for that.  Formatting floats and doubles into UTF-8 may be another, though `vsnprintf` is not the most elegant API around, and the `CVarArgs` tax is higher than I expected.  At least the current PR manages to isolate that operation so we can continue looking for better ways to express it.

> Another fear I have is that we're setting the efficiency bar impossibly high...

My first focus for Swift Protobuf was correctness.  As I learned what that involved, I wrote a lot of ugly code filled with special cases that eventually got refactored into a clearer form that better expressed the real problems at hand.

Now that we have a rich test suite to help protect correctness, I've shifted my attention to performance.  As before, eliminating unnecessary overhead has introduced a few awkward workarounds, but as I better understand what work is really needed (and conversely, what work can be avoided), I plan to continue looking for clearer ways to express the code.  And yes, I do expect some of that refactoring to yield ground on performance.

As long as we do continue to look for ways to improve the clarity of the code, I'm confident we'll find them.  And now that we've managed to hide the decoding and encoding internals within the library, I think we'll have the luxury post-1.0 of continuing to refactor our internal logic in fairly dramatic ways without impacting our clients.

Sources/SwiftProtobuf/DoubleFormatter.swift
tbkka: I reworked things so that the format string is separately allocated and initialized (rather than trying to break up the one large work buffer here).

Hopefully, this makes things clearer.

thomasvl:
Side comment -

One thing I've always sorta liked in the C++ impl is the `CodedInputStream`/`CodeOutputStream` models.  It allows the library to wrap something that writes to buffers, file descriptors, etc. meaning they can read/write directly with files, networking, etc. and not just in memory.

In our quest for performance, we've done all decoding/encoding around having the full payload in memory.  Which while ok for clients, might hurt things from a server usage pov.  I've got some ideas about how to speed out binary serialization, but they would never work in a streaming model, so I'm hesitant to actually put effort in to them.

Anyway, my fear is are we optimizing to the point that if we ever need to support streaming for server usage, we'll have to duplicate lots of code and/or accept what appears to be large performance hits because things were pushed too specific initially.

I haven't looked in detail, but the C++ has a few methods with _NoVirtual_ in the names and have very carefully crafted things so buffer only stuff can still be really fast, but still support the reading/writing external things.  There might be similar things we can do to split out specializations for in memory vs. more generic.

allevato:
> migrating a lot of the simple pointers to buffer pointers

Coincidentally, I have a branch sitting around that does exactly that for the JSON decoder. Unfortunately a couple PRs that have gone in since I started, along with this one, will make merge conflicts numerous 😕 

By moving to buffer pointers, my hope is that we can actually rewrite the decoder to be expressed in terms of `Sequence` instead. (We'd have to eliminate the backtracking that happens as we tokenize, but we're just talking basic DFAs and would end up improving performance slightly anyway.) Since buffer pointers provide that conformance we'd get that for free, but then it would also give us the ability to decode streaming data sources that can also be expressed that way, as @thomasvl mentions.

> Swift's standard String support is heavily biased towards UTF-16 and has a lot of compromises in its UTF-8 handling right now. JSON and Text formatting and parsing are inherently UTF-8 based, and it's not yet entirely clear to me how best to express UTF-8 operations in Swift.

I'm not sure it's any more biased toward or against UTF-16 than UTF-8; the `_StringCore` type used internally is written to efficiently handle either UTF-8 or UTF-16 encoded text and—at least if we start from an empty string and start building an encoding result up—if we never append anything that isn't ASCII, it will never shift to the UTF-16 representation.

As for decoding, that *does* get trickier because it depends where the string comes from. If it's a pure Swift string that contains text-format or JSON, then there's a chance that it's already in the UTF-8-only core representation, but if it's wrapping a Foundation `NSString` that came from, say, a file API, it's much harder to reason about its representation.

Part of the problem is that the UTF-8 and UTF-16 views on `String` aren't mutable, so there's no convenient way to build up a `String` by appending UTF-8 code units. If anything, I would say that `String` is biased toward Unicode scalar representation, because that's the only view on `String` that implements the full set of mutating collection operations.

All that being said, let's go ahead and move forward with this. But let's also keep an eye on opportunities like what @thomasvl mentioned about streaming possibilities; the binary coding could definitely benefit from being able to support that model for encoding and decoding, so figuring out the right way to express it generally in terms of another protocol without assuming the whole thing is in memory and using pointers throughout would be very beneficial. For JSON and text it would also be nice, but I think it's not as critical there.

PR #393, opened 2017-03-20
Add an isA() to Google_Protobuf_Any.

thomasvl:
Allows developers to quickly check the type of an Any, so they can
see if it something they actually want to unpack and inspect.


tbkka:
Very elegant.

PR #395, opened 2017-03-20
Turn unpackTo() as being a reset and add extension support.

thomasvl:
- Document that it is a reset and not a merge.
- Update code paths to always be a reset instead of a mix of resets and merges.
- Support Extensions on unpackTo().

### DIFF
--- a/Sources/SwiftProtobuf/AnyMessageStorage.swift
+++ b/Sources/SwiftProtobuf/AnyMessageStorage.swift
@@ -101,30 +101,32 @@ internal class AnyMessageStorage {
     return encodedType == M.protoMessageName
   }
 
-  func unpackTo<M: Message>(target: inout M) throws {
+  func unpackTo<M: Message>(target: inout M, extensions: ExtensionSet?) throws {
     guard isA(M.self) else {
       throw AnyUnpackError.typeMismatch
     }
-    var protobuf: Data?
+
+    // Cached message is correct type, copy it over.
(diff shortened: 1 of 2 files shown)


tbkka:
I agree that the current behavior is a reset and I think that reset is probably the right behavior.  But it is an odd function signature for a reset method.

I wonder if we should just make this method `internal` and only support `init(unpackingAny:)` as the public API?

thomasvl:
> I wonder if we should just make this method internal and only support init(unpackingAny:) as the public API?

I'm fine with that, I was also debating if we wanted to have it return the message instead (`unpack<M: Message>() -> M`)

tbkka:
I could see a method that returns the unpacked message.  However, I try to avoid methods that are generic only on the return type, as I find them awkward to use.  Of course, the fix for that would be something like `unpack<M: Message>(as: M.Self) -> M` which is also a little odd.

Hmmm....  I think I'm inclined to either leave it as-is or make it `internal`.

allevato:
I agree that methods generic only on their return type are awkward to use and should generally be avoided.

If we fear that users might be confused that `unpackTo` does a merge because of the signature, an initializer would be the cleanest way to go—it gets around the generic return type design issue while making it clear that what you get back is a new value.

thomasvl:
New commit up dropping unpackTo and just using the initializer.

tbkka:
Looks good.

allevato:
👍

Issue #359, opened 2017-03-06
Investigate Any behaviors for consistency


thomasvl:
Capturing this so we look to double check the behaviors, at quick glance some parts seem like they aren't right -

- `hashValue` looks like it will come up with different values if an `Any` was initialized with a message vs. json vs. binary, even if the _message_ is the same in all cases.
- `_protobuf_generated_isEqualTo` looks like has a partial impl, but it needs to be consistent with `hashValue`.
- `_protobuf_generated_decodeField` seems like it is directly accessing `_value`, which may mean we don't output binary if the object was created from JSON input.
- The comments in a few places talk about delayed decoding of JSON to avoid doing it if the output is going to be JSON.  But when serialized binary is desired, because we travers once to collect sizes, we might actually do the JSON decode twice to generate the serialized form.

tbkka:
Re: `hashValue` and `isEqualTo` -- I'm not convinced there is a completely robust solution here.  For example, if you decode one object from binary, another from JSON, and don't have the message available, then I don't think you can determine whether the objects actually represent the same value.  (Honestly, I suspect the best answer might be for `Any` to be neither `Equatable` nor `Hashable` and for any-valued fields to be completely omitted from equality tests and hash calculations.)

I'm not sure I follow your concern about `decodeField`.  Any objects override `decodeJSON` and `traverse()` forces the deferred decode as required, so the specific concern you have here should never arise.

I think you are right about the double-decode for JSON-to-binary transcoding, but I'm not sure how important it is.  I've long assumed that transcoding is actually the least common case, so we should optimize first for encode/decode to the object form and then for same-format recoding.  In both cases, the deferred decoding should be the best approach.

Another question to consider:  Should we defer TextFormat decoding as well?

thomasvl:
Re leaving Any out of Equality/Hash - I'm a little worried there since I would fully expect someone to use Any within other messages (since it is the proto3 syntax replacement for Extensions).  So not having it in quality would greatly limit the value of Equality on messages in general.

Re decodeField - yea, sorry, fired this off to quickly, I was thinking about encoding, not decoding; so there isn't an issue here.

[...]

PR #397, opened 2017-03-20
Tweak/Document Any's hashValue and equality support.

thomasvl:
- typeURL seems to be the only safe thing to use in hashValue.
- Update docs for equality support, but it sorta seems like with current
  types we might not be able to do it unless we go down the binary path
  only.

### DIFF
--- a/Sources/SwiftProtobuf/AnyMessageStorage.swift
+++ b/Sources/SwiftProtobuf/AnyMessageStorage.swift
@@ -331,6 +331,16 @@ extension AnyMessageStorage {
       return true
     }
 
+    // If both have contentJSON, and they exactly match; the messages are equal.
+    // Because there could be map in the message (or the JSON could just be in a different
+    // order), the fact that the JSON isn't the same doesn't always mean the messages
+    // aren't equal.
+    if let myJSON = _contentJSON, let otherJSON = other._contentJSON, myJSON == otherJSON {
+      return true
+    }
+
+    // Out of options; to do more compares, the states conversions would have to be
+    // done to do comparisions.  Give up and say they aren't equal.
     return false
   }
 }
--- a/Sources/SwiftProtobuf/AnyMessageStorage.swift
+++ b/Sources/SwiftProtobuf/AnyMessageStorage.swift
@@ -284,6 +284,18 @@ extension AnyMessageStorage {
   }
 }
 
+/// The obvious goal for Hashable/Equatable conformance would be for
+/// hash and equality to behave as if we always decoded the inner
+/// object and hashed or compared that.  Unfortunately, Any typically
+/// stores serialized contents and we don't always have the ability to
+/// deserialize it.  Since none of our supported serializations are
+/// fully deterministic, we can't even ensure that equality will
+/// behave this way when the Any contents are in the same
+/// serialization.
+///
+/// As a result, we can only really perform a "best effort" equality
+/// test.  Of course, regardless of the above, we must guarantee that
+/// hashValue is compatible with equality.
 extension AnyMessageStorage {
   var hashValue: Int {
     var hash: Int = i_2166136261
@@ -326,7 +338,9 @@ extension AnyMessageStorage {
 
     // If both have serialized data, and they exactly match; the messages are equal.
     // Because there could be map in the message, the fact that the data isn't the
-    // same doesn't always mean the messages aren't equal.
+    // same doesn't always mean the messages aren't equal. Likewise, the binary could
+    // have been created by a library that doesn't order the fields, or the binary was
+    // created using the appending ability in of the binary format.
     if let myValue = _valueData, let otherValue = other._valueData, myValue == otherValue {
       return true
     }
--- a/Sources/SwiftProtobuf/AnyMessageStorage.swift
+++ b/Sources/SwiftProtobuf/AnyMessageStorage.swift
@@ -15,6 +15,9 @@
 
 import Foundation
 
+private let i_2166136261 = Int(bitPattern: 2166136261)
+private let i_16777619 = Int(16777619)
+
 fileprivate func serializeAnyJSON(for message: Message, typeURL: String) throws -> String {
   var visitor = try JSONEncodingVisitor(message: message)
   visitor.startObject()
@@ -283,14 +286,18 @@ extension AnyMessageStorage {
 
 extension AnyMessageStorage {
   var hashValue: Int {
-    var hash: Int = 0
-    hash = (hash &* 16777619) ^ _typeURL.hashValue
-    if let v = _valueData {
-      hash = (hash &* 16777619) ^ v.hashValue
+    var hash: Int = i_2166136261
+    if !_typeURL.isEmpty {
+      hash = (hash &* i_16777619) ^ _typeURL.hashValue
     }
-    if let m = _message {
-      hash = (hash &* 16777619) ^ m.hashValue
-    }
+    // Can't use _valueData for a few reasons:
+    // 1. Since decode is done on demand, two objects could be equal
+    //    but created differently (one from JSON, one for Message, etc.),
+    //    and the hashes have to be equal even if we don't have data yet.
+    // 2. map<> serialization order is undefined. At the time of writing
+    //    the Swift, Objective-C, and Go runtimes all tend to have random
+    //    orders, so the messages could be identical, but in binary form
+    //    they could differ.
     return hash
   }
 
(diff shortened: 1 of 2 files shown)


thomasvl:
This PR is also for some discussion around how to deal with Any's.  Since proto3 syntax pushes them as a replacement for extensions, I sorta worry about not getting equality correct causing problems for folks.  The usage of Any on [github seems pretty high](https://github.com/search?utf8=%E2%9C%93&q=%22google.protobuf.Any%22+extension%3Aproto).

tbkka:
"Getting equality correct" is problematic even if you do limit to "the binary path only."

Without deterministic serialization, binary comparison can only guarantee that objects test equal if they were deserialized from the same original message.  And we can do that much today without the binary-only restriction.  And deterministic serialization isn't something you can rely on, as @acozzette explained here:
       https://github.com/google/protobuf/issues/2830

The binary-only restriction would also mean that Any cannot be used for JSON if you don't have the schema. The ongoing debate about proto3 unknown fields shows that people really do want to support pass-through use cases.

Rather than breaking the JSON pass-through support for Any in order to try to get equality that can never work 100% anyway, I would rather push for an option that allowed reading/writing Any to JSON by serializing the binary-coded value as a bytes field (without using the specialized JSON Any format).  That would support JSON/binary transcoding and JSON pass-through for Any without requiring schema for every Any object, all while still keeping the Any contents in binary form to improve the equality behavior.  I suspect such a behavior would have to be optional to continue supporting JSON-only implementations.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -259,27 +266,35 @@ extension AnyMessageStorage {
...
+    // removes issues like map<> serialization order, some other protocol buffer
+    // implementation details/bugs around serialized form order, etc.; but that
+    // would also greatly slow down equality tests.
+    //
+    // Despite the problems with binary serialization, that may be the best bet
+    // as looking at most other language implementations around Any they don't
+    // do lazy decode and instead keep all the `value` fields in binary form and
+    // end up just comparing those binary fields when comparing Anys.
+
+    // If both have messages, check if they are the same.
```
thomasvl: btw - any ideas how we might be able to make this work?  So far the only thing I've come up with is to push _p_g_isEqualtTo(other:) all the way back to Message (or something on Message that calls it) make it do the check the types also.

tbkka: That would work, though I'm not thrilled about adding an extra type check to every equality test.

We might be able to drill down to the right `==` implementation using generics.  I think it requires two steps to first get the `_MessageImplementationBase` types for one of the arguments, then verify the type of the other.  Maybe something like this?
```
private func testEqual(lhs: Message, rhs: Message) -> Bool {
   if let lhs = lhs as? _MessageImplementationBase {
      return testEqualMessages(lhs: lhs, rhs: rhs)
   }
}

private func testEqualMessages<M: _MessageImplementationBase>(lhs: M, rhs: Message) -> Bool {
   if let rhs = rhs as? M {
      return lhs == rhs
   }
   return false
}
```

tbkka: This seems to work.   It adds the heterogenous equality as a separate method on `Message`.  This avoids adding the type check to `==` but at the cost of adding another method to the `Message` protocol:
```
diff --git a/Sources/SwiftProtobuf/Message.swift b/Sources/SwiftProtobuf/Message.swift
index 518820a..b58df2e 100644
--- a/Sources/SwiftProtobuf/Message.swift
+++ b/Sources/SwiftProtobuf/Message.swift
@@ -112,8 +112,10 @@ public protocol Message: CustomDebugStringConvertible {
   /// debugging, for conformance with the `CustomDebugStringConvertible`
   /// protocol.
   var debugDescription: String { get }
+
+  func isEqualTo(message: Message) -> Bool
 }
 
 public extension Message {
 
   /// If the generated code needs to provide its own implementation, usually
@@ -197,6 +199,13 @@ public extension Message {
 
 public extension _MessageImplementationBase {
+  public func isEqualTo(message: Message) -> Bool {
+    if let other = message as? Self {
+      return self == other
+    }
+    return false
+  }
+
   public static func ==(lhs: Self, rhs: Self) -> Bool {
     return lhs._protobuf_generated_isEqualTo(other: rhs)
   }
 }

```

thomasvl: Yea, adding something to Message was all I could come up with.  Added this in the new commits.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -236,31 +243,41 @@ extension AnyMessageStorage {
...
+    if let myMsg = _message, let otherMsg = other._message, type(of: myMsg) == type(of: otherMsg) {
+      // Since the messages are known to be same type, we can claim both equal and
+      // not equal based on the equality comparison.
+      return myMsg.isEqualTo(message: otherMsg)
+    }
+
+    // If both have serialized data, and they exactly match; the messages are equal.
+    // Because there could be map in the message, the fact that the data isn't the
+    // same doesn't always mean the messages aren't equal.
```
tbkka: There are more issues with binary serialization than just `map`.  Protobuf recommends (but does not require) fields be in numeric order.  Unknown fields are routinely re-ordered as they get decoded/recoded.  Even repeated array fields can get reordered in some cases (the unknown enum problem).  The same message could get decoded/recoded by different systems and end up with very different binary forms for any of these reasons.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -236,31 +243,41 @@ extension AnyMessageStorage {
...
     if let myValue = _valueData, let otherValue = other._valueData, myValue == otherValue {
       return true
     }
 
+    // If both have contentJSON, and they exactly match; the messages are equal.
+    // Because there could be map in the message (or the JSON could just be in a different
+    // order), the fact that the JSON isn't the same doesn't always mean the messages
+    // aren't equal.
+    if let myJSON = _contentJSON, let otherJSON = other._contentJSON, myJSON == otherJSON {
```
tbkka: A more robust approach would do an anonymous decode (using `Google_Protobuf_Value` or even Foundation's `JSONSerialization`) and compare those.  That would protect against field ordering, numeric formatting, and other insignificant variations in the JSON text.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -236,31 +243,41 @@ extension AnyMessageStorage {
...
+      return true
+    }
+
+    // Out of options; to do more compares, the states conversions would have to be
+    // done to do comparisions.  Give up and say they aren't equal.
```
tbkka: I see one more possibility here:  If the `typeURL` is for a registered type, we could decode both objects and compare them.  The code to do that is already in this file, but it's a little scattered around.  I would suggest landing this as-is, then coming back to refactor the code for that in a subsequent PR.

tbkka:
This looks good to me.  At least, I think you have the right structure here now.  I would go ahead and merge this as-is, then consider whether you want to tackle further improvements in this area now or open issues to track them for later work.

Sources/SwiftProtobuf/AnyMessageStorage.swift
thomasvl: Yup, stole your comment for this whole extension from the other PR and added a little more here and in hashValue.

Sources/SwiftProtobuf/AnyMessageStorage.swift
thomasvl: Yea, we can look at that in follow up, but I'm also hesitant to but too much work in here because there's always a chance someone is crazy and uses the message as a dictionary key or stores them in a set, so the on demand conversion would suddenly get very costly.

Sources/SwiftProtobuf/AnyMessageStorage.swift
thomasvl: Same comment as above, making equality that _expensive_ worries me a little since it could be used in places where it is somewhat hidden/removed from the developer.

tbkka: > ... making equality that expensive worries me ...

Yeah.  That's a good point.  I suspect this area will require some experimenting.  For example, maybe our decoders are fast enough that it's not an issue?  Or alternatively, we might be able to cache the conversions?  I've also considered tinkering to see when/if it might be worth eagerly decoding to the message form sometimes; that would also affect the economics here.

thomasvl: Seem my [comment/question on the issue](https://github.com/apple/swift-protobuf/issues/359#issuecomment-288485802).  I'm wondering if an API on Any would help here, as could provide a way for someone to put their uses on Any into a state where things could more reliable work.

PR #398, opened 2017-03-20
Use a DisptachQueue to make the registry thread safe.

thomasvl:
Fixes https://github.com/apple/swift-protobuf/issues/376

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
@@ -47,6 +47,9 @@ internal func typeName(fromURL s: String) -> String {
   return s[typeStart..<s.endIndex]
 }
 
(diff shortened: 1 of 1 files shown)


Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
```diff
@@ -47,6 +47,9 @@ internal func typeName(fromURL s: String) -> String {
   return s[typeStart..<s.endIndex]
 }
 
+fileprivate var serialQueue = DispatchQueue.init(label: "com.apple.swift.typeRegistry")
```
allevato: Drop the `.init`; just `DispatchQueue(label: ...)`. It's cleaner.

thomasvl: Doh, done.

PR #400, opened 2017-03-20
Make extension field names be simple String, allow generating them dynamically

tbkka:
This changes the `fieldName` property on extension objects to be a simple `String` property (instead of `StaticString`).  (This was enabled by #390, which isolated the handling of extension field names.)

It then changes how extensions get generated so that the `fieldName` is computed by appending the extension field name proper to the name of the surrounding scope.  A typical extension object now looks like this:
```
 let ProtobufUnittest_Extensions_optional_sint32_extension = SwiftProtobuf.MessageExtension<OptionalExtensionField<SwiftProtobuf.ProtobufSInt32>, ProtobufUnittest_TestAllExtensions>(
   _protobuf_fieldNumber: 5,
-  fieldName: "protobuf_unittest.optional_sint32_extension",
+  fieldName: _protobuf_package + ".optional_sint32_extension",
   defaultValue: 0
 )
```

This was motivated by Issue #177, which questioned whether we could eliminate the duplicated package and message names when writing extension objects.  This shows that it's possible, but it actually makes the generated/compiled Swift code slightly larger.  (`unittest.pb.swift.o` goes from 6054536 bytes unstripped to 6064584 bytes)

My inclination:  Change the `fieldName` property to `String` as a general API cleanup, but leave the generated names as constants instead of computing them (basically, keep the first commit here but drop the second one).   Thoughts?

### DIFF
--- a/Sources/protoc-gen-swift/ExtensionGenerator.swift
+++ b/Sources/protoc-gen-swift/ExtensionGenerator.swift
@@ -22,12 +22,12 @@ struct ExtensionGenerator {
     let descriptor: Google_Protobuf_FieldDescriptorProto
     let generatorOptions: GeneratorOptions
     let path: [Int32]
-    let protoPath: String
     let protoPackageName: String
     let swiftDeclaringMessageName: String?
     let context: Context
     let comments: String
     let fieldName: String
+    let fieldNamePath: String
     let apiType: String
     let swiftFieldName: String
     let swiftHasPropertyName: String
@@ -78,10 +78,13 @@ struct ExtensionGenerator {
         self.apiType = descriptor.getSwiftApiType(context: context, isProto3: false)
         self.comments = file.commentsFor(path: path)
         self.fieldName = descriptor.isGroup ? descriptor.bareTypeName : descriptor.name
-        if let parentProtoPath = parentProtoPath {
-            self.protoPath = parentProtoPath + "." + fieldName
+        if let parentProtoPath = parentProtoPath, !parentProtoPath.isEmpty {
+            var p = parentProtoPath
+            assert(p.hasPrefix("."))
+            p.remove(at: p.startIndex)
+            self.fieldNamePath = p + "." + fieldName
         } else {
-            self.protoPath = fieldName
+            self.fieldNamePath = fieldName
         }
 
         let baseName: String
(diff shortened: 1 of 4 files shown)


tbkka:
If I change `ExtensionGenerator` to only compute extension names if the scope is longer than 40 characters, then `unittest.pb.swift.o` does not change size at all.  But in that case, I only see about a half-dozen uses of the new capability across all our reference protos.

Sources/protoc-gen-swift/ExtensionGenerator.swift
```diff
@@ -115,10 +115,13 @@ struct ExtensionGenerator {
         let scope = swiftDeclaringMessageName == nil ? "" : "static "
         let traitsType = descriptor.getTraitsType(context: context)
 
-        var fieldNamePath = self.protoPath
-        if fieldNamePath.hasPrefix(".") {
-            fieldNamePath.remove(at: fieldNamePath.startIndex)
-            fieldNamePath = "\"\(fieldNamePath)\""
```
thomasvl: Actually, why do we only quote the string if it started with a '.', seems like if we didn't do that pruning then we don't quote the value here.  Should the `p.print("fieldName: \(fieldNamePath),\n")` always add the quotes instead?

thomasvl:
Since extensions hopefully aren't that common, I'd say we're probably ok going with the slightly longer strings.

Sources/protoc-gen-swift/ExtensionGenerator.swift
tbkka: By convention, the package name inside of protoc always begins with a '.'.  So this path always gets taken.

I should probably just assert the presence of the leading '.' instead of making this conditional.

tbkka:
@thomasvl:  Any thoughts on changing the API so that `fieldName` is `String` instead of `StaticString`?

thomasvl:
Using `String` seems fine.  Since the first arg to the initializer has the leading underscore, we've already marked the method as private, so we can always change it in the future.

thomasvl:
The `TextFormatEncoder` and `TextFormatEncoderVisitor` changes seem worth doing also.

Sources/protoc-gen-swift/ExtensionGenerator.swift
tbkka: I put the quotes into the `p.print` below.

I also moved the calculation of the fully-scoped field name into the initializer.

thomasvl:
ps - might need to tweak the default merge comment since they aren't generated dynamically.

PR #401, opened 2017-03-21
Missing any.pb.swift in the iOS, tvOS and watchOS targets

erikt:
The iOS, tvOS or watchOS targets can not be built, because Xcode's Build phase, Compile Sources, is missing the reference to any.pb.swift for these targets.


thomasvl:
Oops, sorry about that and thanks!

Issue #108, opened 2016-10-18
Should we drop the leading_detached_comments


thomasvl:
Java, C#, and ObjC are the languages in the google/protobuf sources that pull over comments, but they only pull over the `leading_comments` and `trailing_comments`.  They all ignore the `leading_detached_comments`.

There is some value in the `leading_detached_comments`, but the catch is the proto file structure doesn't come through.  Detached comments are bound to the item that they appeared before; but proto files don't have a defined order, they mix enums/messages/extensions in the content of the file however the file's author chose.  About the only thing we know is they made some sense at the point they appeared in the proto file, but they could apply to the thing right after them or they could apply to everything after them.  But the generator sees a `FileDescriptor` where all the enums/messages/extensions have grouped into repeated fields, and the file order is lost.

Given a file like:

``` protobuf

// Foo and Bar are for…

message Foo { }

enum Bar {}

// Baz and Mumble are for…

message Baz {}

enum Mumble {}

message Blah {}
```

Common convention is to generate the Enums first, the source would end up like:

``` c++
enum Bar {}
enum Mumble {}

// Foo and Bar are for…

class Foo {}

// Baz and Mumble are for…

class Baz {}
class Blah {}
```

After the reordering, the comments don't make as much sense.

It is likely also worth mentioning that some languages don't bring the comments over at all.  Even for developers that work in the languages that support pulling over comments, I've seen several not realize this happens; so the proto file comments (even when attached to things) aren't that useful.

So we need to decide if we think it is worth bringing the leading comments over or not.

tbkka:
If I remember correctly, the motivation for this was to preserve the copyright messages at the beginning of the file.

In particular, the Swift Protobuf runtime library directly includes Swift sources generated from Google-copyrighted protos.  Those protos include a copyright message that requires the copyright message itself to be preserved.  In the generated descriptor we get from protoc, that copyright message becomes a "leading detached comment" attached to the first syntactic element in the descriptor.   Unfortunately, there's no guarantee that the first syntactic element is a `syntax` element (legacy proto2 files lack this), so I opted to preserve leading detached comments everywhere.

We could do this in other ways, of course.  For example, the detached comments could be rendered with ```//``` instead of ```///``` which would leave them out of the docstring.

thomasvl:
Detached comments already are left out of the docstring (they are `//`).

[...]

PR #405, opened 2017-03-21
Drop leading_detached_comments except where trying to get copyright.

thomasvl:
The other generators that pull over comments don't seem to bring over leading detached
comments, so follow in their lead.

Fixes https://github.com/apple/swift-protobuf/issues/108

### DIFF
--- a/Sources/protoc-gen-swift/FileGenerator.swift
+++ b/Sources/protoc-gen-swift/FileGenerator.swift
@@ -218,7 +218,7 @@ class FileGenerator {
         }
     }
 
-    func commentsFor(path: [Int32]) -> String {
+    func commentsFor(path: [Int32], includeLeadingDetached: Bool = false) -> String {
         func escapeMarkup(_ text: String) -> String {
             // Proto file comments don't really have any markup associated with
             // them.  Swift uses something like MarkDown:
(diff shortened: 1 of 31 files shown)


thomasvl:
I figured this was an easy thing to confirm if we want to do what the issue landed on.  If we want to drop it and _won't fix_ the issue, that's fine also.

Sources/protoc-gen-swift/FileGenerator.swift
```diff
@@ -285,7 +283,8 @@ class FileGenerator {
         // the file is an empty path. That never seems to have comments on it.
         // https://github.com/google/protobuf/issues/2249 opened to figure out
         // the right way to do this since the syntax entry is optional.
-        let comments = commentsFor(path: [Google_Protobuf_FileDescriptorProto.FieldNumbers.syntax])
+        let comments = commentsFor(path: [Google_Protobuf_FileDescriptorProto.FieldNumbers.syntax],
+                                   includeLeadingDetached: true)
```
tbkka: Should probably comment _why_ we're including leading detached comments here.

thomasvl: Done

Issue #293, opened 2017-02-22
Should most uses of `ExtensionSet` change to a protocol


thomasvl:
This is something I've been looking at doing to for ObjC, so capturing it for discussion before our 1.0 API locks us down for a while.

The current api around decoding messages directly uses `ExtensionSet`, but should it use a protocol instead (say `ExtensionResolver`?)?  The idea being this protocol would just have the methods needed to look up `Extensions`.

The generation would still make the helper `ExtensionSet`s they do now, but it would allow a developer to write a custom resolver that could be a giant switch, cascade of if/else, etc. where they hard cold all of the extension their app needs.  There by removing the startup time for building the `ExtensionSet` and potentially having better lookup time for resolving extensions.

tbkka:
@thomasvl: I have a working rough implementation of this idea; it was pretty simple.  Most interestingly, it did not touch the generated code, though I did have to jigger a few public APIs to accept the new protocol instead of the struct type.

I like that this opens up options for us to explore post-1.0.  In particular, the current `ExtensionSet` lookup code is a little hampered by the fact that types are not Hashable, which leads us to walk lists of types to try to find matches.   With this in place, we could explore having the code generator emit tailored implementations of the protocol instead.  That might be a big improvement.

@allevato:  Any ideas about naming?  My current code uses `ExtensionCollection` for the protocol and `ExtensionSet` for the concrete implementation.  Maybe you have a better idea?

tbkka:
After thinking about it overnight, I'm going with `ExtensionMap` for the protocol (since the protocol just has methods for looking up extension objects given the message being extended and the field number), and `SimpleExtensionMap` for the current implementation.

tbkka:
Closed via #406

PR #406, opened 2017-03-21
Make extension lookup go through a protocol

tbkka:
This implements #293.

It introduces a new `ExtensionMap` protocol that has two methods for looking up information about extensions:
* A method that looks up the extension number given the message being extended and the name
* A method that returns the extension object given the message being extended and the number

The old `ExtensionSet` struct has been renamed `SimpleExtensionSet`.  It provides a basic implementation of this protocol and is used by the code generator.

This will allow developers to optimize extension lookup by providing tailored implementations of the protocol (for example, that used switch statements for faster lookup).  We could even have our code generator build such tailored implementations if we wanted.

### DIFF
--- a/Sources/SwiftProtobuf/ExtensionMap.swift
+++ b/Sources/SwiftProtobuf/ExtensionMap.swift
@@ -1,0 +1,38 @@
+// Sources/SwiftProtobuf/ExtensionMap.swift - Extension support
+//
+// Copyright (c) 2014 - 2017 Apple Inc. and the project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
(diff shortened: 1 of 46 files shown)


Sources/SwiftProtobuf/ExtensionMap.swift
```diff
@@ -0,0 +1,38 @@
...
+///
+/// A set of extensions that can be passed into deserializers
+/// to provide details of the particular extensions that should
+/// be recognized.
+///
+// -----------------------------------------------------------------------------
+
+/// A collection of extension objects.
+///
+/// An `ExtensionMap` is used during decoding to look up
```
thomasvl: I'm fine with the name, but I'll float another possible name: ExtensionRegistry.

PR #409, opened 2017-03-21
Text encoder should never fail, improve Any formatting

### LINKED ISSUE #229: Encoding to text format can't throw; remove spurious `throws`
Encoding to text format shouldn't throw—the only places where it currently throws are because of protocol requirements in the `FieldType` methods or when a field name fails to resolve (in which case we should be returning the field number as a string instead).

We can't clean up the API 100% because `TextEncodingVisitor` conforms to a protocol with throwing methods and even though the concrete implementations of those methods can be made non-throwing, since `traverse` operates on `Visitor` polymorphically, it can't know that the concrete visitor it's given doesn't throw. However, we can clean up as much as we can and then swallow the never-thrown errors at key points. For example,

* `serializeTextValue` in the `FieldType` protocol should be made non-throwing, because none of its concrete implementations except for `Message` are declared as throwing
* `Message.serializeTextValue` can be made non-throwing by wrapping its implementation in a swallowing `do/catch`
* Update `TextEncodingVisitor.protoFieldName` to return the field number instead of throwing if the field name is not found
* Now, none of the calls inside `TextEncodingVisitor.visit*` methods can throw; those `try`s can be removed and the methods can be made non-throwing
* Finally, make `Message.serializeText` non-throwing by wrapping its implementation in a swallowing `do/catch`

tbkka:
Fixes #229:  The Text format encoder is intended for debug use, so it should never fail.

Earlier work fixed Text format field naming to correctly use numbers if names were unavailable.  This fixes the remaining failure case of Any objects that contain undecoded JSON.  It also expands the logic for printing Any objects to make them generally more legible.

Any objects now get printed in one of three formats, depending on the content. The first two are also used by Google's C++ implementation.

1. Verbose form is used if the Any holds an unserialized message or if we can decode the contents.  A message with field `any_field` containing an Any containing an Any containing a Duration might print as:

```
any_field {
  [type.googleapis.com/google.protobuf.Any] {
    [type.googleapis.com/google.protobuf.Duration] {
      seconds: 123
      nanos: 234567890
    }
  }
}
```
    
2. If we have binary contents that we cannot decode, we use a plain form that just dumps the Any as a regular object with two fields:

```    
any_field {
  type_url: "type.googleapis.com/some.unknown.type"
  value: "\\022\\007\\b{\\020\\322\\361\\354o"
}
```

3. If we have JSON contents that we cannot decode, we dump it in a similar format, but comment out the `json` field to avoid confusing other Text format decoders:

```    
any_field {
  type_url: "type.googleapis.com/some.unknown.type"
  #json: "{\"optionalInt32\":7}"
}
```


thomasvl:
Having looked at the cl yet, but capturing it while I see it:

>```
>any_field {
>  type_url: "type.googleapis.com/some.unknown.type"
>  #json: "{\"optionalInt32\":7}"
>}
>```

Do we need quotes (and escaping) within that comment, or should we drop them to simplify the output?

Reference/google/protobuf/any.pb.swift
```diff
@@ -187,15 +187,7 @@ struct Google_Protobuf_Any: SwiftProtobuf.Message, SwiftProtobuf._MessageImpleme
   }
 
   func traverse<V: SwiftProtobuf.Visitor>(visitor: inout V) throws {
-    try withExtendedLifetime(_storage) { (_storage: _StorageClass) in
```
thomasvl: I guess something regressed in another push?

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -162,35 +162,81 @@ internal class AnyMessageStorage {
...
-  // Called before the message is traversed to do any error preflights.
-  // Since traverse() will use _value, this is our chance to throw
-  // when _value can't.
-  func preTraverse() throws {
+  // Specialized traverse() implementation for Any
+  // This does various checks depending on the output format
+  // and contents, then traverses the storage fields.
+  func traverse<V: Visitor>(visitor: inout V) throws {
+    // Encoding to Text format for debugging is special:
+    if visitor is TextFormatEncodingVisitor {
```
thomasvl: This feels a little wrong to me.  TextFormatDecoder already knows about Any, does it make more sense to keep the same dependency and make the TextFormat encoding know about Any and things off within that instead of having to do this test here slowing down all formats?

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -211,13 +257,58 @@ internal class AnyMessageStorage {
...
+    }
+    decoder.scanner = subDecoder.scanner
+    if try decoder.nextFieldNumber() != nil {
+      // Verbose any can never have additional keys.
+      throw TextFormatDecodingError.malformedText
+    }
+  }
+}
+
+/// Hashable and Equatable conformance
```
thomasvl: History tends to be easier to follow if changes like this comment aren't rolled into unrelated commits.

Reference/google/protobuf/any.pb.swift
```diff
@@ -187,15 +187,7 @@ struct Google_Protobuf_Any: SwiftProtobuf.Message, SwiftProtobuf._MessageImpleme
...
-      try _storage.preTraverse()
-      if !_storage._typeURL.isEmpty {
-        try visitor.visitSingularStringField(value: _storage._typeURL, fieldNumber: 1)
-      }
-      if !_storage._value.isEmpty {
-        try visitor.visitSingularBytesField(value: _storage._value, fieldNumber: 2)
-      }
-    }
+    try _storage.traverse(visitor: &visitor)
     try unknownFields.traverse(visitor: &visitor)
```
thomasvl: What ends up happening on textFormat if there were unknown fields?  Does the output end up making sense in the current code flow?  If you change how traverse is picked of, this issue might go away.

tbkka:
```
any_field {
 type_url: "type.googleapis.com/some.unknown.type"
 #json: "{\"optionalInt32\":7}"
}
```

> Do we need quotes (and escaping) within that comment, or should we drop them to simplify the output?

My thinking was to format it exactly like a regular `bytes` field except that it's commented-out.  In particular, that would make it easier down the road to support it in the decoder as well (if we decided to do so).

Sources/SwiftProtobuf/AnyMessageStorage.swift
thomasvl: Thinking a little move about this, having to do this check within Any proves that our visitor api can't be used to write the TextFormat support.  Checkout outside is what someone else would likely have to do.  (yes, there's the catch that they any doesn't expose everything to do this, but it still provides some what of a comparison)

tbkka: I've moved the check into the Text encoder.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -162,35 +162,81 @@ internal class AnyMessageStorage {
...
+  // This prefers the more-legible "verbose" format if it can
+  // use it, otherwise will fall back to simpler forms.
+  private func textTraverse(visitor: inout TextFormatEncodingVisitor) {
+    if let msg = _message {
+      emitVerboseTextForm(visitor: &visitor, message: msg, typeURL: _typeURL)
+    } else if let valueData = _valueData {
+      if let messageType = Google_Protobuf_Any.messageType(forTypeURL: _typeURL) {
+        // If we can decode it, we can write the readable verbose form:
+        do {
+          let m = try messageType.init(serializedData: valueData)
```
tbkka: Added `partial: true` to this call

Sources/SwiftProtobuf/AnyMessageStorage.swift
tbkka: ... and restored the previous code generation for Any.

Sources/SwiftProtobuf/AnyMessageStorage.swift
tbkka: Removed unrelated comment.

Reference/google/protobuf/any.pb.swift
tbkka: I believe it actually does end up correct, but I'll add some tests today to verify this.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -162,6 +162,36 @@ internal class AnyMessageStorage {
     throw AnyUnpackError.malformedAnyField
   }
 
+
```
thomasvl: Just checking since it's hard to tell in github's diff UI, this is just a move within the file?

tbkka: Yes.  I moved the text encode/decode support into an extension below, to match how we had grouped the JSON special support.

PR #410, opened 2017-03-22
Derive the reserved word lists from the Swift language spec

tbkka:
This adds collections of Swift reserved words to
PluginLibrary/SwiftLanguage that are taken directly from
the language spec.  It then redefines the reserved word
lists used to filter struct, enum, and field names in terms
of these.

This should clarify why various words are reserved by the
code generator and hopefully make it easier to maintain
these lists over time.  In particular, the lists of Swift
reserved words can just be copied/updated from the langauge
spec as it evolves.


tbkka:
This updates `unittest_swift_naming.proto` with some additional field names that were overlooked earlier.

I also need to check and update the messages and enum cases to make sure every Swift reserved word is covered there as well.

Sources/protoc-gen-swift/ReservedWords.swift
```diff
@@ -211,60 +127,23 @@ func sanitizeFieldName(_ s: String) -> String {
...
+
+    // We don't need to protect all of these keywords, just the ones
+    // that interfere with enum cases:
+    // names = names.union(PluginLibrary.swiftKeywordsReservedInParticularContexts)
+    names.insert("associativity")
+    names.insert("dynamicType")
+    names.insert("optional")
+    names.insert("required")
+
+    names = names.union(PluginLibrary.swiftKeywordsUsedInDeclarations)
```
thomasvl: Maybe a comment why this only adds 3 of the sets instead of the 5 the others add?

tbkka: Done.

PR #411, opened 2017-03-23
Documentation cleanup

allevato:
This PR also renames the `(Binary|JSON|TextFormat)TypeAdditions` files to `Message+(Binary|JSON|TextFormat)Additions` to better reflect their content now that the old metatypes are gone.


Sources/SwiftProtobuf/Enum.swift
```diff
@@ -16,19 +16,26 @@
 // -----------------------------------------------------------------------------
 
 /// Generated enum types conform to this protocol, which provides the
-/// hashability requirement for enums as well as the name mapping requirement
```
allevato: Here I removed the prose about the name mapping requirement because it's not part of the public API, so it doesn't need to be in the public docs.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -115,11 +115,10 @@ public protocol Message: CustomDebugStringConvertible {
 }
 
 public extension Message {
-
-  /// If the generated code needs to provide its own implementation, usually
-  /// because the underlying `.proto` file uses proto2 syntax, it will provide
-  /// its own implementation. Generally, users of the generated code should
-  /// not.
+  /// Generated proto2 messages that contain required fields (or nested messages
```
thomasvl: Extensions also come into play (if the message is extendable, then an extension could be added that is proto2)

Sources/SwiftProtobuf/TextFormatTypeAdditions.swift
```diff
@@ -15,46 +15,54 @@
...
-        if let any = self as? Google_Protobuf_Any {
-            any._storage.textTraverse(visitor: &visitor)
-        } else {
-            try! traverse(visitor: &visitor)
-        }
-        return visitor.result
+  /// Returns a string containing the Protocol Buffer text format serialization
+  /// of the message.
+  ///
+  /// Unlike binary encoding, presence of required fields is not enforced when
```
thomasvl: So we're little inconsistent, you mention isInitialized here, but not no init(textFormatString:extensions:).  How about we drop all references from isInitialized in the TextFormat and JSON and just make the phrasing in binary clear that it is specific to that one format?

Sources/SwiftProtobuf/UnknownStorage.swift
```diff
@@ -16,7 +16,12 @@
 
 import Foundation
 
+/// Contains any unknown fields in a decoded message (that is, fields that were
+/// sent on the wire but were not recognized by the generated message
+/// implementation).
```
thomasvl: Do you want to mention it also has fields where the data type was wrong for the message's definition?

Sources/SwiftProtobuf/Decoder.swift
```diff
@@ -43,89 +43,88 @@ import Foundation
...
+  // but someone might want to track OneOf conflicts in the future.
+  mutating func handleConflictingOneOf() throws
 
-    // Get the next field number.  For JSON and Text, the decoder
-    // translates name to number at this point, based on information
-    // it obtained from the message when it was initialized.
-    mutating func nextFieldNumber() throws -> Int?
+  // Get the next field number.  For JSON and Text, the decoder
+  // translates name to number at this point, based on information
+  // it obtained from the message when it was initialized.
```
thomasvl: Document that it should return nil at the end of the input?

Sources/SwiftProtobuf/AnyUnpackError.swift
```diff
@@ -14,23 +14,29 @@
...
-  /// There was something else wrong...
+
+  /// The `Any` message was malformed in some other way not covered by the other
+  /// error cases.
   case malformedAnyField
-  /// Decoding JSON or Text format requires the message type
+
+  /// Decoding JSON or text format requires the message type
   /// to have been compiled with textual field names.
   case missingFieldNames
```
tbkka: I think this case can be removed; it's never actually used in the code.  (Missing field names is a JSON- or Text-level error and throws either `JSONDecodingError.missingFieldNames` or `TextFormatDecodingError.missingFieldNames` respectively.)

Sources/SwiftProtobuf/BinaryTypeAdditions.swift
```diff
@@ -1,114 +0,0 @@
-// Sources/SwiftProtobuf/BinaryTypeAdditions.swift - Per-type binary coding
```
tbkka: Thanks for renaming this file.

Sources/SwiftProtobuf/Decoder.swift
```diff
@@ -43,89 +43,88 @@ import Foundation
...
 public protocol Decoder {
-    // Called by a Oneof when it already has a value and is being
-    // asked to accept a new value.  Some decoders specify that oneof
-    // decoding must fail in this case.  Those decoders throw a
-    // suitable error from this function; other decoders make it a
-    // no-op.  No existing decoder actually needs this to be mutating,
-    // but someone might want to track OneOf conflicts in the future.
-    mutating func handleConflictingOneOf() throws
+  // Called by a Oneof when it already has a value and is being
+  // asked to accept a new value.  Some decoders specify that oneof
```
tbkka: I think "Some decoders..." here should be "Some formats require oneof decoding to fail in this case."   (This behavior is required by the format specification, our "decoder" is just an implementation.)

Sources/SwiftProtobuf/Enum.swift
tbkka: I would drop the text about the "hashability requirement" as well, since that's obvious from the declaration conforming to `Hashable`.

Sources/SwiftProtobuf/Enum.swift
```diff
@@ -16,19 +16,26 @@
...
+  ///
+  /// For proto2 enums, this initializer will fail if the raw value does not
+  /// correspond to a valid enum value. For proto3 enums, this initializer never
+  /// fails; unknown values are created as instances of the `UNRECOGNIZED` case.
+  ///
+  /// - Parameter rawValue: The raw integer value from which to create the enum
+  ///   value.
   init?(rawValue: Int)
 
+  /// The raw integer value of the enum value.
```
tbkka: Should this elaborate on the proto3 `UNRECOGNIZED` handling?

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -0,0 +1,133 @@
...
+import Foundation
+
+/// Binary encoding and decoding methods for messages.
+public extension Message {
+  /// Returns a `Data` value containing the Protocol Buffer binary format
+  /// serialization of the message.
+  ///
+  /// - Parameters:
+  ///   - partial: Binary serialization requires all `required` fields be
+  ///     present; if `false` (the defaul), this method throws
```
tbkka: defaul -> default

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -0,0 +1,133 @@
...
+  ///     `BinaryEncodingError.missingRequiredFields` if any were missing. When
```
tbkka: Suggest:  ... if `false` (the default), this method will call `Message.isInitialized` before encoding to verify that all required fields are present.  The `Message.isInitialized` method will throw `BinaryEncodingError.missingRequiredFields` if it finds any required fields are not set.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  private func serializeBinary(
+    into pointer: UnsafeMutablePointer<UInt8>
+  ) throws {
+    var visitor = BinaryEncodingVisitor(forWritingInto: pointer)
+    try traverse(visitor: &visitor)
+  }
+
+  internal func serializedDataSize() throws -> Int {
```
tbkka: I personally would include basic documentation here for the benefit of anyone who wants to work on the code someday.  How about
```
/// (Internal API) Computes the size of the binary serialized data.
///
/// This is used by `serializedData()` to determine the correct size of the
/// output buffer so that the low-level encoding can proceed at full speed without bounds checks.
```

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
allevato: Tweaked a bit since `isInitialized` is a property, not a throwing method.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
allevato: Done (similar).

Sources/SwiftProtobuf/Message.swift
allevato: Done.

Sources/SwiftProtobuf/UnknownStorage.swift
allevato: Done.

allevato:
> Google_Protobuf_Any+Extensions has a Message extension also, do you want to move that out into a file to go with your file renaming?

Done.

PR #412, opened 2017-03-23
Move helper to be non instance, and fix variable usage.

thomasvl:


Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -32,6 +32,16 @@ fileprivate func serializeAnyJSON(for message: Message, typeURL: String) throws
   return visitor.stringResult
 }
 
+fileprivate func emitVerboseTextForm(visitor: inout TextFormatEncodingVisitor, message: Message, typeURL: String) {
+  let url: String
+  if typeURL.isEmpty {
+    url = buildTypeURL(forMessage: message, typePrefix: defaultTypePrefix)
+  } else {
+    url = typeURL
```
thomasvl: This was url = _typeURL below, reality was we only ever passed _typeURL, so it didn't matter, but since it's passed it should use what is passed.  Making it non instance base helps ensure something like that doesn't happen again.

PR #414, opened 2017-03-23
Use an enum for the internal state of AnyMessageStorage.

thomasvl:
Reduces the risk of every getting into poorly defined state.


thomasvl:
Trying to support having more than one of these set would gonna be tricky as each place we check the state, we'd have to get the orders "just right" from a preference pov, etc.  So just going to the true state seemed like a better approach to keep the code simple.

PR #415, opened 2017-03-23
Clean up Any support of required fields.

thomasvl:
- Expose `partial` on the helper to create a Google_Protobuf_Any from a Message,
  and do the validation there.
- With AnyMessageStorage always use partial support since validation was done
  up front or should have been done by whoever sent us the payload. In almost
  all of these cases, there's no good way to deal missing required fields any
  way, so just let things slide.
- Remove the isInitialized check in preTraverse. This was wrong in that it was
  causing required fields to get checked for TextFormat and JSON encoding.

### DIFF
--- a/Sources/SwiftProtobuf/AnyMessageStorage.swift
+++ b/Sources/SwiftProtobuf/AnyMessageStorage.swift
@@ -126,7 +126,7 @@ internal class AnyMessageStorage {
 
     switch state {
     case .binary(let data):
-      target = try M(serializedData: data, extensions: extensions)
+      target = try M(serializedData: data, extensions: extensions, partial: true)
 
     case .message(let msg):
       if let message = msg as? M {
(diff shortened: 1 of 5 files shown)


Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -173,13 +173,12 @@ internal class AnyMessageStorage {
...
-      // A message could be checked when set, but that isn't always
-      // clean in the places it gets decoded from some other form, so
-      // validate it here.
-      if !m.isInitialized {
-        throw BinaryEncodingError.missingRequiredFields
-      }
+    case .message:
+      // When set from a developer provided message, partial support
+      // is done. Any message that comes in from another format isn't
+      // checked, and transcoding the isInitializd requirement is
```
allevato: s/isInitializd/isInitialized/

thomasvl: Done

PR #418, opened 2017-03-24
Cleanup documentation for WKTs.

allevato:
I should have done the reïndentation in a separate commit, but once I started writing the docs and having Xcode start doing 2 instead of 4, I was already too deep into my edits. Focus on the darker highlighted parts.

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Duration+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Duration+Extensions.swift
@@ -20,191 +20,215 @@ private let minDurationSeconds: Int64 = -maxDurationSeconds
 private let maxDurationSeconds: Int64 = 315576000000
 
 private func parseDuration(text: String) throws -> (Int64, Int32) {
-    var digits = [Character]()
-    var digitCount = 0
-    var total = 0
-    var chars = text.characters.makeIterator()
-    var seconds: Int64?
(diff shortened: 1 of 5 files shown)


Sources/SwiftProtobuf/Google_Protobuf_Duration+Extensions.swift
```diff
@@ -20,191 +20,215 @@ private let minDurationSeconds: Int64 = -maxDurationSeconds
...
-        let nd = round((value - sd) * TimeInterval(nanosPerSecond))
-        let (s, n) = normalizeForDuration(seconds: Int64(sd), nanos: Int32(nd))
-        self.init(seconds: s, nanos: n)
-    }
+  public typealias FloatLiteralType = Double
+
+  /// Creates a new `Google_Protobuf_Duration` from a floating point literal
+  /// that is interpreted as a duration in seconds, rounded to the nearest
+  /// nanosecond.
+  public init(floatLiteral value: Double) {
```
thomasvl: If it matters, this one you didn't seem to add a parameter comment for.

allevato: I started to, but then I noticed in other files I mostly didn't document the parameters of `ExpressibleBy*Literal` initializers. In a sense it's unnecessary since you would never call the initializer directly to begin with. I waffle back and forth on it.

PR #419, opened 2017-03-24
Ext naming

thomasvl:


thomasvl:
This is more progress on #271

PR #422, opened 2017-03-24
Improved reserved word handing for extension names

thomasvl:


Sources/protoc-gen-swift/ReservedWords.swift
```diff
@@ -278,6 +278,15 @@ private let reservedEnumCases: Set<String> = [
             "self",
         ]
 
+/*
+ * Many Swift reserved words can be used as Extension names if we put
```
allevato: Nit: ///-style comments

thomasvl: This file has the other style, so I'm following that at the moment.  ☹️

PR #426, opened 2017-03-25
Allow perf harness to compare multiple revisions of SwiftProtobuf

allevato:
Whew. I got inspired yesterday, and I should have done this a long time ago; it wasn't as bad as I thought it would be.

The harness now takes `-c` options that let you specify git revisions that should be benchmarked and compared to the current working tree. They can be specified as anything the git tools allow—I imagine branch names (probably `master`) will be most useful but you can also use commit hashes and whatnot.

`-c` also accepts the string `c++`, which means compare against C++ (the old behavior, and also what happens if you don't use any `-c` options). You can specify `-c` multiple times, each one becomes a series in the plot, and the current working tree is always last; we're not limited to two anymore (we probably won't use this very much, unless we want to compare, say, `master`, the working tree, and C++ all at once).

This works by checking out the local repo history at the given revisions into temporary locations and building the harnesses there. This means it uses the generator and the runtime at that revision—so even if you're changing generator code, you end up generating the right thing for each revision and it just works.

It should be noted that this will only work for commits *going forward from the time this is merged*; because of changes I made to the harness generating scripts to separate the generation from the running (to isolate API changes), the harness makes assumptions about the scripts that will be available in the checkout.

I tested it by adding a `usleep(1)` in the binary decoder (I can't account for the seemingly nontrivial noise elsewhere in the two branches, though):
harness

This was generated with the command line:
```
$ Performance/perf_runner.sh -c c++ -c revision-perf -p false 100 "fixed32"
```

### DIFF
--- a/Performance/perf_runner.sh
+++ b/Performance/perf_runner.sh
@@ -242,11 +242,6 @@ fi
 
 readonly field_count="$1"; shift
 if [[ "$1" == "all" ]]; then
-  # if [[ "$comparison" != "c++" ]]; then
-  #   echo "ERROR: Commit-based comparison cannot be used with 'all'."
-  #   echo
-  #   usage
-  # fi
   readonly requested_field_types=( \
     int32 sint32 uint32 fixed32 sfixed32 \
     int64 sint64 uint64 fixed64 sfixed64 \
@@ -255,6 +250,17 @@ if [[ "$1" == "all" ]]; then
 else
   readonly requested_field_types=( "$@" )
 fi
+
+# ---------------------------------------------------------------
+# Set up a hook to cleanup revision comparison checkouts when the script
+# completes.
+declare -a CLEANUP_WHEN_DONE
+function cleanup_revision_checkouts() {
+  if [[ "${#CLEANUP_WHEN_DONE[@]}" -ne 0 ]]; then
+    rm -rf "${CLEANUP_WHEN_DONE[@]}"
+  fi
+}
+trap cleanup_revision_checkouts EXIT HUP INT QUIT TERM
 
 # ---------------------------------------------------------------
 # Pull in language-specific runners.
@@ -323,6 +329,7 @@ EOF
         # Check out the commit to a temporary directory and create its _generated
         # directory. (Results will still go in the working tree.)
         tmp_checkout="$(mktemp -d -t swiftprotoperf)"
+        CLEANUP_WHEN_DONE+=("$tmp_checkout")
         git --work-tree="$tmp_checkout" checkout "$comparison" -- .
         mkdir "$tmp_checkout/Performance/_generated"
 
@@ -334,8 +341,6 @@ EOF
         harness_swift="$tmp_checkout/Performance/_generated/harness_swift"
         results_trace="$script_dir/_results/$field_count fields of $field_type (swift)"
         run_swift_harness "$tmp_checkout" "$comparison" "$commit_results"
-
-        rm -r "$tmp_checkout"
       else
         echo
         echo "==== Found cached results for Swift ($comparison) ===================="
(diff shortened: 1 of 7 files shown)


Performance/perf_runner.sh
```diff
@@ -41,22 +41,52 @@ readonly GOOGLE_PROTOBUF_CHECKOUT=${GOOGLE_PROTOBUF_CHECKOUT:-"$script_dir/../..
...
+
+    Currently supported field types:
+        int32, sint32, uint32, fixed32, sfixed32,
+        int64, sint64, uint64, fixed64, sfixed64,
+        float, double, string,
+        ...and repeated variants of the above.
+
+        Additionally, you can specify "all" to run the harness multiple
+        times with all of the (non-repeated) field types listed above.
+        ("all" is not compatible with revision comparisons.)
```
thomasvl: Especially for revision comparisons, should we look at adding a "mixed" or something that is a message with say two dozen fields of differing types and also uses messages within messages to get some depth?

Performance/perf_runner.sh
```diff
@@ -189,6 +242,11 @@ fi
 
 readonly field_count="$1"; shift
 if [[ "$1" == "all" ]]; then
+  # if [[ "$comparison" != "c++" ]]; then
```
thomasvl: delete since it is commented out?

Performance/perf_runner.sh
```diff
@@ -217,39 +274,95 @@ if [[ ! -f "$results_js" ]]; then
...
+      harness_cpp="$script_dir/_generated/harness_cpp"
+      results_trace="$script_dir/_results/$field_count fields of $field_type (cpp)"
+      run_cpp_harness "$harness_cpp"
+    else
+      commit_hash="$(git rev-parse $comparison)"
+      commit_results="$script_dir/_results/${commit_hash}_${field_type}_${field_count}.js"
+
+      # Check to see if we have past results from that commit cached. If so, we
+      # don't need to run it again.
+      if [[ ! -f "$commit_results" ]]; then
```
thomasvl: What if it was testing a different field count/etc.?

Performance/perf_runner.sh
```diff
@@ -217,39 +274,95 @@ if [[ ! -f "$results_js" ]]; then
...
+        echo
+        echo "==== Building/running Swift harness ($comparison) ===================="
+        echo
+
+        # Check out the commit to a temporary directory and create its _generated
+        # directory. (Results will still go in the working tree.)
+        tmp_checkout="$(mktemp -d -t swiftprotoperf)"
```
thomasvl: do you want an exit hook to delete this directory?

Early on, set up a hook:
```bash
  CLEANUP_WHEN_DONE=()
  function cleanup() {
    rm -rf "${CLEANUP_WHEN_DONE[@]}"
  }
  trap cleanup EXIT HUP INT QUIT TERM
```

The this directory creation becomes:

```bash
  tmp_checkout="$(mktemp -d -t swiftprotoperf)"
  CLEANUP_WHEN_DONE+=("$tmp_checkout")
```

Then no mater now the script dies/exits (unless it is an exec), it will ensure everything added to CLEANUP_WHEN_DONE gets removed at the end.

Performance/perf_runner.sh
allevato: There's a TODO in there to do this :)  So yes.

Performance/perf_runner.sh
allevato: The filename includes the field type and count in addition to the hash.

Performance/perf_runner.sh
allevato: I was trying to think of a clean way to do that but my bash-fu isn't that strong—I like your approach.

Performance/perf_runner.sh
allevato: Done.

Issue #388, opened 2017-03-17
StorageClass: copy initializers instead of a copy method


thomasvl:
Comment by @tbkka on #387

>Bit of an aside: Is there any benefit to using copy initializers instead of a copy method like this? The method form needs to initialize every field before copying everything over; would a copy initializer avoid that?

tbkka:
Actually, I don't think this is a 1.0 blocker.  It only affects how the generated struct uses the generated Storage object; we can change it later without any problems.

PR #427, opened 2017-03-27
Switch from a clone model to a copy initializer.

thomasvl:
Should avoid doing an init() followed by assigning all the properties.

Fixes https://github.com/apple/swift-protobuf/issues/388.


Reference/google/protobuf/any_test.pb.swift
```diff
@@ -66,20 +66,18 @@ struct ProtobufUnittest_TestAny: SwiftProtobuf.Message, SwiftProtobuf._MessageIm
 
     init() {}
 
-    func copy() -> _StorageClass {
-      let clone = _StorageClass()
-      clone._int32Value = _int32Value
-      clone._anyValue = _anyValue
-      clone._repeatedAnyValue = _repeatedAnyValue
-      return clone
+    init(storage source: _StorageClass) {
```
tbkka: Maybe `init(copying:)`?

allevato: Swift doesn't seem to have any guidance on naming of copy initializers for classes since there aren't any that I could find in stdlib.

I recommended `storage:` because it mirrors the way the `[NSString initWithString:]` copy initializer gets imported into Swift as `init(string:)`, but ¯\\\_(ツ)\_/¯

tbkka: I've been noting the trend towards using participle phrases as arguments (`mutating: foo` or `unpacking: bar`) and realized that the trend could apply here as well.  But I don't have a strong opinion either way.  And since this is a private internal method, the argument naming is not a critical concern.

allevato: Yeah, I don't have a strong preference for either one, honestly.

@thomasvl originally had `copying:` anyway, so let's go with that :)

thomasvl: Done.

PR #429, opened 2017-03-27
Swift 3.1 compile time checks.

thomasvl:


Sources/SwiftProtobuf/JSONIntegerConverting.swift
```diff
@@ -49,31 +48,39 @@ internal protocol JSONIntegerConverting {
 extension Int64: JSONIntegerConverting {
 
   init?(safely value: Double) {
```
allevato: What if we rename these to `init?(exactly:)` and put the whole definition within a version check, and change the call sites? Then it'll just work everywhere.

thomasvl: Done as a new commit just incase…

PR #433, opened 2017-03-28
Xcode support

thomasvl:


thomasvl:
With this and the PR the other day, all Xcode builds are working/passing again.

PR #434, opened 2017-03-28
Update JSON decoding to use UnsafeBufferPointer.

allevato:
This has no appreciable impact on JSON decoding performance.

I tried doing the same for `TextFormatScanner`, but things got... weird. `nextFieldNumber` nearly doubled in execution time (and instruction count) and I couldn't figure out why. From looking at Hopper, the version at `master` is very nicely inlined and compact—even the call to `skipWhitespace` gets cleanly integrated into the body of the function. It's not immediately clear to me what's unique about that function that the optimizer isn't able to handle the UBP version, so I punted on it for the time being.

### DIFF
--- a/Sources/SwiftProtobuf/JSONScanner.swift
+++ b/Sources/SwiftProtobuf/JSONScanner.swift
@@ -90,7 +90,8 @@ private func decodeBytes(base64String s: String) -> Data? {
     n <<= 6
     switch digit {
     case asciiUpperA...asciiUpperZ: n |= Int(digit - asciiUpperA); bits += 6
-    case asciiLowerA...asciiLowerZ: n |= Int(digit - asciiLowerA + 26); bits += 6
+    case asciiLowerA...asciiLowerZ:
+      n |= Int(digit - asciiLowerA + 26); bits += 6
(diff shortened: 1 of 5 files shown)


Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -220,73 +222,81 @@ private func decodeString(_ s: String) -> String? {
...
-                }
-                throw JSONDecodingError.malformedNumber
-            case asciiBackslash:
-                return nil
-            default:
-                return 0
+private func parseBareUInt(
+  source: UnsafeBufferPointer<UInt8>,
+  index: inout UnsafeBufferPointer<UInt8>.Index,
+  end: UnsafeBufferPointer<UInt8>.Index
```
tbkka: I thought `parseBareUInt` was always called with `end` referring to the end of the input.  If I'm right, then `end` here may be redundant, you can just use `source.endIndex` within this function and lose the extra argument.

tbkka:
BTW, I'm not surprised that you had problems with `TextFormatScanner.nextFieldNumber`.  That function has proven much more brittle (performance-wise) than the JSON version for some reason.

Issue #420, opened 2017-03-24
Support other string encodings when decoding JSON from Data?


allevato:
If a user makes a request to some web service for some JSON, the API is likely to return `Data` rather than `String`. We currently support decoding directly from `Data`, but we assume that it's encoded as UTF-8.

Foundation's `JSONSerialization` has some interesting logic for decoding `Data` where it makes a best guess at sniffing out the encoding (looking for a BOM, or otherwise checking for zero-bytes in appropriate places), and then wraps the get-next-byte/index-advancing logic in such a way that the main parsing loop becomes encoding-agnostic. It supports UTF-8, UTF-16LE, UTF-16BE, UTF-32LE, and UTF-32BE.

We should look into doing this as well; I have some thoughts on the design rolling around in my head that I hope to code up soon.

This isn't 1.0 critical; it's additive and would not break public API.

thomasvl:
Should we spin a 1.0 task to do a little encoding sniffing and add an error if the data appears to be an encoding we don't support?

allevato:
The documentation at least states that the JSON is expected to be UTF-8 encoded, so I don't think it's *urgent* to do so.

If someone gets bitten by this, I think we can add the fix then, or better yet, have the rest of the work described by this PR done instead.

tbkka:
According to the RFC, encoding sniffing for JSON is actually very simple.  The first character in a JSON-encoded data is always an ASCII character, and JSON only permits UTF-8, UTF-16, and UTF-32.  So you only need to look at the first four bytes and notice the pattern of zero bytes.

The RFC also makes the point that a lot of implementations only accept UTF-8, so I think the odds of someone hitting this are pretty low.

allevato:
Right—I realize I forgot to leave the link to `JSONSerialization` where they do exactly that: https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSJSONSerialization.swift#L229

Adopting that should be fairly straightforward, if we want to do it.

tbkka:
Almost 7 years later, and we've had no requests for this.  Maybe it's not worth the complexity?

allevato:
It didn't seem critical enough to jump on when I filed the issue, so I think we can close it. It's reasonable that if someone wants to decode from JSON and they have text in a different encoding, they need to transcode it first. Especially since protobuf-JSON is more specific than just arbitrary JSON; it would be weird to write a backend that generated protobuf-JSON but encoded it in something other than UTF-8.

PR #435, opened 2017-03-29
Add jsonUTF8Data() for symmetry with init(jsonUTF8Data:)

tbkka:
This fills an obvious hole in the JSON API.

It also gives a noticable performance bump encoding JSON objects with
many subobjects, since appending Data is faster than appending String.

### DIFF
--- a/Sources/SwiftProtobuf/Message+JSONAdditions.swift
+++ b/Sources/SwiftProtobuf/Message+JSONAdditions.swift
@@ -34,6 +34,26 @@ public extension Message {
     return visitor.stringResult
   }
 
+  /// Returns a Data containing the UTF-8 JSON serialization of the message.
+  ///
+  /// Unlike binary encoding, presence of required fields is not enforced when
+  /// serializing to JSON.
+  ///
+  /// - Returns: A Data containing the JSON serialization of the message.
+  /// - Throws: `JSONEncodingError` if encoding fails.
+  func jsonUTF8Data() throws -> Data {
+    if let m = self as? _CustomJSONCodable {
+      let string = try m.encodedJSONString()
+      let data = string.data(using: String.Encoding.utf8)! // Cannot fail!
+      return data
+    }
+    var visitor = try JSONEncodingVisitor(message: self)
+    visitor.startObject()
+    try traverse(visitor: &visitor)
+    visitor.endObject()
+    return visitor.dataResult
+  }
+
   /// Creates a new message by decoding the given string containing a
   /// serialized message in JSON format.
   ///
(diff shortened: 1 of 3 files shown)


Sources/SwiftProtobuf/Message+JSONAdditions.swift
```diff
@@ -34,6 +34,26 @@ public extension Message {
...
   }
 
+  /// Returns a Data containing the UTF-8 JSON serialization of the message.
+  ///
+  /// Unlike binary encoding, presence of required fields is not enforced when
+  /// serializing to JSON.
+  ///
+  /// - Returns: A Data containing the JSON serialization of the message.
+  /// - Throws: `JSONEncodingError` if encoding fails.
+  func jsonUTF8Data() throws -> Data {
```
thomasvl: Should jsonString() just call this and create the string from the data?  The visitor's stringResult might also be able to get removed.

allevato: +1

Sources/SwiftProtobuf/Message+JSONAdditions.swift
```diff
@@ -34,6 +34,26 @@ public extension Message {
...
+    if let m = self as? _CustomJSONCodable {
+      let string = try m.encodedJSONString()
+      let data = string.data(using: String.Encoding.utf8)! // Cannot fail!
```
allevato: In the future, it would be nice to get rid of calls to `string.data(using:)` because they're also pretty slow. I have some thoughts on how to do this for encoding that are related to the stream-based stuff I'd like to do for decoders in #420, but since it involves the use of some specific `String` APIs, I'm inclined to wait until Swift 4 is further along in development because those APIs are going to change a great deal and anything I do now might be wasted effort.

tbkka: I think `_CustomJSONCodable` should encode into a `JSONEncoder` instead of just building and returning a String.  That would be a much better fit for streaming encoding, especially for things like `ListValue` and `Struct`.

But that's for a future PR... ;-)

Sources/SwiftProtobuf/Message+JSONAdditions.swift
tbkka: I reimplemented `jsonString()` in terms of `jsonUTF8Data`.  This is not sufficient to remove `stringResult` since that's used in lots of other places (especially by the various implementations of `_CustomJSONCodable.encodedJSONString()`).

I also changed the performance harness to use `jsonUTF8Data` instead of `jsonString`.

PR #436, opened 2017-03-29
Fix up the copyright headers and helper script.

thomasvl:
Also move the script into a sub dir since most people will never need it.


thomasvl:
Not sure what policy Apple has on when to update copyrights in general, I didn't change all the 2016s to 2017s, so running the script takes a little manual review right now not to drop the 2017s that are already in new files from this year.

PR #439, opened 2017-03-30
FLT_DIG and DBL_DIG need to be Int32 to match other C imports

tbkka:
Fixes build warning on Linux


tbkka:
This gets me back to the same six test failures (all due to Swift/Linux treating strings that only contain control characters as equal to the empty string).

PR #441, opened 2017-03-30
Disable known-to-fail tests until the next version of Swift Linux is released

tbkka:
These two sets of tests are known to fail on Swift Linux 3.1 and earlier because of

    https://bugs.swift.org/browse/SR-4218

so only run them on non-Linux or Linux Swift later than 3.1.  Hopefully, this bug will be fixed by then.


tbkka:
With this change, I now see 0 test failures on Ubuntu with Swift 3.1 Release.

PR #445, opened 2017-04-03
Hopefully start making the generated source a little more approachable

thomasvl:


thomasvl:
Note: in the _Files changed_ view, using the _### files_ dropdown to filter to "Sources/protoc-gen-swift" might help see the changes that were needed to explain the diffs in the generated files themselves.  Some ivars had to become `fileprivate` so the extension has access to them.

Assuming this makes sense, there's other things we can continue to move down, this being those two protocols happened to make for a logical checkpoint.

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -167,7 +164,7 @@ class MessageGenerator {
     }
   }
 
-  func generateNested(printer p: inout CodePrinter, file: FileGenerator, parent: MessageGenerator?) {
+  func generateMainStuct(printer p: inout CodePrinter, file: FileGenerator, parent: MessageGenerator?) {
```
tbkka: Should be `generateMainStruct` (with an 'r')

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -314,6 +295,44 @@ class MessageGenerator {
     }
   }
 
+  func generateRuntimeSupport(printer p: inout CodePrinter, file: FileGenerator, parent: MessageGenerator?) {
```
tbkka: I presume the next PR would split this further into `generateProtoNameProviding` and `generateMessageImplementationBase` or some such?

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -314,6 +295,44 @@ class MessageGenerator {
...
+    p.print("\n")
+    p.print("// Support for the runtime.\n")
```
tbkka: This comment seems gratuitous.

Sources/protoc-gen-swift/MessageGenerator.swift
thomasvl: Doh, I stared that that a few times saying something was wrong, but failed to notice it.  :(

Sources/protoc-gen-swift/MessageGenerator.swift
thomasvl: Removed.  I was also debating just doing one comment before all these later decls (from the FileGenerator), but I'm not sure it's worth it, so just removing it.

Sources/protoc-gen-swift/MessageGenerator.swift
thomasvl: Hadn't really thought about it, but makes sense.

Issue #447, opened 2017-04-03
Regression in floating-point handling in JSON (uppercase exponent `E`)


Nicien:
I'm using swift protobuf version v0.9.29 with XCode 8.3 (8E162)

```{"mileage":6.6678272E7}```
The parsing of this JSON fails (`JSONDecodingError.failure`)
because the textual representation of the floating-point number contains an uppercase `E`.

I dig in the source and found that the function `JSONScanner.parseBareDouble` is looking only for the lowercase character `e`.

The comment of the function `parseBareDouble` refers to the RFC 7159.
RFC 7159 says [Page 6]
```
An exponent part begins with the letter E in upper or lower case,
   which may be followed by a plus or minus sign.
```

I see no reason to refuse a floating-point with an uppercase `E`.

Debug steps:

`JSONDecoder.decodeSingularFloatField()` calls
`JSONScanner.nextFloat()` (line 737) calls
`JSONScanner.parseBareDouble`` stop the parsing before the letter `E` because it's uppercase.

Then:
`JSONDecoder.nextFieldNumber()` calls
`try scanner.skipRequiredComma()` (line 40)
finally `JSONScanner.skipRequiredCharacter` fail on the letter `E`

Vivien

tbkka:
Good catch!  Apparently, we forgot to test this case, so it slipped through.

allevato:
Fixed by #448.

PR #448, opened 2017-04-03
fix floating-point handling in JSON which contains uppercase exponent E

Nicien:
This PR fix a regression in floating-point handling in JSON
See issue #447


tbkka:
Thanks!

While you're here, could you please add a test so this doesn't get broken again in the future?

This could be as simple as adding the following just after line 335 in https://github.com/apple/swift-protobuf/blob/master/Tests/SwiftProtobufTests/Test_JSON.swift#L335
```
        assertJSONDecodeSucceeds("{\"singleDouble\":\"1.5E1\"}") {$0.singleDouble == 15}
```

Nicien:
no problem, test added

tbkka:
Thank you!

PR #450, opened 2017-04-04
Clear doesn't return anything.

thomasvl:
Likely a copy/paste that snuck in at some point.


tbkka:
Oh.  Good catch!  I'm a little surprised the compiler didn't complain about that..

PR #454, opened 2017-04-04
Split the oneof Enum generation.

thomasvl:
Move the fileprivate impl details down so the enum defintion is shorter
and more likely just want someone reading the source needs.

### DIFF
--- a/Sources/protoc-gen-swift/MessageGenerator.swift
+++ b/Sources/protoc-gen-swift/MessageGenerator.swift
@@ -224,7 +224,7 @@ class MessageGenerator {
     p.print("\(visibility)var unknownFields = SwiftProtobuf.UnknownStorage()\n")
 
     for o in oneofs {
-      o.generateNested(printer: &p)
+      o.generateMainEnum(printer: &p)
     }
 
     // Nested enums
@@ -304,6 +304,10 @@ class MessageGenerator {
     generateMessageImplementationBase(printer: &p)
     p.outdent()
     p.print("}\n")
+
+    for o in oneofs {
+      o.generateRuntimeSupport(printer: &p)
+    }
 
     // Nested enums and messages
     for e in enums {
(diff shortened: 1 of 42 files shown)


thomasvl:
This leaves the `Equatable` conformance and `==` generation on the main Enum.  We could move this down, but it also sorta seems like something that we might want to keep up top so it is more discoverable, thoughts?

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -327,6 +327,10 @@ class MessageGenerator {
     // Nested messages
     //
 
+    for o in oneofs {
+      o.generateRuntimeSupport(printer: &p)
```
tbkka: Just a thought:  `generateExtensions` might be a more descriptive name here.

thomasvl: _Extensions_ is always an overloaded terms in protos and Swift…  If we have a better name let's bulk do it after all these PRs land.

thomasvl:
@tbkka any thoughts on https://github.com/apple/swift-protobuf/pull/454#issuecomment-291546088 (Equatable and ==)?

Sources/protoc-gen-swift/MessageGenerator.swift
tbkka: Good point.  😛

tbkka:
I think I agree with leaving it where it is.  The generated code for `==` is pretty straightforward, so doesn't provide much in the way of visual clutter.

Sources/protoc-gen-swift/MessageGenerator.swift
thomasvl: generateRuntimeExtensions?  generateSwiftExtensions?  generateRuntimeExtras?

PR #455, opened 2017-04-04
Shorten the code generated for fields in a oneof.

thomasvl:
The non oneof cases already collapse down to a line for the getter
and a line for the setter, so this follows in that style and tries to
generate more compact source so someone looking at the generated code
can more easily see the message's fields vs. the generated boilerplate
to make it all happen.


Reference/conformance/conformance.pb.swift
```diff
@@ -98,28 +98,20 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message {
 
   var protobufPayload: Data {
     get {
-      if case .protobufPayload(let v)? = payload {
-        return v
-      }
+      if case .protobufPayload(let v)? = payload { return v }
       return Data()
```
tbkka: Someday, we should look at `Data` and message-valued fields here and have them return a static empty object instead of constructing a new one.

thomasvl: Opened #457 as an enhancement for this.

PR #458, opened 2017-04-04
Add comment marker for the Swift Extnesions

thomasvl:
- Move the _protobuf_package down to the other overhead things.
- Add a comment to mark where the support stuff starts.


thomasvl:
Here is what Xcode's jump bar displays this as:  
screen shot

Issue #457, opened 2017-04-04
Investigate static values for field defaults


thomasvl:
The getters for proto2 syntax creates `Data()`, strings, and message instances for when the value wasn't set.  It might make sense for some of these to have a static single instance that we use so the creation doesn't have to happen.

tbkka:
A few random thoughts:

* The obvious implementation would add a `staticInstance` to messages and use that to initialize fields.  But that creates another API connection between separately-generated messages that I'd like to avoid.

* For message instances, could we somehow have the default initializer hide the optimization?
```
    public init() {
        self = Self.staticInstance
    }
```
I don't think the above literally works (the private `staticInstance` has to be built somehow), but maybe there's a way to game this.

* Of course, maybe the compiler is already smart enough to do the above for us?

* A static Data() instance could be exposed from the library somewhere?  Or as a single `fileprivate` at the top or bottom of any generated file that needed it?

* I hadn't thought about strings.  I had always assumed that `""` was very fast, but I've not actually measured it.

Obviously, any of the above would want some perf testing to see if it was actually faster.

allevato:
So it turns out the `Data` value type overlay in Swift is... pretty dumb. And `String` has some weird edges too. (I smell some pull requests!)

In the macOS implementations of Foundation, asking for an empty `NSData` or `NSString` always returns the same singleton object; that makes sense because it's always immutable. But in Swift, I've observed the following:

* `Data` doesn't do anything special for the empty case—`let d = Data()` always ends up `malloc`ing a zero-length buffer. However, `let d2 = d` has them both share the same buffer until one is mutated, as we'd expect.

* `let s = ""` (i.e., specifically creating a string from the empty literal) always allocates a new `_StringCore` with a zero-length buffer. *However,* `String()` always creates instances that share the *same* empty storage. This is presumably because they go through different initializers, and the one that initializes from a literal doesn't handle the zero-length special case. Probably an oversight.

* Arrays and dictionaries don't have this problem: when initialized empty, they always share a common empty storage.

So, we potentially have some performance gains if we manage our own empty `Data` singleton until the language fixes that for us. Then, here's what else we can do:

* For messages without heap storage: nothing. Copying a static empty instance of the message isn't going to be any faster than creating a new one.
[...]

PR #459, opened 2017-04-05
Add message with different field types to perf harness.

allevato:
Borne out of discussion in #457.

This supports both proto2 and proto3 (proto2 includes a group). We may want to have a message that does deeper nesting, but that can be done in a separate PR.

I had to reduce the run count from 1000 to 100 in the heterogeneous case because Instruments was prohibitively slow (as in, about 10 minutes).

### DIFF
--- a/Performance/Harness.cc
+++ b/Performance/Harness.cc
@@ -35,7 +35,6 @@ using std::vector;
 Harness::Harness(std::ostream* results_stream) :
     results_stream(results_stream),
     measurement_count(10),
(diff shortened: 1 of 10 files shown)


allevato:
> I'm realizing we also don't cover any type of map<> usage.

Good point. I think I'll leave that for a separate PR though, because I want to put a little more consideration into how we do it. (Is it enough to just use one or a few key-value pairings, or do we need to support more, etc.)

Issue #40, opened 2016-10-04
Validate ReservedWords.swift


thomasvl:
Taking a quick peek, it seems like some of the lists include system reserved and some include things the library itself uses.  Might make sense to document which are which, and see if we can come up with a way to harvest the ones reserved because the library uses them directly from the protocols so we ensure they stay up to date (for example, does `isEmpty` still belong in the lists?).

tbkka:
Finding a clean way to document this list would be good.

`isEmpty` certainly does not belong.

It would also be good to find ways to reduce the number of reserved names.  For example, I've considered moving `protoPackageName` and related metadata into a sub-struct.

thomasvl:
Capturing while I think of it: Can we use the Mirror within the plugin to avoid manually building all of the lists?  Inspect some of the proto classes the plugin, collecting their properties and then filter the list to remove the properties added by the proto definition (`protoFieldNames`). That might help remove some of the manual collection of names.

tbkka:
There's the complication that Swift Protobuf currently overrides the standard reflection (so that the presence or absence of the storage class is invisible to clients).  We could work around that, of course, but it still wouldn't work because:

Swift 3 reflection does not (yet) enumerate methods or computed properties, only stored properties.

When/if that changes, we can certainly investigate this approach.  It's a good idea.

thomasvl:
@tbkka I don't suppose you have any info we can add to the file on how those lists were built?  I'm guessing it comes somewhat from the [Keywords and Punctuation](Rehttps://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/LexicalStructure.html#//apple_ref/doc/uid/TP40014097-CH30-ID413) section of the language.

tbkka:
The original lists were ad hoc copies from another protobuf implementation.  As such, they're actually a mix of C++, Obj-C, and Swift keywords.  (ugh)  They need to be compared to the Keywords and Punctuation section of the Swift spec.

thomasvl:
Will work on this once I finish out the other name changes/visibility changes so we can collect the names the library will use/add to the Enum and Message namespaces.

thomasvl:
Hand off this since it look @tbkka is getting to it before me.

tbkka:
PR #410 made some progress on this.  The lists of Swift keywords and names that we avoid are now derived from lists that directly follow the Swift language spec.

[...]

PR #460, opened 2017-04-06
Issue #40: More testing of field/enum/message names

tbkka:
This introduces an automatic test of field/enum/message names that harvests a word list from the SwiftProtobuf library source code, builds protos from it, and compiles those as part of the test suite.

I've also expanded some of the other naming tests to verify that you can't accidentally shadow basic features (such as "has"/"clear" properties, or the "serializedData()" method) with a poorly-chosen field name.

Based on this, I've reworked some of the name sanitization to protect some of the issues uncovered.

Caveat:  The four large protos added to the test build increase the time for a clean compile of the test suite from 2m30s to over 4m on my laptop.  (The new protos are rebuilt as part of `make regenerate`, so usually they don't have to be recompiled at all.)


Makefile
```diff
@@ -374,14 +378,99 @@ regenerate-plugin-protos: build ${PROTOC_GEN_SWIFT}
...
+Protos/mined_words.txt: Sources/SwiftProtobuf/*
+	@echo Building $@
+	@cat Sources/SwiftProtobuf/* | \
+	grep ' \(public\|func\|var\) ' | \
+	grep -v ' \(private\|internal\) ' | \
+	sed -e 's/[^a-zA-Z0-9_]/ /g' | \
+	tr " " "\n" | \
+	sed -e 's/^_//' | \
+	sort -f | \
+	uniq -i | \
```
thomasvl: sort -u instead of going through uniq?

Protos/generated_swift_names_enum_cases.proto
```diff
@@ -0,0 +1,611 @@
+syntax = "proto3";
```
thomasvl: Do you want the copyright on these and/or something saying it is generated?

Sources/protoc-gen-swift/ReservedWords.swift
```diff
@@ -47,47 +46,45 @@ private let reservedTypeNames: Set<String> = {
...
-func sanitizeMessageTypeName(_ s: String) -> String {
+private func sanitizeTypeName(_ s: String, disambiguator: String) -> String {
     if reservedTypeNames.contains(s) {
-        return s + "Message"
+        return s + disambiguator
     } else if isAllUnderscore(s) {
-        return s + "Message"
+        return s + disambiguator
+    } else if s.hasSuffix(disambiguator) {
+        let e = s.index(s.endIndex, offsetBy: -disambiguator.characters.count)
```
thomasvl: maybe a comment on what this one is covering so when someone stumbles on it they don't have to figure it out?

Sources/protoc-gen-swift/ReservedWords.swift
```diff
@@ -47,47 +46,45 @@ private let reservedTypeNames: Set<String> = {
...
-            "jsonFieldNames",
-            "unknownFields",
+        "debugDescription",
+        "decoder",
+        "description",
+        "dynamicType",
+        "fieldNumber",
+        "hashValue",
+        "isInitialized",
+        "other",
```
thomasvl: other?  decoder?  visitor? where do they come from?

Should we generally comment on this list being from Message.swift?

Sources/protoc-gen-swift/ReservedWords.swift
```diff
@@ -20,15 +20,14 @@ import PluginLibrary
 private let reservedTypeNames: Set<String> = {
     () -> Set<String> in
     var names: Set<String> = [
-        "anyTypeURL",
+        "SwiftProtobuf",
```
thomasvl: Maybe a general comment on where this come from and why?  i.e. - call out the functions that have to block sub types, etc.

Sources/protoc-gen-swift/ReservedWords.swift
```diff
@@ -20,15 +20,14 @@ import PluginLibrary
...
+        "decodeMessage",
         "description",
         "dynamicType",
         "hashValue",
-        "isEmpty",
         "isEqual",
-        "jsonFieldNames",
+        "protoMessageName",
         "traverse",
         "unknownFields",
```
thomasvl: Shouldn't Extensions also go in this list?

Sources/protoc-gen-swift/ReservedWords.swift
```diff
@@ -109,7 +106,11 @@ private let reservedFieldNames: Set<String> =  {
 /// It appends "_p" to any name that can't be
 /// used as a field name in Swift source code.
 func sanitizeFieldName(_ s: String, basedOn: String) -> String {
-    if reservedFieldNames.contains(basedOn) {
+    if basedOn.hasPrefix("clear") || s.hasPrefix("clear") {
```
thomasvl: I don't believe you need to check s in here, just the basedOn.

Sources/protoc-gen-swift/ReservedWords.swift
tbkka: `other`, `decoder`, and `visitor` are local variables within `isEqual`, `decodeMessage`, and `traverse` respectively.  Those local variables shadow properties with those names.

Makefile
tbkka: Done.

Protos/generated_swift_names_enum_cases.proto
tbkka: Ouch.  Makefile blunder left those out.  Fixed.

Sources/protoc-gen-swift/ReservedWords.swift
allevato: Should we try to be smarter here and have a list of words that are just locals like this, and prepend `self.` to their references in those functions? That way we don't waste names that otherwise don't have to be lost otherwise.

tbkka: I'm going to rework the generator to prefix the field references in these methods with `self.`  That's a more robust fix that will also protect us if we need to add other local variables to these methods in the future.

Sources/protoc-gen-swift/ReservedWords.swift
tbkka: Yep, you're right.  Fixed.

Sources/protoc-gen-swift/ReservedWords.swift
tbkka: Done.

Sources/protoc-gen-swift/ReservedWords.swift
tbkka: Good catch!  Fixed.

thomasvl:
Just checking, doesn't look like new comments have shown up, am I reading that correctly?

Sources/protoc-gen-swift/ReservedWords.swift
thomasvl: I guess the other option would be to make our generated code use names that will never be fields or sub types "arg_visitor", etc.

tbkka:
> doesn't look like new comments have shown up, am I reading that correctly?

I started on Friday but didn't finish.  I'll submit the rest of this today.

Sources/protoc-gen-swift/ReservedWords.swift
tbkka: Done.

Sources/protoc-gen-swift/ReservedWords.swift
tbkka: Done.

Sources/protoc-gen-swift/ReservedWords.swift
tbkka: I've implemented the `self.` option instead.  It's pretty simple for the code generator:  It basically just has to prefix each field with either `storage.` or `self.` depending on whether it's using a CoW backing class or not.

tbkka:
I filled in a bunch of comments and in the process caught a few more oversights (the test protos for naming didn't have extensions or required fields, so never tripped over `Extensions` or `isInitialized`).

Sources/protoc-gen-swift/ReservedWords.swift
```diff
@@ -19,18 +19,37 @@ import PluginLibrary
...
+
+    // Subtype of many messages, used to scope nested extensions
+    names.insert("Extensions")
+
+    // Subtypes are static references, so can conflict with static
+    // class properties:
+    names.insert("protoMessageName")
+
+    // Methods on Message that we need to avoid shadowing.
+    // XXX Testing shows we do not need to avoid `serializedData` or
```
thomasvl: XXX?

PR #461, opened 2017-04-06
Fixed oneof isInitialized generation

akievsky:
Previously oneOf was generated with case "None" and oneOf var was non-optional.
Currently, it's generated without "None" case, but optional var, so when isInitialized is generated and all cases in oneOf are messages and have required fields, "default" is not generated which leads to compiler error "Switch must be exhaustive" (nil case is not covered). Looks like having oneOf as optional var requires "default" to be generated always


thomasvl:
Can you add to one of the testing protos something that covers this so we can ensure it doesn't get broken?

allevato:
> Looks like having oneOf as optional var requires "default" to be generated always

Well, either that or a `case nil`/`case .none`.

Can you add a .proto file to the Protos/ directory that demonstrates the problem, so that future work ensures that it continues to work?

Then, can you regenerate the test protos and reference protos in case any of them change because of this generation change? Run `make regenerate` and `make reference` to do that.

akievsky:
Added proto file and generated reference, but not sure about "test protos"

allevato:
You'll need to add the path to your new proto in the `TEST_PROTOS` list in the Makefile (around line 81) and then do `make regenerate` after that.

The reference protos are just used for diffing purposes—it needs to be in `TEST_PROTOS` if you want it to be compiled with the tests, even though we don't have any unit tests that exercise it. But that's enough to make sure we don't break this case.

Makefile
```diff
@@ -129,7 +129,8 @@ TEST_PROTOS= \
 	Protos/unittest_swift_reserved_ext.proto \
 	Protos/unittest_swift_runtime_proto2.proto \
 	Protos/unittest_swift_runtime_proto3.proto \
-	Protos/unittest_swift_startup.proto
+	Protos/unittest_swift_startup.proto \
+    Protos/unittest_swift_oneof_all_required.proto
```
allevato: Can you tab-indent this to match the other lines, instead of spaces, please?

allevato:
Looks good—thanks for catching a tricky edge case!

PR #466, opened 2017-04-06
Performance improvements for String/Data empty values.

allevato:
I've replaced `""` with `String()` throughout the runtime and generated code, and `Data()` with a static singleton `Internal.emptyData`.

The speedup for strings is about 1.3x across the board. The speedup for data varies, with a whopping 14x speedup for creating empty messages with lots of `Data` fields.

Click below for pretty graphs:
Perf visualization

Partially addresses #457 (still need to do statics for non-empty defaults).


thomasvl:
I wonder if we should add a few greps to the Makefile to confirm we don't accidentally add back in Data() and "" usage?

PR #467, opened 2017-04-06
Extension generation factoring

thomasvl:
Move all the generated extension support into a sub section of the file instead of mixing it through the messages and file in general.  It comes after all the types in the file, but before the runtime overhead bits.

### DIFF
--- a/Reference/google/protobuf/unittest_mset.pb.swift
+++ b/Reference/google/protobuf/unittest_mset.pb.swift
@@ -122,15 +122,6 @@ struct ProtobufUnittest_TestMessageSetExtension1: SwiftProtobuf.Message {
 
   var unknownFields = SwiftProtobuf.UnknownStorage()
 
-  struct Extensions {
-
-    static let message_set_extension = SwiftProtobuf.MessageExtension<OptionalMessageExtensionField<ProtobufUnittest_TestMessageSetExtension1>, Proto2WireformatUnittest_TestMessageSet>(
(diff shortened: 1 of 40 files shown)


Reference/google/protobuf/unittest_mset.pb.swift
```diff
@@ -311,11 +295,35 @@ extension Proto2WireformatUnittest_TestMessageSet {
...
+/// this .proto file. It can be used any place an `SwiftProtobuf.ExtensionMap` is needed
+/// in parsing, or it can be combined with other `SwiftProtobuf.SimpleExtensionMap`s to create
+/// a larger `SwiftProtobuf.SimpleExtensionMap`.
 let ProtobufUnittest_UnittestMset_Extensions: SwiftProtobuf.SimpleExtensionMap = [
   ProtobufUnittest_TestMessageSetExtension1.Extensions.message_set_extension,
   ProtobufUnittest_TestMessageSetExtension2.Extensions.message_set_extension
 ]
 
+extension ProtobufUnittest_TestMessageSetExtension1 {
+  struct Extensions {
```
thomasvl: btw - since this is just a namespace for these, should we make it an `enum` instead of a `struct`?  (Since I think we did that in spots within the library for style)

allevato: +1

thomasvl: Done.

PR #470, opened 2017-04-06
Speed up CodePrinter a bit.

allevato:
Apply the ol' `UnicodeScalar`-instead-of-`Character` optimization again, this time to `CodePrinter`. After a few repeated runs, this manages to speed up regeneration by a few seconds (from about ~20 on average to ~14-15 on my machine).

### DIFF
--- a/Sources/protoc-gen-swift/CodePrinter.swift
+++ b/Sources/protoc-gen-swift/CodePrinter.swift
@@ -12,42 +12,73 @@
 /// source code text.
 ///
 // -----------------------------------------------------------------------------
-import Foundation
 
+/// Prints code with automatic indentation based on calls to `indent` and
+/// `outdent`.
 struct CodePrinter {
-    private(set) var content = ""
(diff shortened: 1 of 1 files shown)


Sources/protoc-gen-swift/CodePrinter.swift
```diff
@@ -12,42 +12,73 @@
...
+              contentsOf: scalars[index..<newLineIndex])
+          }
+          content.unicodeScalars.append("\n")
 
-    mutating private func resetIndent() {
-        currentIndent = (0..<currentIndentDepth).map { Int -> String in return "  " } .joined(separator:"")
+          atLineStart = true
+          index = scalars.index(after: newLineIndex)
+        } else {
+          // We reached the end of the string, so just print copy over whatever
```
thomasvl: "print copy" -> "copy" ?

allevato: Woops.

PR #472, opened 2017-04-07
Tweak generated field order.

thomasvl:


thomasvl:
Slight aside - should we include any field comments on the cases of the oneof enum?  Currently we only print them on the fields themselves, but if someone just uses the oneof access, it seems like having them there might also be useful.

allevato:
Seems like a good idea to include the comments there, too.

PR #474, opened 2017-04-07
Include field comments on the oneof enum cases.

thomasvl:


Reference/conformance/conformance.pb.swift
```diff
@@ -230,11 +230,28 @@ struct Conformance_ConformanceResponse: SwiftProtobuf.Message {
   var unknownFields = SwiftProtobuf.UnknownStorage()
 
   enum OneOf_Result: Equatable {
```
thomasvl: I wonder if we should put some canned comment on this to better tie it to the oneof field and/or the oneof comment?

> Used for the oneof field [name].

or

> Used for the oneof field [name].
>
> [oneof field comment]

allevato: The names of the cases are the same as the names of the fields, right? So I don't think it's necessary, especially since the oneof enum is always tightly coupled to the message it's in.

thomasvl: So for this one, the property is _result_, we stick the _OneOf\__ on the front for the enum's name.

thomasvl: Not sure I follow your comment though, yes the cases match the fields, so are you saying we shouldn't include the comments on the cases (i.e. - this PR)?

allevato: Oh, sorry, I thought you meant put comments on the *cases*.

Having a comment on the enum might be valuable. Having it on the property is less valuable since the type is easily findable from the property definition, but the property isn't always easily findable from the enum definition.

thomasvl: Repeated the oneof comment on the enum also.  Hopefully this will let Xcode's help/tips provide some info in all cases.

PR #475, opened 2017-04-08
Re-add Foundation import to Google_Protobuf_Any extension

VFUC:
`import Foundation` is required due to the use of `DispatchQueue`.
This was removed in #417 but makes builds fail on Linux, where explicit import is required.

FYI I've tagged the commit as `0.9.30` to use with SwiftPM on our CI, please let me know if this is an issue. 

Cheers 🙂

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
@@ -12,6 +12,8 @@
 ///
 // -----------------------------------------------------------------------------
 
+// Explicit import of Foundation is necessary on Linux,
+// don't remove unless obsolete on all platforms
 import Foundation
 
 public extension Google_Protobuf_Any {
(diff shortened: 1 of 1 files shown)


Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
```diff
@@ -12,6 +12,8 @@
 ///
 // -----------------------------------------------------------------------------
 
+import Foundation
```
thomasvl: Can you add a comment here that we need to keep this for Linux so we don't break things again by accident.

thomasvl:
> FYI I've tagged the commit as 0.9.30 to use with SwiftPM on our CI, please let me know if this is an issue.

I don't think a tag you your rep has any way to make it up in a PR.  If we do cut something as 0.9.30 here, it likely does mean you run the risk of having a different release for that tag then we have.

Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
VFUC: Certainly.

VFUC:
Alright, thanks for the quick answer! I've added the requested comment.

thomasvl:
Thanks!

Issue #369, opened 2017-03-10
Generate better traverse() for some oneofs


thomasvl:
The `traverse()` we currently generated for oneofs ends up passing a start/end range and then the generated code checks that when going to write out the state.  However if there are no other fields/extension ranges intermixed in the oneof field numbers, then there is no need to pass the start/end and no reason to check them when serializing. And that should be the common case.

We also print the cases within the oneof `traverse` in sorted field number, but we'd probably be better off printing them in the order of the enum, so the compiler has a better chance of generating a jump table for the final code.

PR #477, opened 2017-04-11
Oneof cleanup and generation improvements.

thomasvl:
Fixes https://github.com/apple/swift-protobuf/issues/369.

### DIFF
--- a/Sources/protoc-gen-swift/OneofGenerator.swift
+++ b/Sources/protoc-gen-swift/OneofGenerator.swift
@@ -71,14 +71,9 @@ class OneofGenerator {
             //      }
             //      int32 m = 13;
             //    }
+            let sortedOneofFieldNumbers = fieldsSortedByNumber.map { $0.number }
             let firstIndex = parentFieldNumbersSorted.index(of: first)!
-            var isContinuousInParent = true
-            for i in 0..<fields.count {
-                if fieldsSortedByNumber[i].number != parentFieldNumbersSorted[firstIndex + i] {
-                    isContinuousInParent = false
-                    break
-                }
-            }
+            var isContinuousInParent = sortedOneofFieldNumbers == Array(parentFieldNumbersSorted[firstIndex..<(firstIndex + fields.count)])
             if isContinuousInParent {
                 // Make sure there isn't an extension range in the middle of the fields.
                 //    message AlsoBad {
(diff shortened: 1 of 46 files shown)


Sources/protoc-gen-swift/OneofGenerator.swift
```diff
@@ -49,6 +50,53 @@ class OneofGenerator {
...
+            //      oneof o {
+            //        int32 a = 1;
+            //        int32 z = 26;
+            //      }
+            //      int32 m = 13;
+            //    }
+            let firstIndex = parentFieldNumbersSorted.index(of: first)!
+            var isContinuousInParent = true
+            for i in 0..<fields.count {
+                if fieldsSortedByNumber[i].number != parentFieldNumbersSorted[firstIndex + i] {
```
tbkka: I find this logic a little confusing.  It would be clearer (to me) to use the same logic as you use for extension checks below, namely:  Are any of the parent field numbers between our first and last field numbers?

thomasvl: The catch is the parent list includes the fields in this oneof, so it would be a little more complicated.

Another solution (which I didn't try coding) could be to see if sortedFields == parentFields[firstIndex..<firstIndex+fields.count].

tbkka: Or subtract our fields from the parent fields (set logic), then do the same check as extensions?

thomasvl: Was able to use the slice, how does this seem?

Sources/protoc-gen-swift/OneofGenerator.swift
```diff
@@ -71,14 +71,9 @@ class OneofGenerator {
...
+            let sortedOneofFieldNumbers = fieldsSortedByNumber.map { $0.number }
             let firstIndex = parentFieldNumbersSorted.index(of: first)!
-            var isContinuousInParent = true
-            for i in 0..<fields.count {
-                if fieldsSortedByNumber[i].number != parentFieldNumbersSorted[firstIndex + i] {
-                    isContinuousInParent = false
-                    break
-                }
-            }
+            var isContinuousInParent = sortedOneofFieldNumbers == Array(parentFieldNumbersSorted[firstIndex..<(firstIndex + fields.count)])
```
thomasvl: Compiler doesn't like the compare without the Array() around it (saying you can't compare an Array with and ArraySlice.

PR #483, opened 2017-04-20
Clean up/speed up some Context descriptor lookups.

thomasvl:
- Remove unused swiftNameForProtoName on Context.
- Rename getMessageForPath(path:) to messageDescriptor(forTypeName:) and have
  it use the already built catch so it isn't always linear.
- Remove the linear lookup of Message types.
- Add helper for enum descriptor lookup off Context.
- Drop Context's fileByProtoName, isn't used.
- Stop tracking the parents, wasn't used.
- Stop tracking what type names were groups at the Context level. Groups are
  still messages and the knowledge they are a group is inspected on the field
  when generating, so this was just wasted work.


thomasvl:
There are still lots of other lookups that are linear because they build data up along the way, but that can be a different cleanup.

thomasvl:
Actually, I'll drop this and just do https://github.com/apple/swift-protobuf/issues/484 which will expand on what this happened to notice.

PR #488, opened 2017-04-25
Remove the hand made naming tests.

thomasvl:
Protos/generated_swift_names_*.proto should cover these, so remove
the manually created ones in Protos/unittest_swift_naming.proto.


tbkka:
I'm not at all certain that the generated version does cover everything here.  Have you checked?

thomasvl:
> I'm not at all certain that the generated version does cover everything here. Have you checked?

nope, I was hoping you had.  😄   I'll do some greps this morning.

thomasvl:
Ok, the lists in this file aren't in sync, so there are already some gaps just between each other.  But merging them all and then comparing them to what ends up in `mined_words.txt`, the gaps in `mined_words.txt` are:

1. With the exception of `swiftKeywordsWithNumberSign`, the other word sets in `Sources/PluginLibrary/SwiftLanguage.swift`.
2. A few things like _UNRECOGNIZED_ which clearly matters for enum values.
3. A bunch that seem to be lifted from someplace (maybe ObjC or C++ tests?); and I'm not sure if they really matter to Swift: `union`, `volatile`, `wchar_t`, etc.

For _1_ we could data mine from the source file or switch to generating those lists.  Not sure what we want to do about _2_ and _3_.

Should we open a new issue and move this all there until someone wants to make a run at it?

tbkka:
I did a little grepping myself and didn't find nearly as much overlap as I expected.

I'm happy to put this off a bit.

thomasvl:
Closing now that the info is moved into an issue.

PR #490, opened 2017-04-26
Generate comments for the has*/clear* extension methods.

thomasvl:
Like https://github.com/apple/swift-protobuf/commit/66756fe65af10d3f246c9d89ed98215d78e832aa, but for extensions. Since Extension names tend to be long, the methods weren't folded onto single lines.


thomasvl:
Another option would be to not use the Swift name, but instead use the protobuf name for the extension.  If we want to use the protobuf one, then we might want to add a comment on the property directly to include the name there.

Sources/protoc-gen-swift/ExtensionGenerator.swift
```diff
@@ -170,11 +170,13 @@ struct ExtensionGenerator {
         p.print("set {setExtensionValue(ext: \(swiftFullExtensionName), value: newValue)}\n")
         p.outdent()
         p.print("}\n")
+        p.print("/// Returns true if extension `\(swiftFullExtensionName)`\n/// has been explicitly set.\n")
```
tbkka: Style: Should there be a blank line before this comment?

Sources/protoc-gen-swift/ExtensionGenerator.swift
```diff
@@ -170,11 +170,13 @@ struct ExtensionGenerator {
...
         p.print("\(generatorOptions.visibilitySourceSnippet)var \(swiftHasPropertyName): Bool {\n")
         p.indent()
         p.print("return hasExtensionValue(ext: \(swiftFullExtensionName))\n")
         p.outdent()
         p.print("}\n")
+        p.print("/// Clears the value of extension `\(swiftFullExtensionName)`.\n/// Subsequent reads from it will return its default value.\n")
```
tbkka: And here?

Sources/protoc-gen-swift/ExtensionGenerator.swift
thomasvl: In the regular fields, we currently don't; so I was copying that here.  My thought was just trying to keep things compact if someone does crack open the code to skim things.  I'm inclined to say we let it be in this form until we have a chance to get specific feedback from developers.

tbkka:
> If we want to use the protobuf one, then we might want to add a comment...

I've long wondered if we should add a comment to each property with the proto definition that it was generated from:
```
     // Proto definition: `optional sint64 total_input_errors = 12;`
     var totalInputErrors: Int64 = 0;
```

Sources/protoc-gen-swift/ExtensionGenerator.swift
tbkka: Sounds fine.

thomasvl:
> I've long wondered if we should add a comment to each property with the proto definition that it was generated from:
>```
> // Proto definition: `optional sint64 total_input_errors = 12;`
> var totalInputErrors: Int64 = 0;
>```

So there's a few possible catches:
- Would one only do that when there is no proto file comment to bring over?  Would you add it along with the proto comment and hope it doesn't become noise in the file?
- If one does this, is it the minimal info (like your example), or should one try to reconstruct the full decl?  i.e. - any default value decl, any field options (even though the library doesn't currently capture/expose them), etc.

tbkka:
Here's a straw man:
* I would propose appending it to an existing comment (with a blank line).
* I would start with the minimal info and consider expanding it later.

PR #491, opened 2017-04-26
Add "id" to the list of uppercase names.

thomasvl:
Like "HTTP", "ID" should be all caps in names for Apple's
naming conventions.

### DIFF
--- a/Protos/unittest_swift_naming.proto
+++ b/Protos/unittest_swift_naming.proto
@@ -784,7 +784,7 @@ enum EnumFieldNames {
     unowned = 89;
     weak = 90;
     willSet = 91;
-    id = 92;
+//    id = 92;
     _cmd = 93;
     out = 96;
     bycopy = 98;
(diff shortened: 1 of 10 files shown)


thomasvl:
I got asked about why the generated names were _Id_ instead of _ID_ for fields like `int64 request_id = 1;`, so I figured if we are going to make the change, we should do it before 1.0 since it does change what we generate out of proto files.

Lopdo:
For the record, I am strongly opposed to this change. There is great divide between opinions on Id vs ID. I am in the camp claiming ID should be used when talking about documents (Identification Documents) and Id/id when talking about identifiers (single work, no reason to capitalise). According to swift naming conventions all combinations have their use, ID/id/Id are all valid capitalisations, depending on where they are and what they mean. Forcing everything to ID is imho wrong

Protos/unittest_swift_naming.proto
```diff
@@ -784,7 +784,7 @@ enum EnumFieldNames {
     unowned = 89;
     weak = 90;
     willSet = 91;
-    id = 92;
+//    id = 92;
```
tbkka: I recommend just removing this.

PR #496, opened 2017-04-27
Add encoding/decoding for top-level Arrays of Message types from JSON

zwaldowski:
JSON supports putting both objects and arrays at the top-level. For developers using Protobuf JSON as a bridge to an existing service, it may be necessary to ingest one of these top-level arrays directly as an array of messages.

### DIFF
--- a/Sources/SwiftProtobuf/JSONEncodingVisitor.swift
+++ b/Sources/SwiftProtobuf/JSONEncodingVisitor.swift
@@ -30,6 +30,16 @@ internal struct JSONEncodingVisitor: Visitor {
       return encoder.stringResult
   }
 
+  /// Creates a new visitor for serializing a message of the given type to JSON
+  /// format.
+  init(type: Message.Type) throws {
+    if let nameProviding = type as? _ProtoNameProviding.Type {
+      self.nameMap = nameProviding._protobuf_nameMap
+    } else {
+      throw JSONEncodingError.missingFieldNames
+    }
+  }
+
   /// Creates a new visitor that serializes the given message to JSON format.
   init(message: Message) throws {
     if let nameProviding = message as? _ProtoNameProviding {
@@ -37,6 +47,14 @@ internal struct JSONEncodingVisitor: Visitor {
     } else {
       throw JSONEncodingError.missingFieldNames
     }
+  }
+
+  mutating func startArray() {
+    encoder.startArray()
+  }
+
+  mutating func endArray() {
+    encoder.endArray()
   }
 
   mutating func startObject() {
--- a/Sources/SwiftProtobuf/Array+JSONAdditions.swift
+++ b/Sources/SwiftProtobuf/Array+JSONAdditions.swift
@@ -1,0 +1,85 @@
+// Sources/SwiftProtobuf/Array+JSONAdditions.swift - JSON format primitive types
+//
+// Copyright (c) 2014 - 2017 Apple Inc. and the project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/master/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+///
+/// Extensions to `Array` to support JSON encoding/decoding.
+///
+// -----------------------------------------------------------------------------
+
+import Foundation
+
+/// JSON encoding and decoding methods for arrays of messages.
+public extension Array where Iterator.Element: Message {
+  /// Returns a string containing the JSON serialization of the messages.
+  ///
+  /// Unlike binary encoding, presence of required fields is not enforced when
+  /// serializing to JSON.
+  ///
+  /// - Returns: A string containing the JSON serialization of the messages.
+  /// - Throws: `JSONEncodingError` if encoding fails.
+  func jsonString() throws -> String {
+    let data = try jsonUTF8Data()
+    return String(data: data, encoding: String.Encoding.utf8)!
+  }
+
+  /// Returns a Data containing the UTF-8 JSON serialization of the messages.
+  ///
+  /// Unlike binary encoding, presence of required fields is not enforced when
+  /// serializing to JSON.
+  ///
+  /// - Returns: A Data containing the JSON serialization of the messages.
+  /// - Throws: `JSONEncodingError` if encoding fails.
+  func jsonUTF8Data() throws -> Data {
+    var visitor = try JSONEncodingVisitor(type: Iterator.Element.self)
+    visitor.startArray()
+    for v in self {
+        visitor.startObject()
+        try v.traverse(visitor: &visitor)
+        visitor.endObject()
+    }
+    visitor.endArray()
+    return visitor.dataResult
+  }
+
+  /// Creates a new array of messages by decoding the given string containing a
+  /// serialized array of messages in JSON format.
+  ///
+  /// - Parameter jsonString: The JSON-formatted string to decode.
+  /// - Throws: `JSONDecodingError` if decoding fails.
+  public init(jsonString: String) throws {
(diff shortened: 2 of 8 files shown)


Sources/SwiftProtobuf/JSONEncodingVisitor.swift
```diff
@@ -30,6 +30,16 @@ internal struct JSONEncodingVisitor: Visitor {
...
+  init(type: Message.Type) throws {
+    if let nameProviding = type as? _ProtoNameProviding.Type {
+      self.nameMap = nameProviding._protobuf_nameMap
+    } else {
+      throw JSONEncodingError.missingFieldNames
+    }
+  }
+
   /// Creates a new visitor that serializes the given message to JSON format.
   init(message: Message) throws {
```
zwaldowski: I'm not sure if it makes sense to keep this initializer; the previous protocol extensions can call `JSONEncodingVisitor(type: Self.self)` instead, but I was trying not to touch code outside the scope of this change.

Sources/SwiftProtobuf/Message+JSONAdditions.swift
```diff
@@ -93,3 +93,72 @@ public extension Message {
   }
 }
 
+
+public extension Array where Iterator.Element: Message {
```
tbkka: This should go in a separate file called `Array+JSONAdditions.swift`

tbkka:
The code per se looks pretty good to me, although I could always do with more tests.  ;-)

The question is really just whether this is something that belong in SwiftProtobuf itself.  I'd be interested in hearing what people think about this.

Sources/SwiftProtobuf/Message+JSONAdditions.swift
allevato: This PR looks good! But my main concern about this change is the extensions on `Array` itself. Even though they're constrained to `SwiftProtobuf.Message` here, what would happen if someone else extended `Array` in a non-mutually-exclusive way (such as without a generic constraint at all) to add something like `jsonString` or `init(jsonString:)`? My guess is that the most specifically constrained one would get called, but either way, something would be inaccessible.

Or maybe I'm just worried for nothing, because this is how extensions work in Swift today and there's no way to work around this (you can't selectively import extensions) as the language is designed today. The only other option I see is mangling the names (`protobufJSONString`), which also sucks because (1) we explicitly decided not to do that in the messages themselves so (2) it would make the names inconsistent.

Thoughts?

Sources/SwiftProtobuf/JSONEncodingVisitor.swift
allevato: And we respect you greatly for it. 😄

tbkka: I agree; let's consider that kind of code consolidation in a separate PR.

Sources/SwiftProtobuf/Message+JSONAdditions.swift
zwaldowski: Addressed.

tbkka: This is my biggest concern as well.  Other languages that support similar extension facilities (Javascript) have had real problems from folks extending common system classes this way.  And I see many, many people trying to add JSON extensions to core Swift classes, so this is a real possibility, especially for people transitioning from other systems to SwiftProtobuf.

Another option might be to add static methods to the Message type:
```
     static func arrayFrom(jsonUTF8Data: Data) throws -> [Self] { ... }

     static func jsonUTF8Data(array: [Self]) throws -> Data { ... }
```

Then folks could access this behavior as
```
    let decodedArray =  try MyMessage.arrayFrom(jsonUTF8Data: data);
    let jsonData = try MyMessage.jsonUTF8Data(array: someMessageArray);
```

This may be a little less obvious to some people, but would avoid the name-collision problem.

We could also provide generic functions in the `SwiftProtobuf` namespace rather than tying it into the Message type.

Thoughts?

tbkka: Yet another option:  We could support this by providing a facility to initialize message from `JSONSerialization`-decoded data.  That would provide a reusable building block for folks decoding all sorts of JSON structures.

zwaldowski: Hm. Good point. I understand the concern, and share it.

On the one hand, it is somewhat sandboxed by being a constrained extension. If there were a legitimate ambiguity scenario, it's not impossible to shift things out to a file that doesn't import `SwiftProtobuf`.

One the other hand, a static funcs definitely seems not terrible, if that seems within the API flavor of the library.

zwaldowski: Re: [this](https://github.com/apple/swift-protobuf/pull/496#discussion_r113752518), I'd like to hear a little more of what you're thinking there.

zwaldowski:
Re [this](https://github.com/apple/swift-protobuf/pull/496#issuecomment-297775281), I'm open to keeping it a local extension, but writing it to library-quality was limited by access control. I imagine it's desired to keep all the decoder bits internal. What are our options for an abstraction point?

My project is working to incrementally adopt Protobuf, and our initial workaround for this scenario was to manipulate the incoming JSON to match a more ideal Protobuf layout (`{"response": \(jsonString)}`). My whole goal here it to do _not that_, for sure.

Sources/SwiftProtobuf/Message+JSONAdditions.swift
allevato: We should not provide interoperability with `JSONSerialization`. Our library is an *implementation of protocol buffers*, which has a very specific JSON format. Supporting anything outside that is out of scope.

However, the case in this PR seems allowable because we're talking about an array of objects that have themselves been encoded to protobuf JSON spec. The spec doesn't say anything about that, so we have more flexibility here. If it's useful to people writing APIs that work with arrays of objects, that arguably has value.

On this specific issue (extensions), I'm leaning toward the static methods. I'm not *crazy* about it, but it sounds like a better option.

Rationale: If I were designing a library whose *entire purpose* was to add extensions to Swift's common types, I'd feel differently—people who didn't want those extensions would just not use the library. But since we're a project that aims to provide specific functionality that doesn't currently do that, it seems like the right choice here is to not force extensions on users that they can't separate from the rest of the library. (If we had submodules that weren't implicitly exported to their parent, we could do something nice here, but we don't have submodules at all yet.)

tbkka: > ... a facility to initialize message from JSONSerialization-decoded data. ...

Just as we have decoder methods to initialize a message from a JSON String or from JSON UTF8 data, we could initialize a message from the output of `JSONSerialization`.

That would allow people to do things like:
```
    let decodedJSON = JSONSerialization.jsonObject(with: data)
    let decodedJSONArray = ... verify it's an array ...
    let messageArray = try decodedJSONArray.map { try MyMessage(jsonObject: $0) }
```

This approach has an obvious disadvantage:  It would require adding what is essentially an entire new decoder and encoder to SwiftProtobuf and it would be slower than our current JSON support.

But it also has the advantage that people could use `JSONSerialization` to tear apart essentially arbitrary JSON (not just arrays) and then leverage SwiftProtobuf to type-check and bind pieces of that into generated messages.

To be clear:  This is just brainstorming right now.  I'm not actually proposing we do this.

tbkka: I'm inclined to adding the static methods to `Message` as well.

zwaldowski: Addressed.

zwaldowski:
Updated. Let me know if there's anything else! :)

allevato:
Looks great—thanks for your contribution!

PR #497, opened 2017-04-27
Start moving generation over to Descriptors

thomasvl:
This is a checkpoint to try and keep down how easy the changes are the follow, hopefully that works out in the end.  🤞 

This gets the new classes mostly wired in with changing to many things.  Comment extraction is the main thing changed.  Next step after this will be to start having the Descriptors vend type info and that's where things will change a lot more within the Generators.


Sources/PluginLibrary/Array+Extensions.swift
```diff
@@ -0,0 +1,22 @@
...
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/master/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+
+import Foundation
+
+extension Array {
+  /// Like map, but calls the transform with the index and value.
+  func enumeratedMap<T>(_ transform: (Int, Element) throws -> T) rethrows -> [T] {
```
allevato: Why not just use `yourArray.enumerated().map { ... }` at the call site instead? Your transform will have to change slightly because it'll take a *single* argument that is a tuple containing the index and element, instead of taking them as separate elements, but otherwise, it's the same.

Sources/PluginLibrary/Descriptor+Extensions.swift
```diff
@@ -0,0 +1,107 @@
...
+// https://github.com/apple/swift-protobuf/blob/master/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+
+import Foundation
+import SwiftProtobuf
+
+extension FileDescriptor: ProvidesSourceCodeLocation {
+  // True if this file should perserve unknown enums within the enum.
+  public var hasPreservingUnknownEnumSemantics: Bool {
```
allevato: Super-nit: `hasUnknownEnumPreservingSemantics` might read a little cleaner.

Sources/PluginLibrary/Descriptor.swift
```diff
@@ -56,35 +57,70 @@ public final class DescriptorSet {
...
+
   public let proto: Google_Protobuf_FileDescriptorProto
-  public var name: String { return self.proto.name }
+  public var name: String { return proto.name }
+  public var package: String { return proto.package }
+
+  public let syntax: Syntax
+
+  public var dependencies: [String] { return proto.dependency }
+  public var publicDependencies: [String] { return proto.publicDependency.map { dependencies[Int($0)] } }
```
allevato: Would these benefit from being `lazy` so that the map is only computed once when it's first requested?

Sources/PluginLibrary/Descriptor.swift
```diff
@@ -98,46 +134,66 @@ public final class FileDescriptor {
     self.services.forEach { $0.bind(file: self, registry: registry) }
   }
 
+  // TODO(thomasvl): Eventually hide this and just expose it info off the descriptors so
+  // paths aren't needed externally.
+  public func sourceCodeInfoLocation(path: [Int32]) -> Google_Protobuf_SourceCodeInfo.Location? {
+    guard let location = locationMap[HashableArray(path)] else {
```
allevato: Could we reuse `Foundation.IndexPath` here? (We'd need to clean up the `Int32` to `Int` conversions.)

Sources/PluginLibrary/Descriptor.swift
```diff
@@ -98,46 +134,66 @@ public final class FileDescriptor {
...
+      return nil
+    }
+    return location
+  }
+
+  // Lazy so this can be computed on demand, as the imported files won't need
+  // comments during generation.
+  private lazy var locationMap: [HashableInt32Array:Google_Protobuf_SourceCodeInfo.Location] = {
```
allevato: This `typealias` doesn't seem to buy you much when you use the type-inferred initializer elsewhere.

Sources/PluginLibrary/Array+Extensions.swift
thomasvl: macOS app on my laptop -

```swift
extension Array {
  /// Like map, but calls the transform with the index and value.
  func enumeratedMap<T>(_ transform: (Int, Element) throws -> T) rethrows -> [T] {
    var i: Int = -1
    return try map {
      i += 1
      return try transform(i, $0)
    }
  }
}

struct Blah {
  let title: String
  let index: Int
  init(title: String, index: Int) {
    self.title = title
    self.index = index
  }
}

@NSApplicationMain
class AppDelegate: NSObject, NSApplicationDelegate {

  func applicationDidFinishLaunching(_ aNotification: Notification) {
    let l = ["abc", "def", "ghi", "jkl"]

    var start = DispatchTime.now()
    for _ in 0...5000000 {
      let _ = l.enumerated().map {
        return Blah(title: $0.element, index: $0.offset)
      }
    }
    var end = DispatchTime.now()
    var elapsed = end.uptimeNanoseconds - start.uptimeNanoseconds
    var intervalSeconds = Double(elapsed) / 1000000000
    print("Using enumerated: \(intervalSeconds)")

    start = DispatchTime.now()
    for _ in 0...5000000 {
      let _ = l.enumeratedMap {
        return Blah(title: $1, index: $0)
      }
    }
    end = DispatchTime.now()
    elapsed = end.uptimeNanoseconds - start.uptimeNanoseconds
    intervalSeconds = Double(elapsed) / 1000000000
    print("Using enumeratedMap: \(intervalSeconds)")
  }
}
```

Release build:
Using enumerated: 3.694987967
Using enumeratedMap: 0.961241992

Debug build:
Using enumerated: 20.038512905
Using enumeratedMap: 8.521299144

Want me to add that in a comment within the extension?

Sources/PluginLibrary/Descriptor.swift
thomasvl: In theory, but I'm trying not to make everything lazy and let some things be computed when I figured the generators can do some caching themselves.

Since imports cause the graph to get walked from one node to a completely different one, it isn't easy for a generator to wrap them with caching without rebuilding the whole graph themselves.  So I'm trying to make the graph building and the Swift types the things that end up cached so avoid generating them over an over if a type is reused in different places.

Sources/PluginLibrary/Descriptor+Extensions.swift
thomasvl: Done.  Blindly pulled the C++ naming without much thought.

Sources/PluginLibrary/Descriptor.swift
thomasvl: Done.

Sources/PluginLibrary/Array+Extensions.swift
allevato: ☹️ 

I wonder why that's so much slower... `enumerated()` returns a lazily-evaluated sequence so it's not like it's creating a whole separate intermediate list.

Yeah, leave it and comment it then.

Sources/PluginLibrary/Descriptor.swift
allevato: SG.

Sources/PluginLibrary/Descriptor.swift
thomasvl: Done.

Sources/PluginLibrary/Array+Extensions.swift
thomasvl: Comment added.

PR #498, opened 2017-05-01
Move naming helpers into the PluginLib

thomasvl:


Sources/PluginLibrary/NamingUtils.swift
```diff
@@ -10,8 +10,183 @@
...
+  "rawValue",
+  "self",
+]
+
+
+private func isAllUnderscore(_ s: String) -> Bool {
+  if s.isEmpty {
+    return false
+  }
+  for c in s.characters {
```
allevato: Nit: if you're concerned about performance, iterate over `s.unicodeScalars`, since I'm not sure if my character-literal speedup PR made it into Swift 3.1 😄

thomasvl: Yea, this is just pulling over the existing code.  There's another block coming over next which is similar, a followup pass will likely help, but some of the general improvements of using descriptors is likely going to dwarf these.

PR #501, opened 2017-05-04
Correct TextFormat for Group Extension Fields.

thomasvl:
When used as extension fields, the format of the group name is the
lowercase form.

You can find some examples in google/protobuf's
src/google/protobuf/testdata/text_format_unittest_extensions_data_pointy.txt -

```
[protobuf_unittest.optionalgroup_extension] {
  a: 117
}
```

### DIFF
--- a/Reference/google/protobuf/unittest.pb.swift
+++ b/Reference/google/protobuf/unittest.pb.swift
@@ -7899,7 +7899,7 @@ let ProtobufUnittest_Extensions_optional_bytes_extension = SwiftProtobuf.Message
 
 let ProtobufUnittest_Extensions_OptionalGroup_extension = SwiftProtobuf.MessageExtension<OptionalGroupExtensionField<ProtobufUnittest_OptionalGroup_extension>, ProtobufUnittest_TestAllExtensions>(
   _protobuf_fieldNumber: 16,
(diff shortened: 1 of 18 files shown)


thomasvl:
Also did up a test with protoc to confirm:

foobar.proto:
```protobuf
syntax = "proto2";

package foo.bar;

message MyMessage {
  extensions 1 to max;
}

extend MyMessage {
  optional group Group1Extension = 1 {
    optional int32 a = 2;
  }
}

message ScopingMessage {
  extend MyMessage {
    optional group Group2Extension = 3 {
      optional int32 b = 4;
    }
  }
}
```

foobar.pbascii:
```
[foo.bar.group1extension] {
  a: 1
}
[foo.bar.ScopingMessage.group2extension] {
  b: 2
}
```

Running:
```bash
cat foobar.pbascii \
  | ./protoc --encode=foo.bar.MyMessage foobar.proto \
  | ./protoc --decode=foo.bar.MyMessage foobar.proto
```

produces the same as in input (lowercase), and if you try to change the input to use Group1Extension and Group2Extension, protoc actually errors saying they are unknown.

thomasvl:
Tagging 1.0 since the error with protoc seems to imply what we current generate for text format won't be parsable by other implementations.

PR #503, opened 2017-05-04
Fix reserved word escaping for Extension names.

thomasvl:


Sources/protoc-gen-swift/ReservedWords.swift
```diff
@@ -244,7 +244,7 @@ func sanitizeMessageScopedExtensionName(_ s: String, skipBackticks: Bool = false
   // Since thing else is added to the "enum Extensions" for scoped
   // extensions, there is no need to have a reserved list.
   if reservedEnumCases.contains(s) {
-    return "\(s)"
```
thomasvl: Helps if I hadn't just made a new and identical string out of the old one…  ☹️

PR #506, opened 2017-05-08
Naming wiring

thomasvl:


Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -275,10 +275,12 @@ extension Google_Protobuf_FieldDescriptorProto {
 struct MessageFieldGenerator {
     private let fieldDescriptor: FieldDescriptor
     private let generatorOptions: GeneratorOptions
+    private let namer: SwiftProtobufNamer
 
     var descriptor: Google_Protobuf_FieldDescriptorProto { return fieldDescriptor.proto }
     let oneof: Google_Protobuf_OneofDescriptorProto?
     let jsonName: String?
+    let hasFieldPresence: Bool
```
tbkka: If this always echoes the descriptor value, then I would express it as:
```
    var hasFieldPresence: Bool {return fieldDescriptor.hasFieldPresence}
```

Similarly for most of the other broken-out fields here.

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -130,18 +113,18 @@ class MessageGenerator {
...
-        p.print("\(visibility)static let protoMessageName: String = _protobuf_package + \".\(protoMessageName)\"\n")
+        p.print("\(visibility)static let protoMessageName: String = \(parent.swiftFullName).protoMessageName + \".\(descriptor.name)\"\n")
+    } else if !descriptor.file.package.isEmpty {
+        p.print("\(visibility)static let protoMessageName: String = _protobuf_package + \".\(descriptor.name)\"\n")
     } else {
-        p.print("\(visibility)static let protoMessageName: String = \"\(protoMessageName)\"\n")
+        p.print("\(visibility)static let protoMessageName: String = \"\(descriptor.name)\"\n")
     }
 
     let usesHeadStorage = storage != nil
```
tbkka: Just noticed:  Did you mean this to be `usesHeapStorage`?

thomasvl: Yea, I spotted this today also. It will get cleaned up in the larger changes to field generation.

Sources/protoc-gen-swift/MessageFieldGenerator.swift
thomasvl: So I'm mixed about this.  Making it a computed property means the call go through every time it is used.  Making it a let and assigning it within init means it is called once and cached locally.  So there's actually a big difference in what happens under the hood.

I'd love it if there was a way that wasn't `lazy` to say "this value comes from this sub object", and not force the split world of the decl and collecting in init.

Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -473,7 +466,7 @@ struct MessageFieldGenerator {
     }
 
     func generateHasProperty(printer p: inout CodePrinter, usesHeapStorage: Bool) {
-        if isRepeated || isMap || oneof != nil || (isProto3 && !isMessage) {
+        if !hasFieldPresence {
```
allevato: Nit: `guard hasFieldPresence else { return }`? (I like to use guard for early-returns when possible.)

Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -483,7 +476,7 @@ struct MessageFieldGenerator {
     }
 
     func generateClearMethod(printer p: inout CodePrinter, usesHeapStorage: Bool) {
-        if isRepeated || isMap || oneof != nil || (isProto3 && !isMessage) {
+        if !hasFieldPresence {
```
allevato: Ditto

Sources/protoc-gen-swift/MessageFieldGenerator.swift
tbkka: Performance is not a primary concern for code generation, so I'd prefer whatever is clearest.

thomasvl: Proto generation for a lot of Google's products is a non trivial part of a lot of our builds.  Anyway, this is just a checkpoint to get the namer support in for fields.  I'm still working on a pretty large change to how the message/fields generate so a lot of this will change in follow ups.

Sources/protoc-gen-swift/MessageFieldGenerator.swift
thomasvl: Done

Sources/protoc-gen-swift/MessageFieldGenerator.swift
thomasvl: Done

Sources/protoc-gen-swift/MessageGenerator.swift
thomasvl: Throwing this in an added commit for other comments.

PR #509, opened 2017-05-08
Work around https://bugs.swift.org/browse/SR-1927

thomasvl:
Fixes https://github.com/apple/swift-protobuf/issues/508


thomasvl:
New version up that immediately corrects the index so all uses are correct.

PR #511, opened 2017-05-09
Tbkka documentation

tbkka:
A pass through (most of) the documentation to update it to reflect the current system, clarify discussions, and eliminate some duplication between different documents.


Documentation/INTERNALS.md
```diff
@@ -165,18 +192,16 @@ cases:
  * If there are any fields containing a message or group type
  * If there are more than 16 total fields
 
+This logic will doubtless change in the future:
```
thomasvl: Can't tell, did you mean for this to be a heading of a bullet list or something like that?

Documentation/INTERNALS.md
```diff
@@ -187,28 +212,26 @@ Here is the actual first part of the generated code for `message Foo` above:
...
 public struct Foo: ProtobufGeneratedMessage {
-  public var protoMessageName: String {return "Foo"}
-  public var protoPackageName: String {return ""}
-  public var jsonFieldNames: [String: Int] {return [
-    "fooField": 1,
-  ]}
-  public var protoFieldNames: [String: Int] {return [
-    "fooField": 1,
-  ]}
+  static let protoMessageName: String = "Foo"
```
thomasvl: Does it make sense to just put prose in with references to the generated properties/functions and not try to repeat the generated code since it will surely be stale quickly?

Documentation/INTERNALS.md
```diff
@@ -226,9 +249,8 @@ message Foo {
 }
 ```
 
-This generates a storage class, of course. When the serializer invokes
-`traverse()` on the struct, it simply passes the visitor to `traverse()` on the
-storage class, which looks like this:
+This generates a storage class, of course.
+The storage class and the generated `traverse()` looks like this:
```
thomasvl: Is it worth pointing out the traverse goes in field order (hence the complications with extensions and oneofs)?

Documentation/API.md
```diff
@@ -108,10 +125,23 @@ want to override the default generated behavior for any reason:
 Proto enums are translated to Swift enums in a fairly straightforward manner.
 The resulting Swift enums conform to the `SwiftProtobuf.Enum` protocol which extends
 `RawRepresentable` with a `RawValue` of `Int`.
+The name of the Swift enum is copied directly from the name in the proto file.
+If that name would cause problems for the generated code, the word `Enum` will
+be appended to the name.
+
+Proto2 enums simply have a case for each enum value.
```
thomasvl: This seems partial, did you want to talk about proto3 syntax files having the unknown values vs. pushing things into the unknown fields?

thomasvl: Scratch that, I see it below.

Documentation/API.md
```diff
@@ -198,41 +239,123 @@ invisible to the user.  The property will be serialized if it has a non-default
...
 **Proto2 `required` fields** Required fields behave the same as
 optional fields, except that serialization or deserialization may
 fail if the field is not provided.
 
+To illustrate the handling of proto2 fields, consider the following
+short example:
+```protobuf
+syntax = "proto2";
+message ExampleProto2 {
+   int32 item_count = 1 [default = 12];
```
thomasvl: needs `optional` or `required` since it is proto2 syntax.

Documentation/API.md
```diff
@@ -198,41 +239,123 @@ invisible to the user.  The property will be serialized if it has a non-default
...
+**Singular message fields** generate simple properties of the corresponding
+Swift struct type.
+The fields are initialized with default instances of the struct.
+(This initialization is usually done lazily the first time you read such
+a field.)
+Message fields generate `has` and `clear` methods as above for both proto2
+and proto3.
+
+**Proto2 groups** act exactly like messages in all respects, except that they cannot be
+directly serialized on their own.
```
thomasvl: Actually they can, they are just just messages.  The difference is the binary format is different when it is serialized as a field.

Documentation/INTERNALS.md
tbkka: I didn't want to call that much attention to it.  I've rephrased it some.

Documentation/INTERNALS.md
tbkka: I think even stale code examples here are informative.

Documentation/INTERNALS.md
tbkka: Good point!   I've added a couple of sentences about that.

Documentation/API.md
tbkka: Re-reading, I realize I didn't complete that section; it now contrasts proto2 handling and clarifies that JSON deals with text and numeric enum values differently.

Documentation/API.md
tbkka: Oops!  Fixed.

Documentation/API.md
tbkka: Good catch!  Fixed.

PR #512, opened 2017-05-10
Use a more portable construct for testing the protoc version

tbkka:
The [[ ... ]] bash-ism apparently doesn't work in Makefiles on Ubuntu.


thomasvl:
I'll bet it was dependent on the the fact that /bin/sh on macOS is actually still bash vs. a bourne shell.

PR #516, opened 2017-05-10
Silence the grep used for the version check.

thomasvl:
Otherwise is generated output, and if you happen to have something like
"GREP_OPTIONS=--colour=auto" in your env, you get this color block when
this action happen.


thomasvl:
@tbkka I believe this option on grep is standard, so it should work fine on Linux also.

thomasvl:
Switch to /dev/null just to be safe.  tx.

PR #518, opened 2017-05-10
Update conformance and plugin documentation

tbkka:
Another pass over the Conformance check and Plugin documents.

This is mostly just wordsmithing and reformatting.

The only substantial change is to restructure the Conformance testing discussion to encourage people to just use the `make test` target (which now runs conformance).


tbkka:
The per-language user guides on developer.google.com have a nice structure that I'd like to mimic someday.  But my immediate goal is just to correct our existing docs to unblock the 1.0 release.

Issue #480, opened 2017-04-19
Linux support/coverage for PluginLibraryTests


thomasvl:
Looking at the Makefile, LinuxMain only includes the tests from SwiftProtobufTests.

tbkka:
Fixed in #519

PR #519, opened 2017-05-11
Test and fix issues with Swift 3.0.1+ (Linux) and Xcode 8.1+ (macOS)

tbkka:
I set up a test harness to grind through the full test suite with every Swift Linux release from 3.0.1 onwards.  There turned out to just be a handful of issues:

   * The specialized Linux version of FileIo.readFileData was just broken. The macOS version works fine on Linux, so just use it.
   * I refactored the UTF8-to-String conversion to DTRT with both Linux <3.1 and Linux 3.1+.
   * I fixed up the Makefile a tiny bit
   * I went ahead and tested with the latest dev snapshot, which complained about redundant generic constraints, so I cleaned those up.

There's another issue with Swift Linux 3.0 that I've not looked carefully at yet.


Makefile
```diff
@@ -278,7 +278,8 @@ install: build
 	${INSTALL} ${PROTOC_GEN_SWIFT} ${BINDIR}
 
 clean:
-	swift build --clean
+	-swift build --clean
```
thomasvl: What does the leading "-" on these do?

Sources/SwiftProtobuf/StringUtils.swift
```diff
@@ -50,21 +62,27 @@ internal func utf8ToString(bytes: UnsafePointer<UInt8>, count: Int) -> String? {
     if count == 0 {
         return String()
     }
-#if os(Linux)
-    // As of March, 2017, the NSString(bytes:length:encoding:)
-    // initializer incorrectly stops at the first zero character for
-    // Linux versions of Swift:
+#if !os(Linux) || swift(>=3.2)
```
allevato: Why 3.2? That doesn't actually exist. Wouldn't >=4.0 be equivalent?

Sources/SwiftProtobuf/Internal.swift
```diff
@@ -40,7 +40,7 @@ public enum Internal {
 
   /// Helper to loop over dictionary with values that are Messages to see if
   /// they are all initialized (see Message.isInitialized for what that means).
-  public static func areAllInitialized<K: Hashable>(_ mapToMessages: [K: Message]) -> Bool {
+  public static func areAllInitialized<K>(_ mapToMessages: [K: Message]) -> Bool {
```
allevato: I'm surprised the compiler actually allows this, but I just checked and it does. So I guess they're inferring the `Hashable` conformance now based on the context where it's used in the argument?

Makefile
tbkka: Tells make to ignore these lines if they fail.
So whichever one is not supported on the local platform will not stop the build.  (I considered just dropping this part, since `rm -rf .build` does the same thing.)

Sources/SwiftProtobuf/StringUtils.swift
tbkka: I really wanted `>3.1.1` but that doesn't work.

Has there been any announcement about the next version?  If you know the next version will be 4.0, we could use that here.

Sources/SwiftProtobuf/Internal.swift
tbkka: Yep.  This works in Swift Linux 3.0.1 and later.  I was a little surprised as well; the type-checker is clearly smarter than I gave it credit for.

I think this should work for Xcode 8.2; I'll track that down and try it.

tbkka:
I just tested this branch in Xcode 8.1 and 8.2.  With two minor corrections to the Xcode project, this code works fine there as well.  In particular, those versions of the compiler are also able to handle the simplified generic constraints.

I'm downloading Xcode 8.3 as I type this to test it there as well.

tbkka:
@allevato and @thomasvl asked whether any of these changes (such as eliminating duplicate constraints) affect building with older Xcode.

A:  No.

tbkka:
I just saw #480 which explained why I've been seeing different numbers of tests on the two platforms.  I added a fix for that, so we're now getting full testing on Linux and macOS.

Sources/SwiftProtobuf/StringUtils.swift
allevato: All I know is what's been publicly talked about on the mailing lists—I've heard core team members talk about 4.0 but I've never heard them reference a possible 3.2. If they were planning to do another 3.x point-release with bug fixes, I'd expect that they'd at least have a branch for it on Github though.

PR #520, opened 2017-05-11
Add a generic FieldGenerator interface and move Oneof items to it.

thomasvl:
- Add FieldGenerator for a interface for field generation.
- Add FieldGeneratorBase for some common things.
- Move MessageGenerator and MessageStorageClassGenerator over to using
  the FieldGenerator protocol.
- Add a custom impl of FieldGenerator within OneofGenerator so it can
  forward all work to the OneofGenerator and provide basic caching of
  computed values.
- Overhaul Oneof generation in terms of the new classes.
- Remove the oneof field support from MessageFieldGenerator.

### DIFF
--- a/Reference/unittest_swift_fieldorder.pb.swift
+++ b/Reference/unittest_swift_fieldorder.pb.swift
@@ -429,10 +429,12 @@ struct Swift_Protobuf_OneofTraversalGeneration: SwiftProtobuf.Message, SwiftProt
       try visitor.visitSingularInt32Field(value: v, fieldNumber: 113)
     }
     try self.oConflictField?.traverse(visitor: &visitor, start: 126, end: 127)
+    try self.oConflictExtensionsStart?.traverse(visitor: &visitor, start: 201, end: 202)
     try visitor.visitExtensionFields(fields: _protobuf_extensionFieldValues, start: 202, end: 203)
-    try self.oConflictExtensionsStart?.traverse(visitor: &visitor, start: 201, end: 227)
+    try self.oConflictExtensionsStart?.traverse(visitor: &visitor, start: 226, end: 227)
+    try self.oConflictExtensionsEnd?.traverse(visitor: &visitor, start: 301, end: 302)
     try visitor.visitExtensionFields(fields: _protobuf_extensionFieldValues, start: 325, end: 326)
-    try self.oConflictExtensionsEnd?.traverse(visitor: &visitor, start: 301, end: 327)
+    try self.oConflictExtensionsEnd?.traverse(visitor: &visitor, start: 326, end: 327)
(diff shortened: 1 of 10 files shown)


thomasvl:
Sorry about the size of this, there was no good way to checkpoint along the way to provide incremental chunks without adding a bunch of code that then got removed.

The MessageFieldGenerator still needs a final pass to convert to descriptors, but now that its job is much more defined/limited, it should be easier.

Reference/unittest_swift_fieldorder.pb.swift
```diff
@@ -429,10 +429,12 @@ struct Swift_Protobuf_OneofTraversalGeneration: SwiftProtobuf.Message, SwiftProt
       try visitor.visitSingularInt32Field(value: v, fieldNumber: 113)
     }
     try self.oConflictField?.traverse(visitor: &visitor, start: 126, end: 127)
+    try self.oConflictExtensionsStart?.traverse(visitor: &visitor, start: 201, end: 202)
     try visitor.visitExtensionFields(fields: _protobuf_extensionFieldValues, start: 202, end: 203)
-    try self.oConflictExtensionsStart?.traverse(visitor: &visitor, start: 201, end: 227)
```
thomasvl: Note: with was a bug in the old code.  The generation trying to interleave oneofs and extensions wasn't complete.  Since the new code path uncovered the error in the old generation code; with the added benefit the new code is actually simplier/easier to follow.

Sources/protoc-gen-swift/FieldGenerator.swift
```diff
@@ -0,0 +1,117 @@
...
+  /// The generated code should return false the field isn't set.
+  func generateRequiredFieldCheck(printer: inout CodePrinter)
+
+  /// Generate any support needed to this field's value is initialized.
+  /// The generated code should return false if it isn't set.
+  func generateIsInitializedCheck(printer: inout CodePrinter)
+}
+
+/// Simple base class for FieldGenerators that also provides fieldMapNames.
+class FieldGeneratorBase {
```
allevato: Is there a reason that the subclasses declare their conformance to `FieldGenerator` instead of having this base do it?

thomasvl: The base would have to implement all the methods in the protocol (with dummies that are then overridden).

So the other option here was to instead do provide `fieldMapNames` via an extension, but that would mean the FieldGenerator would have to expose the fieldDescriptor, but I was trying to avoid that so it is clear the MessageGenerator shouldn't be inspecting the fields in any way.

allevato: Ugh, right.

PR #521, opened 2017-05-11
Generate a traverse for each subgroup.

thomasvl:
When the oneof isn't continuous in parent message, generate a different
traverse method for each "group", this avoid the need for conditionals
within the cases.

Name the traverses based off the field number(s) in them so reading of
the message traversal is a little more clear also.

If a traversal switch would only have one case, use the if case syntax
instead.


tbkka:
Does this impact compiled code size at all?  Looks like it might be a net win for long consecutive oneof groups since it drops the conditions.

thomasvl:
> Does this impact compiled code size at all? Looks like it might be a net win for long consecutive oneof groups since it drops the conditions.

Sorry, didn't think to check.

My guess it it will also come down to what the optimizer decides to do since the methods are fileprivate, it can probably also inline things easier now.

thomasvl:
Oh, and since we already had the optimization to now pass the start/end when the oneof was completely continuous, I don't expect this to kick in that often, the previous one is likely to cover most proto files.

PR #524, opened 2017-05-15
Finish the move to Descriptors and remove no longer used code

thomasvl:


Sources/PluginLibrary/Descriptor+Extensions.swift
```diff
@@ -74,6 +74,22 @@ extension FieldDescriptor: ProvidesLocationPath, ProvidesSourceCodeLocation {
     path.append(Int32(index))
   }
 
+  /// Returns true if the type can be used for a Packed field.
+  static func isPackable(type: Google_Protobuf_FieldDescriptorProto.TypeEnum) -> Bool {
```
thomasvl: Style wise, I'm not sure what was best here, I went with a static on FieldDescriptor because while this is sorta an attribute of the Type enum, it is only relevant within the context of when it is a repeated field.  I can move it over to an extension if that seems better.

Sources/PluginLibrary/Descriptor.swift
```diff
@@ -388,7 +388,17 @@ public final class FieldDescriptor {
...
 
-  public var isPacked: Bool { return proto.options.packed }
+  /// Should this field be packed format.
+  public var isPacked: Bool {
+    // This logic comes from the C++ FieldDescriptor::is_packed() impl.
+    guard isPackable else { return false }
+    if file.syntax == .proto2 {
+      return proto.hasOptions && proto.options.packed
+    } else {
+      return !proto.hasOptions || !proto.options.hasPacked || proto.options.packed
```
allevato: It might be worth a comment here that calls out the fact that packed is the default for proto3 with a link to <https://developers.google.com/protocol-buffers/docs/proto3#specifying-field-rules> to save future generations a "🤔"

thomasvl: Added a bunch more in the way of comments.

PR #528, opened 2017-05-15
Revisit enum name stemming.

thomasvl:
Remove the requirement that all enum values having to be prefixed
with the enum name to do any stemming. This should help in two ways:

- If a new enum value is added that doesn't share the prefix, it won't
  suddenly change all the existing names breaking existing code.
- If there is an alias (that could be deprecated) for a old name that
  didn't have the prefix, the existence of it doesn't prevent the newer
  names from getting the stemming and being more "swifty".

Note: The CSharp generator does this, it prefix strips any value it can,
without requiring all of the value to have the enum type as the prefix.

Reminder - This has never been generic common prefix stripping, it is
only about stripping the enum name off the front of the value.  That
specific case is because most protobuf enum follow that pattern since
global C++ enum values are global scope, so a common prefix is needed
on them.


thomasvl:
@tbkka & @allevato thoughts on this before we're locked down with 1.0?

PR #530, opened 2017-05-15
Add some more doc pointers to the generated code.

thomasvl:


thomasvl:
@tbkka / @allevato - I believe this is what we kicked around a while back.  Since the docs are updated, seems like it makes sense to add the pointers.

Reference/conformance/conformance.pb.swift
```diff
@@ -1,10 +1,10 @@
...
- * Source: conformance/conformance.proto
- *
- */
+// DO NOT EDIT.
+//
+// Generated by the Swift generator plugin for the protocol buffer compiler.
+// Source: conformance/conformance.proto
+//
+// For information on using the generated types, please see the documenation:
+//   https://github.com/apple/swift-protobuf/blob/master/Documentation/API.md
```
tbkka: Minor quibble:  I would just refer to https://github.com/apple/swift-protobuf/ here and let the README point people to more specific documentation.  In particular, these generated files get checked into source control and can live there for years before they're regenerated.

Reference/conformance/conformance.pb.swift
```diff
@@ -140,6 +140,9 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message {
 
   init() {}
 
+  /// Used by the SwiftProtobuf decoding machinery, not generally used directly.
+  /// `init(serializedData:)`, `init(jsonUTF8Data:)`, and other decoding initializers
+  /// are defined in `Message` and `Message+*Additions` files.
```
tbkka: I think this comment needs to specifically mention the "SwiftProtobuf library" somewhere.  Perhaps

> Used by the decoding initializers in the SwiftProtobuf library. ...

Or perhaps:

> ... are defined in the SwiftProtobuf library.  See the `Message` and Message+*Additions` files.

Reference/conformance/conformance.pb.swift
```diff
@@ -154,6 +157,9 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message {
     }
   }
 
+  /// Used by the SwiftProtobuf encoding machinery, not generally used directly.
+  /// `Message.serializedData()`, `Message.jsonUTF8Data()`, and other serializer methods
+  /// are defined in `Message` and `Message+*Additions` files.
```
tbkka: As above, we should be clear that these "Message and Message+Additions files" are located in the SwiftProtobuf library source code.

Reference/conformance/conformance.pb.swift
thomasvl: Done.

Reference/conformance/conformance.pb.swift
thomasvl: Basically used both.

PR #534, opened 2017-05-16
Inline oneof traverse

thomasvl:
In working on the oneof parsing, it dawned on me, the helpers we have on the oneof likely are more a result of the old generator code vs. actually being needed from a behavior pov.  So this takes the helpers out of the `traverse` code paths, and instead inlines the groups directly within the message's traverse.  Thoughts?

@tbkka @allevato

### DIFF
--- a/Reference/conformance/conformance.pb.swift
+++ b/Reference/conformance/conformance.pb.swift
@@ -168,7 +168,7 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message {
       try visitor.visitSingularBytesField(value: v, fieldNumber: 1)
     case .jsonPayload(let v)?:
       try visitor.visitSingularStringField(value: v, fieldNumber: 2)
-    default: break
+    case nil: break
     }
     if self.requestedOutputFormat != .unspecified {
       try visitor.visitSingularEnumField(value: self.requestedOutputFormat, fieldNumber: 3)
@@ -325,7 +325,7 @@ struct Conformance_ConformanceResponse: SwiftProtobuf.Message {
       try visitor.visitSingularStringField(value: v, fieldNumber: 5)
     case .serializeError(let v)?:
       try visitor.visitSingularStringField(value: v, fieldNumber: 6)
-    default: break
+    case nil: break
     }
     try unknownFields.traverse(visitor: &visitor)
   }
--- a/Reference/conformance/conformance.pb.swift
+++ b/Reference/conformance/conformance.pb.swift
@@ -163,7 +163,13 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message {
   /// other serializer methods are defined in the SwiftProtobuf library. See the
   /// `Message` and `Message+*Additions` files.
   func traverse<V: SwiftProtobuf.Visitor>(visitor: inout V) throws {
-    try self.payload?.traverse(visitor: &visitor)
+    switch self.payload {
+    case .protobufPayload(let v)?:
+      try visitor.visitSingularBytesField(value: v, fieldNumber: 1)
+    case .jsonPayload(let v)?:
+      try visitor.visitSingularStringField(value: v, fieldNumber: 2)
+    default: break
+    }
     if self.requestedOutputFormat != .unspecified {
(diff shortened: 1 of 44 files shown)


Reference/conformance/conformance.pb.swift
```diff
@@ -163,7 +163,13 @@ struct Conformance_ConformanceRequest: SwiftProtobuf.Message {
   /// other serializer methods are defined in the SwiftProtobuf library. See the
   /// `Message` and `Message+*Additions` files.
   func traverse<V: SwiftProtobuf.Visitor>(visitor: inout V) throws {
-    try self.payload?.traverse(visitor: &visitor)
+    switch self.payload {
```
thomasvl: These could get tweaked to be:

```swift
if let o = self.payload {
  switch o {
  ...
```
And then the `default: break` would only be needed when the oneof has to get split up because of intermixed other fields.

thomasvl: #535 is a parallel PR with that form so both can be looked at to decide if we want to do either one.

allevato: I prefer this one. "if let x ... switch x" is an anti-pattern as far as I'm concerned, since the pattern matching in Swift is capable of resolving the optionality at the same time as the cases themselves.

Tests/SwiftProtobufTests/unittest_swift_fieldorder.pb.swift
```diff
@@ -230,16 +230,26 @@ struct Swift_Protobuf_TestFieldOrderings: SwiftProtobuf.Message, SwiftProtobuf.E
...
+      if case .oneofInt64(let v)? = _storage._options {
+        try visitor.visitSingularInt64Field(value: v, fieldNumber: 60)
+      }
       if let v = _storage._myFloat {
         try visitor.visitSingularFloatField(value: v, fieldNumber: 101)
       }
-      try _storage._options?.traverse_150(visitor: &visitor)
+      if case .oneofString(let v)? = _storage._options {
+        try visitor.visitSingularStringField(value: v, fieldNumber: 150)
+      }
```
tbkka: This is very impressive.  I like how you've handled the interleaving here.

Reference/conformance/conformance.pb.swift
tbkka: Personally, I prefer the "if let x ... switch x" pattern here because:   We have no handling for the `.none` case; and It doesn't visually repeat the optional unwrap on every case.  But I don't feel that strongly about it.

If you do leave it this way, I suggest changing the `default: break` to `case .none: break`   The former breaks the compiler's exhaustiveness checking, hiding possible bugs.

thomasvl: I'm confused, there is no `.none` case, if nothing is set on the oneof, then the value is `nil`.  Hence all the `?` on the case lines.

allevato: `nil` is the same as `.none` since we're switching on an `Optional`. You can also write `case nil`, in fact. Similarly, the pattern `.foo?` is equivalent to `.some(.foo)`.

thomasvl: Uploading new comments now with a `case nil`.   `default` could still get added if any group has >1 field in it and there is >1 group.

Sources/protoc-gen-swift/OneofGenerator.swift
```diff
@@ -404,8 +404,10 @@ class OneofGenerator {
                 p.print("try visitor.visitSingular\(f.protoGenericType)Field(value: v, fieldNumber: \(f.number))\n")
                 p.outdent()
             }
-            // Covers any cases not in this group as well as nil (no oneof field being set).
-            p.print("default: break\n")
+            p.print("case nil: break\n")  // Cover not being set.
```
thomasvl: Any preference, should this be the first case in the switch or the last?

thomasvl:
@tbkka @allevato Do the two new commits look good?

tbkka:
They do look good.

allevato:
+1. Having the `nil` case be last seems right to me.

PR #535, opened 2017-05-16
Inline oneof traverse alternate

thomasvl:
This is #534 but with the `if let` around the switches.


thomasvl:
Closing this, going with the other PR.

PR #536, opened 2017-05-16
Fix oneof generation to check groups for required fields.

thomasvl:


thomasvl:
Added a second commit with a test case so this can't regress.

Tests/SwiftProtobufTests/Test_Required.swift
```diff
@@ -60,6 +60,18 @@ class Test_Required: XCTestCase, PBTestHelpers {
...
+        var msg2 = ProtobufUnittest_OneOfContainer()
+        XCTAssertTrue(msg2.isInitialized)
+        msg2.option3 = ProtobufUnittest_OneOfContainer.Option3()
+        XCTAssertFalse(msg2.isInitialized)
+        msg2.option4 = 1
+        XCTAssertTrue(msg2.isInitialized)
+        msg2.option3 = ProtobufUnittest_OneOfContainer.Option3()
+        XCTAssertFalse(msg2.isInitialized)
+        msg2.option3.a = 1
+        XCTAssertTrue(msg2.isInitialized)
```
tbkka: Interesting test.  It hadn't occurred to me that there were any cases where setting a new value might cause a previously-initialized message to be no longer initialized.  Nice work.

Issue #531, opened 2017-05-15
Investigate using backticks on Enum and Message names as well as field names


thomasvl:
EnumValues and Extension names uses backticks to avoid having to add suffixes in a bunch of cases, but this same sorta logic wasn't done in other place.

Doing this would allow on Enum and Message type names would mean we don't have to "mangle" them; and in most uses of the type, it will be fully qualified, so the backticks won't have to be used (by the generator or developer's code).  So from an overall readability pov there shouldn't be much of a hit, and things will be clearly with the closer mapping to the source .proto files.

On field names, all access of the fields will likely be prefixed with an object (`msg.name`), so no backticks are needed as the type provides the context so the Swift term isn't a problem.

thomasvl:
Tagging this with 1.0 Milestone since it would change apis.  I'm hoping to get this, but it goes at the bottom of the list compared to true behavior things recently found.

thomasvl:
So I made a run at Message and Enum names; the `generated_swift_names_*.proto` files did they job and made just about everything blow up.  The failure is the backticks work when the types are declared, but unlike the enum cases, they don't work when fully qualified; the compiler still seems to try to interpret them, which sorta makes not appending "Enum" or "Message" to them far less interesting/useful.

I'll take a look at field names to see if that does any better.

thomasvl:
Closing out it was done for fields, and the initial attempt for enum and message names didn't go well.

PR #541, opened 2017-05-17
Use backticks on some field names.

thomasvl:


thomasvl:
This will basically fix #531, since it doesn't look like backticks are really doable for enum and message names.

Sources/PluginLibrary/NamingUtils.swift
```diff
@@ -70,6 +69,32 @@ fileprivate let reservedTypeNames: Set<String> = {
...
+ * Many Swift reserved words can be used as fields names if we put
+ * backticks around them:
+ */
+fileprivate let quotableFieldNames: Set<String> = {
+  () -> Set<String> in
+  var names: Set<String> = []
+
+  // We don't need to protect all of these keywords, just the ones
+  // that interfere with enum cases:
+  // names = names.union(swiftKeywordsReservedInParticularContexts)
```
allevato: Can we do something with these commented lines (and below)? It's not clear why they're here.

Sources/PluginLibrary/NamingUtils.swift
```diff
@@ -392,6 +416,10 @@ public enum NamingUtils {
       return s + "_p"
     } else if reservedFieldNames.contains(basedOn) {
       return s + "_p"
+    } else if basedOn == s && quotableFieldNames.contains(basedOn) {
+      // backticks are only used on the base names, if we're sanitizing based on something else
+      // this is skiped (the "hasFoo" doesn't get backticks just because the "foo" does).
```
allevato: s/skiped/skipped/

Sources/PluginLibrary/NamingUtils.swift
thomasvl: Done.  Turns out I never cycled back to finish validation, these ones didn't need to be here.

Sources/PluginLibrary/NamingUtils.swift
thomasvl: Done.

PR #542, opened 2017-05-18
Generator option for mapping proto files to Swift modules

### LINKED ISSUE #294: Story for generating code being used in Swift modules
The current code generation assumes the runtime will come from a module named `SwiftProtobuf` and that all generated code will end up in a single module.  But reality likely won't match that.

Something gRPC likely will have some protos it will bundle, and then the code they generate for a service will likely get built into another module.  So those _service_ will likely need something like _import gRPC_ to get access to the protos shipped there.  And to avoid collisions, any proto that is in _gRPC_ should likely get references in the generated code as _gRPC.[PROTO_NAME]_ (just like we do for the _SwiftProtobuf_ references.

This can also quickly cascade, one package might expose one set of protos, another package might need to reference those, a third provides some independent protos, and then a forth comes in and has protos that reference the ones from two and three.

Initially it might make sense to think of trying to providing a mapping that uses [proto packages](https://developers.google.com/protocol-buffers/docs/proto#packages) to decide what Swift package the code will be in; but that constraint is likely to cause problems.  A collection of proto files might all have the same proto package, but they could be needed by different independent modules, so they want to be in different Swift modules.

For reference the ObjC generator used a [generator options for this](https://github.com/google/protobuf/tree/master/objectivec#objective-c-generator-protoc-options).  Swift being different, the `generate_for_named_framework` doesn't seem like it will apply, as the module the code will end up in hopefully will never be needed because it won't be imported and the types shouldn't need scoping.  So something like `named_framework_to_proto_path_mappings_path` may make the most sense, some form of input let the generator figure out what modules would needed to be imported by each file for the referenced types and how to prefix the types to avoid collisions.

thomasvl:
Fixes #294 

I'm especially happy with the last two commits since it shows how easily this fit into the overhaul of the generator code and the shows it working nicely on all the testing files.


Documentation/PLUGIN.md
```diff
@@ -123,6 +123,40 @@ The possible values for `Visibility` are:
...
+This options allow you to provide a mapping of module names the and proto files that
+will end up in each one. This data is used during generation to then `import` the
+module and scope the types. This option takes an path to where to read a TextFormat
+mapping:
+
+```
+$ protoc --swift_opt=ProtoPathModuleMappings=[path.asciipb] --swift_out=. foo/bar/*.proto
+```
+
+The format of that mapping files defined in
```
allevato: s/files/file is/

Sources/PluginLibrary/SwiftProtobufNamer.swift
```diff
@@ -16,9 +16,31 @@ import Foundation
...
+  public convenience init() {
+    self.init(protoFileToModuleMappings: ProtoFileToModuleMappings(), targetModule: "")
+  }
+
+  /// Initializes a a new namer.  All names will be generated as from the pov of the
+  /// given file using the provided file to module mapper.
+  public convenience init(
+    currentFile file: FileDescriptor,
+    protoFileToModuleMappings mappings: ProtoFileToModuleMappings
+   ) {
```
allevato: Extra leading space

Tests/PluginLibraryTests/Test_ProtoFileToModuleMappings.swift
```diff
@@ -0,0 +1,217 @@
...
+import XCTest
+@testable import PluginLibrary
+
+// Support equality to simplify testing of getting the correct errors.
+extension ProtoFileToModuleMappings.LoadError: Equatable {
+  public static func ==(lhs: ProtoFileToModuleMappings.LoadError, rhs: ProtoFileToModuleMappings.LoadError) -> Bool {
+    switch (lhs, rhs) {
+    case (.entryMissingModuleName(let l), .entryMissingModuleName(let r)): return l == r
+    case (.entryHasNoProtoPaths(let l), .entryHasNoProtoPaths(let r)): return l == r
+      case (.duplicateProtoPathMapping(let l1, let l2, let l3),
```
allevato: Extra leading space

Documentation/PLUGIN.md
thomasvl: Done

Sources/PluginLibrary/SwiftProtobufNamer.swift
thomasvl: Done

Tests/PluginLibraryTests/Test_ProtoFileToModuleMappings.swift
thomasvl: Done.

Documentation/PLUGIN.md
```diff
@@ -123,6 +123,40 @@ The possible values for `Visibility` are:
 * `Public`: The visibility on the types is set to `public` so the types will
   be exposed outside the module they are compiled into.
 
+
+##### Generation Option: `ProtoPathModuleMappings` - Swift Module names for proto paths
+
+By default, the sources generated assume all the generated types will be in the same
+module. However, since protos can reference types from another proto file, there is
+the possibility that those generated files might need to end up different modules.
```
tbkka: "... from another proto file, those generated files might end up in different modules."

Documentation/PLUGIN.md
```diff
@@ -123,6 +123,40 @@ The possible values for `Visibility` are:
...
+This options allow you to provide a mapping of module names the and proto files that
```
tbkka: This option allows you to specify that the code generated from the proto files will be distributed in multiple modules.

Documentation/PLUGIN.md
```diff
@@ -123,6 +123,40 @@ The possible values for `Visibility` are:
...
+will end up in each one. This data is used during generation to then `import` the
+module and scope the types. This option takes an path to where to read a TextFormat
```
tbkka: "... takes the path of a file providing the mapping:"

Documentation/PLUGIN.md
```diff
@@ -123,6 +123,40 @@ The possible values for `Visibility` are:
...
+}
+mapping {
+  module_name: "OtherModule"
+  proto_file_path: "mumble.proto"
+  proto_file_path: "other/file.proto"
+}
+```
+
+The `proto_file_path` values should be just like the `import` paths in the proto files,
+they are the how `protoc` sees the file during generation.
```
tbkka: "The `proto_file_path` values here should match the paths used in the proto file `import` statements."

Documentation/PLUGIN.md
```diff
@@ -123,6 +123,40 @@ The possible values for `Visibility` are:
 * `Public`: The visibility on the types is set to `public` so the types will
   be exposed outside the module they are compiled into.
 
+
+##### Generation Option: `ProtoPathModuleMappings` - Swift Module names for proto paths
+
+By default, the sources generated assume all the generated types will be in the same
```
tbkka: "By default, the code generator assumes all of the resulting Swift files will be put into the same module."

Reference/google/protobuf/test_messages_proto3.pb.swift
```diff
@@ -499,212 +499,212 @@ struct ProtobufTestMessages_Proto3_TestAllTypes: SwiftProtobuf.Message {
   }
 
   /// Well-known types
-  var optionalBoolWrapper: Google_Protobuf_BoolValue {
-    get {return _storage._optionalBoolWrapper ?? Google_Protobuf_BoolValue()}
+  var optionalBoolWrapper: SwiftProtobuf.Google_Protobuf_BoolValue {
+    get {return _storage._optionalBoolWrapper ?? SwiftProtobuf.Google_Protobuf_BoolValue()}
```
tbkka: Nice!

Documentation/PLUGIN.md
thomasvl: Done

Documentation/PLUGIN.md
thomasvl: Done

Documentation/PLUGIN.md
thomasvl: Done

Documentation/PLUGIN.md
thomasvl: Done.

Documentation/PLUGIN.md
thomasvl: Done

PR #543, opened 2017-05-20
Another attempt to explain why optional != Optional

tbkka:
It occurred to me recently that part of the confusion rests in the fact that protobuf binary encoding does omit "unset" fields, which naturally leads to comparisons with nullable Optionals.  So here's another way to explain this distinction by first clarifying that fields omitted from the serialized form are not actually unset, it's just that their value is not transferred.

### DIFF
--- a/Documentation/INTERNALS.md
+++ b/Documentation/INTERNALS.md
@@ -59,12 +59,7 @@ message Foo {
 }
 ```
 
-The original implementation of Swift Protobuf generated properties with
-Swift optional type, but that obvious approach proved problematic.
-In practice, testing whether a field is set is fairly uncommon and using
-Swift optionals directly tended to lead to abuse of the Swift `!` operator
-in many cases where the user knew that certain fields would always be set.
-Instead we generate fields that use Swift optionals internally (to track
+In this case, we generate fields that use Swift optionals internally (to track
 whether the field was set on the message) but expose a non-optional
 value for the field and a separate `hasXyz` property that can be used to
 test whether the field was set:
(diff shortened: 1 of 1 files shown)


Documentation/INTERNALS.md
```diff
@@ -95,25 +90,46 @@ the generated getters.
...
+only indicates whether the field value was explicitly set;
+the field itself still always has a value.
+
+This is fundamentally different than Swift Optionals.
+A Swift Optional is "nullable"; it can explicitly represent
+the absence of a value.
+Proto2 optional fields are not nullable; they always have a value.
+This is a subtle but important difference between the two.
+
+In my own code, I've found that I use proto2 optionals
```
thomasvl: Maybe rephrase this to the less first person?

Documentation/INTERNALS.md
```diff
@@ -95,25 +90,46 @@ the generated getters.
...
+Most of the time, I just use whatever value is there,
+relying on the default when no value was set explicitly.
+But sometimes, I do need to verify that the other side
+explicitly provided the values I expect.
+In these cases, it is far easier for me to assert
+that the received object `has` the expected fields
+in just one place, then use the current values throughout,
+than to unwrap Swift Optionals at every access.
 
 **Proto2 required fields:**
```
thomasvl: It might make sense to point out that "optional" in protobuf is within the context of: "optional", "required", and "repeated", and that context really makes the most sense from the pov of the original binary encoding.  "required" had to be explicitly set or binary encoding was not allowed.  The message is still _valid_ without a "required" field, and it can be included via an option to store/reload it (like a _draft_), but it can be serialized to send to another process.

Documentation/INTERNALS.md
tbkka: Done.

Documentation/INTERNALS.md
tbkka: I suspect that would actually just muddy the discussion:  proto2 "optional" is not simply "not required".

I have however expanded the next section discussing proto2 required fields to clarify exactly when the enforcement of required fields occurs.

PR #544, opened 2017-05-22
Generate a fully qualified name for the extension field type.

thomasvl:


thomasvl:
Looking at other extension stuff, I noticed the type reference in the generated code wasn't fully qualified, which I believe we always want to do.

Issue #502, opened 2017-05-04
Extension support requires the default value twice


thomasvl:
`class MessageExtension` takes a `defaultValue` to init.  So every instance generated gets passed one.

The `extension ExtensibleMessage` provides `getExtensionValue` which checks `_protobuf_extensionFieldValues` and if the extension isn't found, returns the default value.

But then when generating the Swift extension to expose the field on the message, the getter calls `getExtensionValue` and if it isn't a repeated field extension it also generates `?? _the_default_value_again_`.

So for the non repeated cases, have two code paths to create the default instance meaning general bloat.

The instance captured in the MessageExtension has the down side of having to be created during startup of the extension, so avoiding it might actually help startup time for the extension.  But the current flows require it because of the difference between repeated and non repeated extensions.

In anycase, seems like we should revisit this so we can clean things up and only use the default in one of the two cases instead of both.

Marking this as 1.0 Milestone since it like is an api change.

thomasvl:
fyi - #549 moved where the default was needed to avoid needing them at startup.

But #546 was what fixed this by only have the default needed once.

PR #546, opened 2017-05-22
Avoid the generated code needing to capture the Extension default values twice.

thomasvl:


thomasvl:
This ends up needing the defaultValue in two places:
1. Default protocol imp for extension lookup in ExtensibleMessage.
1. During the creation of a new field for decoding (decode is on the extension field).
Ideally, I'd change 1 so it returns nil and let the generator add the ?? default value.  That would let the default value not get created unless it was needed.  But the catch is 2.  We'd have to figure out how to redo decoding so it doesn't have to be done on the field, and instead can just deal with values and let the field get created after the fact.  The catch being extensions that are a message themselves, they would have to see if there in an existing value to do the merging if the field number is found twice.

thomasvl:
Should I bump `compatibilityVersion` as part of this since there is some shuffling of what is generated?

thomasvl:
It might make sense to move the decode from the ExtensionFields to MessageExtension( or MessageExtensionBase).  But just doing that would mean generating a custom decode for *every* extension, so we'd likely need to insert subclasses of MessageExtension for the generator to use like the ExtensionField ones so the decode methods really could be written once (and compile/linked once).

thomasvl:
Anyway, this is pretty good progress on #502, and so lets get this in, and I'll see if I can find something to deal with the default values in the next few days.

tbkka:
This is a nice cleanup.

Please do bump `compatibilityVersion` since this does change the API between the library and the generated code.

thomasvl:
Added a commit to bump `compatibilityVersion`

PR #548, opened 2017-05-22
Performance: `bytes` fields in JSON & TextFormat

tbkka:
Dusting off an old branch that's been sitting around here for a while:  This speeds up 'bytes' decode for JSON and TextFormat by about 4x, and speeds up JSON encode by about 3x.

The JSON/Text decoders here avoid expensive incremental appends by first validating and measuring the field contents, then allocating an output Data directly to the right size, and finally parsing directly into the allocated storage.

The JSON encoder just has a small rework to sidestep the overhead of Data's iterator.

### DIFF
--- a/Sources/SwiftProtobuf/JSONScanner.swift
+++ b/Sources/SwiftProtobuf/JSONScanner.swift
@@ -79,6 +79,7 @@ private func fromHexDigit(_ c: UnicodeScalar) -> UInt32? {
   }
 }
 
+// Decode the RFC 4648 section 4 Base 64 encoding.
 let base64Values: [Int] = [
 /* 0x00 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
 /* 0x10 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
@@ -101,14 +102,17 @@ let base64Values: [Int] = [
 /// Returns a `Data` value containing bytes equivalent to the given
 /// Base64-encoded string, or nil if the conversion fails.
 ///
-/// Notes on Google's implementation:
+/// Notes on Google's implementation (Base64Unescape() in strutil.cc):
 ///  * Google's C++ implementation accepts arbitrary whitespace
 ///    mixed in with the base-64 characters
 ///  * Google's C++ implementation ignores missing '=' characters
 ///    but if present, there must be the exact correct number of them.
-///  * Google's C++ implementation accepts both "regular" and
-///    "web-safe" base-64 variants (it seems to prefer the
-///    web-safe version as defined in RFC 4648
+///
+/// Note: Google's C++ code seems to allow many base-64 extensions
+/// (including websafe and '.' as padding), but the Java version just
+/// uses uses Guava's BaseEncoding.base64() (which only supports the
+/// RFC4648 standard encoding) and the conformance test explicitly
+/// requires us to reject '-' and '_' characters.
 private func parseBytes(
   source: UnsafeBufferPointer<UInt8>,
   index: inout UnsafeBufferPointer<UInt8>.Index,
--- a/Sources/SwiftProtobuf/JSONScanner.swift
+++ b/Sources/SwiftProtobuf/JSONScanner.swift
@@ -79,41 +79,143 @@ private func fromHexDigit(_ c: UnicodeScalar) -> UInt32? {
   }
 }
 
+let base64Values: [Int] = [
+/* 0x00 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+/* 0x10 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+/* 0x20 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 62, -1, -1, -1, 63,
+/* 0x30 */ 52, 53, 54, 55, 56, 57, 58, 59, 60, 61, -1, -1, -1, -1, -1, -1,
+/* 0x40 */ -1,  0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14,
+/* 0x50 */ 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, -1, -1, -1, -1, -1,
+/* 0x60 */ -1, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, 36, 37, 38, 39, 40,
+/* 0x70 */ 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, -1, -1, -1, -1, -1,
+/* 0x80 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+/* 0x90 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+/* 0xa0 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+/* 0xb0 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+/* 0xc0 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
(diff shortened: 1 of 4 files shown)


Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -357,26 +272,153 @@ internal struct TextFormatScanner {
...
+                       // Out of range octal: three digits and first digit is greater than 3
+                      throw TextFormatDecodingError.malformedText
+                    }
+                    p += 1
+                  }
+                }
+                count += 1
+              case asciiLowerX: // 'x' hexadecimal escape
+                // C standard allows any number of digits after \x
+                // We ignore all but the last two
```
thomasvl: What is the comment about ignoring all but the last two about?

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -357,26 +272,153 @@ internal struct TextFormatScanner {
...
-        while p != end {
-            let c = p[0]
-            if c == terminator {
-                let s = utf8ToString(bytes: start, count: p - start)
-                p += 1
-                skipWhitespace()
-                return s
+    /// Scan a string that encodes a byte field, return a count of
+    /// the number of bytes that should be decoded from it
+    private mutating func validateAndCountBytesFromString(terminator: UInt8) throws -> Int {
```
thomasvl: So if you return a tuple (or overload and use a negative result), you could include a flag to say if there are any escape characters.  Then, if there are no escapes (likely the common case), parseBytesFromString can become a simple copy of that may bytes.  i.e. - since you pre scan, might as well collect a hint if you can optimize the later work.

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -357,26 +272,153 @@ internal struct TextFormatScanner {
...
+      throw TextFormatDecodingError.malformedText
+    }
+
+    /// Protobuf Text format uses C ASCII conventions for
+    /// encoding byte sequences, including the use of octal
+    /// and hexadecimal escapes.
+    ///
+    /// Assumes that validateAndCountBytesFromString() has already
+    /// verified the correctness.  So we get to avoid error checks here.
+    private mutating func parseBytesFromString(terminator: UInt8, into data: inout Data) {
```
thomasvl: Why use the terminator vs. using the count you collected?

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -357,26 +272,153 @@ internal struct TextFormatScanner {
...
+                    }
+                    p += 1
+                  }
+                }
+                count += 1
+              case asciiLowerX: // 'x' hexadecimal escape
+                // C standard allows any number of digits after \x
+                // We ignore all but the last two
+                var hexDigits = 0
+                while p != end && fromHexDigit(p[0]) != nil {
```
thomasvl: This seems wrong.  How would one use a hex digit and happen to need to follow it by something that is also hex?  "\x07for fun"

Tokenizer::ParseStringAppend() seems to say the C++ only accepts 1 or 2 following characters.

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: The count is the number of bytes after decoding.  I could add a separate count for the number of bytes before decoding, but I don't think it would actually simplify things.  Using the same terminator checks means that the validation and parsing functions have fundamentally identical structures.

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: I should probably have said "we interpret the result as a number modulo 256".

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -79,41 +79,143 @@ private func fromHexDigit(_ c: UnicodeScalar) -> UInt32? {
...
     }
-    if i % 4 == 3 {
-      out.append(UInt8(truncatingBitPattern: n >> 16))
-      if bits >= 16 {
-        out.append(UInt8(truncatingBitPattern: n >> 8))
-        if bits >= 24 {
-          out.append(UInt8(truncatingBitPattern: n))
+    source.formIndex(after: &index)
+
+    // Count the base-64 digits
```
tbkka: Good point.  I should experiment.  Because you would have to find the close quote first, your idea would boil down to:
* Scan for the close quote (faster than my current first pass)
* Size the output to 3/4 of that
* Parse
* Truncate if necessary (probably not, assuming whitespace is uncommon)
This might also be marginally simpler than my current code, which would be another plus.

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: Another good idea.  I'll try this as well.

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: Oops.  I mis-read that.  Somehow I thought they were following the C/C++ standards (which do allow any number of hex digits).  I'll fix that.

P.S.  The answer to your question:  In C or C++ (which allow any number of hex digits), you would either use an octal escape `"\007for fun"` or use an escape representation of the following character as well: `"\x07\x66or fun"`.

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -79,41 +79,143 @@ private func fromHexDigit(_ c: UnicodeScalar) -> UInt32? {
   }
 }
 
+let base64Values: [Int] = [
```
tbkka: It's the regular one only.

I went back and looked more carefully; although the C++ libraries support a lot of base64 variants, the conformance test and Java versions make clear that only the regular base64 encoding should be supported.

I've updated the comments here to clarify.

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -79,41 +79,143 @@ private func fromHexDigit(_ c: UnicodeScalar) -> UInt32? {
...
-      return nil
+///
+/// Notes on Google's implementation:
+///  * Google's C++ implementation accepts arbitrary whitespace
+///    mixed in with the base-64 characters
+///  * Google's C++ implementation ignores missing '=' characters
+///    but if present, there must be the exact correct number of them.
+///  * Google's C++ implementation accepts both "regular" and
+///    "web-safe" base-64 variants (it seems to prefer the
+///    web-safe version as defined in RFC 4648
```
tbkka: I added a ref to some of the relevant code bits, and removed the reference to the "web-safe" base64 version, since it's not actually supported.

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: I experimented with this a bit, but it doesn't seem to be any faster in practice, so I've left it as-is.

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -79,41 +79,143 @@ private func fromHexDigit(_ c: UnicodeScalar) -> UInt32? {
...
+                    n = 0
+                }
+            } else {
+                switch digit {
+                case asciiDoubleQuote:
+                    source.formIndex(after: &index)
+                    break digits
+                case asciiSpace:
+                    break
+                case 61: // Count padding
```
tbkka: Done.

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: This worked out well.  Thanks for the idea!

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: Comment is now gone.

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: Fixed the behavior here to match C++ protobuf for both strings and bytes fields.

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -619,9 +621,15 @@ internal struct TextFormatScanner {
...
-        result = Data(count: n)
-        parseBytesFromString(terminator: c, into: &result)
+        var sawBackslash = false
+        let n = try validateAndCountBytesFromString(terminator: c, sawBackslash: &sawBackslash)
+        if sawBackslash {
+          result = Data(count: n)
+          parseBytesFromString(terminator: c, into: &result)
+        } else {
+          result = Data(bytes: p, count: n)
+          p += n + 1
```
tbkka: Done.

PR #549, opened 2017-05-23
Extension default handling changes

thomasvl:


thomasvl:
The first commit avoids creating a field and copying the default into it just to decode over it.

The second move where we actually need/create the default, so it doesn't happen at startup, and instead only when the extension is looked up.  This does cause a default to get created each time it is needed, but if we do #457 for the message fields in general, it could also be applied here.

@tbkka / @allevato thoughts?

thomasvl:
ps - the startup concerns are mainly for non macOS uses; those other platforms are where developers are much more likely to be sensitive to the time it takes users to get to their app in a usable state.

Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -1007,11 +1007,22 @@ internal struct BinaryDecoder: Decoder {
...
+            }
             if consumed {
-              values[fieldNumber] = fieldValue
+                if fieldValue != nil {
+                    values[fieldNumber] = fieldValue
+                } else {
+                    // Really things should never get here as the required data
+                    // should always be available or the decode of the extension
+                    // would throw.
+                    throw BinaryDecodingError.truncated
```
allevato: If we should never get here, should we make this a `fatalError` instead of a throw?

thomasvl: I debated that, but the only part haven't fully unwound is what would happen on a binary blob if someone crafted what should be a repeated-packed-pod extension, and then on the wire sent the tag and a length of zero for the data.  I don't think that would blow up in the wire decode, and I currently pick off no values being decoded in `init?`, but that can't directly throw because it doesn't know which type of decoding error to throw.  Things like that have always sorta been a grey area in the spec, it shouldn't happen, but I also can't detect it at the field decode level because the empty list case can happen if the wire type was wrong and that concept is specific to the binary decoder.

I don't think any of the optional extensions could ever end up here as they should always throw their format specific error when they can't extract data (or consumed here would be false for wrong wire types).

We could change the repeated decoded to never return nil, as when doing a merging decode (the other decode method in the ExtensionFields), there is no check to ensure they did decode something.

Should I just tweak the two comments to capture some of these ramblings instead?

allevato: Yeah, we probably should—that way if this line ever *does* get executed, we have a better idea of what might have gone wrong without having to remember all of these details.

Actually... since Swift errors don't come with backtraces, figuring out that it was *this* truncated error vs. some other truncated error gets a little annoying (you'd have to place a breakpoint and wait for things to go south). I wonder if we should introduce a new error case, perhaps with a string reason, to handle these kinds of internal/illegal-state errors that we don't want to bring down the whole world?

thomasvl: In realizing the different in repeated decoding, I decided to make those two behaviors consistent for safety.  So one new commit for that change, and then another commit to add custom errors so if this ever happened it is more trackable.

Reference/google/protobuf/unittest_mset.pb.swift
```diff
@@ -304,7 +304,7 @@ struct ProtobufUnittest_RawMessageSet: SwiftProtobuf.Message {
 
 extension Proto2WireformatUnittest_TestMessageSet {
   var ProtobufUnittest_TestMessageSetExtension1_messageSetExtension: ProtobufUnittest_TestMessageSetExtension1 {
-    get {return getExtensionValue(ext: ProtobufUnittest_TestMessageSetExtension1.Extensions.message_set_extension)}
+    get {return getExtensionValue(ext: ProtobufUnittest_TestMessageSetExtension1.Extensions.message_set_extension) ?? ProtobufUnittest_TestMessageSetExtension1()}
```
tbkka: Double-checking that I understand where you're headed:

I think you're saying that you want to soon change this to be
```
      ... ?? ProtobufUnittest_TestMessageSetExtension1.defaultInstance
```
So that each generated message publishes a single `defaultInstance` to use here and in other similar situations?

We could make that change post-1.0 _except_ for the message types in the library itself.

thomasvl: Yea, that's what we had suggested in #457.  Extensions could also make use of it.  The changes in this PR are more about avoiding the MessageExtension having to have a copy of the default, as that is a startup cost and a memory cost.

thomasvl: @tbkka this PR ok?  The defaultInstance thing would be independent, and is tracked in that other issue since it seems like extensions and fields should be common on this.

tbkka: Apologies for being unresponsive.  Yes, this is fine.  (My question earlier was purely for clarification, not an objection.)

PR #550, opened 2017-05-24
Document building release and static linking the runtime.

thomasvl:
Without the the runtime being a static link, the plugin they copy could
probably suddenly stop working if they update their Xcode.  So play it
safe and document doing static linkage instead.

### DIFF
--- a/Documentation/PLUGIN.md
+++ b/Documentation/PLUGIN.md
@@ -34,11 +34,11 @@ Building the plugin should be simple on any supported Swift platform:
 ```
 $ git clone https://github.com/apple/swift-protobuf
 $ cd swift-protobuf
-$ swift build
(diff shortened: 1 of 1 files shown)


thomasvl:
> Should we add a small shell script (`build.sh`) to automatically pass in the preferred arguments?

There's an _install_ target in the `Makefile`, it reuses the _build_ there, so it is _DEBUG_.  We could change that one?  (or drop it in favor of something else)

allevato:
For me personally, I wouldn't expect to find a build target for a Swift project in a Makefile, and ours is big enough that I doubt most users wouldn't scour through it to find the one they're looking for. A separate small shell script might make it more obvious.

thomasvl:
Landing, we can also revisit a build.sh and the makefile later.

PR #551, opened 2017-05-25
Regenerate.  #548 likely caused some of the name shuffling.

thomasvl:


tbkka:
Yeah, this is a problem with the generated name protos; since they mine the SwiftProtobuf library sources for possibly-dangerous words, they're very sensitive to any code change in the library.

PR #552, opened 2017-05-25
Add a static defaultInstance to the _StorageClasses.

thomasvl:
Start each message with this static instance so there is no heap
allocation until the value is actually mutated. This should be
the biggest win when getting a message field that hasn't been
set as it avoids a heap allocation in that read only case.


thomasvl:
I believe this is the last thing on https://github.com/apple/swift-protobuf/issues/457.

thomasvl:
There is one potential follow up from a performance pov, but i'm not convinced the numbers will be more than noise, and it comes at the cost of a little more codegen.

Current generation is:

```swift
  fileprivate mutating func _uniqueStorage() -> _StorageClass {
    if !isKnownUniquelyReferenced(&_storage) {
      _storage = _StorageClass(copying: _storage)
    }
    return _storage
  }
```

Since every message starts off with the `defaultInstance`, the first mutation of every message is a `_StorageClass(copying:)`.  But before this change, we started with an `init` since we create those heap instances.  So we are trading an `init` for the copy of those default values.  The copy has the potential to be more expensive because it requires reading from another set of addresses and writing into the new instance.  Likewise, the compiler could be smart, and if all the initial values were zero, it might `bzero`/etc. the new instance instead of individual assignments during `init`, where the copy has to be value lookups since they aren't known at compile time.

So we could generate something like:

```swift
  fileprivate mutating func _uniqueStorage() -> _StorageClass {
    if !isKnownUniquelyReferenced(&_storage) {
      if _storage === _StorageClass.defaultInstance {
        _storage = _StorageClass()
      } else {
        _storage = _StorageClass(copying: _storage)
     }
    }
    return _storage
  }
```

Thoughts?

allevato:
I doubt the compiler can be that smart (the `bzero` example) because the fields in the storage class are optionals and optionals have their discriminator set to 1 when they aren't set.

It could be worth exploring the perf side effects of that change separately—I don't know off the top of my head what kind of gains it would have.

tbkka:
Regarding @thomasvl 's idea about restructuring `_uniqueStorage`, I played a while back with the idea of changing that function to not `return _storage` with the idea that it might be more efficient to assert uniqueness and then have clients just use the `_storage` variable directly.  I could find no performance impact at all when I measured it, though, so I've not explored it further.

PR #554, opened 2017-05-25
Support field number parsing in TextFormat

### LINKED ISSUE #176: Add unknown field support to Text encoder/decoder
As pointed out by @thomasvl :  "If we do swap debugDescription to use text format (#148), we likely will want them so the info is complete."

thomasvl:
Fixes #176

### DIFF
--- a/Tests/SwiftProtobufTests/Test_TextFormat_Map_proto3.swift
+++ b/Tests/SwiftProtobufTests/Test_TextFormat_Map_proto3.swift
@@ -47,11 +47,51 @@ class Test_TextFormat_Map_proto3: XCTestCase, PBTestHelpers {
         assertTextFormatDecodeFails("map_int32_int32 {key:1}")
     }
 
+    func test_Int32Int32_numbers() {
+        assertTextFormatDecodeSucceeds("1 {\n  key: 1\n  value: 2\n}\n") {(o: MessageTestType) in
+            return o.mapInt32Int32 == [1:2]
+        }
+        assertTextFormatDecodeSucceeds("1 {key: 1, value: 2}") {(o: MessageTestType) in
+            return o.mapInt32Int32 == [1:2]
+        }
+        assertTextFormatDecodeSucceeds("1 {key: 1; value: 2}") {(o: MessageTestType) in
+            return o.mapInt32Int32 == [1:2]
+        }
+        assertTextFormatDecodeSucceeds("1 {key:1 value:2}") {(o: MessageTestType) in
+            return o.mapInt32Int32 == [1:2]
+        }
+        assertTextFormatDecodeSucceeds("1 {key:1 value:2}\n1 {key:3 value:4}") {(o: MessageTestType) in
+            return o.mapInt32Int32 == [1:2, 3:4]
+        }
+        assertTextFormatDecodeSucceeds("1 [{key:1 value:2}, {key:3 value:4}]") {(o: MessageTestType) in
+            return o.mapInt32Int32 == [1:2, 3:4]
+        }
+        assertTextFormatDecodeSucceeds("1 [{key:1 value:2}];1 {key:3 value:4}") {(o: MessageTestType) in
+            return o.mapInt32Int32 == [1:2, 3:4]
+        }
+        assertTextFormatDecodeFails("1 [{key:1 value:2},]")
+        assertTextFormatDecodeFails("1 [{key:1 value:2}")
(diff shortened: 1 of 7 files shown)


Tests/SwiftProtobufTests/Test_TextFormat_Map_proto3.swift
```diff
@@ -47,11 +47,82 @@ class Test_TextFormat_Map_proto3: XCTestCase, PBTestHelpers {
...
+        }
+        assertTextFormatDecodeSucceeds("1 {1:1 2:2}\n1 {1:3 2:4}") {(o: MessageTestType) in
+            return o.mapInt32Int32 == [1:2, 3:4]
+        }
+        assertTextFormatDecodeSucceeds("1 [{1:1 2:2}, {1:3 2:4}]") {(o: MessageTestType) in
+            return o.mapInt32Int32 == [1:2, 3:4]
+        }
+        assertTextFormatDecodeSucceeds("1 [{1:1 2:2}];1 {1:3 2:4}") {(o: MessageTestType) in
+            return o.mapInt32Int32 == [1:2, 3:4]
+        }
```
tbkka: Should we include a case with `,` and `;` separators within the map elements?  E.g.
```
     1 [{1:1,2:2},{1:3;2:4}]
```

tbkka:
How close are we now to being able to support the following?  If this works, we could expand the textFormat helper in TestHelpers to get pretty good coverage of our ability to round-trip unknown fields in TextFormat.
```
    let empty = Google_Protobuf_Empty(serializedData: message.serializedData()) // All unknown
    let withNumbers = empty.textFormatString() // Encode with numbers
    let decodedFromNumbers = MessageTestType(textFormatString: withNumbers)
    XCTAssertEqual(message, decodedFromNumbers)
```

Tests/SwiftProtobufTests/Test_TextFormat_proto3.swift
```diff
@@ -1389,4 +1396,87 @@ class Test_TextFormat_proto3: XCTestCase, PBTestHelpers {
...
+            + "}\n"
+            + "19 {\n"
+            + "  c: 88\n"
+            + "}\n"
+            + "20 {\n"
+            + "  d: -9\n"
+            + "}\n"
+            + "21: BAZ\n"
+            + "22: FOREIGN_BAZ\n"
+            + "23: IMPORT_BAZ\n"
```
tbkka: I think enums should also be supported as numbers; should we use that here?

thomasvl: Can you spin an issue for that, I had assumed that was already support, and it seems independent of supporting using field numbers for input.

thomasvl:
>How close are we now to being able to support the following? If this works, we could expand the textFormat helper in TestHelpers to get pretty good coverage of our ability to round-trip unknown fields in TextFormat.
>
>```swift
>    let empty = Google_Protobuf_Empty(serializedData: message.serializedData()) // All unknown
>    let withNumbers = empty.textFormatString() // Encode with numbers
>    let decodedFromNumbers = MessageTestType(textFormatString: withNumbers)
>    XCTAssertEqual(message, decodedFromNumbers)
>```

I left a comment in the `nextFieldNumber` code, comparing to the C++, it supports an option to non error and skip the value.  It uses the option for both the named or number case.  For the field name case, this makes sense because if you don't know the name, there's nothing you can do.  For field numbers, you'd think you might be able to push it to unknowns; but reality is different.  Given "1: 1", is that a bool, int32, sint32, int64, sint64, etc.?  There isn't really any way to know.

Tests/SwiftProtobufTests/Test_TextFormat_Map_proto3.swift
thomasvl: I just duplicated what was already here with names.  I'll admit I hadn't realized `;` was valid in general.

Is this another thing to spin into a sep issue since it doesn't seem specific to field numbers in TextFormat?

Tests/SwiftProtobufTests/Test_TextFormat_proto3.swift
thomasvl: Opened #555

Tests/SwiftProtobufTests/Test_TextFormat_Map_proto3.swift
thomasvl: Opened #556

Issue #444, opened 2017-04-03
Add a message depths to parsing


thomasvl:
Java, C++, etc. have a limit to how many deep messages can be nested when parsing.

https://github.com/google/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L62
> `private static final int DEFAULT_RECURSION_LIMIT = 100;`

https://github.com/google/protobuf/blob/master/src/google/protobuf/io/coded_stream.cc#L86
> `int CodedInputStream::default_recursion_limit_ = 100;`

...

In some cases they use a default and allow a per parse override (usually on the input stream).  At quick glance, I don't see anything in the spec requiring this, but I believe it it meant sorta as a safety net when parsing invalid input.  On the client side, this likely isn't too important since clients will likely be talking to trusted servers via https, but on the server side, it may well be more important to help prevent DOS attacks.

thomasvl:
fyi - I think I saw a depth limit for the JSON code pathes also.

PR #560, opened 2017-05-26
JSON Decode message nesting depth limit

thomasvl:
JSON side of #444


Sources/SwiftProtobuf/JSONDecoder.swift
```diff
@@ -24,12 +25,15 @@ internal struct JSONDecoder: Decoder {
     throw JSONDecodingError.conflictingOneOf
   }
 
-  internal init(source: UnsafeBufferPointer<UInt8>) {
-    self.scanner = JSONScanner(source: source)
+  internal init(source: UnsafeBufferPointer<UInt8>, options: JSONDecodingOptions?) {
```
allevato: Instead of making this optional, how about making it non-optional with a default value of `JSONDecodingOptions()`? (Throughout.)

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -652,9 +654,27 @@ internal struct JSONScanner {
...
+      throw JSONDecodingError.messageDepthLimit
+    }
+  }
+
+  private mutating func decrementRecursionDepth() {
+    recursionBudget += 1
+    // This should never happen, if it does, something is probably corrupting memory, and
+    // simply throwing doesn't make much sense.
+    if recursionBudget > recursionLimit {
+      fatalError("Some how JSONDecoding unwound more objects than it started")
```
allevato: s/Some how/Somehow/

thomasvl: Done.

Sources/SwiftProtobuf/JSONDecoder.swift
thomasvl: Done

thomasvl:
New commits up, with a slight reorder to avoid having to regenerate again (i.e. - updated that commit)

PR #561, opened 2017-05-26
Fix bytes hash with Swift 3.0.x

tbkka:
Apparently, Data.hashValue was broken at one time.

Note:  This still uses the native Data.hashValue with Swift 3.1 and later.

### DIFF
--- a/Sources/SwiftProtobuf/HashVisitor.swift
+++ b/Sources/SwiftProtobuf/HashVisitor.swift
@@ -79,7 +79,15 @@ internal struct HashVisitor: Visitor {
 
   mutating func visitSingularBytesField(value: Data, fieldNumber: Int) throws {
     mix(fieldNumber)
+#if swift(>=3.1)
     mix(value.hashValue)
+#else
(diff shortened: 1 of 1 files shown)


Sources/SwiftProtobuf/HashVisitor.swift
```diff
@@ -79,7 +79,15 @@ internal struct HashVisitor: Visitor {
 
   mutating func visitSingularBytesField(value: Data, fieldNumber: Int) throws {
     mix(fieldNumber)
+#if swift(>=3.1)
     mix(value.hashValue)
+#else
```
thomasvl: Is there a swift bug we can link to to explain this?

tbkka: I'll take a look and see if I can find something relevant to comment.

tbkka:
I found the bug report on bugs.swift.org and added a comment in PR #569.

thomasvl:
> I found the bug report on bugs.swift.org and added a comment in PR #569.

Thanks!

PR #564, opened 2017-05-30
Move binary message decoding completely within the BinaryDecoder.

thomasvl:
While working on the recursion limit for binary, I was having to share some knowledge over in the extension, and realized this could be moved since it was a left over from when we needed two different message behaviors.


thomasvl:
@allevato do I need to bump the version number stamped in the generated sources for this?

allevato:
@thomasvl Should we bother with that pre-1.0?

thomasvl:
Wasn't sure, we bumped it with the last api change.

allevato:
Then I guess for consistency, do it here. It probably doesn't matter though.

tbkka:
Where is the API change here?  I see some shuffling of internal methods, but nothing that affects the API between generated code and library.  The version lock is intended to protect running generated code against an incompatible library; changes that do not affect the generated code can't require updating the version lock.

thomasvl:
The removal of `_protobuf_mergeSerializedBytes` was what I wasn't sure of.  Since it is an extension and it gets called from a specialized method, I wasn't sure exactly how that would interact.

Thinking about it again, I guess it gets bound at compile, so as long as they recompile after updating SwiftProtobuf things would be ok?

tbkka:
The `_protobuf_mergeSerializedBytes` method was internal, so it's removal doesn't impact the API at all.

PR #568, opened 2017-05-30
Add a message depths to parsing binaries

### LINKED ISSUE #444: Add a message depths to parsing

thomasvl:
Serialized binary side of #444

Fixes #444


Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
```diff
@@ -96,13 +98,17 @@ internal struct TextFormatEncodingVisitor: Visitor {
               var bytes = Internal.emptyData
               try decoder.decodeSingularBytesField(value: &bytes)
               bytes.withUnsafeBytes { (p: UnsafePointer<UInt8>) -> () in
-                  var testDecoder = BinaryDecoder(forReadingFrom: p, count: bytes.count)
+                  var testDecoder = BinaryDecoder(forReadingFrom: p,
+                                                  count: bytes.count,
+                                                  parent: decoder)
```
tbkka: Should we plan to add recursion depth limits to the TextFormat decoder as well?  If so, would it use a separate `Options` struct?

thomasvl: Since TextFormat should never be a _wire format_, I wasn't planning on added it.

My thought was to keep independent ones, just like the Error object, so they don't have to stay in sync and we don't have to try to document what applies where.  Also, if someone did want to write a custom decoder, they might need their own options so a custom options object wouldn't help them.

Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -46,12 +50,47 @@ internal struct BinaryDecoder: Decoder {
...
         fieldStartP = p
         self.extensions = extensions
+        self.options = options
+        recursionBudget = options.messageDepthLimit
+    }
+
+    internal init(
+      forReadingFrom pointer: UnsafePointer<UInt8>,
+      count: Int,
+      parent: BinaryDecoder
```
allevato: Looking more closely at this (I don't remember if the JSON decoder had the same thing), since `BinaryDecoder` is a `struct`, it seems like this effectively copies the decoder when it's passed in as a parent, and then discards the copy afterwards (assuming the compiler isn't smart enough to elide the copy). Do we know if this has a negative performance impact?

thomasvl: JSONDecoder does the same thing.

I'm trying to check, but running `Performance/perf_runner.sh -c 40211b54097692154f0636005cddfe9f2c09dfad` seems to be changing files in my local repo, so I'm not sure if I'm running it correctly or not.

thomasvl: screen shot 2017-05-30 at 2 56 24 pm

Which seems to say binary didn't change, but JSON go faster with this change?

thomasvl: Adding a commit to make the parent inout and then comparing that to this CL seems to be reporting everything was faster at this commit, so the inout slows things down.  So I'm not completely sure what to say here.  Do we want to get this in and then you can look?  Strangely, if I add a commit that isn't pushed off my local machine and then run comparing to the previous commit, the script always exists with a change staged and local edits.  The staged changes seem to be without my last comment and the unstagged changes seem to to be that last commit, so it's like a reverse/restore is happening via the script?

allevato: Those differences look like noise, probably. I'd say go ahead and submit this since there doesn't seem to be a significant impact.

As for the modification, all the checkouts should be happening in a temp location but maybe there's an error in the script somewhere...

PR #571, opened 2017-05-30
Add a BinaryDecodingOption to discard unknown fields.

thomasvl:
As part of the change to proto3 syntax, some other languages are
providing means to drop unknowns while parsing and/or an api to
strip unknowns from the object graph. Stripping from the graph
would require generating a new method (since we don't have
reflection), but it is pretty easy to add an option now to drop
the unknowns during parsing.

- While at it, tweak the repeated enum parsing to not create the array
  until it is needed (since it likely isn't needed the majority of the
  time).
- Add tests to cover the option.


Sources/SwiftProtobuf/BinaryDecodingOptions.swift
```diff
@@ -21,5 +21,18 @@ public struct BinaryDecodingOptions {
...
+  /// Discard unknown fields while parsing.
+  ///
+  /// The Protobuf binary format allows unknown fields to be still parsed
+  /// so the schema can be expanded without requiring all readers to be updated.
+  /// This works in part by haivng any unknown fields preserved so they can
+  /// be relayed on without loss. For a while the proto3 syntax definition
+  /// called for unknown fields to be dropped, but that lead to problems in
+  /// some case. The default is to follow the spec and keep them, but setting
+  /// this option to `true` allows a developer to strip them during a parse
+  /// incase they have a specific need to drop the unknown fields from the
```
tbkka: "in case" needs a space.

tbkka: We should probably include a sentence specifying the default as part of the doc comment.  In this case,
```
By default, the parser does not discard unknown fields.
```

thomasvl: Doh, added a sentence at the start like we have for messageDepthLimit.

PR #573, opened 2017-05-31
Update the collection done for mined_words.txt.

thomasvl:
Strip out comments from lines matching.

Use extended regexes with grep to use word breaks instead of spaces when
matching, the greps were failing to match lines that start with the terms
which means the type level types/functions. The mostly still go collected
when the got used as a type in some other code, but this ensure they are
collected.


thomasvl:
@tbkka can you double check that this will work on linux?

PR #575, opened 2017-05-31
Group the Swift Extensions for Proto Extensions

### LINKED ISSUE #492: Group the proto extensions by extendee
When we generate the Swift extension for a protobuf extension, it is a new `extension [NAME] { ... }` block each time, but in the proto file, you actually can declare them in a single `extend [NAME] { ... }` block.  It would probably make for slightly better code the generated code did one Swift extension for all the extensions to the class.

thomasvl:
_tl;dr;_ -

```protobuf
extend MyMessage {
  optional int32 a = 100;
  optional string b = 101;
}
```

used to generate:

```swift
extension MyMessage {
  var a: Int32 { ... }
  var hasA: Bool { ... }
  func clearA() { ... }
}

extension MyMessage {
  var b: String { ... }
  var hasB: Bool { ... }
  func clearB() { ... }
}
```  

but now it generates:

```swift
extension MyMessage {
  var a: Int32 { ... }
  var hasA: Bool { ... }
  func clearA() { ... }

  var b: String { ... }
  var hasB: Bool { ... }
  func clearB() { ... }
}
```  

Fixes #492

### DIFF
--- a/Sources/protoc-gen-swift/ExtensionGenerator.swift
+++ b/Sources/protoc-gen-swift/ExtensionGenerator.swift
@@ -46,9 +46,46 @@ class ExtensionGenerator {
         }
 
         func generateMessageSwiftExtensions(printer p: inout CodePrinter) {
-            for e in extensions {
+            guard !extensions.isEmpty else { return }
+
+            // Reorder the list so they are grouped by the Message being extended, but
+            // maintaining the order they were within the file within those groups.
+            let grouped: [ExtensionGenerator] = extensions.enumerated().sorted {
+                // When they extend the same Message, use the original order.
+                if $0.element.containingTypeSwiftFullName == $1.element.containingTypeSwiftFullName {
+                    return $0.offset < $1.offset
+                }
+                // Otherwise, sort by the Message being extended.
+                return $0.element.containingTypeSwiftFullName < $1.element.containingTypeSwiftFullName
+            }.map {
+                // Now strip off the original index to just get the list of ExtensionGenerators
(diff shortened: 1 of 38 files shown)


Sources/protoc-gen-swift/ExtensionGenerator.swift
```diff
@@ -19,6 +19,101 @@ import PluginLibrary
 import SwiftProtobuf
 
 class ExtensionGenerator {
+
+    // Helper to hold all the extensions when generating things across all
+    // the extensions within a given file.
+    class Set {
```
allevato: This should probably be renamed so it doesn't ever risk colliding with the stdlib type of the same name (e.g., if someone wanted to use Swift's `Set` elsewhere in the `ExtensionGenerator` scope).

thomasvl: Chatting offline, it might make sense to flip the model, make `ExtensionSet` the outer class and `ExtensionGenerator` the inner one.  That also can kill `register(extensions:)` and instead have the generators created by the `ExtensionSet` on demand.

But doing that will will be a pretty large churn on the ExtensionGenerator file, want me to do that in this PR (as an added commit) or do it as a followup PR?

allevato: How about renaming to `ExtensionSet` in this PR and flipping in the next?

thomasvl: That works.

PR #576, opened 2017-05-31
Move all extension generation into ExtensionSetGenerator.

thomasvl:


thomasvl:
This is the followup from #575

PR #577, opened 2017-06-02
README.md rework

tbkka:
-> Moves sales pitch to the front
-> Clarifies the options for including the library
-> Expands the Quick Start with Swift client code examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,15 +3,13 @@
 # Swift Protobuf
 
 ---
-
+>
 > Status: We are converging on a 1.0 release.  The current 0.9.903 tag should
 > meet all of our goals for 1.0.  Please try it and let us know!
-
-> (Note: There is a small chance of further API breakage, which is why
-> we're not yet calling this `1.0RC`.  As far as we know, the biggest
-> remaining item is to complete the documentation and ensure it
-> matches the current behavior.  Any help is much appreciated.)
-
+>
+> (Note: There is a very small chance of further API breakage, which is why
+> we're not yet calling this `1.0RC`.)
+>
 ---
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -105,7 +144,7 @@ directory.
 To install, just copy this one executable into a directory that is
```
thomasvl: Right above here, should we tweak this to suggest building release and linking the runtime static like we did on the other page?

README.md
```diff
@@ -181,50 +232,45 @@ message BookInfo {
...
+of serialization and deserialization capabilities:
+```swift
+// Create a BookInfo object and populate it:
+var info = BookInfo()
+info.id = 1734
+info.title = "Really Interesting Book"
+info.author = "Jane Smith"
+
+// As above, but generating a read-only value:
+let info2 = BookInfo. with {
```
thomasvl: drop the space in there? `BookInfo.with {`

README.md
```diff
@@ -181,50 +232,45 @@ message BookInfo {
...
+    $0.id = 1735
+    $0.title = "Even More Interesting"
+    $0.author = "Jane Q. Smith"
+  }
+
+// Serialize to binary protobuf format:
+let binaryData: Data = info.serializedData()
```
thomasvl: I believe most of these need `try`, up to you if you want to also include examples of error handling.

tbkka: Ouch!  Good catch!  Fixed.

README.md
tbkka: Done.

README.md
tbkka: Done.

PR #579, opened 2017-06-02
Document a release process.

thomasvl:


Documentation/INTERNALS.md
```diff
@@ -451,3 +451,94 @@ TODO
 
 TODO
 
+# Releasing
+
+When doing a release:
+
+1. Examine what has changed
+
+   Githubs compre UI does a reasonable job here.  [Open that UI](https://github.com/apple/swift-protobuf/compare)
```
tbkka: s/compre/compare/

Documentation/INTERNALS.md
```diff
@@ -451,3 +451,94 @@ TODO
...
+   is _Draft a new release_.
+
+   The tag should be `v[a.b.c]` where the number *exactly* matches one you examined in
+   `Sources/SwiftProtobuf/Version.swift`.
+
+   For the description call out any major things in that release.  Usually a short summary and
+   then a reference to the pull request for more info is enough.
+
+   _Note:_ If we end up including a prebuilt `protoc-gen_swift` (or any other binaries) in the
+   releases; they also can be attached at this point.
```
tbkka: I'm not thrilled about getting into the business of shipping binaries.  The packaging folks already do this very well.

Documentation/INTERNALS.md
```diff
@@ -451,3 +451,94 @@ TODO
...
+   When looking at a individual commit, at the top github will show that it was a commit on master
+   and include a reference '#XYZ', this tells you what pull request it was part of.  This is useful
+   for doing the release notes in the next section.
+
+1. Validate the versions numbers
+
+   1. Inspect `Sources/SwiftProtobuf/Version.swift` and ensure the number is what you expect for
+      the next release.
+
+      Normally we try to bump _master_ to a new revision after each release, so this number may
```
tbkka: This makes me a little nervous:  SwiftPM works directly off of git tags (not "releases") so will happily let people grab tags from master.  Master may have (hopefully temporary) API breakage or other significant problems.  So tagging master and then inheriting that version number in a release branch may be tricky.  I'm not sure whether there are good alternatives, though.

Documentation/INTERNALS.md
```diff
@@ -451,3 +451,94 @@ TODO
...
+
+   For the description call out any major things in that release.  Usually a short summary and
+   then a reference to the pull request for more info is enough.
+
+   _Note:_ If we end up including a prebuilt `protoc-gen_swift` (or any other binaries) in the
+   releases; they also can be attached at this point.
+
+1. Publish the CocoaPod
+
+   1. Do a final sanity check that CocoaPods is happy with the release you just made, in the project
```
tbkka: Great!  I'm not that familiar with CocoaPods; this helps a lot.

Documentation/INTERNALS.md
```diff
@@ -451,3 +451,94 @@ TODO
 
 TODO
 
+# Releasing
```
tbkka: I think it might make more sense to have this as a separate RELEASE.md document.

Documentation/INTERNALS.md
thomasvl: Done

Documentation/INTERNALS.md
thomasvl: Dropping this comment.  I had it because #7 is open about providing them.  Do we want to just close that as not gonna do it?

Documentation/INTERNALS.md
thomasvl: I'm not following.

github's "releases" are basically a tag with some web ui and the auto creation of tar/zip files of the bits for folks to download.

There are two other patterns I've seen for this: 

- Use even revisions for the _pending_ state and odd for the _released_ state.
- Use zero as the master branch, and all release have a non zero revision.

Both allow you to always tell if some was using `master` directly or not.  But both add complexity in how you cut a release.

Thoughts?

Documentation/INTERNALS.md
thomasvl: Made RELEASING.md

Documentation/INTERNALS.md
tbkka: That would be my preference, yes.

Documentation/INTERNALS.md
tbkka: Not really a problem today, since I don't expect us to have any API-breaking changes even in master for a little while.  Once that happens, we'll have to consider how to manage the divergence between master and the 1.x release branch.

Documentation/RELEASING.md
```diff
@@ -0,0 +1,90 @@
+# Releasing Swift Protobuf
+
+---
+
+When doing a release:
+
+1. Examine what has changed
+
+   Githubs compare UI does a reasonable job here.  [Open that UI](https://github.com/apple/swift-protobuf/compare)
```
allevato: s/Githubs/GitHub's/

thomasvl: Done

PR #582, opened 2017-06-06
Swift 4 compat part1

tbkka:
These few changes address compatibility issues when building our code with various recent Swift 4 development snapshots.   None of them impair building with Swift 3.

The three changes are:
  * When slicing a string, explicitly mark the result as a String.  Swift 4 string slices now return a "StringSlice" type here that needs to be explicitly converted to a String; this is a no-op on Swift 3.
  * Publish the default typeURL value:  since this is used as a default value in public methods, Swift 4 insists that we make it public.  I've made the name more verbose as part of this change.
  * Rework the `withExtendedLifetime` calls to explicitly deconstruct the tuple.  The Swift 4 dev snapshots are fussier about this than Swift 3 was.

Admittedly, Swift 4 is not final, and any of these may be unnecessary with the final version of Swift 4 when it ships.

Note:  There is one remaining issue that still prevents building with the Swift 4 snapshots, but that breaks Swift 3 compatibility so will require additional effort to correctly conditionalize that change.  I'll submit that separately once I have it ready.

ToDo: I've not yet submitted regenerated protos from the `withExtendedLifetime` change.

ToDo:  I haven't yet run this change through all of the Swift environments I've collected.

I'll remove the "WIP" in the title when I've completed the ToDo's mentioned above.


Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
```diff
@@ -45,7 +45,7 @@ internal func typeName(fromURL s: String) -> String {
     }
   }
 
-  return s[typeStart..<s.endIndex]
+  return String(s[typeStart..<s.endIndex])
```
thomasvl: You mention it is a no op for Swift 3, is there any real cost to it for Swift 4 we need to worry about?

Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
```diff
@@ -32,7 +32,7 @@ internal func typeName(fromMessage message: Message) -> String {
   return messageType.protoMessageName
 }
 
-internal let defaultTypePrefix: String = "type.googleapis.com"
+public let defaultAnyTypeURLPrefix: String = "type.googleapis.com"
```
thomasvl: There's a part of me that would like to keep this private.  I wonder why default args have to be public.  If we put the raw string value in each place it wouldn't be exposing a symbol.

Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
tbkka: On Swift 4, converting a StringSlice to a String does seem to be a real operation.  But I don't think there's any "cost" to worry about.  I think that it's the same total work as with Swift 3, Swift 4 just exposes it as two separate operations (marking the slice and copying it into a String).  I don't think we want to return a StringSlice instead, so we don't really have a choice anyway.

Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
tbkka: I could back this out for now.  (Or maybe keep the name change but leave the symbol `internal` for now?)  As I mentioned earlier, Swift 4 isn't final, so this may not be really needed.

tbkka: I've switched this back to `internal` for now.

tbkka:
I think this is ready to land now.

PR #584, opened 2017-06-07
Add Swift 4 string decoder

tbkka:
This takes advantage of the `String(decoding:as:)` initializer
in the current Swift 4 development snapshots to improve
String decoding speed on Linux.

It's only a minor performance win on macOS, but it
does finally give us consistent code for this on both
platforms.

I've also reorganized this file to try to make it easier to
follow.  It now has clearly identified sections for Swift 4,
Swift 3/macOS, and Swift 3/Linux.


Sources/SwiftProtobuf/StringUtils.swift
```diff
@@ -8,32 +8,115 @@
...
+
+// Swift's support for working with UTF8 bytes directly has
+// evolved over time.  The following tries to choose the
+// best option depending on the version of Swift you're using.
+
+#if swift(>=4.0)
+
+///////////////////////////
+//
+// Swift 4 (all platforms)
```
thomasvl: Do you want to use `// MARK: - Blah blah` so Xcode helps group them in the jump bar?

Sources/SwiftProtobuf/StringUtils.swift
```diff
@@ -8,32 +8,115 @@
...
+    }
+    let codeUnits = UnsafeBufferPointer<UInt8>(start: bytes, count: count)
+    let sourceEncoding = Unicode.UTF8.self
+
+    // Verify that the UTF-8 is valid.
+    var p = sourceEncoding.ForwardParser()
+    var i = codeUnits.makeIterator()
+  Loop:
+    while true {
+      switch p.parseScalar(from: &i) {
```
thomasvl: fyi - Mixing 4 space and 2 space indent.

Sources/SwiftProtobuf/StringUtils.swift
```diff
@@ -8,32 +8,115 @@
...
     if let s = s {
         return String._unconditionallyBridgeFromObjectiveC(s)
     }
     return nil
 }
-#endif
 
-#if !os(Linux) || swift(>=3.2)
-// Deliberately empty.  macOS and Linux >=3.2 don't need the slow version.
 #else
```
thomasvl: Do you want to use elseif os(linux) and let the build break if we ever get an unexpected platform for < swift 4?

Sources/SwiftProtobuf/StringUtils.swift
```diff
@@ -8,32 +8,115 @@
...
 
-#if !os(Linux) || swift(>=3.1)
-fileprivate func fastUtf8ToString(bytes: UnsafePointer<UInt8>, count: Int) -> String? {
+internal func utf8ToString(bytes: UnsafePointer<UInt8>, count: Int) -> String? {
+    if count == 0 {
+        return String()
+    }
+    // On Apple platforms, the Swift 3 version of Foundation has a String
+    // initializer that works for us:
     let s = NSString(bytes: bytes, length: count, encoding: String.Encoding.utf8.rawValue)
```
thomasvl: Can this just be `encoding: .utf8` (or something short like that)?

Sources/SwiftProtobuf/StringUtils.swift
tbkka: Good idea.  Done.

Sources/SwiftProtobuf/StringUtils.swift
tbkka: Reindented the whole file (since I'm rewriting it all anyway).

Sources/SwiftProtobuf/StringUtils.swift
tbkka: Unfortunately not, because the `NSString` encoding argument here takes a number, not an enum.

Sources/SwiftProtobuf/StringUtils.swift
tbkka: Great idea!  Done.

PR #588, opened 2017-06-08
Add a type to wrap the "core" of the plugin.

thomasvl:
- Kill off Context (was basically empty now).
- Add a helper type (GeneratorPlugin) to have functions for each parts
  of the "flow" of the plugin so main.swift is a long run on block.
- Put the two generation code paths (files and stdin) onto common code
  so error handing, etc. is common.
- Helper for the binary name instead of coding it in multiple places.
- Move help, etc. about the plugin out of Version so it is tied to
  the code that parses args.


Sources/protoc-gen-swift/main.swift
```diff
@@ -20,86 +20,251 @@
...
-    response = CodeGeneratorResponse(error: "Unknown generation parameter '\(name)'")
-  } catch GenerationError.invalidParameterValue(let name, let value) {
-    response = CodeGeneratorResponse(
-      error: "Unknown value for generation parameter '\(name)': '\(value)'")
-  } catch GenerationError.wrappedError(let message, let e) {
-    response = CodeGeneratorResponse(error: "\(message): \(e)")
-  } catch let e {
-    response = CodeGeneratorResponse(error: "Internal Error: \(e)")
+    let response = generate(request: request)
+    if !sendReply(response: response) {
```
allevato: Nit: Write `if !x ... { return }` as `guard x else { return }` instead.

thomasvl: Done.

PR #594, opened 2017-06-08
Don't visit unknowns unless there is data.

thomasvl:
A few of the Visitors to some level of work around the zero length (trying to append it, etc.), this way we generally avoid they having to be defensive against zero length.


thomasvl:
Another option here might be to make the unknown storage on the message optional and don't create it until the instance needs it.

PR #596, opened 2017-06-09
MessageSet support

thomasvl:
This should be complete.  I'm getting it out so other eyes can look at it for general code quality, I need to make another read through it and maybe follow up with some more tests around crappy binary input and how it will be handled.  In parallel, we're trying to get the behaviors better defined in general, as the C++ and Java seem to differ and have some interesting choices.


thomasvl:
This is ready to go. Our unknown handling doesn't currently match the C++, but I don't think part of its behavior is correct; so I'll continue to push on that with the Protobuf Team and we can revisit that when the behavior is better defined.

Sources/SwiftProtobuf/SelectiveVisitor.swift
```diff
@@ -0,0 +1,268 @@
...
+}
+
+/// Default impls for everything so things using this only have to write the
+/// methods they expect.  Asserts to catch developer errors, but becomes
+/// nothing in release to keep code size small.
+///
+/// NOTE: This is an impl for *everything*. This means the default impls
+/// provided by Visitor to bridge packed->repeated, repeated->singular, etc
+/// won't kick in.
+internal extension SelectiveVisitor {
```
allevato: The name of this thing sounds like it behaves differently than described (it sounds to me like it filters, or something), but a better one doesn't immediately come to mind. 🤷‍♀️

thomasvl: yea, I didn't want to use _Base_ as it isn't something most should need.  _Minimal_, _SubSet_, etc. all didn't seem to work.

thomasvl: I guess _Custom_ might work, but at one level all of them are custom…

PR #597, opened 2017-06-15
Make union() not return a new instance.

thomasvl:
The two insert() methods are mutating and modify the instance.  union() was tagged as
mutating, but it didn't return a new instance, so change it to actually be mutating
since that seems more consistent.

### DIFF
--- a/Sources/SwiftProtobuf/SimpleExtensionMap.swift
+++ b/Sources/SwiftProtobuf/SimpleExtensionMap.swift
@@ -73,13 +73,17 @@ public struct SimpleExtensionMap: ExtensionMap, ExpressibleByArrayLiteral, Custo
         }
     }
 
-    public mutating func union(_ other: SimpleExtensionMap) -> SimpleExtensionMap {
-        var out = self
+    public mutating func formUnion(_ other: SimpleExtensionMap) {
         for (_, list) in other.fields {
             for (_, e) in list {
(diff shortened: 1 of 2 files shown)


tbkka:
Swift's standard Set.union() returns a new instance.  We should remain compatible with that.

Set.formUnion() adds the given set to the current instance.

I would be happy with your change as long as we edit the name to `formUnion` for compatibility with the Set API, or we could just remove the `mutating` marker.  Either would be an improvement.

thomasvl:
> Swift's standard Set.union() returns a new instance. We should remain compatible with that.
>
> Set.formUnion() adds the given set to the current instance.
>
> I would be happy with your change as long as we edit the name to formUnion for compatibility with the Set API, or we could just remove the mutating marker. Either would be an improvement.

No problem, the `mutating` label was what confused me, lets just bring it in sync with `Set` by having both so folk can pick the performance they are after.

thomasvl:
ps - since our type doesn't actually have "set" in the name, I'm not completely sure we need to follow `Set`.  One could also stretch to `Dictionary` which would be `merge` and `merging`.

thomasvl:
New commit up.

tbkka:
Using `Dictionary` naming conventions would be okay as well.  I just don't want to contradict the usage in the standard library.

PR #599, opened 2017-06-15
Add a union initializer for SimpleExtensionMap

thomasvl:
Since a developer might need to use extensions defined in a few different files, providing an initializer to create the combined registry seemed to make sense.


thomasvl:
> We should include some simple tests for this new functionality, of course.

Yea, in doing the other union additions I realized we don't generally have tests for SimpleExtensionMap.  I've got a note to myself to backfill them.

PR #601, opened 2017-06-15
SimpleExtensionMap

thomasvl:
- Add tests
- Hopefully simplify some of the internal (and lower the memory requirements by not storing a copy of something already in the AnyMessageExtension).
- Speed up union logic.


Tests/SwiftProtobufTests/Test_SimpleExtensionMap.swift
```diff
@@ -0,0 +1,219 @@
...
+    map.insert(ext1)
+    assert(map: map, contains: [ext1])
+
+    map.insert(ext2)
+    assert(map: map, contains: [ext2, ext1])
+
+    map.insert(ext3)
+    assert(map: map, contains: [ext3, ext2, ext1])
+
+    map.insert(ext4)
```
tbkka: Comment:  `// ext4 has the same message and number as ext2, so should replace it`

thomasvl: Added it to all the places so if someone has to look at any individual test case it will be clear.

PR #602, opened 2017-06-15
Fix Float round-trip accuracy in JSON & Text format

tbkka:
We originally copied Google's C++ implementation and used 8 digits for
float.  Unfortunately, about 1.5% of all floats require 9 digits.

This adds a test for one specific value that requires 9 digits, and adds tests that verify round-trip
correctness for 10,000 random floats and doubles.

Also, stop referring to FLT_DIG and DBL_DIG in the code.  Those
constants refer to concepts that aren't particularly helpful here.

### DIFF
--- a/Tests/SwiftProtobufTests/Test_JSON.swift
+++ b/Tests/SwiftProtobufTests/Test_JSON.swift
@@ -439,6 +439,7 @@ class Test_JSON: XCTestCase, PBTestHelpers {
         assertRoundTripJSON {$0.singleFloat = 0.0001}
         assertRoundTripJSON {$0.singleFloat = 0.00001}
         assertRoundTripJSON {$0.singleFloat = 0.000001}
+        assertRoundTripJSON {$0.singleFloat = 1.00000075e-36}
         assertRoundTripJSON {$0.singleFloat = 1e-10}
         assertRoundTripJSON {$0.singleFloat = 1e-20}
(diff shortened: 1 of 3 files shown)


Tests/SwiftProtobufTests/Test_JSON.swift
```diff
@@ -481,6 +482,20 @@ class Test_JSON: XCTestCase, PBTestHelpers {
         }
     }
 
+    func testSingleDouble_roundtrip() throws {
+        for _ in 0..<10000 {
+            let d = drand48()
+            assertRoundTripJSON {$0.singleDouble = d}
```
thomasvl: If these were to fail, will be be able to tell what the value was?  i.e. - is the only logging by way of TextFormat or something that would be going though the formatting code?

tbkka:
> ... float/double are smaller ...

Float is always IEEE 754 single-precision binary32 and Double is always IEEE 754 double-precision binary64 on all platforms, regardless of the default integer size.

Tests/SwiftProtobufTests/Test_JSON.swift
tbkka: If it's wrong, it will only be wrong by `.nextUp` or `.nextDown`, so finding the failing value won't be that hard.

In practice, problems in the formatting code will cause lots of failures; it won't be hard to find problems once we know they're there.

PR #603, opened 2017-06-20
Avoid extra buffering when writing out sub messages.

thomasvl:
Instead of creating a complete Data for the submessage, compute the
length (which would have been done anyway inside of serializedData()),
write out that varint, and then traverse the message to write its
content.  This avoids the allocation of the Data for the sub message
and then having to copy it into the real Data being created.

### DIFF
--- a/Sources/SwiftProtobuf/BinaryEncodingVisitor.swift
+++ b/Sources/SwiftProtobuf/BinaryEncodingVisitor.swift
@@ -18,7 +18,7 @@ import Foundation
 /// Visitor that encodes a message graph in the protobuf binary wire format.
 internal struct BinaryEncodingVisitor: Visitor {
 
-  private var encoder: BinaryEncoder
+  var encoder: BinaryEncoder
 
   /// Creates a new visitor that writes the binary-coded message into the memory
   /// at the given pointer.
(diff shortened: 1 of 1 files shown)


Sources/SwiftProtobuf/BinaryEncodingVisitor.swift
```diff
@@ -18,7 +18,7 @@ import Foundation
 /// Visitor that encodes a message graph in the protobuf binary wire format.
 internal struct BinaryEncodingVisitor: Visitor {
 
-  private var encoder: BinaryEncoder
+  var encoder: BinaryEncoder
```
allevato: Does this need to be promoted to internal when the only change in this PR is within this type/file?

thomasvl: I was thinking since it is used in the other type at the end, it did, but maybe not.  But I'm about to use it in the next PR for some validation from another file.

tbkka:
Nice!  What's the performance impact?

thomasvl:
> Nice! What's the performance impact?

Nothing I'd claim.  The first time I ran them, it said every test was faster before my change (including JSON and TextFormat, and those shouldn't change).  Second run said everything but JSON encode was faster after my change.  So I'd say for what the script checks, it is all in the noise range.  The real "wins" likely are if there are deeply nested messages, as that's why the extra Data usage could have added up.

PR #604, opened 2017-06-20
Sanitycheck serialize behavior

thomasvl:


Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -34,18 +34,15 @@ public extension Message {
     let requiredSize = try serializedDataSize()
     var data = Data(count: requiredSize)
     try data.withUnsafeMutableBytes { (pointer: UnsafeMutablePointer<UInt8>) in
-      try serializeBinary(into: pointer)
+      var visitor = BinaryEncodingVisitor(forWritingInto: pointer)
+      try traverse(visitor: &visitor)
+      if requiredSize != visitor.encoder.distance(pointer: pointer) {
```
thomasvl: I'm torn here, if this fires, it means there is a bug within the library, so should we just make this an assert instead of making a developer exposed error?

tbkka: I'm inclined to make it an assert.

allevato: Agreed—don't give users the opportunity to catch our bugs. (It's also easy to silently ignore errors using `try?`, and we'd want this to show up so people let us know.)

thomasvl:
Changed to an assert.

PR #608, opened 2017-06-21
Add a helper for "delimited message" reading/writing.

thomasvl:
The Java implementation long ago added the concept of a "delimited message" when
reading/writing to a stream. Normally, you can concatenate the serialized form
together and it is still one message; but this form used a varint to write
the length and then write the message, allowing multiple messages to be
written distinctly.

In github.com/google/protobuf/pull/710 Google's protobuf team has
suggested generally supporting this, but the think it should be as
utilities rather than directly on the Message classes. The logic
being there could be multiple ways to "frame" messages for different
transports, and adding them all to Message can lead to bloat/confusion
on the over all api surface.

For these reasons, the api here is static functions scoped to an enum to
provide logical grouping, but not pollute the Message's namespace. Swift
extensions on InputStream/OutputStream could also be done, but since lots
of uses of protocol buffers might never need this, it seemed better not to
pollute those namespaces either.


Sources/SwiftProtobuf/DelimitedMessageUtil.swift
```diff
@@ -0,0 +1,215 @@
...
+  /// before trying to read/write to it.
+  case unknownStreamError
+
+  /// While reading/writing to the stream, less than the expected bytes was
+  /// read/written.
+  case truncated
+}
+
+/// Helper methods for reading/writing messages with a length prefix.
+public enum DelimitedMessageUtil {
```
thomasvl: I am in no means in love with this name (or the file name), and will happily change it to something that we think is better.  😄

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
```diff
@@ -0,0 +1,215 @@
...
+// -----------------------------------------------------------------------------
+///
+/// Helpers to read/write message with a length prefix.
+///
+// -----------------------------------------------------------------------------
+
+import Foundation
+
+/// Additional errors for delimited message handing.
+public enum DelimitedMessageError: Error {
```
thomasvl: Style wise I debated scoping this under the enum but then the logic name would just be "Error", and I wasn't sure if that lead to naming confusion as folks read the code.  If that seems better, I'll move it back.

tbkka: This is the same thinking that led us to separate the other Error enums as top-level.

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
allevato: What about `DelimitedMessages`?  Then you have `DelimitedMessages.write/parse`, which reads nicely.

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
allevato: I'm in favor of nesting; it would only add a couple characters to the catch sites and many users might just do a catch-all or `try?` anyway and it won't affect them.

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
```diff
@@ -0,0 +1,215 @@
...
+  ///   - to: The `OutputStream` to write the message too.  The stream is
+  ///     is assumed to be ready to be written too.
+  ///   - partial: If `false` (the default), this method will check
+  ///     `Message.isInitialized` before encoding to verify that all required
+  ///     fields are present. If any are missing, this method throws
+  ///     `BinaryEncodingError.missingRequiredFields`.
+  /// - Throws: `BinaryEncodingError` if encoding fails; throws
+  ///           `DelimitedMessageError` for some writing errors, and the
+  ///           underlying OutputStream.streamError for a stream error.
+  public static func write(
```
allevato: Should we call this `serialize` to align with the naming used elsewhere?

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
```diff
@@ -0,0 +1,215 @@
...
+  ///     fields.
+  ///   - partial: If `false` (the default), this method will check
+  ///     `Message.isInitialized` before encoding to verify that all required
+  ///     fields are present. If any are missing, this method throws
+  ///     `BinaryEncodingError.missingRequiredFields`.
+  ///   - options: The BinaryDecodingOptions to use.
+  /// - Throws: `BinaryDecodingError` if decoding fails; throws
+  ///           `DelimitedMessageError` for some rading errors, and the
+  ///           underlying InputStream.streamError for a stream error.
+  public static func parse<M: Message>(
```
allevato: `merge(into: ...)`?

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
```diff
@@ -0,0 +1,215 @@
...
+
+  /// Write a single size-delimited message from the given stream. Delimited
+  /// format allows a single file or stream to contain multiple messages,
+  /// whereas normally writing multiple non-delimited messages to the same
+  /// stream would cause them to be merged. A delimited message is a varint
+  /// encoding the message size followed by a message of exactly that size.
+  ///
+  /// - Parameters:
+  ///   - message: The message to be written.
+  ///   - to: The `OutputStream` to write the message too.  The stream is
```
allevato: s/too/to/

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
```diff
@@ -0,0 +1,215 @@
...
+  ///     is assumed to be ready to be written too.
```
allevato: s/too/to/

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
```diff
@@ -0,0 +1,215 @@
...
+  ///   - partial: If `false` (the default), this method will check
+  ///     `Message.isInitialized` before encoding to verify that all required
+  ///     fields are present. If any are missing, this method throws
+  ///     `BinaryEncodingError.missingRequiredFields`.
+  /// - Throws: `BinaryEncodingError` if encoding fails; throws
+  ///           `DelimitedMessageError` for some writing errors, and the
+  ///           underlying OutputStream.streamError for a stream error.
```
allevato: Backticks around `OutputStream.streamError`

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
```diff
@@ -0,0 +1,215 @@
...
+  /// - Parameters:
+  ///   - message: The message to be written.
+  ///   - to: The `OutputStream` to write the message too.  The stream is
+  ///     is assumed to be ready to be written too.
+  ///   - partial: If `false` (the default), this method will check
+  ///     `Message.isInitialized` before encoding to verify that all required
+  ///     fields are present. If any are missing, this method throws
+  ///     `BinaryEncodingError.missingRequiredFields`.
+  /// - Throws: `BinaryEncodingError` if encoding fails; throws
+  ///           `DelimitedMessageError` for some writing errors, and the
```
allevato: s/errors, and/errors, or/

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
```diff
@@ -0,0 +1,215 @@
...
+  ///
+  /// - Parameters:
+  ///   - message: The message to be written.
+  ///   - to: The `OutputStream` to write the message too.  The stream is
+  ///     is assumed to be ready to be written too.
+  ///   - partial: If `false` (the default), this method will check
+  ///     `Message.isInitialized` before encoding to verify that all required
+  ///     fields are present. If any are missing, this method throws
+  ///     `BinaryEncodingError.missingRequiredFields`.
+  /// - Throws: `BinaryEncodingError` if encoding fails; throws
```
allevato: Replace semicolon with comma

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
```diff
@@ -0,0 +1,215 @@
...
+  ///   - extensions: An `ExtensionMap` used to look up and decode any
+  ///     extensions in this message or messages nested within this message's
+  ///     fields.
+  ///   - partial: If `false` (the default), this method will check
+  ///     `Message.isInitialized` before encoding to verify that all required
+  ///     fields are present. If any are missing, this method throws
+  ///     `BinaryEncodingError.missingRequiredFields`.
+  ///   - options: The BinaryDecodingOptions to use.
+  /// - Returns: The message read.
+  /// - Throws: `BinaryDecodingError` if decoding fails; throws
```
allevato: semicolon -> comma

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
```diff
@@ -0,0 +1,215 @@
...
+  ///           `DelimitedMessageError` for some rading errors, and the
```
allevato: s/rading errors, and/reading errors, or/

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
```diff
@@ -0,0 +1,215 @@
...
+  ///   - extensions: An `ExtensionMap` used to look up and decode any
+  ///     extensions in this message or messages nested within this message's
+  ///     fields.
+  ///   - partial: If `false` (the default), this method will check
+  ///     `Message.isInitialized` before encoding to verify that all required
+  ///     fields are present. If any are missing, this method throws
+  ///     `BinaryEncodingError.missingRequiredFields`.
+  ///   - options: The BinaryDecodingOptions to use.
+  /// - Throws: `BinaryDecodingError` if decoding fails; throws
+  ///           `DelimitedMessageError` for some rading errors, and the
```
allevato: Same corrections as `parse`

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
tbkka: In those cases, of course, we put the error enum in a separate file.

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
tbkka: Or perhaps `DelimitedMessageStream`?  In the sense that this is something you write messages to or read them from.

Another structure would be to have `DelimitedMessageOutputStream` and `DelimitedMessageInputStream` that you used like so:
```
  let binaryStream: OutputStream
  let messageStream = DelimitedMessageOutputStream(stream: binaryStream)
  messageStream.write(message)
```

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
thomasvl: The one difference I see is those other cases, the Errors are the only exposed part of the types.  The Encoders/Decoders are all internal to the library so nesting the Error would get in the way to actually exposing it.

This is one case where the error could be scoped to the public type it goes along with.

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
thomasvl: I was sorta modeling the [C++](https://github.com/google/protobuf/blob/master/src/google/protobuf/util/delimited_message_util.h) here, when we do look at more streaming interfaces for serialization, I wasn't sure if we'd have more types that should logically support this also.

I worry a little about wrapping the stream, as it could appear the wrapper "takes over" the stream and they aren't allowed to make calls on it, as opposed to we're just providing some helpers to read/wrote messages onto the existing types.

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
thomasvl: Done

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
thomasvl: Done

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
thomasvl: Done

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
thomasvl: Done

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
thomasvl: Done

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
thomasvl: Done

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
thomasvl: Done.

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
thomasvl: Done

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
thomasvl: Since the methods are on the scoping type, I was worried "merge" could be seen as an input or output operation.  But maybe not?

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
thomasvl: Hm, I was focused on the write/parse side of things, but yea, maybe we need to capture "serialize" in here to better capture which format it is.  We could also capture this in the helper type - `DelimitedBinarySerization` ?

Sources/SwiftProtobuf/DelimitedMessageUtil.swift
thomasvl: Raising up from the now "outdated comment" I added this too:

We could also capture better that is specific to the binary serialization format with something like `DelimitedBinarySerialization` (or something like that).  And that would tie things together instead of trying to get those names into the function names themselves.

thomasvl: Maybe `SerializedDelimited`?

allevato: `BinaryDelimited`? Then `BinaryDelimited.serialize` and `BinaryDelimited.parse`? I like the symmetry of having `serialize` in the method name, since that's how the other messages are done. I don't think we use `write` anywhere else in public APIs, do we?

thomasvl:
New commits up with naming shuffles.  I merge them since having the files renamed seems to make the diffs even more noisy.

Issue #585, opened 2017-06-07
Swift 4 compatibility


tbkka:
I've identified a handful of issues trying to build SwiftProtobuf with the current Swift 4 development snapshots.  We do of course want to be compatible with Swift 4 when it ships, so most of these we'll probably have to fix.

- [x] `withExtendedLifetime` tuple deconstruction should be more explicit (addressed in #582)
- [x] String slices need to be explicitly converted to String (addressed in #582)
- [x] Public functions seem to want public default values (changed in an early draft of #582 but backed out until we figure out how to best address this)
- [x] SR-4218 leads to test failures on Linux for JSON and TextFormat strings with control characters (we disabled these tests on Linux for Swift 3 and were hoping to reenable the tests as soon as this bug was fixed -- disabled again in #611 for Swift 4.0)
- [x] Should use `String(decoding:as:)` to convert UTF-8 to String -- this is faster than the `NSString` initializer we used with Swift 3 and seems to work consistently on Linux and macOS. (addressed in #584)
- [x] `truncatingBitPattern:` was renamed in Swift 4 (addressed in #611)

tbkka:
PR #611 addresses the still-broken SR4218 issue and the name change for `truncatingBitPattern`.  This gets us very close to being able to build with either Swift 3 or 4 from the same source.

tbkka:
So at this point, I think only the default argument problem remains to get something that builds with the current Swift 4 development snapshots.

I asked the Swift folks, and they said that in the future, they want to inline default arguments at the caller, which requires them to be public.  So we could either expose the necessary symbol or repeat the literal string.  I think there is only one public function that uses that symbol as a default right now, so I don't think it makes a big difference either way.

thomasvl:
Yea, go ahead and make the constant public given the direction they are going.

tbkka:
PR #612 implements this.  It also addresses a couple of other issues -- compatibility with older Swift compilers on Linux, and an ambiguity that was flagged as a warning in the new compilers.

tbkka:
I believe we've now resolved all the oustanding issues with Swift 4 per se.

I'll close this.  (I'm also going to look separately at #610.)

PR #611, opened 2017-07-08
Swift4 compat

tbkka:
More incremental progress on Issue #585.

This addresses the renamed truncating initializers by introducing a simple layer of indirect functions (lowercase 'int(truncating:)', etc).  Hopefully, this can be ripped out again when we drop Swift 3.x support (maybe not until Swift 5 ships?).

It also disables certain String tests that tickle a bug that is still present in snapshots of Linux Swift 4.0.  Hopefully, the bug will get fixed in Swift 4.1.

### DIFF
--- a/Sources/SwiftProtobuf/MathUtils.swift
+++ b/Sources/SwiftProtobuf/MathUtils.swift
@@ -38,3 +38,98 @@ internal func div<T : SignedInteger>(_ a: T, _ b: T) -> T {
     assert(b > 0)
     return a >= 0 ? a / b : (a + 1) / b - 1
 }
+
+// Swift's standard library changed the name of the initializer
+// that truncates an integer to a narrower type.  To continue
+// to support Swift 3, we've introduced these functions to
+// handle that.
+//
+#if swift(>=4.0)
+//
+// Swift 4 (prerelease) changed this initializer to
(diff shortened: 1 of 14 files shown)


Sources/SwiftProtobuf/MathUtils.swift
```diff
@@ -38,3 +38,98 @@ internal func div<T : SignedInteger>(_ a: T, _ b: T) -> T {
     assert(b > 0)
     return a >= 0 ? a / b : (a + 1) / b - 1
 }
+
+// Swift's standard library changed the name of the initializer
+// that truncates an integer to a narrower type.  To continue
+// to support Swift 3, we've introduced these functions to
+// handle that.
+//
+#if swift(>=4.0)
```
thomasvl: So rather then having something we have to go touch the code to clean up, why not follow what we did Int64 and UInt64. There we provided the init?(exactly:) for older versions of swift, and nothing was needed going forward.  This means no shim on newer versions, and the "extra" code compiles away until we trim it.  - https://github.com/apple/swift-protobuf/blob/master/Sources/SwiftProtobuf/JSONIntegerConverting.swift#L18

tbkka: Good point.  Done.

PR #612, opened 2017-07-10
Swift 4 linux compat

tbkka:
This fixes a couple of minor compatibility issues:

* Exposes `defaultAnyTypeURLPrefix` for compatibility with Swift 4
* Uses an explicit NSData-to-Data conversion for compatibility with Swift Linux
* Comments out the body of one test that won't work at all with Swift Linux < 3.1

There are still some warnings with the latest dev snapshots I need to sort through.  But I think this fixes the last of the actual build errors and test failures.


Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
```diff
@@ -16,6 +16,8 @@
 // don't remove unless obsolete on all platforms
 import Foundation
 
+public let defaultAnyTypeURLPrefix: String = "type.googleapis.com"
```
tbkka: I moved this since this file has the only use of it in a public API (the only other file that references this constant is purely internal).

Tests/SwiftProtobufTests/Test_BinaryDelimited.swift
```diff
@@ -15,6 +15,7 @@ import SwiftProtobuf
 class Test_BinaryDelimited: XCTestCase {
 
   func testEverything() {
+#if swift(>=3.1)
```
thomasvl: Don't we still want this always running on macOS?

tbkka: `OutputStream.toMemory` and `XCTAssertNoThrow` are available only starting in Swift 3.1 (on Linux, at least).  This is the only test that uses those, so I've just commented out the whole thing.  We could work around those missing functions if you think it's important to run this test on older Swift.

tbkka: I unconditionally enabled it on OSX and iOS.

tbkka:
I included a fix for a potential bug that was being flagged by the latest compiler snapshots.

PR #617, opened 2017-07-18
Mark all visitMapField() methods as "throws"

### LINKED ISSUE #610: Xcode9b2 build failing (with Carthage)

tbkka:
Note:  The Swift language does not require this; it's
not really a bug in our code.

But, it seems to be tickling a bug within the Swift compiler,
so it seems reasonable to make the code more consistent here.

This seems to fix #610 (builds failing with Carthage on Xcode9b2).


Adam0000:
Thanks for the workaround. In case you're wondering, only the change in SelectiveVisitor was needed to stop the Swift compiler from crashing.

PR #623, opened 2017-07-26
Update protos from google/protobuf and update for new conformance tests

thomasvl:


Sources/Conformance/main.swift
```diff
@@ -60,32 +60,39 @@ func buildResponse(serializedData: Data) -> Conformance_ConformanceResponse {
         return response
     }
 
-    let parsed: ProtobufTestMessages_Proto3_TestAllTypes?
+    var msgType: SwiftProtobuf.Message.Type
```
allevato: I think you can make this a `let` since the compiler can determine that it was definitely assigned before first use.

tbkka:
Yay!  More Tests!  This is very exciting...

thomasvl:
> Yay! More Tests! This is very exciting...

😄  the proto2 syntax stuff includes getting conformance to start looking at MessageSet for all languages.

Sources/Conformance/main.swift
thomasvl: Done.

PR #624, opened 2017-07-28
Perf: Use DoubleFormatter.utf8ToDouble() for parsing doubles.

tbkka:
The crux of this change is just to add a DoubleFormatter instance
to JSONScanner and move `parseBareFloatString()` into JSONScanner
so it can use that object for the parsing.  (TextFormat already
worked this way.)

Along the way, I changed the name to `parseBareDouble()` to reflect
that it now returns a `Double?` instead of a `String?`, and simplified
a bunch of places that can now avoid juggling the no-longer-needed
String intermediate.

This is a significant performance boost when parsing JSON objects that
contain float or double fields: `perf_runner 100 double`
is almost 3x faster with this change.


Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -706,6 +398,311 @@ internal struct JSONScanner {
...
+          throw JSONDecodingError.malformedNumber
+        case asciiBackslash:
+          return nil
+        default:
+          return 0
+        }
+      }
+      return 0
+    case asciiOne...asciiNine: // 1...9
+      var n = UInt64(c - 48)
```
thomasvl: s/48/asciiZero/  (you did that below)

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -706,6 +398,311 @@ internal struct JSONScanner {
...
+  // is used by callers that are parsing quoted numbers.  See nextSInt()
+  // and nextUInt() below.
+  private func parseBareUInt(
+    source: UnsafeBufferPointer<UInt8>,
+    index: inout UnsafeBufferPointer<UInt8>.Index,
+    end: UnsafeBufferPointer<UInt8>.Index
+  ) throws -> UInt64? {
+    let start = index
+    let c = source[index]
+    source.formIndex(after: &index)
```
thomasvl: I think the code might get a little simpler if you skip this.  In the zero case before, advance it to check the next character, but in the double path, you reset it, so this was wasted.  Likewise, the case where it is a read number, you probably won't need need to start the number with the initial value, and can just use the loop direct seeding `digit` with zero.

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -706,6 +398,311 @@ internal struct JSONScanner {
...
+      }
+      return 0
+    case asciiOne...asciiNine: // 1...9
+      var n = UInt64(c - 48)
+      while index != end {
+        let digit = source[index]
+        switch digit {
+        case asciiZero...asciiNine: // 0...9
+          let val = UInt64(digit - asciiZero)
+          if n >= UInt64.max / 10 {
```
thomasvl: is this double check needed or can this all be done in one check?

```
if n > UInt64.max / 10 && val > UInt64.max % 10 { throw ... }
```

?

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -706,6 +398,311 @@ internal struct JSONScanner {
...
+  //
+  // This uses parseBareUInt() to do the heavy lifting;
+  // we just check for a leading minus and negate the result
+  // as necessary.
+  //
+  // As with parseBareUInt(), if it encounters a "\" backslash
+  // character, it returns a nil.  This is used by callers that are
+  // parsing quoted numbers.  See nextSInt() and nextUInt() below.
+
+  private func parseBareSInt(
```
thomasvl: parseBareSInt64?

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -706,6 +398,311 @@ internal struct JSONScanner {
...
+    source: UnsafeBufferPointer<UInt8>,
+    index: inout UnsafeBufferPointer<UInt8>.Index,
+    end: UnsafeBufferPointer<UInt8>.Index
+  ) throws -> Int64? {
+    if index == end {
```
thomasvl: Why is this needed, but parseBareUInt didn't need to validate their inputs in the same way?

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -706,6 +398,311 @@ internal struct JSONScanner {
...
+    let c = source[index]
+    if c == asciiMinus { // -
+      source.formIndex(after: &index)
+      // character after '-' must be digit
+      let digit = source[index]
+      if digit < asciiZero || digit > asciiNine {
+        throw JSONDecodingError.malformedNumber
+      }
+      if let n = try parseBareUInt(source: source, index: &index, end: end) {
+        let limit: UInt64 = 0x8000000000000000 // -Int64.min
```
thomasvl: Down below you did UInt64(bitPattern: Int64.max), any reason you can't do a named constant here to avoid the need for a comment?

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -706,6 +398,311 @@ internal struct JSONScanner {
...
+    source: UnsafeBufferPointer<UInt8>,
+    index: inout UnsafeBufferPointer<UInt8>.Index,
+    end: UnsafeBufferPointer<UInt8>.Index
+  ) throws -> Double? {
+    // RFC 7159 defines the grammar for JSON numbers as:
+    // number = [ minus ] int [ frac ] [ exp ]
+    let start = index
+    var c = source[index]
+    if c == asciiBackslash {
+      return nil
```
thomasvl: Same comment, still not clear on why backslash comes up in theres.

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -706,6 +398,311 @@ internal struct JSONScanner {
...
+      if index == end {
+        index = start
+        throw JSONDecodingError.truncated
+      }
+      c = source[index]
+      if c == asciiBackslash {
+        return nil
+      }
+    } else if c == asciiUpperN { // Maybe NaN?
+      // Return nil, let the caller deal with it.
```
thomasvl: Document this on the function?

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -706,6 +398,311 @@ internal struct JSONScanner {
...
+      return nil
+    }
+
+    if c == asciiUpperI { // Maybe Infinity, Inf, -Infinity, or -Inf ?
+      // Return nil, let the caller deal with it.
```
thomasvl: Same, document it?

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -706,6 +398,311 @@ internal struct JSONScanner {
...
+    }
+
+    // Integer part can be zero or a series of digits not starting with zero
+    // int = zero / (digit1-9 *DIGIT)
+    switch c {
+    case asciiZero:
+      // First digit can be zero only if not followed by a digit
+      source.formIndex(after: &index)
+      if index == end {
+        if let d = numberFormatter.utf8ToDouble(bytes: source, start: start, end: index) {
```
thomasvl: At this point, isn't the number always zero?  i.e.- why do we need to call the numberFormatter?

(i guess you need to error for -0?)

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -706,6 +398,311 @@ internal struct JSONScanner {
...
+  //
+  // This implements the full grammar defined by the JSON RFC 7159.
+  // Note that Swift's string-to-number conversions are much more
+  // lenient, so this is necessary if we want to accurately reject
+  // malformed JSON numbers.
+  //
+  // This is used by nextDouble() and nextFloat() to parse double and
+  // floating-point values, including values that happen to be in quotes.
+  // It's also used by the slow path in parseBareSInt() and parseBareUInt()
+  // above to handle integer values that are written in float-point notation.
```
thomasvl: Also calling out why backslash returns nil here would help, when first reading the code, it seemed like these should be errors, not return nils.

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -706,6 +398,311 @@ internal struct JSONScanner {
...
+  // called for both unquoted numbers and for numbers stored in quoted
+  // strings.  In the latter case, the caller is responsible for
+  // consuming the leading quote and verifying the trailing quote.
+  //
+  // If the number is in floating-point format, this uses a slower
+  // and less accurate approach: it identifies a substring comprising
+  // a float, and then uses Double() and UInt64() to convert that
+  // string to an unsigned intger.
+  //
+  // If it encounters a "\" backslash character, it returns a nil.  This
```
thomasvl: So I had lots of comments at first about the backslash, because it wasn't until the final calling code that it made sense; so I'd suggest beefing up this comment.  The important detail that wasn't clear to me is this is called _first_ on the raw bytes in the JSON, so there is a chance a \uXXXX could be encountered when the double was quoted; so by returning nil, it is a signal that this couldn't directly parse the number and if it was a quoted string, that decode _must_ be done before and the result of that decode can then be tried for parsing again.

tbkka:
@thomasvl:  Great feedback!  I had just moved this code into the struct without taking a close look at it.  I'll go through and clean it up.

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: Just above, it says:
> ... called for both unquoted numbers and for numbers
> stored in quoted strings.  In the latter case, the caller is
> responsible for consuming the leading quote and
> verifying the trailing quote.

I'll play with rewording here to see if I can make it clearer.

thomasvl: Yea, I saw that, but when I first read it, I took it pretty literally – skip the quote and check for the matching one at the end (after the parse).  The reality is this will return nil on the backslash, so they have to read & decode the quoted string and try to parse that result.

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: I'll change it to the single check:
```
   if n > UInt64.max || (n == UInt64.max && val > UInt64.max % 10)
```

thomasvl: Not sure that's right, if `n` is 3 shy of `UInt64.max` and `val` is 0-2, things are fine, but 3-9 would overflow. You likely need the /10 on the max still.

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: I could write this in other ways, but I'm not sure what you would find clearer.  Which of these would you prefer?
```
   UInt64(1) << UInt64(63)
   UInt64(Int64.max) + 1
```

thomasvl: Hm, neither of those matches what you comment was.  I guess something like UInt64(bitPattern: -Int64.max) doesn't work either.  I guess stick with what you have then.

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -706,6 +398,311 @@ internal struct JSONScanner {
...
+  // is used by callers that are parsing quoted numbers.  See nextSInt()
+  // and nextUInt() below.
+  private func parseBareUInt(
```
tbkka: Done.

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: Done.

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: Done.

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: Good catch!  parseBareUInt did need to validate in the same way; I filled in the missing tests for this.

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: Changed to start with zero (as you suggested above).

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: Good catch!  It is always zero here.

I've also filled in some tests to verify that we do actually correctly parse 0 and -0.

tbkka:
I pushed a bunch of changes, including new tests for various truncation scenarios, +0 vs. -0, and UInt64.

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -706,6 +398,325 @@ internal struct JSONScanner {
...
+      return 0
+    case asciiOne...asciiNine: // 1...9
+      var n = 0 as UInt64
+      while index != end {
+        let digit = source[index]
+        switch digit {
+        case asciiZero...asciiNine: // 0...9
+          let val = UInt64(digit - asciiZero)
+          if (n > UInt64.max / 10
+              || (n == UInt64.max / 10 && val > UInt64.max % 10)) {
```
thomasvl: Forgive me, thinking this though…

What we want is to when:
   `n * 10 + value > UInt64.max`
so just shifting value:
   `n * 10 > UInt64.max - value`
But we still need to ensure the `n * 10` won't overflow, so I think this is:
   `(n > UInt64.max / 10) || (n * 10 > UInt64.max - value)`
I'm not sure if that's any better.

Heck:
```
let newN = n *& 10 +& value
if newN < n {
  throw JSONDecodingError.numberRange
}
n = newN
```

Maybe @allevato has ideas or we just stick with what you have.

_math is hard…_

allevato: Explicit overflow checks might make the code clearer.

Something like this (untested):

```swift
let (nTimes10, multiplyOverflow) = n.multiplyWithOverflow(10)
if multiplyOverflow { throw ... }
let (n, addOverflow) = nTimes10.addWithOverflow(val)
if addOverflow { throw ... }
n = newN
```

Could be factored out into a helper function.

tbkka:
```
(n > UInt64.max / 10) || (n * 10 > UInt64.max - value)
```
I like this form.  Especially since `UInt64.max / 10` is a constant and `n * 10` can be reused immediately below.  That's less total work than my current form.

I've also updated the same code in TextFormatScanner.

Tests/SwiftProtobufTests/Test_JSON.swift
```diff
@@ -299,6 +327,67 @@ class Test_JSON: XCTestCase, PBTestHelpers {
...
+        // Quoted or unquoted numbers, positive or zero
+        assertJSONDecodeSucceeds("{\"singleUint64\":1}") {$0.singleUint64 == 1}
+        assertJSONDecodeSucceeds("{\"singleUint64\":\"1\"}") {$0.singleUint64 == 1}
+        assertJSONDecodeSucceeds("{\"singleUint64\":0}") {$0.singleUint64 == 0}
+        assertJSONDecodeSucceeds("{\"singleUint64\":\"0\"}") {$0.singleUint64 == 0}
+        // Protobuf JSON does not accept leading zeros
+        assertJSONDecodeFails("{\"singleUint64\":01}")
+        assertJSONDecodeFails("{\"singleUint64\":\"01\"}")
+        // But it does accept exponential (as long as result is integral)
+        assertJSONDecodeSucceeds("{\"singleUint64\":4.294967295e9}") {$0.singleUint64 == UInt32.max}
```
thomasvl: @tbkka this doesn't compile since it is comparing an `UInt64` with a `UInt32`.  Did you want to use `UInt64(UInt32.max)` or to update the test to be the 64bit value?

thomasvl: I tried changing the line to be:

```swift
assertJSONDecodeSucceeds("{\"singleUint64\":1.8446744073709551615e19}") {$0.singleUint64 == UInt64.max}
```

And it fails.  `JSONScanner.swift:483` -

```swift
            if let u = UInt64(exactly: d) {
              return u
            }
```

Seems to be failing so it decides it can't parse in that form.

tbkka: Ouch.  Fix forthcoming...

(Seems that Swift 4 does accept this, so I'm going to take a few minutes to run some tests through older compilers to see if there are any other similar gaffes.)

tbkka:
When we see a floating-point format for an integer field, we parse it via Double, which cannot accurately represent `1.8446744073709551615e19`.  If we want to support that, we would have to use Float80 as the intermediate form when parsing floating-point text into an integer field.

thomasvl:
Should we land a "quick fix" to get things building/testing again, and open an issue to track the longer term fix here?  (I'm afraid to ask what the other google/protobuf languages do around this…)

tbkka:
Just a couple of minutes....  ;-)

tbkka:
Quick fix committed in #631

tbkka:
It would be a trivial change for us to use Float80 as the intermediate in this particular case, which would give us better/different behavior when parsing certain JSON documents.  It probably doesn't impact compatibility either way, though, since other protobuf JSON implementations should always write out quoted integer forms in this case (which preserves full precision) and non-protobuf JSON implementations are generally limited to Double precision anyway.

Note that the JSON RFC very carefully declines to limit floating-point numbers to Double precision.  It does, however, point out that most implementations are limited to Double for numeric values.

PR #627, opened 2017-07-31
Perf: Rework Text and JSON enum handling to take full advantage of NameMap

tbkka:
On decode, we were converting the UTF-8 enum name into a String and
then NameMap was converting it back to UTF-8 for lookup.  The code now
short-circuits that, looking up the UTF-8 key directly on the NameMap.

This is a little tricky for JSON because a quoted enum key is allowed
to have backslash escapes.  This should be very rare since proto enum
names are always simple ASCII, so we can optimize by first trying a fast
path that returns UTF-8 for a key with no such escapes, then falling back
to a slower String decode if we must.

On encode, JSON repeated fields (including enums) were spending a lot
of unnecessary time converting constant strings such as "[" and ","
to/from UTF8.  This commit fixes that, with sizable performance gains
for all repeated field types in JSON.

I've also edited the Performance harness to support `enum` as a field
type.

### DIFF
--- a/Sources/SwiftProtobuf/Enum.swift
+++ b/Sources/SwiftProtobuf/Enum.swift
@@ -69,4 +69,12 @@ extension Enum {
     }
     self.init(rawValue: number)
   }
+
+  internal init?(rawUTF8: UnsafeBufferPointer<UInt8>) {
+    guard let nameProviding = Self.self as? _ProtoNameProviding.Type,
+      let number = nameProviding._protobuf_nameMap.number(forJSONName: rawUTF8) else {
+      return nil
+    }
+    self.init(rawValue: number)
+  }
 }
--- a/Sources/SwiftProtobuf/Enum.swift
+++ b/Sources/SwiftProtobuf/Enum.swift
@@ -70,6 +70,13 @@ extension Enum {
     self.init(rawValue: number)
   }
 
+  /// Internal convenience initializer that returns the enum value with the
+  /// given name, if it provides names.
+  ///
+  /// Since the text format and JSON names are always identical, we don't need
+  /// to distinguish them.
+  ///
+  /// - Parameter name: Buffer holding the UTF-8 bytes of the desired name.
   internal init?(rawUTF8: UnsafeBufferPointer<UInt8>) {
     guard let nameProviding = Self.self as? _ProtoNameProviding.Type,
       let number = nameProviding._protobuf_nameMap.number(forJSONName: rawUTF8) else {
--- a/Sources/SwiftProtobuf/JSONEncoder.swift
+++ b/Sources/SwiftProtobuf/JSONEncoder.swift
@@ -86,14 +86,10 @@ internal struct JSONEncoder {
         data.append(contentsOf: buff)
     }
 
-    /// Append a `_NameMap.Name` to the JSON text.  As with
-    /// StaticString above, a `_NameMap.Name` provides pre-converted
+    /// Append a `_NameMap.Name` to the JSON text surrounded by quotes.
+    /// As with StaticString above, a `_NameMap.Name` provides pre-converted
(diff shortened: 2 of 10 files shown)


Sources/SwiftProtobuf/Enum.swift
```diff
@@ -69,4 +69,12 @@ extension Enum {
     }
     self.init(rawValue: number)
   }
+
+  internal init?(rawUTF8: UnsafeBufferPointer<UInt8>) {
```
thomasvl: Doc Comment?

Sources/SwiftProtobuf/JSONEncoder.swift
```diff
@@ -94,6 +94,12 @@ internal struct JSONEncoder {
         data.append(contentsOf: name.utf8Buffer)
     }
 
+    internal mutating func appendQuoted(name: _NameMap.Name) {
```
thomasvl: Doc Comment

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -1151,7 +1151,7 @@ internal struct JSONScanner {
       throw JSONDecodingError.truncated
     }
     if currentByte != asciiDoubleQuote {
-      throw JSONDecodingError.malformedString
+      return nil
```
thomasvl: Why isn't this an error any more?

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -216,6 +216,25 @@ internal struct TextFormatScanner {
         }
     }
 
+    private mutating func parseUTF8Identifier() -> UnsafeBufferPointer<UInt8>? {
```
thomasvl: Why optional, seems like this will always return something, it might just have a count of zero?

Sources/SwiftProtobuf/Enum.swift
tbkka: Done.

Sources/SwiftProtobuf/JSONEncoder.swift
tbkka: Done.  Curiously, the `append(name:)` just above is now unused, so I removed that and slid the doc comment down to apply to `appendQuoted(name:)`

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: Nice catch!

While changing this, I noticed that `parseIdentifier()` just below should also be non-optional.  While fixing that, I realized that this exact loop was being repeated in several places, so I edited them to use `parseUTF8Identifier()`, simplifying a number of points in the code.

Thanks!

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: It wasn't really an error before:  before this PR, the only caller of this method already checked this condition.

To clarify, I've just renamed this method to `nextOptionalKey()` to indicate that it returns `nil` if the text doesn't look like a quoted key.  It's now used as the fast path for both field name and enum parsing.  In both of those cases, we can use the `NameMap` to quickly resolve UTF8 bytes without converting to a String.

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -848,18 +838,14 @@ internal struct TextFormatScanner {
         if p == end {
             throw TextFormatDecodingError.malformedText
         }
-        let c = p[0]
         let start = p
```
thomasvl: if this moves into the one case, the default doesn't need to reassign it.

tbkka: Even better, 'start' here is no longer needed at all.  Removed.

PR #628, opened 2017-08-01
EnumValue stemming/naming improvements

### LINKED ISSUE #621: Enums improperly stemmed and conflicting
A bug I encountered while trying out a recent version of the code on some Google-internal protos. Consider the following enum:

```proto
enum FooBar {
  FOO_BAR_BAZ = 0;
  QUUX = 1;
  BAZ = 2;
}
```

Our current stemming logic produces the following Swift enum cases:

```swift
baz = 0
quux = 1
baz = 2
```

Obviously, this doesn't compile. _(Why would someone design an enum like that to begin with?_ 🤷‍♂️  _But that ship has already sailed.)_

So, our stemming logic needs to change. Some options:

* Instead of stemming each case in isolation, stem in the context of all the cases and check for conflicts.
* Always stem "all-or-nothing", the logic we had before.
* Something else.

thomasvl:
Fixes #621

### DIFF
--- a/Sources/PluginLibrary/Descriptor.swift
+++ b/Sources/PluginLibrary/Descriptor.swift
@@ -272,6 +272,7 @@ public final class EnumValueDescriptor {
   public weak var file: FileDescriptor! { return enumType.file }
 
   public var name: String { return proto.name }
+  public var fullName: String
   public var number: Int32 { return proto.number }
 
   public private(set) weak var aliasOf: EnumValueDescriptor?
@@ -285,6 +286,9 @@ public final class EnumValueDescriptor {
     self.index = index
     self.enumType = enumType
     aliasOf = aliasing
+
+    let fullName = "\(enumType.fullName).\(proto.name)"
+    self.fullName = fullName
   }
 }
 
--- a/Sources/PluginLibrary/NamingUtils.swift
+++ b/Sources/PluginLibrary/NamingUtils.swift
@@ -326,63 +326,61 @@ public enum NamingUtils {
     return prefix + "_"
   }
 
-  /// Remove the proto prefix from the given string.  A proto prefix means
-  /// underscores and letter case are ignored.
-  ///
-  /// - Precodition: The two strings must only be 7bit ascii.
-  ///
-  /// - Returns: nil if nothing can be stripped, otherwise returns the stripping
-  ///            string.
-  static func strip(protoPrefix prefix: String, from: String) -> String? {
-    let prefixChars = prefix.lowercased().unicodeScalars
-    precondition(prefixChars.count == prefix.lengthOfBytes(using: .ascii))
-    var prefixIndex = prefixChars.startIndex
-    let prefixEnd = prefixChars.endIndex
+  /// Helper a proto prefix from strings.  A proto prefix means underscores
+  /// and letter case are ignored.
+  struct PrefixStripper {
+    private let prefixChars: String.UnicodeScalarView
 
-    let fromChars = from.lowercased().unicodeScalars
-    precondition(fromChars.count == from.lengthOfBytes(using: .ascii))
-    var fromIndex = fromChars.startIndex
-    let fromEnd = fromChars.endIndex
+    init(prefix: String) {
+      self.prefixChars = prefix.lowercased().replacingOccurrences(of: "_", with: "").unicodeScalars
+    }
 
-    while (prefixIndex != prefixEnd) {
-      if (fromIndex == fromEnd) {
-        // Reached the end of the string while still having prefix to go
-        // nothing to strip.
-        return nil
-      }
+    /// Strip the prefix and return the result, or return nil if it can't
+    /// be stripped.
+    func strip(from: String) -> String? {
+      var prefixIndex = prefixChars.startIndex
+      let prefixEnd = prefixChars.endIndex
 
-      if prefixChars[prefixIndex] == "_" {
+      let fromChars = from.lowercased().unicodeScalars
+      precondition(fromChars.count == from.lengthOfBytes(using: .ascii))
+      var fromIndex = fromChars.startIndex
+      let fromEnd = fromChars.endIndex
+
+      while (prefixIndex != prefixEnd) {
+        if (fromIndex == fromEnd) {
+          // Reached the end of the string while still having prefix to go
+          // nothing to strip.
+          return nil
+        }
+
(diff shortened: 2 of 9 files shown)


Sources/PluginLibrary/Descriptor.swift
```diff
@@ -285,6 +286,9 @@ public final class EnumValueDescriptor {
     self.index = index
     self.enumType = enumType
     aliasOf = aliasing
+
+    let fullName = "\(enumType.fullName).\(proto.name)"
```
allevato: Does this work better as a `lazy` property (or do you still have performance concerns there)?

Sources/PluginLibrary/NamingUtils.swift
```diff
@@ -326,63 +326,61 @@ public enum NamingUtils {
...
+    }
+
+    /// Strip the prefix and return the result, or return nil if it can't
+    /// be stripped.
+    func strip(from: String) -> String? {
+      var prefixIndex = prefixChars.startIndex
+      let prefixEnd = prefixChars.endIndex
+
+      let fromChars = from.lowercased().unicodeScalars
+      precondition(fromChars.count == from.lengthOfBytes(using: .ascii))
```
allevato: This is just a sanity check that the case name we got was ASCII-only, right? Did you have any issues where this wasn't the case?

Sources/PluginLibrary/NamingUtils.swift
```diff
@@ -326,63 +326,61 @@ public enum NamingUtils {
...
+      var fromIndex = fromChars.startIndex
+      let fromEnd = fromChars.endIndex
+
+      while (prefixIndex != prefixEnd) {
+        if (fromIndex == fromEnd) {
```
allevato: Nit: I would personally use `guard`s for many of these early-return checks, but I don't have a strong opinion about it.

Sources/PluginLibrary/SwiftProtobufNamer.swift
```diff
@@ -81,18 +82,84 @@ public final class SwiftProtobufNamer {
...
-    let baseName = enumValue.name
-    if let stripped = NamingUtils.strip(protoPrefix: enumValue.enumType.name, from: baseName) {
-      let camelCased = NamingUtils.toLowerCamelCase(stripped)
-      if isValidSwiftIdentifier(camelCased) {
-        return NamingUtils.sanitize(enumCaseName: camelCased)
+  /// Compute the short names to use for the values of this enum.
+  private func computeRelativeNames(enum e: EnumDescriptor) {
+    let stripper = NamingUtils.PrefixStripper(prefix: e.name)
+
+    /// Determine the initial canidate name for the name before
```
allevato: Nit: `//`

Sources/PluginLibrary/SwiftProtobufNamer.swift
```diff
@@ -81,18 +82,84 @@ public final class SwiftProtobufNamer {
...
+      if let stripped = stripper.strip(from: baseName) {
+        let camelCased = NamingUtils.toLowerCamelCase(stripped)
+        if isValidSwiftIdentifier(camelCased) {
+          return camelCased
+        }
+      }
+      return NamingUtils.toLowerCamelCase(baseName)
+    }
+
+    // Bucked based on canidate names to check for duplicates.
```
allevato: `Bucketed`? and `candidate`

Sources/PluginLibrary/SwiftProtobufNamer.swift
```diff
@@ -81,18 +82,84 @@ public final class SwiftProtobufNamer {
...
+    for (camelCased, enumValues) in canidates {
+      // If there is only one, sanitize and cache it.
+      guard enumValues.count > 1 else {
+        enumValueRelativeNameCache[enumValues.first!.fullName] =
+          NamingUtils.sanitize(enumCaseName: camelCased)
+        continue
+      }
+
+      // There are two possible cases:
+      // 1. There is the main entry and then all aliase for it that
```
allevato: aliases

Sources/PluginLibrary/Descriptor.swift
thomasvl: I've been trying to keep `lazy` to a minimum in here pending some performance analysis on generation, as there is an easy way to thread the plugin and generate files in parallel, but all of the lazy usage would have to go.

Sources/PluginLibrary/NamingUtils.swift
thomasvl: Yea, was just a sanity check, and I'm just assuming protoc will do this instead.

Sources/PluginLibrary/NamingUtils.swift
thomasvl: So I've been trying to figure out a pattern for `guard` vs `if`, the two possible reasons I've come up with is if a `let` is needed for the rest of the code, then `guard` seems to make more sense.  The other is when the logic gate reads better with an `else` meaning use the guard.  But yea, it is hard to say when one is better in a lot of cases.  ☹️

Sources/PluginLibrary/SwiftProtobufNamer.swift
thomasvl: Even though it documents the nested function?

Sources/PluginLibrary/SwiftProtobufNamer.swift
thomasvl: Done.

Sources/PluginLibrary/SwiftProtobufNamer.swift
thomasvl: Done.

Sources/PluginLibrary/SwiftProtobufNamer.swift
allevato: Woops, I didn't even realize that was a nested function—I thought it was just a comment inside this function. Nvm.

Sources/PluginLibrary/NamingUtils.swift
allevato: To me, `guard` is a nice visual indicator of "this checks a condition that demands an early exit from this scope (or more rarely, loop continue)"; it highlights special cases well whether or not there's a binding involved. (Though it absolutely makes the use of that binding easier as well by avoiding extra levels of nesting.)

PR #630, opened 2017-08-01
Shim XCTAssertNoThrow for older versions of Swift.

thomasvl:
XCTAssertNoThrow got added in Xcode 8.2, but there isn't a good way
to test the Xcode version (or XCTest version), so instead use the
Swift version to figure out when to shim it so the test builds/runs.


thomasvl:
@tbkka did you have to disable these tests on linux because of an issue with older versions of Foundation there or was it just because of this method?  Just wondering if we should update the lower conditionals.

tbkka:
If I remember correctly, Linux also had problems with the other Foundation stuff here.

PR #631, opened 2017-08-01
Xcode compatibility fixes...

tbkka:
Several minor fixes that correct warnings and compile errors seen in various versions of Xcode.  With this, we now build and pass tests on Xcode 8.1 (Swift 3.0.1), Xcode 8.2 (Swift 3.0.2), Xcode 8.3 (Swift 3.1) and a pre-release version of Xcode 9 (Swift 4).

### DIFF
--- a/Tests/SwiftProtobufTests/Test_BinaryDelimited.swift
+++ b/Tests/SwiftProtobufTests/Test_BinaryDelimited.swift
@@ -15,7 +15,7 @@ import SwiftProtobuf
 class Test_BinaryDelimited: XCTestCase {
 
   func testEverything() {
-#if swift(>=3.1) || os(OSX) || os(iOS)
+#if swift(>=3.1)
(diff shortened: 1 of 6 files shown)


tbkka:
I just reverted the change to disable the test that used XCTAssertNoThrow on macOS, since Thomas just landed a better fix.

Tests/SwiftProtobufTests/Test_BinaryDelimited.swift
```diff
@@ -15,7 +15,7 @@ import SwiftProtobuf
 class Test_BinaryDelimited: XCTestCase {
 
   func testEverything() {
-#if swift(>=3.1) || os(OSX) || os(iOS)
+#if swift(>=3.1)
```
thomasvl: Don't think you want this with my last change now pushed.

PR #633, opened 2017-08-01
Perf: pre-size arrays for binary packed fields

tbkka:
Decoding packed varint fields was spending a lot of time appending to the arrays.

With this change, binary decoding of `100 'repeated int32'` is now on par with the C++ implementation.


tbkka:
We would not lose it entirely:  A stream based parser should still work in generously-sized chunks, so this should still be able to retrieve chunks of bytes, size to handle all of that data, and repeat.

PR #635, opened 2017-08-02
Use _visitRepeated consistently for all repeated field types

tbkka:
This fixes the callback signature to avoid capture ambiguity and updates the repeated enum and message handlers to use the generic facility.

### DIFF
--- a/Sources/SwiftProtobuf/JSONEncodingVisitor.swift
+++ b/Sources/SwiftProtobuf/JSONEncodingVisitor.swift
@@ -134,7 +134,11 @@ internal struct JSONEncodingVisitor: Visitor {
     encoder.putBytesValue(value: value)
   }
 
-  private mutating func _visitRepeated<T>(value: [T], fieldNumber: Int, encode: (T) -> ()) throws {
+  private mutating func _visitRepeated<T>(
+    value: [T],
+    fieldNumber: Int,
+    encode: (inout JSONEncoder, T) throws -> ()
+  ) throws {
     try startField(for: fieldNumber)
     var comma = false
     encoder.startArray()
@@ -143,7 +147,7 @@ internal struct JSONEncodingVisitor: Visitor {
           encoder.comma()
       }
       comma = true
-      encode(v)
+      try encode(&encoder, v)
     }
     encoder.endArray()
   }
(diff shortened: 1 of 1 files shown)


Sources/SwiftProtobuf/JSONEncodingVisitor.swift
```diff
@@ -134,7 +134,11 @@ internal struct JSONEncodingVisitor: Visitor {
     encoder.putBytesValue(value: value)
   }
 
-  private mutating func _visitRepeated<T>(value: [T], fieldNumber: Int, encode: (T) -> ()) throws {
+  private mutating func _visitRepeated<T>(
+    value: [T],
+    fieldNumber: Int,
+    encode: (inout JSONEncoder, T) throws -> ()
```
thomasvl: Why pass the encoder as inout vs. just use the ivar like before?

tbkka: This fixes a subtle error in the old code that I've just recently figured out, thanks to some new diagnostics that are appearing in Swift 4.

The old code was structured like this:
```
struct Foo {
   private mutating func tool(closure: () -> ()) {
      ... stuff that mutated self ...
      closure()
   }
   mutating func client() {
      _tool {
           ... code that mutates self ...
      }
   }
}
```

Here's my understanding:

If this were a class and had reference semantics, then it would be clear that the passed-in closure and the code in the `tool` method are both mutating the same object.  That intuition is what guided the first implementation here.

But for structs, that's not how it _should_ operate.  Instead, it should behave as though the closure is constructed with captured copies of all the referenced values before being passed to the `tool` method.  In that case, the code in `tool` and the code in the closure would be mutating different objects, and in fact the mutations performed in the closure would be lost.

The Swift compiler doesn't actually work this way today, of course, else our original code here would not have worked.  Instead, the compiler acts like it is usually inlining everything in the obvious way (almost as if `tool` were a macro).  But the Swift 4 compiler has recently started warning about some occurrences of this (presumably before they fix the semantics in some later version).

The solution, of course, is for the closure to accept the relevant objects to mutate as inout params, as I've done in this change.  Then it's clear that the inside of `tool` and the closure are, in fact, intended to mutate the same structs.

BTW, the compiler warning seems to be inconsistent in the current Swift 4 snapshots; so far I've only seen it in release builds and even then only when there are additional complications -- such as generic arguments -- in the code.  I've only seen them because I've been running the performance harness regularly and it builds in release mode.  Our current `make test` cannot currently build in release mode, so cannot catch these kinds of issues.

thomasvl: Thank's for the detailed reply.

PR #636, opened 2017-08-03
Perf:  direct parsing of bool in text format

tbkka:
This avoids a redundant parsing/conversion of String objects, with significant perf improvement for parsing boolean fields.


Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -720,6 +722,28 @@ internal struct TextFormatScanner {
         return d
     }
 
+    private mutating func skipOptionalKeywordCaseSensitive(bytes: [UInt8]) -> Bool {
+        let start = p
+        for b in bytes {
+            if p == end || p[0] != b{
```
thomasvl: space before '{'

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -720,6 +722,28 @@ internal struct TextFormatScanner {
...
+                p = start
+                return false
+            }
+            p += 1
+        }
+        if p == end {
+            return true
+        }
+        let c = p[0]
+        if ((c >= asciiUpperA && c <= asciiUpperZ)
```
thomasvl: Is this check complete enough?  what about a number ("true2you")?  Would it be safer to ensure this is followed by whitespace, a comma, ']', or '}'?  Since that's normally breaks a token in text format?

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -817,15 +841,27 @@ internal struct TextFormatScanner {
             throw TextFormatDecodingError.malformedText
         }
         let c = p[0]
+        p += 1
         switch c {
-        case asciiZero, asciiOne, asciiLowerF, asciiUpperF, asciiLowerT, asciiUpperT:
-            switch parseIdentifier() {
-            case "0", "f", "false", "False":
+        case asciiZero:
+            return false
```
thomasvl: Does this allow "0123" to become "0foo:1" to become "false" and then would "foo" get parsed as a field name?

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -817,15 +841,27 @@ internal struct TextFormatScanner {
...
+        case asciiOne:
+            return true
```
thomasvl: Same concern?

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -817,15 +841,27 @@ internal struct TextFormatScanner {
...
+               || ((p[0] < asciiUpperA || p[0] > asciiUpperZ)
+                   && (p[0] < asciiLowerA || p[0] > asciiLowerZ)) {
                 return false
-            case "1", "t", "true", "True":
+            }
+        case asciiLowerT, asciiUpperT:
+            let rue = [asciiLowerR, asciiLowerU, asciiLowerE]
+            if p == end
+               || skipOptionalKeywordCaseSensitive(bytes: rue)
+               || ((p[0] < asciiUpperA || p[0] > asciiUpperZ)
```
thomasvl: Same concern.

Seems like in all these cases, you might want to set a proposed value, and then ensure the next thing is whitespace or a TextFormat delimiter to the malformed inputs ("0foo:1", etc.), don't extract a subset and then parse the rest of this token as a field.

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: Done.

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: Fixed (and tested).

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: Fixed (and tested).

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: Fixed (and tested)

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: Fixed (and tested).

tbkka:
I added tests for the issues you raised -- which verified that the previous code was, in fact, broken.

I then reworked the code along the lines you suggested.

PR #642, opened 2017-08-15
Support logging the directory via an env variable.

thomasvl:
To help in debugging generation issues, add an env variable that
will get the request logged so it is easier to capture a request
from a build system/script.


tbkka:
Q: Should this be documented in Documentation/PLUGIN.md?

thomasvl:
> Q: Should this be documented in Documentation/PLUGIN.md?

I don't think there because folks using the plugin to generate their code should need it.  But if we wanted INTERNALS to cover some of the plugin design, then we could add it there?

PR #650, opened 2017-08-25
Speed up .pb.swift generation

allevato:
These changes make .pb.swift generation by about 200–250% (using Instruments on some of our internal protos).

Some speed-ups came from rewriting name-transforming functions to use `UnicodeScalar` instead of `Character` and avoiding excess string creation. Some more performance got squeezed out of `CodePrinter` by dropping the index-scanning-and-slicing code and simply iterating over and appending the scalars passed into `print`. Appending single scalars one-by-one turns out to be faster after all.

### DIFF
--- a/Sources/PluginLibrary/NamingUtils.swift
+++ b/Sources/PluginLibrary/NamingUtils.swift
@@ -171,7 +171,7 @@ fileprivate func isAllUnderscore(_ s: String) -> Bool {
   if s.isEmpty {
     return false
   }
-  for c in s.characters {
+  for c in s.unicodeScalars {
     if c != "_" {return false}
   }
   return true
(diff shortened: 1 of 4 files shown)


Sources/PluginLibrary/NamingUtils.swift
```diff
@@ -198,90 +198,74 @@ fileprivate func sanitizeTypeName(_ s: String, disambiguator: String) -> String
...
-        current = last
-      }
-      if c != "_" {
-        out.append(current)
-        current = c
+        current = makeUnicodeScalarView(from: scalar)
       }
+    } else if last == "_" {
+      out.append(current)
+      current = makeUnicodeScalarView(from: last)
```
thomasvl: What happens on an invalid character, it gets dropped, but does it get the states right?  i.e. before what happened to "abc$xyz" and what happens now?

As long as we only ever send field names, enums, etc. here it shouldn't happen.  Just not sure if we want to trap it in some way as an added safety net.

allevato: I just tested it and characters outside a-z, A-Z, 0-9, and _ still cause it to split correctly around them, because `lastIsUpper` and `lastIsLower` still get updated for every character, so we still split around them when we encounter the next character.

Sources/PluginLibrary/NamingUtils.swift
```diff
@@ -198,90 +198,74 @@ fileprivate func sanitizeTypeName(_ s: String, disambiguator: String) -> String
...
 
-fileprivate let digits: Set<String> = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9"]
 
 fileprivate func splitIdentifier(_ s: String) -> [String] {
-  var out = [String]()
-  var current = ""
-  var last = ""
+  var out: [String.UnicodeScalarView] = []
+  var current = String.UnicodeScalarView()
+  var last: UnicodeScalar = "\0"
```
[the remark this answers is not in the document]
allevato: Done.

PR #662, opened 2017-09-18
Update Xcode project files for Xcode 9

hauptanja:


SwiftProtobuf.xcodeproj/project.pbxproj
```diff
@@ -2264,7 +2280,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
```
thomasvl: Why was this raised?

Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -238,7 +238,7 @@ internal struct BinaryDecoder: Decoder {
                 if bodyBytes % itemSize != 0 || itemCount > UInt64(Int.max) {
                     throw BinaryDecodingError.truncated
                 }
-                value.reserveCapacity(value.count + Int(extendingOrTruncating: itemCount))
+                value.reserveCapacity(value.count + Int(truncatingIfNeeded: itemCount))
```
thomasvl: This probably isn't right, and it likely means an update to https://github.com/apple/swift-protobuf/blob/master/Sources/SwiftProtobuf/MathUtils.swift#L42-L89 instead so the code will still compile with older Xcodes also.

thomasvl:
@tbkka Looks like some of the extensions to bridge to swift 4 need an update.

allevato:
As @thomasvl alluded to above, please use conditional Swift version checks to ensure that the code still compiles in Swift 3.x as well. We don't (yet) have plans to drop compatibility with it.

thomasvl:
Take a look at `Sources/SwiftProtobuf/JSONIntegerConverting.swift`, the way we support current Swift and old versions define an extension for the old versions that provides the new name.  This way the code can be updated to the current name, but still build with older versions of the language.

Sources/SwiftProtobuf/BinaryDecoder.swift
tbkka: Yep.  Changing to the newest Swift 4 syntax is fine, but without updating MathUtils.swift, this entirely breaks Swift 3 compatibility.

tbkka:
@hauptanja -- Please look at PR #663.  This just has the change from `extendingOrTruncated` to `truncatingIfNeeded` (tracking the recent change to this initializer in Swift 4), along with the necessary changes to MathUtils.swift to preserve Swift 3 compatibility.  Note that this _does_ rename the initializer used everywhere -- hopefully someday we can just remove the compatibility shim from MathUtils.swift someday when we decide to drop Swift 3 support and it will "just work."

PR #663 does not try to change the xcodeproj files.  I think we probably should update those, and would like to continue to use this PR #662 to discuss those.

First, can you explain how you updated the xcodeproj files?  Did you just let Xcode do the update or did you make specific changes?  (It may be appropriate to "just use Xcode's changes", but I want to understand what that is first.)  Did you try regenerating the xcodeproj files using the `swift package` subcommand to see if it did the same thing?

hauptanja:
@tbkka I applied the recommended settings from Xcode. I did not try regenerating xcodeproj files.

SwiftProtobuf.xcodeproj/project.pbxproj
```diff
@@ -2090,6 +2096,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				APPLICATION_EXTENSION_API_ONLY = YES;
+				"CODE_SIGN_IDENTITY[sdk=appletvos*]" = "";
```
tbkka: I suspect we don't want to include a code signing identity in our Xcode project files.

thomasvl:
#671 should cover what was left here.

PR #663, opened 2017-09-19
Change `extendingOrTruncating` to `truncatingIfNeeded`

tbkka:
Swift 4 prerelease (until Xcode 9 beta 4 at least) has been using the
former name for the truncating initializer and just recently switched
to the latter term.

Unlike PR#662, this also updates MathUtils.swift so that we
preserve compatibility with Swift 3.  It also omits any changes
to the xcodeproj files.

Tested using command-line tools with Xcode 8.1, 8.2., 8.3, and 9.0 GM Seed (9A235).

### DIFF
--- a/Sources/SwiftProtobuf/MathUtils.swift
+++ b/Sources/SwiftProtobuf/MathUtils.swift
@@ -43,46 +43,46 @@ internal func div<T : SignedInteger>(_ a: T, _ b: T) -> T {
 //
 // Swift 3 called this initializer "truncatingBitPattern";
 // Swift 4 (prerelease) changed this initializer to
-// "extendingOrTruncating".
+// "truncatingIfNeeded".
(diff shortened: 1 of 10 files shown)


Sources/SwiftProtobuf/MathUtils.swift
```diff
@@ -43,46 +43,46 @@ internal func div<T : SignedInteger>(_ a: T, _ b: T) -> T {
 //
 // Swift 3 called this initializer "truncatingBitPattern";
 // Swift 4 (prerelease) changed this initializer to
```
thomasvl: Maybe drop the "(prerelease)" part since I doubt it is changing again?

tbkka: Done.

Issue #665, opened 2017-09-20
Cocoapods and Swift versioning?


tbkka:
Just pushed a 0.9.905 tag in response to Issue #664.  However, `pod spec lint` is failing with the following warning/error:
```
 -> SwiftProtobuf (0.9.905)
    - WARN  | swift_version: The validator for Swift projects uses Swift 3.0 by default, if you are using a different version of swift you can use a `.swift-version` file to set the version for your Pod. For example to use Swift 2.3, run: 
    `echo "2.3" > .swift-version`

Analyzed 1 podspec.

[!] The spec did not pass validation, due to 1 warning (but you can use `--allow-warnings` to ignore it).
```

I'm not sure what this really means.  It gives the impression that every podspec should specify an exact `swift_version`, but Since our code builds with Swift 3 and Swift 4, I don't think we should need to do that?

thomasvl:
The CocoaPods approach seems to be to declare you branches as only supporting one swift version, and you pick which a version to use the swift version desired.  I'll dig into their docs again to see what I can find; http://blog.cocoapods.org/CocoaPods-1.1.0/ has some info, but talks about fears with Swift language changes, and we're sorta past that point now generally (and we worked around some of it).

thomasvl:
Found some more info/history on https://github.com/CocoaPods/CocoaPods/issues/6791.

thomasvl:
@tbkka where version of cocoapods do you have installed?  `pod --version` for me is _1.2.1_, and I'm not getting any warning on `pod spec lint SwiftProtobuf.podspec`.

thomasvl:
I'm gonna go ahead and do the push in the mean time with my version since it seem like it is happy.

thomasvl:
Spec is live, I'm curious now if the warning is specific to a cocoapods version you had?

thomasvl:
Actually, I'll be it is version specific, `pod spec lint --help` and `pod trunk push --help` both document a `--swift-version=VERSION` argument that takes precedence over a `.swift-version` file.  So given we want to support multiple, I vote we continue to leave off the file and we can always test with multiple if we wanted to be paranoid.

tbkka:
@thomasvl:  Sounds like a plan.  I'll close this and we'll wait to see if there are any real problems from omitting the swift-version.

PR #666, opened 2017-09-20
Set master to version 0.9.906 for ongoing work

tbkka:
This should probably be merged only after the cocoa pods issue mentioned in PR #665 is addressed.


thomasvl:
>This should probably be merged only after the cocoa pods issue mentioned in PR #665 is addressed.

You can actually land this.  The push can be down with the podspec from the tag since it has the correct version.

PR #667, opened 2017-09-20
Obey common env vars to locate 'swift' and 'swiftc' commands

tbkka:
The Swift compiler folks have established some conventions for working with multiple versions of Swift on the same system.

### DIFF
--- a/Performance/runners/swift.sh
+++ b/Performance/runners/swift.sh
@@ -16,6 +16,14 @@
 #
 # -----------------------------------------------------------------------------
 
+# Use xcrun if it's present (macOS) or omit it if it's not (Linux)
+XCRUN=`which xcrun`
+if [ -n "${XCRUN:-}" ]; then
+    XCRUN="${XCRUN} -sdk macosx"
+fi
+# How to run 'swiftc': Use SWIFT_EXEC if it's set or just plain "swiftc"
+SWIFTC=${SWIFT_EXEC:-swiftc}
+
 function run_swift_harness() {
   # Wrapped in a subshell to prevent state from leaking out (important since
   # this gets run multiple times during a comparison).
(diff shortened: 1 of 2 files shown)


Performance/runners/swift.sh
```diff
@@ -16,6 +16,14 @@
 #
 # -----------------------------------------------------------------------------
 
+# Use xcrun if it's present (macOS) or omit it if it's not (Linux)
+XCRUN=`which xcrun`
+if [ -n "${XCRUN:-}" ]; then
+    XCRUN="${XCRUN} -sdk macosx"
+fi
+# How to run 'swiftc': Use SWIFT_EXEC if it's set or just plain "swiftc"
+SWIFTC=${SWIFT_EXEC:-swiftc}
```
thomasvl: The swiftc handing is different between the two scripts, there a reason for that or should it be the same?

tbkka: It's for two different programs:  `swift` in one and `swiftc` in the other.  The `SWIFT_EXEC` variable specifies the full path to `swiftc`.  To get the `swift` command, we look in the same directory for a different filename.

Someone else pointed out that as written, these scripts don't handle spaces in pathnames.  I'll take a quick stab at fixing that.

thomasvl: The spaces will be harder, you'll likely need to use a bash array since you want XCRUN to pick up arguments, etc.

tbkka: After thinking about it a bit, I don't think we need to worry much about `XCRUN`, since it's very unlikely that anyone will have that anywhere other than the standard `/usr/bin/xcrun`.  The `XCRUN` variable is really only to handle macOS (has it) vs. Linux (doesn't have it).  So I really only need to worry about space-containing paths for `SWIFT` and `SWIFTC`, which is a lot simpler.

PR #670, opened 2017-09-29
Repeated field decode performance for Text & JSON

tbkka:
This provides almost a 2x performance improvement when decoding repeated fields that use array notation in Text and JSON formats.  (JSON always uses `[...]` array notation; TextFormat only uses it for packed fields.)

The change simply counts the array elements in advance so we can resize the array once.  For arrays of certain primitive types, this count can be as simple as counting comma characters.

### DIFF
--- a/Sources/SwiftProtobuf/JSONDecoder.swift
+++ b/Sources/SwiftProtobuf/JSONDecoder.swift
@@ -74,6 +74,8 @@ internal struct JSONDecoder: Decoder {
     if scanner.skipOptionalArrayEnd() {
       return
     }
+    let count = try scanner.countArrayElementsQuickly()
+    value.reserveCapacity(count)
     while true {
       let n = try scanner.nextFloat()
       value.append(n)
@@ -108,6 +110,8 @@ internal struct JSONDecoder: Decoder {
     if scanner.skipOptionalArrayEnd() {
       return
     }
+    let count = try scanner.countArrayElementsQuickly()
+    value.reserveCapacity(count)
     while true {
       let n = try scanner.nextDouble()
       value.append(n)
@@ -150,6 +154,8 @@ internal struct JSONDecoder: Decoder {
     if scanner.skipOptionalArrayEnd() {
       return
     }
+    let count = try scanner.countArrayElementsQuickly()
+    value.reserveCapacity(count)
     while true {
       let n = try scanner.nextSInt()
       if n > Int64(Int32.max) || n < Int64(Int32.min) {
@@ -187,6 +193,8 @@ internal struct JSONDecoder: Decoder {
     if scanner.skipOptionalArrayEnd() {
       return
     }
+    let count = try scanner.countArrayElementsQuickly()
+    value.reserveCapacity(count)
     while true {
       let n = try scanner.nextSInt()
       value.append(n)
@@ -229,6 +237,8 @@ internal struct JSONDecoder: Decoder {
     if scanner.skipOptionalArrayEnd() {
       return
     }
+    let count = try scanner.countArrayElementsQuickly()
+    value.reserveCapacity(count)
     while true {
       let n = try scanner.nextUInt()
       if n > UInt64(UInt32.max) {
@@ -266,6 +276,8 @@ internal struct JSONDecoder: Decoder {
     if scanner.skipOptionalArrayEnd() {
       return
     }
+    let count = try scanner.countArrayElementsQuickly()
+    value.reserveCapacity(count)
     while true {
       let n = try scanner.nextUInt()
       value.append(n)
@@ -356,6 +368,9 @@ internal struct JSONDecoder: Decoder {
     if isMapKey {
       value = try scanner.nextQuotedBool()
     } else {
+      // Unlike other types, booleans are *only* quoted
+      // as map keys.  Reject quoted booleans when we're
+      // not looking for a map key.
       value = try scanner.nextBool()
     }
   }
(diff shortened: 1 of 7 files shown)


Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -1076,4 +1076,25 @@ internal struct TextFormatScanner {
...
+    internal mutating func countArrayElementsQuickly() throws -> Int {
+        let start = p
+        var count = 0
+        while p != end {
+            switch p[0] {
+            case asciiCloseSquareBracket:
+                p = start
+                return count
+            case asciiComma:
+                count += 1
```
tbkka: Are comments a problem here?  A comment might contain extra commas, which would bloat the count here.  The only bad effect would be over-allocating the resulting array.

Fortunately, JSON doesn't allow comments, so this cannot happen there.

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -1425,24 +1477,23 @@ internal struct JSONScanner {
   // they don't know; newer clients may reject the same input due to
   // schema mismatches or other issues.
   private mutating func skipString() throws {
-    if currentByte != asciiDoubleQuote {
-      throw JSONDecodingError.malformedString
+    guard hasMoreContent && currentByte == asciiDoubleQuote else {
+      throw JSONDecodingError.truncated
```
thomasvl: .truncated seems correct if hasMoreContent failed, but if we really have to worry about coming in on something other than a quote, it seems like .malformedString is more correct.  Is the currentByte check more of an assert/precondition that isn't really needed in production code?

Sources/SwiftProtobuf/JSONDecoder.swift
```diff
@@ -74,6 +74,8 @@ internal struct JSONDecoder: Decoder {
     if scanner.skipOptionalArrayEnd() {
       return
     }
+    let count = try scanner.countArrayElementsQuickly()
+    value.reserveCapacity(count)
```
thomasvl: Shouldn't all these use count + value.count also?

allevato:
To echo @thomasvl's general comment above, this is one of those kinds of changes that really concerns me where highly use-case-specific code is being written that slices away some design generality for performance reasons (2x sounds good, but what are the absolute numbers?) and may restrict us in the future from making deeper API changes/improvements because we'll have to rationalize the performance drop.

Perhaps one of the things we should look into sooner rather than later (post 1.0) is reconsidering all of the encoding/decoding based on stream-based APIs instead of the assumption that everything will be in memory at once. That would help us set a baseline for what kind of optimizations we can reasonably expect to make going forward.

Sources/SwiftProtobuf/JSONDecoder.swift
tbkka: Not for JSON.  Text format needs it to handle cases like this:
```
someKey: [1, 2, 3]
someKey: [4, 5, 6]
```
but JSON doesn't permit duplicate keys.

Sources/SwiftProtobuf/TextFormatScanner.swift
thomasvl: Actually, does array mode support newlines?  i.e. -
```
foo: [ 1, 2,
  3, 4]
```
If it doesn't, then there is no way for a comment.  If it does, you could use flag to track being in a comment and once it starts, skip until the end of line.

tbkka:
The code here will still work correctly even if the lookahead fails.  Which is precisely what we want for stream support:  The non-streaming case will be as fast as possible, only stream parsing will be slower because the lookahead might underestimate the count, we won't expand the array enough, and the array logic will end up performing additional expansions when it tries to `append()` beyond the initial allocation.

Would it be clearer if I renamed the methods to `underestimateArrayElementCount{Quickly,Carefully}`?

I've also started adding some discussion to #130 to help us clarify the goals for stream parsing and hash out some possible approaches.

thomasvl:
Personally I'm find with the names you used.  My concern is more just if we switch to a more stream based api, odds are all cases will lose support for this look head (or we're going to have to bloat the code by seeing if the "stream" wraps a buffer and special case it).  The C++ protobuf code does a bunch of this sorta stuff, but it makes some of the parsing code extremely complex and results in much larger binaries.

allevato:
Yeah, if we end up implementing lookahead for a stream-based API, we've really just traded one set of reällocations for another—instead of growing the array of field values dynamically without knowing the upper bound, we have to grow a lookahead buffer dynamically without knowing the upper bound. So that would potentially neutralize the gains here (depending on which is more costly).

I just want to make sure we don't end up in a situation where we can't make important design improvements and API changes because we'll say "it's some factor slower and that's untenable."

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: Good point.  I've split this into two separate checks that throw separate errors.

tbkka:
> ... we have to grow a lookahead buffer dynamically without knowing the upper bound ...

Why?  As I mentioned above, an underestimate is perfectly acceptable, so the code as written does not *require* unbounded lookahead.  But it can benefit from whatever lookahead might be available.  In particular, if the following array is only partially available, it can still scan what is available to yield an underestimate and use that count for the initial capacity.  That count will still improve the performance since it will reduce later expansions.  If no lookahead is available, then the code behaves just as it did before.

Sources/SwiftProtobuf/JSONDecoder.swift
thomasvl: Ah, right.

Sources/SwiftProtobuf/TextFormatScanner.swift
thomasvl: Did you get a chance to double check this support?

tbkka: I'm pretty sure comments can occur here, so this does need to be fixed up to avoid overestimating.

PR #677, opened 2017-10-16
Fix #676, API docs have wrong signature for isEqualTo()

### LINKED ISSUE #676: API docs have wrong signature for isEqualTo()
In the API docs it says: https://github.com/apple/swift-protobuf/blob/9501879a79221a355daf87e52475c3bf7c3ed333/Documentation/API.md#overridable-message-methods

Such that there is a computed variable `public var isEqual(other: Example) -> Bool` that can be overridden in an extension to `Example`, but I can't see how to do that.

I've attempted to write a function with the signature `public func isEqual(other: Example) -> Bool`, but it doesn't get called. There is also no apparent reference to `isEqual` in the code base.

tengyifei:


tbkka:
Thanks!  Let us know if you run into any other problems!

PR #678, opened 2017-10-17
Update some plugin code for Swift 4.

thomasvl:
Warnings:
```
Sources/PluginLibrary/NamingUtils.swift:193:23: 'substring(to:)' is deprecated: Please use String slicing subscript with a 'partial range upto' operator.
Sources/protoc-gen-swift/StringUtils.swift:41:18: 'substring(to:)' is deprecated: Please use String slicing subscript with a 'partial range upto' operator.
Sources/protoc-gen-swift/StringUtils.swift:41:47: 'substring(from:)' is deprecated: Please use String slicing subscript with a 'partial range from' operator.
```
Errors:
```
Sources/PluginLibrary/NamingUtils.swift:369:14: Subscripts returning String were obsoleted in Swift 4; explicitly construct a String from subscripted result
```

### DIFF
--- a/Sources/PluginLibrary/NamingUtils.swift
+++ b/Sources/PluginLibrary/NamingUtils.swift
@@ -190,7 +190,11 @@ fileprivate func sanitizeTypeName(_ s: String, disambiguator: String) -> String 
     // the disambiguator, sanitizing the root, then re-adding the
     // disambiguator:
     let e = s.index(s.endIndex, offsetBy: -disambiguator.characters.count)
-    let truncated = s.substring(to: e)
+    #if swift(>=4.0)
+      let truncated = String(s[..<e])
+    #else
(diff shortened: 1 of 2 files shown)


Sources/PluginLibrary/NamingUtils.swift
```diff
@@ -366,7 +370,11 @@ public enum NamingUtils {
 
       let count = fromChars.distance(from: fromChars.startIndex, to: fromIndex)
       let idx = from.index(from.startIndex, offsetBy: count)
-      return from[idx..<from.endIndex]
+      #if swift(>=4.0)
+        return String(from[idx..<from.endIndex])
+      #else
+        return from[idx..<from.endIndex]
```
tbkka: Personally, I would omit the conditional here and only have the new code, which will work fine on Swift 3.

Issue #125, opened 2016-10-28
A dependency of swift-protobuf also builds the (unwanted) protoc-gen-swift CLI utility


gmosx:
I have a project where I would like to use some Swift code generated from .proto files, using the `protoc-gen-swift` plugin. As described in the documentation, I added the following dependency to my project:

```
dependencies: [
        .Package(url: "https://github.com/apple/swift-protobuf.git", Version(0,9,23))
]
```

However, whenever I try to build my project (using `swift build`) the `protoc-gen-swift` program is also built, further delaying the already longish build process.

Any idea how to avoid building the `protoc-gen-swift` command on projects that only need to use the generated files?

thomasvl:
Looks like SwiftPM can't do this.  The [docs for manifest](https://github.com/apple/swift-package-manager/blob/master/Documentation/Reference.md#package) say:

> ... There should only be one package declared per manifest.

I think we might be able to hack/work around this if we moved `Sources/PluginLibrary/` and `Sources/protoc-gen-swift/` into a top level `protoc-gen-swift` directory.  i.e. - make the root of the project (and Package.swift) just the runtime support, and have a directory inside with plugin itself (that can reference the parent project to get the runtime).  External stuff might not be able to depend on the plugin, but that might be a better state in that everyone wanting to use the runtime doesn't have to end up depending on the plugin also.

@tbkka @allevato thoughts?

thomasvl:
@tbkka should we also bump this up on the list as something to sort out before release?

tbkka:
This isn't something we'll be able to fix without changes to Swift PM.  We should file an issue on Swift PM (if there isn't one already) and track that so we can update when it's possible to do better.

thomasvl:
So the code shuffle suggested wouldn't work?

jcanizales:
> External stuff might not be able to depend on the plugin

I expect that https://github.com/grpc/grpc-swift will eventually need to depend on this plugin, same as with the ObjC versions.

thomasvl:
> I expect that https://github.com/grpc/grpc-swift will eventually need to depend on this plugin, same as with the ObjC versions.

So it will be written in Swift?  I figured you'd be in C++ for any common support between the different grpc impls.

jcanizales:
Yeah, Swift. For this one we're prioritizing lowering the barriers to form
a community. It might even turn out to be easier, despite not having the
resolved Descriptor class to reuse, on account of having string
interpolation for the generation.

[...]

PR #679, opened 2017-10-17
Add a SwiftPM 4 version of the manifest.

thomasvl:
Looking at the SwiftPM docs:
  https://github.com/apple/swift-package-manager/blob/master/Documentation/Usage.md#version-specific-manifest-selection
this naming seems to be what is needed to support SwiftPM 3 getting
the old file while folks using SwiftPM 4 will get the new.

Hopefully this is a step in the direction of supporting folks only
having to build the parts they want, i.e. -
  https://github.com/apple/swift-protobuf/issues/125

### DIFF
--- a/Package@swift-4.swift
+++ b/Package@swift-4.swift
@@ -1,0 +1,34 @@
+// swift-tools-version:4.0
+
+// Package.swift
+//
+// Copyright (c) 2014 - 2017 Apple Inc. and the project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/master/LICENSE.txt
+//
+
+import PackageDescription
+
+let package = Package(
+  name: "SwiftProtobuf",
+  products: [
+    .executable(name: "protoc-gen-swift", targets: ["protoc-gen-swift"]),
+    .library(name: "SwiftProtobuf", type: .static, targets: ["SwiftProtobuf"]),
+  ],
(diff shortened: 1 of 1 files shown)


thomasvl:
I should add, without this, it seemed like `swift build|test` was building with 3.x instead of 4.

thomasvl:
I believe we need to have a SwiftPM4 manifest available (which likely means cutting a point release) for folks to actually get the library built with Swift 4 if they are using SwiftPM themselves.  From my quick look at this, it appears the SwiftPM will use the library in Swift 3 mode if there isn't a manifest saying it supports 4.

thomasvl:
Ah, last paragraph on [swiftlanguageversions](https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescriptionV4.md#swiftlanguageversions)

> If a package does not specify any Swift language versions, the language version to be used will match the major version of the package's Swift tools version. For e.g.: A Swift tools version with a major version of '3' will imply a default Swift language version of '3', and a Swift tools version with a major version of '4' will imply a default Swift language version of '4'.

So by not having a 4 manifest, we're saying we only support Swift 3.

thomasvl:
@tbkka any reason to go with the conditional (what you did in 6314d80) vs. using the proper support for a separate file?  Should we update that conditional to be closer to what this has?

tbkka:
Huzzah!  I'd been looking for the separate file approach, but couldn't find it documented anywhere.

This is *much* better.  Let's merge this and I'll back out my change.

PR #681, opened 2017-10-17
Add a template for new issues.

thomasvl:
Per: https://help.github.com/articles/creating-an-issue-template-for-your-repository/

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,19 @@
+Please be sure to include:
+
+* what OS you are developing on (Linux or macOS, including the version)
+
+* for macOS, what version of Xcode you are using (`xcodebuild -version`),
+  for Linux, what version of Swift (`swift --version`)
+
+* what version of Swift is your code set to compile with (i.e. from project
+  settings, etc.)
+
+* what branch/tag of SwiftProtobuf you are using (1.0.0, etc.)
+
+* if you are getting compile errors, please be sure include all errors/warnings,
+  sometimes the error before the one you are stuck on is important.
+
+* lastly, if it all possible, provide a snippet of `.proto` and/or source
+  that shows the problem.
+
+Thank you!


.github/issue_template.md
```diff
@@ -0,0 +1,7 @@
+Please be sure to include:
+
+* what version of Xcode you are using: `xcodebuild -version`
```
allevato: We should generalize this for Linux as well.

Maybe ask "What operating system you are using" and then "What version of Xcode (if macOS)"?

thomasvl: Ok, i'm confused, I amended the commit, but the PR won't pick it up…

thomasvl: And github finally realizes it…

.github/issue_template.md
```diff
@@ -0,0 +1,9 @@
+Please be sure to include:
+
+* what operating system you are developing on (linux or macOS)
+
+* for macOS, what version of Xcode you are using (`xcodebuild -version`)
+
+* what branch/tag of SwiftProtobuf you are using (1.0.0, etc.)
+
+* what version of Swift is your code set to compile with
```
tbkka: Maybe just "the version of Swift you are using" with a callout to `swift --version`

In any case, I would put this before the SwiftProtobuf version.  There's a natural order here:  either OS -> dev tools -> project   or   project -> dev tools -> OS.

.github/issue_template.md
```diff
@@ -0,0 +1,9 @@
+Please be sure to include:
+
+* what operating system you are developing on (linux or macOS)
```
tbkka: "your operating system (Linux or macOS, including the version)"

.github/issue_template.md
tbkka: I'm not sure if it's worth specifically asking people to try to:
 * copy/paste the specific compiler error or warning they're concerned about
 * Include a short piece of code that demonstrates a problem they're seeing

tbkka:
It's certainly better than what we have.  Go for it!

.github/issue_template.md
thomasvl: done.

.github/issue_template.md
thomasvl: The catch is the version of swift they use does not have to the same version that the library is compiled with (or that `swift --version`) will report.  Xcode settings, Package.swift info, etc all seems to come into play.

About the only thing I thought I was sure about was `swift --version` would only tell us if they had a custom version installed, but even if they do, that doesn't mean Xcode ends up using it.

Anyway, generally tweaked.

thomasvl: Added some more.

PR #683, opened 2017-10-24
swift build does not support --clean

dcow:
I notice this error when building/cleaning `swift-protobuf` with `make`:

```
swift build --clean
error: unknown option --clean; use --help to list available options
make: [clean] Error 1 (ignored)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -279,7 +279,6 @@ install: build
 	${INSTALL} ${PROTOC_GEN_SWIFT} ${BINDIR}
 
 clean:
-	-swift build --clean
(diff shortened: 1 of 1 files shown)


tbkka:
"swift build --clean" is needed by older versions of SwiftPM, "swift package clean" is used by newer versions.  That's why the Makefile uses both (and why both are prefixed with "-" which tells make to ignore any error).

I don't know if we're yet ready to remove the older form; we've committed to supporting a full major version back, so would like to keep support for 3.x.y until 4.x.y is released.

Issue #326, opened 2017-02-25
Enum.init(protoName:)


NachoSoto:
I was relying on this constructor, and it's now gone (see https://github.com/apple/swift-protobuf/pull/320/files#r103062915). Was there a reason for that? Would it be possible to bring it back?

Thanks!

allevato:
Can you describe in more detail why you to be able to translate strings to enum cases outside of the already provided text-format/JSON encoding/decoding processes?

NachoSoto:
Yup. In my particular example, I have this `enum`:
```
enum NotificationCategory {
    PUSH_NOTIFICATION_CATEGORY_1 = 0;
    PUSH_NOTIFICATION_CATEGORY_2 = 1;
   // ... etc
}
```

Then, from my backend, I'm sending notifications to Apple's APNS with that category (`"PUSH_NOTIFICATION_CATEGORY_1"`, etc). In the client I can then easily get the category from the push notification body, and deserialize it that way. 

I could of course send the `rawValue`s instead, and change my app `.plist` to numbers, but that would obscure the meaning of these categories, so it would be a lot better if I could continue sending these raw `String`s.

allevato:
My point of view is going to come as that of an API purist. 😄

* If we consider the main purpose of protocol buffers (fast small binary encoding with type-safe codegen in your favorite language), then the enum case name that you have there is intended for two main purposes: the basis of the generated symbol used in Swift source code, and the string used to represent the value in text-format- and JSON-encoded representations of a *message containing it*. Relying on the enum value name as a string in isolation, if you can't transfer a full message payload in text format or JSON that can parsed through the public APIs, is kind of skirting the edge of the intended use cases.

* On the goal of protobuf is to produce idiomatic code for its languages. Swift enums traditionally don't provide a string/value mapping of their own and don't require it to be useful. The relationship between the proto name and the Swift name isn't terribly relevant to pure Swift code; it's external protobuf metadata.

* One could also argue that sending the numbers is better bandwidth-wise than the names, but I can sympathize with the desire to have more readable names in files like plists that aren't generated from .proto files.

So, in our efforts to slim down the public API to what makes sense for Swift users, we reduced the enum API to only the cases and their raw values. This also serves another purpose—as a longer term goal we've been hoping to refactor the way generated fields/enum values expose their names so that we could provide an option to remove those strings *completely* from the code for users who only needed binary encoding/decoding (#18). There are big wins there for the mobile use cases with respect to binary size.

[...]

PR #684, opened 2017-10-25
Add public convenience property representing the name of the enum value

geor-kasapidi:


allevato:
Thanks for using swift-protobuf!

We've made a conscious decision not to provide public APIs for converting enums to/from their string names—see #326 for much more discussion about this. In short, an API such as this one is not quite the correct venue for it to be expressed; it's essentially a special-case of a descriptor (which we don't support yet), and typically one should think about serializing entire messages into text format if that's what's needed, because the protobuf spec doesn't really make any statements (to my knowledge) about serializing an enum to text in isolation.

geor-kasapidi:
@allevato thanks for your reply! Access to string representation of enum case is useful for API interaction. For example, in my case i have endpoints with protobuf responses and some plain text request parameters (query string, JSON data in body). And i need to pass enum case name as parameter value. As i see, i'm not first person who wants this functionality :) If code from my PR does not coincide with your vision of SwiftProtobuf future, maybe it makes sense to provide some meta information of Message, Enum, etc. For now, i see two options - fork this repo (bad idea), and second - use some reflection:

EricRabil:
Being able to access this would've allowed me to easily log the name of the payload being processed, as I am using `oneof` to enumerate all possible IPC payloads. I think this is a more than acceptable use case, and it's upsetting that the library is still gatekeeping it.

IMO, this data has been sitting in the library for 5 years and counting. It clearly works well enough, even if it is "not quite correct" or a "special-case". If the alternative is reflection, which is not performant, why should this remain an internal API? More importantly, why is the library the only one with easy access to this information?

thomasvl:
@EricRabil are you talking the name of the fields in a `oneof`?  If so that's different (it is field names in gneral), this issue was getting the `case` names from `enum`s.

EricRabil:
Yes, good eye. I should've posted this in #731

thomasvl:
fyi - That's still enum cases.  I don't think we've gotten a general request for field names/oneofs.

EricRabil:
I can make one :D

PR #31, opened 2016-10-03
Fix podspec issue

KyoheiG3:
Excluded of `protoc-gen-swift` and `PluginLibrary`. Resolved https://github.com/apple/swift-protobuf/issues/30 issue.
Thanks.

PR #689, opened 2017-10-30
add back the Sources/PluginLibrary/ files that were removed in #31

cameronehrlich:


thomasvl:
Not OK, we do _not_ want to expose the entire PluginLibrary internals to developers.

tbkka:
`PluginLibrary` has a lot more than just `descriptor.proto`, and those other things should definitely _not_ be included in a general SwiftProtobuf distribution, since they have a very specific and narrow audience.  The purpose of `PluginLibrary` is to expose some tools used by our code generator to people who want to build other code generators.  (Long-term, I would like to generalize more of our code-generation logic and migrate it into `PluginLibrary`.)  If you're building a code generator, you're welcome to explicitly depend on our `PluginLibrary` for that purpose.

If you think `SwiftProtobuf` should include a compiled version of `descriptor.proto`, you should propose just that.

thomasvl:
I'd still rather not do this as most people don't need it, and I don't want to make the CocoaPod expose something that SwiftPM doesn't expose.  We should be consistent about what is valid api exposed in all "package managers".

PR #685, opened 2017-10-25
Sources: remove now deprecated 'characters'

zllak:
With latest Swift version and Xcode (9.1 beta), using `characters` now generates an error
```
Apple Swift version 4.0.1 (swiftlang-900.0.68 clang-900.0.38)
Target: x86_64-apple-macosx10.9
```

PR #691, opened 2017-11-01
Add internal `chars` wrapper

moogle19:
Added a `chars` variable as internal String extension to wrap access to
String characters.
This should fix building for Swift 4.0.2 in XCode 9.1.


moogle19:
@tbkka 
Saw your comment on #685 . Could you please test this PR with Swift 3.1. I currently have no access to a system with Swift 3.1 installed.

tbkka:
#692 is a more ambitious fix.  I've tested it with a variety of Swift 3 compilers, but have not yet tested it with the newest Swift 4 versions.  I'm updating now and hope to finish that today.

One advantage of #692:  Once we stop supporting Swift 3, we can just rip out the conditionals here to leave the Swift 4 version.

PR #692, opened 2017-11-01
Support Swift 4 `characters` deprecation

tbkka:
This is based on @zllak's patch, with added `#if swift` checks to handle older compilers.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,3 +1,4 @@
+// swift-tools-version:3.1
 // Package.swift - description
 //
 // Copyright (c) 2014 - 2016 Apple Inc. and the project authors
@@ -9,6 +10,20 @@
 
 import PackageDescription
 
+#if swift(>=3.1)
+let package = Package(
+  name: "SwiftProtobuf",
+  targets: [
+    Target(name: "PluginLibrary",
+           dependencies: ["SwiftProtobuf"]),
+    Target(name: "protoc-gen-swift",
+           dependencies: ["PluginLibrary", "SwiftProtobuf"]),
+    Target(name: "Conformance",
+           dependencies: ["SwiftProtobuf"]),
+  ],
+  swiftLanguageVersions: [3, 4]
+)
+#else
 let package = Package(
   name: "SwiftProtobuf",
   targets: [
@@ -20,3 +35,4 @@ let package = Package(
            dependencies: ["SwiftProtobuf"]),
   ]
 )
+#endif
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,3 @@
-// swift-tools-version:3.1
 // Package.swift - description
 //
 // Copyright (c) 2014 - 2016 Apple Inc. and the project authors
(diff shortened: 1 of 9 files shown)


Sources/protoc-gen-swift/StringUtils.swift
```diff
@@ -27,7 +31,12 @@ func splitPath(pathname: String) -> (dir:String, base:String, suffix:String) {
       suffix += String(c)
     }
   }
-  if suffix.characters.first != "." {
+#if swift(>=4.0)
+  let validSuffix = suffix.isEmpty || suffix.first == "."
```
thomasvl: Is this the same behavior as before since it didn't check for empty before?

moogle19: If `suffix` is empty it just appended an empty string to `base` before, so it would not make a difference.

moogle19:
@tbkka I think the swift version check must be `>=3.2`

rick-kln:
Something related to the conditional compilation blocks doesn't seem to be working here. 
Carthage build was still failing for us with:
`github "tbkka/swift-protobuf" "zllak-swift401"`

We double checked that Carthage was checking out commit `0abbefcbea7d78aca6f0671f4718fc2f7f03d030` and it is.

Next we cloned and opened the swift-protobuf project at that commit and it doesn't seem to build for us with Xcode 9.1.

captura de pantalla 2017-11-02 a las 16 18 20

thomasvl:
Doesn't Carthage us the checked in project file?  If so, that isn't set to use Swift 4.  So it sounds like Xcode 9.1 in Swift 3.x mode is kicking out the warning when Swift 9.0.x didn't kick out the message in 3.x mode.

rick-kln:
Yeah Carthage does use the checked in project file. Just double checked by running manually to compare the Carthage logs to the errors in Xcode and they're both the same so the problem isn't with our Carthage configuration.

Yeah that sounds correct.

We're trying to build via Carthage using the 4.0.2 toolchain. Obviously when we open the project manually with Xcode the project is still in 3.x mode.

tbkka:
I just added a change to Package.swift that allows it to use the latest language version in all cases.  So attempts to compile with a Swift 4 compiler should now use Swift 4 mode.

tbkka:
I'll update the checks to `>=3.2` to cover the case where the Swift 4 compiler is running in Swift 3 mode.

Package.swift
```diff
@@ -9,6 +10,20 @@
...
+    Target(name: "PluginLibrary",
+           dependencies: ["SwiftProtobuf"]),
+    Target(name: "protoc-gen-swift",
+           dependencies: ["PluginLibrary", "SwiftProtobuf"]),
+    Target(name: "Conformance",
+           dependencies: ["SwiftProtobuf"]),
+  ],
+  swiftLanguageVersions: [3, 4]
+)
+#else
```
thomasvl: Dropping this, right?

tbkka: Done.

Sources/PluginLibrary/NamingUtils.swift
```diff
@@ -189,10 +189,11 @@ fileprivate func sanitizeTypeName(_ s: String, disambiguator: String) -> String
     // conflict.  This can be resolved recursively by stripping
     // the disambiguator, sanitizing the root, then re-adding the
     // disambiguator:
-    let e = s.index(s.endIndex, offsetBy: -disambiguator.characters.count)
-    #if swift(>=4.0)
+    #if swift(>=3.2)
```
thomasvl: fyi - style wise, this one is the only one you are adding that is indented.  I haven't checked what we did in the other case.

tbkka: Looks like there are other places where we currently indent these conditionals.  So I'll leave it as-is for now.

PR #697, opened 2017-11-03
Update docs so folks copy/pasting with package managers will get updates.

thomasvl:
Reading the docs for both ~> 1.0.0 means any 1.0.x but not up to 1.1,
which seems like the safe thing to spell out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ If you're using CocoaPods, add this to your `Podfile` adjusting the `:tag` to
 match the `[tag_name]` you used to build the plugin above:
 
 ```ruby
-pod 'SwiftProtobuf', '~> 1.0.1'
+pod 'SwiftProtobuf', '~> 1.0'
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -191,7 +191,7 @@ If you're using CocoaPods, add this to your `Podfile` adjusting the `:tag` to
 match the `[tag_name]` you used to build the plugin above:
 
 ```ruby
-pod 'SwiftProtobuf', '~> 1.0.1'
+pod 'SwiftProtobuf', '~> 1.0.0'
```
thomasvl: For both of these we can do ~> 1.0 if we're sure 1.1, 1.2, etc. are always gonna be api compatible.

tbkka: I certainly plan to enforce API compatibility within a major version.

thomasvl: Updated to just be `~> 1.0` then.

PR #702, opened 2017-11-14
Add `SwiftProtobufPluginLibrary` product.

kevints:
This restores the `PluginLibrary` module as API for users of
Swift Package Manager versions 4 or greater.


Package@swift-4.swift
```diff
@@ -16,6 +16,7 @@ let package = Package(
   products: [
     .executable(name: "protoc-gen-swift", targets: ["protoc-gen-swift"]),
     .library(name: "SwiftProtobuf", type: .static, targets: ["SwiftProtobuf"]),
+    .library(name: "SwiftProtobufPluginLibrary", targets: ["PluginLibrary"]),
```
thomasvl: Does it still need "SwiftProtobuf" on the front of the name since it is scoped to this package?

kevints: In SwiftPM products share a global namespace (your package declares a dependency on one or more packages, and then your package's targets declare dependencies on targets and products), so the prefix is indeed required. Come to think of it Swift modules live in a global namespace as well, so it might be advisable to rename PluginLibrary to SwiftProtobufPluginLibrary. I'll update the PR with that change.

tbkka: Kevin:  If you could just rename the directory, that seems like the best answer.  (It will require fixing a few other references to get everything to compile cleanly again, of course.)

Thanks for helping with this kind of housekeeping!

kevints: No problem, updated the PR.

gmilos:
@thomasvl it's to do with protobuf and rpc, hmmm.

thomasvl:
@gmilos thanks.  😉 

A lot of what went into the plugin lib, so feel free to reach out with PRs or directly in email about additions/etc.  Having worked on two proto plugins now (ObjC and Swift), I'm hoping I exposed some reasonable helpers, but one never really knows until there are multiple consumers.

kevints:
Thanks for the reviews @thomasvl and @tbkka! Any idea when this patch might be released?

thomasvl:
I've got to do some releases for a few other opensource projects I work on, so I'll try to plow through this along with those today.

thomasvl:
Release 1.0.2 cut (and cocoapod spec pushed)

Issue #708, opened 2017-12-18
Default Configuration for Commandline Builds


rlasante:
I'm working on integrating 1.0.2 with our iOS app and I am running into issues with our jenkins builds of SwiftProtobuf.framework. We have a number of different build configurations that are named differently from the standard 2: `Debug`/`Release`. When Xcode attempts to determine which configuration it should build of dependent frameworks it looks at the app's current configuration and tries to find a matching named configuration in the dependent framework. When it fails to find the match it will then default to `Release`.

Unfortunately there doesn't appear to be any way to specify which configurations should map to which. So if we have an app being built with a configuration: `Dev` then Xcode will look in SwiftProtobuf and determine that neither `Debug` or `Release` match `Dev` so therefore it should build `Release`.

However in the SwiftProtobuf Xcode project under Configurations there's a place to specify what version to use for command-line builds that is set to `Debug`. This means that when xcodebuild can't find `Dev` it then will go and build SwiftProtobuf in `Debug`.

<img width="585" alt="screen shot 2017-12-18 at 1 44 56 pm" src="https://user-images.githubusercontent.com/2073598/34123027-3eab551a-e3fc-11e7-8fd1-1341d4f4670d.png">

This means that we're unable to build a release version of SwiftProtobuf from the command-line without either building our app's `Release` configuration or running a separate xcodebuild command specifically for building SwiftProtobuf.

This link has some info about the issue.
https://www.raizlabs.com/dev/2014/03/specifying-the-build-configuration-for-xcode-project-dependencies/

Is there a reason why SwiftProtobuf defaults to `Debug` on command-line? It seems to me that the better experience at least short term is to default to `Release`

Developing on:
macOS 10.12.6
Xcode 9.0.1 Build version 9A1004
Swift 3.2
SwiftProtobuf 1.0.2

thomasvl:
Switching to Release for the default is likely fine.

Slight aside, using non standard names likely means you'll have these sorta issues with just about anything you try to reuse a project for.

thomasvl:
Fixed via #709

zaidas1m:
Hello!
I don't think this is solved.
I have several configurations and schemes set with different bundle id 
![Captura de pantalla 2021-06-01 a las 12 09 06](https://user-images.githubusercontent.com/65656518/120307395-39e30780-c2d3-11eb-8ab3-3b87f42fac79.png)

[...]

PR #709, opened 2017-12-18
Change Default Commandline Configuration Name to Release

rlasante:
Fix for issue #708 so that the commandline default configuration is Release rather than Debug


thomasvl:
Thanks!

Issue #161, opened 2016-12-06
Install via homebrew


jacobjennings:
A homebrew formula would be a nice addition, especially if it plays nicely with protobuf-swift formula i.e.

brew unlink protobuf-swift
brew link swift-protobuf
protoc ... --swift_out ... ...

azr:
👍

anderscarling:
Just submitted a pull request: Homebrew/homebrew-core#17232

anderscarling:
@tbkka - I just tagged you in the comments in the aforementioned pull req to get this into homebrew. Mentioning you here too, as I have no idea if you will get a notification about a mention in a repo/discussion you are not participating in.

Would appreciate it if you could check it out and maybe help me explain/understand the current release status of swift-protobuf, as I'm trying to championing getting it into the homebrew repo for easy install and usage.

tbkka:
I just tagged SwiftProtobuf 1.0.  That is now an official stable release version.

anderscarling:
That's awesome! Congratulations on getting the release out! :)

tbkka:
Homebrew/homebrew-core#17232  has been merged.

PR #711, opened 2018-01-02
Add Homebrew installation method to README

srvv:
It would be nice to add this option to documentation as it is easier in most cases. #161

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ If you prefer using [Homebrew](https://brew.sh):
 $ brew install swift-protobuf
 ```
 
-This will install `protoc` compliler and Swift code generator plugin.
+This will install `protoc` compiler and Swift code generator plugin.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -135,6 +135,14 @@ directory.
 To install, just copy this one executable into a directory that is
 part of your `PATH` environment variable.
 
+### Alternatively install via Homebrew
+
+If you prefer using [Homebrew](https://brew.sh):
+
+```
+$ brew install swift-protobuf
+```
```
tbkka: I presume this only installs the command-line tool?

We should be more clear about that.

srvv: This is part of 'Building and Installing the Code Generator Plugin' section. Also described what this command does.

srvv:
Fixed.

tbkka:
Excellent!  Thanks!  😎

PR #715, opened 2018-01-20
A couple of fixes for Swift 4.1

NachoSoto:
This breaks compatibility with `Swift 4.0`, I can work on that next week, I just wanted to have an initial version compiling with `4.1` I can start working with.

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
@@ -137,7 +137,7 @@ public extension Google_Protobuf_FieldMask {
   ///   defined using the JSON names for the fields.
   public init?(jsonPaths: String...) {
     // TODO: This should fail if any of the conversions from JSON fails
-    self.init(protoPaths: jsonPaths.flatMap(JSONToProto))
+    self.init(protoPaths: jsonPaths.compactMap(JSONToProto))
   }
 
   // It would be nice if to have an initializer that accepted Swift property
--- a/Sources/SwiftProtobuf/type.pb.swift
+++ b/Sources/SwiftProtobuf/type.pb.swift
@@ -605,8 +605,7 @@ public struct Google_Protobuf_Option: SwiftProtobuf.Message {
     get {return _storage._value ?? Google_Protobuf_Any()}
     set {_uniqueStorage()._value = newValue}
   }
-  /// Returns true if `value` has been explicitly set.
-  public var hasValue: Bool {return _storage._value != nil}
+  
   /// Clears the value of `value`. Subsequent reads from it will return its default value.
   public mutating func clearValue() {_storage._value = nil}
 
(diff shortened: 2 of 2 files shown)


Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -137,7 +137,7 @@ public extension Google_Protobuf_FieldMask {
   ///   defined using the JSON names for the fields.
   public init?(jsonPaths: String...) {
     // TODO: This should fail if any of the conversions from JSON fails
-    self.init(protoPaths: jsonPaths.flatMap(JSONToProto))
+    self.init(protoPaths: jsonPaths.compactMap(JSONToProto))
```
thomasvl: Does this work on older Swift versions or does this need to be conditional?

Sources/SwiftProtobuf/type.pb.swift
```diff
@@ -650,6 +649,11 @@ public struct Google_Protobuf_Option: SwiftProtobuf.Message {
   fileprivate var _storage = _StorageClass.defaultInstance
 }
 
+extension Google_Protobuf_Option {
```
thomasvl: What is this change for?  This is a generated file, it shouldn't get manual edits.

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
NachoSoto: It does not, see PR summary.

Sources/SwiftProtobuf/type.pb.swift
NachoSoto: To workaround a new warning, see Twitter conversation: https://twitter.com/nachosoto/status/954517885375795200

I hope it gets removed...

thomasvl: Lets open an issue for this new warnings, as this one case it just the tip of the iceberg, if the warning sticks, we are going to have to change the generator because this will no doubt happen on other things also.

thomasvl:
The `flatMap` -> `compactMap` change is in #716

tbkka:
I had an idea earlier today that might possibly address the naming issue here.  Consider this simple message:
```
message Foo {
   optional int32 value = 1;
   optional string name = 23;
}
```

We could use the field numbers to name the stored properties in the struct:
```
struct Foo {
   fileprivate var _1: Int32?
   fileprivate var _23: String?
}
```

And then use the field names (and derived forms) only in an extension:
```
extension Foo {
   var value: Int { return _1 ?? 0 }
   var hasValue: Bool { return _1 != nil }
   var name: String { return _23 ?? “” }
   var hasName: Bool { return _23 != nil }
}
```

allevato:
That's a clever idea! But I think it would produce a pretty bad debugging experience, since lldb/Xcode only show you the stored properties when you print/expand a message—are we willing to sacrifice that?

thomasvl:
We need to expand things into some test cases, but I think the use of lack of use of _Storage_ may come into play.  So far it only seems only things using _Storage_ are what are failing, but we'll need to make some messages that don't use storage to confirm what happens.

In the cases where the message uses _Storage_, since those `class`es don't actually conform to protocols, I don't think they will ever trigger the warning, so the raw field names should be fine.

The cases without _Storage_ will take some quick investigation to see how it chokes, but that might be where renaming the fields is required.

I'll try to do some testing later today.

thomasvl:
Moving discussion into #719 rather than on the PR, just so we have an issue not attached to a PR to reference in the code/etc.

PR #720, opened 2018-01-30
Add a simple warning about the prefix option.

thomasvl:


Documentation/API.md
```diff
@@ -130,6 +130,10 @@ in your proto file:
 will generate a struct named `MyFooBar`.
 (Note: `swift_prefix` is only supported by protoc 3.2 or later.)
 
+:warning: Use of the `swift_prefix` is generally discourage as it means the
```
allevato: discouraged

thomasvl: Fixed

Documentation/API.md
```diff
@@ -130,6 +130,10 @@ in your proto file:
...
+message and enum names are even less like the ones generated for other
+languages (most leverage the proto package in some way).
```
tbkka: I don't really understand this last comment.  What "proto package" are you referring to?

Documentation/API.md
```diff
@@ -130,6 +130,10 @@ in your proto file:
 will generate a struct named `MyFooBar`.
 (Note: `swift_prefix` is only supported by protoc 3.2 or later.)
 
+:warning: Use of the `swift_prefix` is generally discouraged as it means the
+message and enum names are even less like the ones generated for other
+languages (most leverage the proto package in some way).
```
tbkka: I don't really understand this last comment.  What "proto package" are you referring to?

Documentation/API.md
thomasvl: The `package` directive is the proto package - https://developers.google.com/protocol-buffers/docs/proto#packages

Documentation/API.md
tbkka: How about this:
```
The `swift_prefix` option has proven problematic in practice.  Because it ignores the `package` directive, it can easily lead to name conflicts and other confusion as your shared proto definitions evolve over time.
```

thomasvl: Works for me.  Just was starting to see some over use of it, so I want to help ensure we warn folks it is more exception as opposed to commonly needed.

@allevato sound good?

allevato: 👍

thomasvl: Updated.

thomasvl: I keep the warning icon.  Look good?

PR #728, opened 2018-03-12
Use automatic linkage for the library product

kevints:
The Swift Package Manager documentation for [products](https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescriptionV4.md#products) recommends that library products use automatic linkage (the default) "so the Package Manager can decide appropriate linkage." Indeed, declaring static linkage slows down builds by forcing the creation of an intermediate static library `libSwiftProtobuf.a`. In automatic linkage SwiftPM elides the creation of this library and links SwiftProtobuf object files directly into the final executable. This has the same effect as using a static library with the advantage of making the overall build faster.

### DIFF
--- a/Package@swift-4.swift
+++ b/Package@swift-4.swift
@@ -15,7 +15,7 @@ let package = Package(
   name: "SwiftProtobuf",
   products: [
     .executable(name: "protoc-gen-swift", targets: ["protoc-gen-swift"]),
-    .library(name: "SwiftProtobuf", type: .static, targets: ["SwiftProtobuf"]),
+    .library(name: "SwiftProtobuf", targets: ["SwiftProtobuf"]),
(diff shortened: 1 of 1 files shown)


thomasvl:
Given we didn't explicitly say static on the plugin lib either, it must have been getting it right, so seems good.  Thanks!

tbkka:
I thought the static lib was being used by the perf harness.  Does that still DTRT without this?

thomasvl:
> I thought the static lib was being used by the perf harness. Does that still DTRT without this?

Hm…

https://github.com/apple/swift-protobuf/blob/master/Performance/runners/swift.sh#L54

The _TODO_ a little above it is also sorta interesting, maybe with this landed it is "done"?

PR #732, opened 2018-03-26
Add swift_version to podspec

davbeck:
This fixes a warning in `pod lib lint`. I’m having trouble finding hard documentation on this, but cocoapods should set the `SWIFT_VERSION` on the framework target. From there, Xcode will handle compatability with different versions of Swift (https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Compatibility.html).

From Cocoapods:
```bash
$ pod lib lint

 -> SwiftProtobuf (1.0.3)
    - WARN  | swift: The validator used Swift 3.2 by default because no Swift version was specified. To specify a Swift version during validation, add the `swift_version` attribute in your podspec. Note that usage of the `--swift-version` parameter or a `.swift-version` file is now deprecated.

[!] SwiftProtobuf did not pass validation, due to 1 warning (but you can use `--allow-warnings` to ignore it).
You can use the `--no-clean` option to inspect any issue.
```

### DIFF
--- a/SwiftProtobuf.podspec
+++ b/SwiftProtobuf.podspec
@@ -7,6 +7,7 @@ Pod::Spec.new do |s|
   s.author = 'Apple Inc.'
   s.source = { :git => 'https://github.com/apple/swift-protobuf.git', :tag => s.version }
 
+  s.swift_version = '3.2'
   s.requires_arc = true
   s.ios.deployment_target = '8.0'
   s.osx.deployment_target = '10.9'


thomasvl:
We talked a little about this in https://github.com/apple/swift-protobuf/blob/master/Documentation/RELEASING.md.

The problem is we actually work with any Swift version (prefer the highest version Xcode supports).  The cocoapods folks seem to be working on improving this as there has been notes about the swift version files/attributes in the last few releases, but the don't seem to have a way to model "can use anything".  And with some of thing things coming in future swift versions (like SE0194), forcing a version would be this can't be support for the developers using a high enough Swift version.

tbkka:
Our policy is to support any Swift version from the latest GM (currently 4.1) to 1 full major version back (3.1).  That of course changes over time.

Putting 3.2 here seems overly conservative.  If the pod spec absolutely must have a Swift version, we should put a relatively recent version here.

davbeck:
If you set it to (for instance) 3.1, projects that compile using that version of Xcode/Swift will work and projects using a newer version of Swift will compile the framework in compatibility mode. If you set it to (for instance) 4.0, it won't compile with versions of Swift less than that.

There is no support in Swift (not cocoapods, the actual compiler) for a framework to support multiple versions of Swift. Source code can use `if #swift` (which it looks like this project uses) but the actual framework has to declare itself.

thomasvl:
So if you look at SwiftPM, by having both versions of the manifest, we get the Swift versions that make sense based on the version a developer is using.  Then is also has a [tag to support a list of versions](https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescriptionV4.md#swiftlanguageversions).

Yes, the compiler doesn't need support for listing a "set" of versions, but the packaging tools need this.  As library authors we make sure our code can work with what a developer wants to use, but cocoapods currently seems to fall short of not allowing us to express this, and picking anything locks folks out (either of older versions or of newer language features).  There doesn't seem to be a good way around this at the moment. 🙁

davbeck:
Which is fine in the broader sense, but when you're talking about Cocoapods specifically, you have to take a stand. Not specifying the version just means that it defaults to 3.2, not that you are gaining any actual extra compatibility with swift 4. Specifying the version will only effect Cocoapod users.

thomasvl:
For reference - [cocoapods support for indicating a range of swift versions](https://github.com/CocoaPods/CocoaPods/issues/7134).

thomasvl:
Fyi - I pinged the CocoaPods issue about supporting version ranges highlighting how the lack of support causes problems for libraries looking to support conditional conformances, new protocols/types (CaseIterable).  If we hard code something in the podspec, it forces folks to adopt a minimum toolchain or it forces the CocoaPod users to not be able to get those features.  Both of which seems sub optimal.

tengyifei:
@thomasvl just FYI CocoaPods is finally going to support specifying multiple versions in 1.7.0: https://github.com/CocoaPods/CocoaPods/pull/8192

https://github.com/CocoaPods/CocoaPods/pull/8192/files#diff-5dab52bd4101d74b36cd3413781a462cR15

tbkka:
It's been a few months; any progress on this?

dnkoutso:
Hi yes I can confirm release 1.7.x supports specifying multiple versions of Swift in the podspec.

jafara:
is this going to happen?

thomasvl:
> is this going to happen?

Anyone can send a PR in to add it, otherwise, we'll hopefully try to look at it next time a release is done.

dnkoutso:
Does compilation work with Swift 4.2 or 5.0? I just tried the normal `SwiftProtobuf.xcodeproj` and I got a bunch of errors.

thomasvl:
> Does compilation work with Swift 4.2 or 5.0? I just tried the normal SwiftProtobuf.xcodeproj and I got a bunch of errors.

We support back to 4.0 (the policy for this project is one major version back, so Swift 5.0 is the current GM, so we support back to 4.0).

As for the break, try syncing again.  https://github.com/apple/swift-protobuf/pull/882 landed about an hour ago to fix something we missed in a passed PR.  The checked in Xcode project doesn't get used much…

dnkoutso:
@thomasvl thanks, my fork was outdated and I can compile in `SwiftProtobuf.xcodeproj` will create a PR with the updated podspec.

PR #738, opened 2018-04-21
Allow 0 to be used for float values in the text protobuf format.

LaurentMazare:
I don't think that there is a formal spec for the protobuf text format. However the main C++ implementation supports using 0 for a floating point value.
Currently the swift implementation does not allow it and fail with a "malformedNumber" error. However any other integer could be used which is not very consistent.
This PR tweaks the current check so that 0 can be used and also adds some unit tests for this behavior.

(note that it's the first time that I look at this codebase as I ran into the issue, so I may have overlooked some good reason to forbid 0 or some easier way to achieve this)

### DIFF
--- a/Sources/SwiftProtobuf/TextFormatScanner.swift
+++ b/Sources/SwiftProtobuf/TextFormatScanner.swift
@@ -674,11 +674,11 @@ internal struct TextFormatScanner {
             c = p[0]
         }
         switch c {
-        case asciiZero: // '0' as first character only if followed by '.'
+        case asciiZero: // '0' as first character is not allowed followed by digit
             p += 1
-            guard p != end else {p = start; return nil}
+            guard p != end else {break}
             c = p[0]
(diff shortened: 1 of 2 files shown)


thomasvl:
Looks like you are making things accept any integer, not just zero as for float values?

Seems like if this is done we should also do this for double.

Does C++ happen to have a unittest to ensure this behavior works?

thomasvl:
Looks like some of their `test_format_unitest_*` files have it:

```
...
optional_float: 111
optional_double: 112
...
```

LaurentMazare:
Thanks for checking that this is indeed covered in the C++ tests and for the quick feedback.
I was probably not very clear but integer values were already accepted for float/double except for 0 (which is the only value that caused the malformedNumber error).

Double parsing seems to use the same function so there is nothing to change to handle these, I just added some dedicated unit tests.

Issue #743, opened 2018-05-15
Malformed string when decoding from JSON


vladiulianbogdan:
Hi,

I have to create SwiftProtobuf messages from JSON. The JSONs are not created from SwiftProtobuf messages (by using `jsonString()` or `jsonUTF8Data()`), instead they are created using JSONSerialization from Foundation.framework.

The problem appears when the JSON has a base64 value which contains a "/" character. JSONSerialization will escape that character with a "\\". When I try to populate a SwiftProtobuf message with the content from JSON, the JSONScanner will report a `JSONDecodingError.malformedString` error because of the unexpected "\\" character.

Example of JSON incorrectly marked as malformed:

```
{"data":"omdtZXNzYWdlZnBhY2tldGdwYXlsb2FkomVpbmRleBg\/ZGRhdGF4gDAwMDEwMjAzMDQwNTA2MDcwODA5MGEwYjBjMGQwZTBmMTAxMTEyMTMxNDE1MTYxNzE4MTkxYTFiMWMxZDFlMWYyMDIxMjIyMzI0MjUyNjI3MjgyOTJhMmIyYzJkMmUyZjMwMzEzMjMzMzQzNTM2MzczODM5M2EzYjNjM2QzZTNm", "id": 0}
```

I had to fix this in one of my forks in order to unblock my project. In parseBytes method from JSONScanner I am also handling the "\" character.

* what OS you are developing on
macOS 10.13.3

* for macOS, what version of Xcode you are using
Xcode 9.2

* what version of Swift is your code set to compile with
Swift 4.0

* what branch/tag of SwiftProtobuf you are using
1.0.3

tbkka:
Thanks for the report!  Yes, we should accept arbitrary backslash escapes here.  That was definitely an oversight.

tbkka:
Please look at #744 and let me know what you think.

tbkka:
I think this is fixed with PR #744.  Let us know if you still have problems.

aarashgoodarzi:
the problem still exists!!

tbkka:
Can you please open a new Issue with more details?  In particular, I think it might help if you could show us the invalid JSON you're seeing.

PR #744, opened 2018-05-15
Issue #743: Correctly handle escaped slashes in JSON byte data

tbkka:
Protobuf JSON is designed to be easily implemented on top of standard
JSON encoders.  For historical reasons, such encoders often encode "/"
(which commonly occurs in base-64 data) as "\/".

I overlooked support for backslash escapes when parsing byte data.

This adds tests and support for all the simple backslash escapes.
It turns out that "\/" is the only one that is legal in base-64 data,
so we simply fail if we encounter \b \f \n \r \t \" \\ or any invalid
sequence.

Note: This does not support \u1234 hexadecimal escapes.  (I don't
expect to ever see those in practice, so I doubt this is a problem.)

### DIFF
--- a/Sources/SwiftProtobuf/JSONScanner.swift
+++ b/Sources/SwiftProtobuf/JSONScanner.swift
@@ -124,7 +124,7 @@ private func parseBytes(
     source.formIndex(after: &index)
 
     // Count the base-64 digits
-    // Ignore unrecognized characters in this first pass,
+    // Ignore most unrecognized characters in this first pass,
     // stop at the closing double quote.
     let digitsStart = index
     var rawChars = 0
@@ -134,14 +134,32 @@ private func parseBytes(
         let digit = source[index]
         if digit == asciiDoubleQuote {
             break
-        } else if digit == asciiPlus || digit == asciiForwardSlash {
-            sawSection4Characters = true
-        } else if digit == asciiMinus || digit == asciiUnderscore {
-            sawSection5Characters = true
-        }
-        let k = base64Values[Int(digit)]
-        if k >= 0 {
-            rawChars += 1
+        } else if digit == asciiBackslash {
+            source.formIndex(after: &index)
+            if index == end {
+                throw JSONDecodingError.malformedString
+            }
+            let escaped = source[index]
+            switch escaped {
+            case asciiLowerU:
+                // TODO: Caller should pre-parse with full string decoder and then try again
+                throw JSONDecodingError.malformedString
+            case asciiForwardSlash:
+                rawChars += 1
+            default:
+                // Reject \b \f \n \r \t \" or \\ and all illegal escapes
+                throw JSONDecodingError.malformedString
+            }
+        } else {
+            if digit == asciiPlus || digit == asciiForwardSlash {
+                sawSection4Characters = true
+            } else if digit == asciiMinus || digit == asciiUnderscore {
+                sawSection5Characters = true
+            }
+            let k = base64Values[Int(digit)]
+            if k >= 0 {
+                rawChars += 1
+            }
         }
         source.formIndex(after: &index)
     }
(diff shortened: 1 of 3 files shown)


Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -172,26 +190,24 @@ private func parseBytes(
...
                 case asciiDoubleQuote:
                     source.formIndex(after: &index)
                     break digits
+                case asciiBackslash:
+                    source.formIndex(after: &index)
+                    let escaped = source[index]
+                    switch escaped {
+                    case asciiForwardSlash:
+                        k = base64Values[Int(escaped)]
+                    default:
```
thomasvl: This can't actually happen since the loop before should have blocked it?

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -134,14 +134,32 @@ private func parseBytes(
...
+            if index == end {
+                throw JSONDecodingError.malformedString
+            }
+            let escaped = source[index]
+            switch escaped {
+            case asciiLowerU:
+                // TODO: Caller should pre-parse with full string decoder and then try again
+                throw JSONDecodingError.malformedString
+            case asciiForwardSlash:
+                rawChars += 1
```
cburrows: Should you set `sawSection4Characters = true` here? Why not just reassign digit and fall through to the else?

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -172,26 +190,24 @@ private func parseBytes(
...
+                    let escaped = source[index]
+                    switch escaped {
+                    case asciiForwardSlash:
+                        k = base64Values[Int(escaped)]
+                    default:
+                        throw JSONDecodingError.malformedString
+                    }
                 case asciiSpace:
-                    break
+                    source.formIndex(after: &index)
```
cburrows: Why this change? It appears to make spaces permissible, and ignored, at most locations in the middle of the base64 data?

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -172,26 +190,24 @@ private func parseBytes(
...
-                    n = 0
-                }
-            } else {
+            var k = base64Values[Int(digit)]
+            if k < 0 {
                 switch digit {
                 case asciiDoubleQuote:
                     source.formIndex(after: &index)
                     break digits
+                case asciiBackslash:
```
cburrows: While we are here, I notice that the protobuf documentation permits the base64 string to be URL-safe, i.e. `+` and `=` and `/` can by `%2B` or whatever. Maybe open another issue?

Sources/SwiftProtobuf/JSONScanner.swift
thomasvl: It is because of the reflow of the loop, this now does a the `continue digits` where before it just did a `break` and so it fell to the code below that did this work.

Sources/SwiftProtobuf/JSONScanner.swift
thomasvl: They are both mixed in the table, the section 4/5 checks are to ensure we only ever get one set and not both (if I'm remembering correctly)

Sources/SwiftProtobuf/JSONScanner.swift
thomasvl: Oops, yup, seems like it should have been done.

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: "URL-safe base-64" is not the same as "URL encoding."  The URL-safe alphabet is the one that uses "-" and "_" instead of "+" and "/".  (This is formalized in RFC4648 section 5 as the "Base 64 Encoding with URL and Filename Safe Alphabet" aka "base64url".)

I've expanded the comments above the table to explain this further.

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: This is based on the Google C++ implementation, which is tolerant of spaces.  On the other hand, [RFC4648 Section 3.3](https://tools.ietf.org/html/rfc4648#section-3.3) urges people to reject base64 text with spaces.  I've chosen to ignore spaces to better match Google's implementation, but this would be easy to change if we had a reason.

Sources/SwiftProtobuf/JSONScanner.swift
cburrows: Ah, my mistake. Thanks.

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: I added a test to document this behavior.

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: Good catch!  I've fixed this and added a test for it.

PR #745, opened 2018-05-22
Revert "Swap the use of IndexPath for a custom Hashable array."

thomasvl:
This reverts commit 7f73940b3f576b5dad05e96665f42b37f5080095.

https://bugs.swift.org/browse/SR-4858 was fixed about a year ago, and the only
issues were on linux, so hopefully we're good dropping this now (atleast I can't
think of any reason for someone to use using a linux swift that is that old).


thomasvl:
@tbkka do you think we need to keep this around?  My thought was unlike macOS where folks are more tied to Xcode updates, for linux, folks really should have moved to newer versions faster, so we can probably trim this sooner that our normal version policy.

tbkka:
I'm okay with this.

In particular, I personally have a higher risk tolerance for the PluginLibrary code.

tbkka:
Follow-up:  I re-ran a test suite that checks SwiftProtobuf with many versions of swift on Ubuntu.  The tests still all pass with Swift as far back as 3.1.

The only issue:  Comments get lost when using a protoc-gen-swift built with Swift earlier than Swift 4.0.  I consider that an acceptable cosmetic issue.

thomasvl:
👍 thanks for confirming.

PR #755, opened 2018-06-14
Fix hashing issues for messages/groups in arrays.

thomasvl:
The original code here landed in f9972067d75361cd37e02723c5ec076c30da077a,
but because of SR-7993 and not having yet landed a Package.swift that
enabled 4.2 support, we weren't hitting this code yet.

Since the visitor api uses Message, it doesn't know they messages are
Hashable, the maps work because they get more complete type info to
pick up the conformance declared on the Base used by for generated
messages.

So...

Going back to manually mixing in the for Arrays, adding a Todo
to revisit this. But this gets master back to working if someone
tries to pull the code into a compiling context that is 4.2+.


thomasvl:
I plan on revisiting this with `hash(into:)` but this is just to fix what was landed in a broken state, but I didn't realized it because of the SwiftPM issue.

PR #764, opened 2018-07-11
Add a 4.2 manifest to enable 4.2 support.

thomasvl:
Note: SR-7993 is now fixed, but depending on what version of swift one uses, this may not fully work.  But this will be working for the GM.


thomasvl:
fyi - I did start a [forums thread](https://forums.swift.org/t/supporting-multiple-toolchain-versions/13599), at the moment, the only way to support multiple toolchains does seem to be to have lots of files like this.  ☹️

PR #768, opened 2018-07-13
Hashable prep

thomasvl:


thomasvl:
@tbkka @allevato I need your thoughts on this one.

I feel like this is a good thing before fully taking on SE0206 because it gets us completely using the conditional conformances added to collections.  But…

It comes at an API cost.  I have changed the signature of some of the `Visitor` methods.  I see two ways out of this:
1. Just accept it, and maybe bump the library to 1.1.x because most folks should be the signatures that are public that we did change.
1. Add methods with the old and new signatures and a default impl of the new back to the old to now break any code.

Option 1 seems better because it avoids the shims, and I think it remains safe because nothing changes for the generated code, so it doesn't require folks to regenerate, etc.

So, specific thoughts on:
- Signature changes ok?  We really only have to make the signature change on the repeated methods, the single ones don't have to have it, but I figured for completeness it might be better to keep them in sync.
- How to handing the api changes?

allevato:
This kind of makes me want to just revisit our whole protocol hierarchy.

We've gone out of our way to provide a `Message` protocol without any associated types or `Self` constraints so that people can create heterogeneous collections of messages. Is this desired so frequently that we're making our lives harder to support it?

What if we just bite the bullet and do what Swift does in the standard library when it comes to wanting these kinds of collections? Let `Message` declare its `Equatable` and `Hashable` requirements and create a type-erasing `AnyMessage` box if someone _really_ wants to handle them in a polymorphic way. I haven't _completely_ thought this through yet so there could be rough edges, but `AnyMessage` could probably still conform to `Message` and forward `traverse` to the boxed message (just as `AnyHashable` conforms to `Hashable` and forwards its operations), so users could still serialize such collections without extra overhead at the call site. They would just have to unbox the underlying message explicitly instead of merely casting it (which they already have to do today anyway).

thomasvl:
I'll admit, as I look at doing SE0206, the fact that we defined `hashValue` already worries me as we sorta want to kill that off also for the new method, but this also has me revisiting some other things…

We used protocols default impls to provide `Equatable` and `Hashable` and did them in such a way that they can be hooked.  With SE0206, I'm guessing we can do the same things with `hash(into:)`, but I also wonder if we're going out of our way to support something that might not be used? And…

The recent Swift Forums thread [_Retroactive Conformances vs. Swift-in-the-OS_](https://forums.swift.org/t/retroactive-conformances-vs-swift-in-the-os/14393) also got me thinking about this.  As the thread mentions, there is potential for extension collisions not just with new system libraries, but also with multiple frameworks.  So as Swift reaching binary stability, are these customization points also opening the door for future issues for developers?  If a third party framework exposes Messages as part of their interface, and some code on top of that framework hooks hashing/equality for something they do with the messages; what happens when the framework is updated and they provide a hashing override also?  Does that suddenly break to code on top because it gets a different hashing/equality behavior?  Are we better off not exposing these hooks and if someone wants to override hashing/equality they just have to box the type like the forums thread talks about?

Anyway, this may be larger than a PR discussion.  Should open an issue for tracking this without it being tied to a PR?  Should we open the issue and continue the discussion, but push this PR ahead to go after full SE0206 support to see if it impacts the discussion and just make sure we resolve this larger question before we try to cut a release?  (or maybe we just cut 1.1.x and then 1.2.x/2.x if we do revisit this hierarchy)

tbkka:
> We've gone out of our way to provide a Message protocol without any associated types or Self constraints so that people can create heterogeneous collections of messages. ...

The real motivation for me was to allow `Message` to be used as a function argument or return type.

thomasvl:
> The real motivation for me was to allow `Message` to be used as a function argument or return type.

What's your thoughts then on how we also try to make things hookable given the approaching binary stability work and potential issues when frameworks and multiple attempts to replace the hooks?

thomasvl:
Dropping this.  It spawning some useful discussion and got us looking at things, but some of what was raised needs to be sorted out before deciding to go forward with this or the other things.

tbkka:
This would be a breaking API change, so we should maybe revisit for a future V2 effort.

PR #772, opened 2018-07-23
Check for unknown fields in the Conformance Requests.

thomasvl:
Error if there are any.  Simply regenerating conformance.pb.swift will
"work around" the failure, but the real thing is to look at the new
fields and add support.

### DIFF
--- a/Sources/Conformance/main.swift
+++ b/Sources/Conformance/main.swift
@@ -60,6 +60,13 @@ func buildResponse(serializedData: Data) -> Conformance_ConformanceResponse {
         return response
     }
 
+    // Detect when something gets added to the conformance request that isn't
+    // supported yet.
+    if !request.unknownFields.data.isEmpty {
+        response.runtimeError = "ConformanceRequest had unknown fields; regenerate conformance.pb.swift and see what support needs to be added."
+        return response
+    }
+
     let msgType: SwiftProtobuf.Message.Type
     switch request.messageType {
     case "":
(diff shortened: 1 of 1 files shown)


Sources/Conformance/main.swift
```diff
@@ -60,6 +60,13 @@ func buildResponse(serializedData: Data) -> Conformance_ConformanceResponse {
         return response
     }
 
+    // Detect when something gets added to the conformance request that isn't
+    // supported yet.
+    if !request.unknownFields.data.isEmpty {
+        response.runtimeError = "ConformanceRequest had unknown fields; regenerate conformance.pb.swift and see what support needs to be added."
```
allevato: Nit: wrap long line

Sources/Conformance/main.swift
```diff
@@ -60,6 +60,13 @@ func buildResponse(serializedData: Data) -> Conformance_ConformanceResponse {
         return response
     }
 
+    // Detect when something gets added to the conformance request that isn't
+    // supported yet.
+    if !request.unknownFields.data.isEmpty {
```
allevato: Nit: `guard request.unknownFields.data.isEmpty else { ... }` is ever-so-slightly clearer since it's an exceptional early exit case.

thomasvl: done

Sources/Conformance/main.swift
thomasvl: done

Issue #799, opened 2018-09-19
Allow to use UNKNOWN case instead of unrecognizedEnumValue


geor-kasapidi:
Now in JSONScanner error will be return if enum case in unknown.

Propose - add setting to JSONDecodingOptions (ex. ignoreUnknownEnumCases) to return .UNKNOWN case instead of error.

Xcode 10, Swift 4.2, SwiftProtobuf - 1.1.2

thomasvl:
I don't think any other languages provide an option for that, and I'm not immediately sure it would be a good thing since changing the enum value like that could put the message in an odd state (depending on how the enum field relates to other fields).  Since proto3 syntax enums support unknown values, it seems like if you need to deal with unknowns, binary might be a better format to use than json?

geor-kasapidi:
It turns out that the behavior is different: in case of json representation of model we have an error, with binary representation - it's ok.

For example, our testers use json transport to validate server data. And we use different inits on SwiftProtobuf.Message for that:

```
extension SwiftProtobuf.Message {
    public init(data: Data, useJSON: Bool) throws {
        if useJSON {
            try self.init(jsonUTF8Data: data)
        } else {
            try self.init(serializedData: data)
        }
    }
}
```

thomasvl:
> It turns out that the behavior is different: in case of json representation of model we have an error, with binary representation - it's ok.

Right, in binary form, we have the raw value and so it can still be modeled.  If you decided to use JSON it can't handing unknown fields or enum values.  That's one of the major limitation of JSON.  If you want to better support your protos evolving with time, binary is the format to use.

thomasvl:
Looks like C++ has picked up an option to use the raw values instead of the names when generating JSON (https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/util/json_util.h#L67-L69), so it might make sense to add that to allow the data to better flow thru JSON.  That seems safer that an option to ignore unknown json enum names (which I still can't find anything else supporting).

tbkka:
I remember implementing support for numeric enums in the decode side -- it's required by the conformance test.   Looks like it even supports unknown values:
   https://github.com/apple/swift-protobuf/blob/a1b3a3a7d458f0a0ac0afb7e61baaac5675c8d63/Sources/SwiftProtobuf/JSONScanner.swift#L1284

So it should only require a little work on the encoder to expose this there.

tbkka:
[...]

PR #800, opened 2018-09-20
Add JSON encoding options

thomasvl:
Doesn't "fix" #799 original request (I'm still hesitant to have an decoding option to include the unknown values since no other languages seem to have it), but it does provide a way (like other languages) to JSON encode enums as ints so they can do better with new enum values since decoders are supposed to accept integer values for enums.


Sources/SwiftProtobuf/Message+JSONAdditions.swift
```diff
@@ -22,9 +22,13 @@ public extension Message {
   /// serializing to JSON.
   ///
   /// - Returns: A string containing the JSON serialization of the message.
+  /// - Parameters:
+  ///   - options: The JSONEncodingOptions to use.
   /// - Throws: `JSONEncodingError` if encoding fails.
-  func jsonString() throws -> String {
-    let data = try jsonUTF8Data()
+  func jsonString(
+    options: JSONEncodingOptions = JSONEncodingOptions()
```
thomasvl: How do we want to handle this from a library version number pov?

tbkka: This change (and the change to `jsonUTF8Data()` below) are certainly backwards source compatible.

But I'm not sure if adding a new optional parameter retains backwards binary compatibility.  @allevato would know.

If it's source and binary backwards compatible, then I see no implications at all for version numbering.

If it's not binary compatible, then it deserves at least a minor version update.

Tests/SwiftProtobufTests/Test_JSONEncodingOptions.swift
```diff
@@ -0,0 +1,104 @@
...
+      $0.optionalEnum = .bar
+    }
+    XCTAssertEqual(try msg1.jsonString(options: asStrings), "{\"optionalEnum\":\"BAR\"}")
+    XCTAssertEqual(try msg1.jsonString(options: asInts), "{\"optionalEnum\":1}")
+
+    let msg2 = ProtobufUnittest_Message3.with {
+      $0.repeatedEnum = [.bar, .baz]
+    }
+    XCTAssertEqual(try msg2.jsonString(options: asStrings), "{\"repeatedEnum\":[\"BAR\",\"BAZ\"]}")
+    XCTAssertEqual(try msg2.jsonString(options: asInts), "{\"repeatedEnum\":[1,2]}")
```
tbkka: I like the way this reads.  Should we bless this by adding some sugar?
```
     let a = try msg.jsonString(options: .enumsAsInts)
```

tbkka:
This looks good.

Sources/SwiftProtobuf/Message+JSONAdditions.swift
thomasvl: Added a commit to bump the version to 1.2.0 now so we don't forget.

Tests/SwiftProtobufTests/Test_JSONEncodingOptions.swift
thomasvl: @tbkka if we only ever had booleans, I think we could use `OptionSet`, and maybe it would work out simpler for the initialization code, but since we already have some that aren't, I don't think we can do something nicer.

If we split the bools into an `OptionSet` and everything else as unique fields, then we might have some middle ground, but given that's a code breaking changes, I'm guessing we don't want to consider that?  (And I'm not sure how I like grouping some in an `OptionSet` and others outside it, but all in a `struct` that still needs to be named `Options`.)

tbkka: Makes sense.  Sounds like we should leave it as-is.

PR #811, opened 2018-11-16
Update SPM Integration Instructions

ryanwilson:
- Adds missing instructions to add `SwiftProtobuf` to the target
- Changes the versioning syntax from "from" to "exact"  in order to better match instructions of using the same library version as the generated code
- Updates the outdated "Version()" string from an old version of SPM

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,12 +177,16 @@ After copying the `.pb.swift` files into your project, you will need to add the
 [SwiftProtobuf library](https://github.com/apple/swift-protobuf) to your
 project to support the generated code.
 If you are using the Swift Package Manager, add a dependency to your
-`Package.swift` file.  Adjust the `Version()` here to match the `[tag_name]`
-you used to build the plugin above:
+`Package.swift` file and import the `SwiftProtobuf` library into the desired
+targets.  Adjust the `"1.2.0"` here to match the `[tag_name]` you used to build
+the plugin above:
 
 ```swift
 dependencies: [
-        .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.1.0")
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -177,12 +177,16 @@ After copying the `.pb.swift` files into your project, you will need to add the
...
-`Package.swift` file.  Adjust the `Version()` here to match the `[tag_name]`
-you used to build the plugin above:
+`Package.swift` file and import the `SwiftProtobuf` library into the desired
+targets.  Adjust the `"1.2.0"` here to match the `[tag_name]` you used to build
+the plugin above:
 
 ```swift
 dependencies: [
-        .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.1.0")
+    .package(url: "https://github.com/apple/swift-protobuf.git", .exact("1.2.0")),
```
thomasvl: I don't think we want `exact` because it means developers have to know to go back and manually update to get bug fixes, etc.

thomasvl: `upToNextMinor` is likely better.

It might make sense to tweak the other docs to call out making sure they are using a protoc plugin that matches the library version they are using.  i.e. - shift the focus to making sure their generator is up to date.  Historically we've had folks using old generators, so that's part of what the other wording was trying to deal with.

@allevato @tbkka - sound good?

cburrows: I think “from:” is the right thing to tell people. The root package in a dep tree can pin its dependencies.

In the case of SwiftProtobuf being an indirect dependency, perhaps more than once, for the root package: of those intermediate projects start expressing their needs this way it’s a huge headache with little upside.

ryanwilson: SGTM, I moved back to `from:`. I didn't add any further instructions regarding always making sure things are up to date.

thomasvl:
Thanks!

PR #818, opened 2018-12-14
Add Package file for Swift 5.

dierksen:


tbkka:
Even better:  once Swift 5 is officially released, we can just drop the old (Swift 3) Package.swift entirely.

thomasvl:
https://github.com/thomasvl/swift-protobuf/tree/swiftpm_tweaks is what I had pending the proposal in that forum thread happening.

> Even better: once Swift 5 is officially released, we can just drop the old (Swift 3) Package.swift entirely.

yup, and by moving the 5+ support to `Package.swift` file would put us on a path to eventually not needing the version specific files.

thomasvl:
My change is up as a #819. It should put things on a path were we don't need to have version specific files in the future.

PR #819, opened 2018-12-14
Swiftpm tweaks

thomasvl:
This is trying to get things into better shape for Swift 5.

@dierksen found that the existing release fail to build because the `Package.swift` we have ends up being assumed to be version 3.  But trunk [SwiftPM already dropped support for that](https://forums.swift.org/t/removing-packagedescription-api-v3-master-branch/14109).

Assuming the ideas [RFC: Tools version with version-specific manifests](https://forums.swift.org/t/rfc-tools-version-with-version-specific-manifests/18224) happens, then we should be generally good with this change.

So factoring in the advice from [swiftLanguageVersions best practices for libraries](https://forums.swift.org/t/swiftlanguageversions-best-practices-for-libraries/18443), I believe the path forward with this change is:

1. Remove our `Swift@swift-#.swift` files as we drop support for those older versions.
2. Only add `.version("#.#")` values to `Package.swift` so we don't require a newer version to understand the non string values.
3. Only create a `Swift@swift-#.swift` when something new gets added to SwiftPM that we want to take advantage of.  And when that happens, the current `Package.swift` would get renamed to capture a version number and the `Package.swift` would get its tools version updated to where the new feature landed that we need to take advantage of.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,35 @@
+// swift-tools-version:4.2
+
+// Package.swift
+//
+// Copyright (c) 2014 - 2018 Apple Inc. and the project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/master/LICENSE.txt
+//
+
+import PackageDescription
+
+let package = Package(
+  name: "SwiftProtobuf",
+  products: [
+    .executable(name: "protoc-gen-swift", targets: ["protoc-gen-swift"]),
+    .library(name: "SwiftProtobuf", targets: ["SwiftProtobuf"]),
+    .library(name: "SwiftProtobufPluginLibrary", targets: ["SwiftProtobufPluginLibrary"]),
+  ],
+  targets: [
+    .target(name: "SwiftProtobuf"),
+    .target(name: "SwiftProtobufPluginLibrary",
+            dependencies: ["SwiftProtobuf"]),
+    .target(name: "protoc-gen-swift",
+            dependencies: ["SwiftProtobufPluginLibrary", "SwiftProtobuf"]),
+    .target(name: "Conformance",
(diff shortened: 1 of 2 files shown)


thomasvl:
@cburrows fyi (since I think you've commented on some the SwiftPM changes in the past)

Package.swift
```diff
@@ -1,6 +1,8 @@
-// Package.swift - description
+// swift-tools-version:4.2
```
cburrows: I find the combinations of these version mechanisms so difficult to reason about. So this seems OK, but can you answer a question? What is the expectation when someone is using Swift tools, say, 4.1.2? If such an spm sees this package declaration at a given version tag, it will decide that it cannot use it, and search versions in reverse until it finds presumably the old Package.swift? Which will be at 1.2.0 (in a typical case where no other package added weird constraints)?

I mean, I think that is what the documentation says will happen I just want to see what you think. And what your intention is.

Also, if I am right, does that mean that spms version 4.0 through 4.1.2 are essentially pinned to SwiftProtobuf 1.2.0, but that spms in v3 are not, due to Package@swift-3.swift?

Or am I just wrong about some mechanism here?

thomasvl: The only docs I can really find about it are https://github.com/apple/swift-package-manager/blob/master/Documentation/Usage.md#handling-version-specific-logic.

As far as your final question, I'm not completely sure what you are asking.  Since SwiftPM also looks for tags with this naming, we always have the option to cut a tag specific to a version and land a `Package` file on there for them if we need it.

As far as spms 4.0 - 4.1.2, since we have a `Package@swift-4.swift`, that will still work for them until we remove it.  And according to our version policy, that won't happen until we get a  5.2 or higher.

thomasvl: No worries.  The @Swift-4.2 and non qualified are a little redundant at the moment, but I didn't see a way around that given we're trying to match older and newer behaviors.  :(

PR #827, opened 2019-01-04
Update the generated protos

cburrows:
Looks like this was left out of 1.3.0, and it's due to the removal of the pre-3.2 code.


thomasvl:
Ugh, thanks.

PR #828, opened 2019-01-05
Correct Swift 4.1 build break

cburrows:
It looks like #822 added a call to Array.firstIndex, which was only
added in Swift 4.2. So builds are broken with Swift 4.1. This adds
firstIndex for Swifts prior to 4.2 to fix them.

### DIFF
--- a/Sources/SwiftProtobufPluginLibrary/Array+Extensions.swift
+++ b/Sources/SwiftProtobufPluginLibrary/Array+Extensions.swift
@@ -36,8 +36,7 @@ extension Array {
   }
 }
 
-#if swift(>=4.2)
-#else  // swift(>=4.2)
+#if !swift(>=4.2)
 extension Array {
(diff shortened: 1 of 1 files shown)


Sources/SwiftProtobufPluginLibrary/Array+Extensions.swift
```diff
@@ -35,3 +35,19 @@ extension Array {
     }
   }
 }
+
+#if swift(>=4.2)
+#else  // swift(>=4.2)
```
thomasvl: `#if !swift(>=4.2)`

?

cburrows: Ah, oops, of course. Done.

thomasvl:
1.3.1 released and CocoaPod also pushed.

PR #833, opened 2019-01-16
supportTVOS

bimawa:
Related with grpc/grpc#14173

Pls, check these changes.

### DIFF
--- a/SwiftProtobuf.podspec
+++ b/SwiftProtobuf.podspec
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.tvos.deployment_target = '9.0'
   s.watchos.deployment_target = '2.0'
 
-  s.cocoapods_version = '>= 1.1.0'
+  s.cocoapods_version = '>= 1.1.1'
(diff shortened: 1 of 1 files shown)


SwiftProtobuf.podspec
```diff
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.tvos.deployment_target = '9.0'
   s.watchos.deployment_target = '2.0'
 
-  s.cocoapods_version = '>= 1.1.0'
+  s.cocoapods_version = '>= 1.1.1'
```
thomasvl: Skimming the link you gave, I don't see why this is needed for tvOS support.  Is there some details someplace that explains it?  The [1.1.1 release notes](https://github.com/CocoaPods/CocoaPods/releases/tag/1.1.1) seem to imply it is a very minor release over 1.1.0.

bimawa: Ups, sry, yes you are right!  Something miss pushing

PR #836, opened 2019-01-23
[DO NOT MERGE] Stress testing the formatter.

allevato:
This PR is being created solely to test the formatter and allow
interested parties to comment directly on the results. The
configuration used here is artificially awkward, in order to
make sure we're hitting as many cases as possible:

- Indents are 6 spaces
- Line length is 80
- Line breaks already in the source are *ignored* (except for single
  blank lines separating neighboring statements/decls). This ensures
  that discretionary line breaks don't hide other potential issues
  in the output.


Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -16,276 +16,298 @@
...
-        }
+/// Custom handling for Text format.
+extension AnyMessageStorage {
+      func decodeTextFormat(
+            typeURL url: String, decoder: inout TextFormatDecoder
+      ) throws {
+            // Decoding the verbose form requires knowing the type.
+            _typeURL = url
+            guard
+                  let messageType = Google_Protobuf_Any.messageType(
```
allevato: This isn't quite what we want. It looks like I allow a break after the `guard` statement, but the other control flow keywords (`if`, etc.) use a space to keep the first clause on the same line. That would fix the weird continuation vs. block clash here.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -301,163 +323,173 @@ extension AnyMessageStorage {
...
-      return true
-    }
-
-    // Out of options. To do more compares, the states conversions would have to be
-    // done to do comparisions; and since equality can be used somewhat removed from
-    // a developer (if they put protos in a Set, use them as keys to a Dictionary, etc),
-    // the conversion cost might be to high for those uses.  Give up and say they aren't equal.
-    return false
-  }
+      #if swift(>=4.2)
```
allevato: Certain `#if` blocks don't get formatted at all, if the current version of the compiler just sends us back an `UnknownStmt/Decl`. Note that these are still indented +2. This should improve with newer versions of the compiler, which have more of the syntax nodes properly filled in.

We do the minimum, of ensuring that the left margin of the block is indented to the correct level.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -301,163 +323,173 @@ extension AnyMessageStorage {
...
+            }
+
+            // If both have serialized data, and they exactly match; the messages are equal.
+            // Because there could be map in the message, the fact that the data isn't the
+            // same doesn't always mean the messages aren't equal. Likewise, the binary could
+            // have been created by a library that doesn't order the fields, or the binary was
+            // created using the appending ability in of the binary format.
+            if case .binary(let myValue) = state,
+                  case .binary(let otherValue) = other.state, myValue
+                  == otherValue
```
allevato: We should probably use a group around the individual clauses to encourage them to wrap as a full unit, and only break internally if needed beyond that.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -301,163 +323,173 @@ extension AnyMessageStorage {
...
+            { return true }
+
+            // If both have contentJSON, and they exactly match; the messages are equal.
+            // Because there could be map in the message (or the JSON could just be in a different
+            // order), the fact that the JSON isn't the same doesn't always mean the messages
+            // aren't equal.
+            if case .contentJSON(let myJSON, _) = state,
+                  case .contentJSON(let otherJSON, _) = other.state, myJSON
+                  == otherJSON
+            { return true }
```
allevato: Do we hate this? (Allowing the body to be on a single line below a prologue that wrapped?)

I'm ambivalent; it's not terrible, and it would take a bit more work to make node wrapping dependent on other nodes. If we were respecting discretionary newlines, of course, the user could put them in and they'd stick.

Sources/SwiftProtobuf/Enum.swift
```diff
@@ -17,77 +17,76 @@
...
+      /// given name, if it provides names.
+      ///
+      /// Since the text format and JSON names are always identical, we don't need
+      /// to distinguish them.
+      ///
+      /// - Parameter name: The name of the enum case.
+      internal init?(name: String) {
+            guard let nameProviding = Self.self as? _ProtoNameProviding.Type,
+                  let number = nameProviding._protobuf_nameMap.number(
+                  forJSONName: name)
```
allevato: Another place where the block indentation level (increased by the left paren) and the continuation line state seem to be interacting oddly...

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -86,22 +83,23 @@ private func fromHexDigit(_ c: UnicodeScalar) -> UInt32? {
...
-/* 0x70 */ 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, -1, -1, -1, -1, -1,
-/* 0x80 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
-/* 0x90 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
-/* 0xa0 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
-/* 0xb0 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
-/* 0xc0 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
-/* 0xd0 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
-/* 0xe0 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
-/* 0xf0 */ -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+      // 0x00
```
allevato: Two problems here:

1. We transform block comments that start a line into line comments; we should probably only do that if it's the **only** thing on the line so that they can still be interspersed like this.
2. We should preserve discretionary horizontal spacing in limited cases, like in multiline array literals.

Performance/generators/cpp.sh
```diff
@@ -71,6 +71,7 @@ using google::protobuf::Descriptor;
 using google::protobuf::DescriptorPool;
 using google::protobuf::TextFormat;
 using google::protobuf::util::BinaryToJsonString;
+using google::protobuf::util::JsonParseOptions;
```
dabrahams: This seems to indicate that the base branch of the PR isn't correct.  Surely the formatter doesn't add lines?!

dabrahams: Also, should we be formatting a `.sh` file?

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -16,276 +16,298 @@
...
+            // Note: Unlike contentJSON below, binary does not bother to capture the
+            // decoding options. This is because the actual binary format is the binary
+            // blob, i.e. - when decoding from binary, the spec doesn't include decoding
+            // the binary blob, it is pass through. Instead there is a public api for
+            // unpacking that takes new options when a developer decides to decode it.
+            case binary(Data)
+            // a message
+            case message(Message)
+            // parsed JSON with the @type removed and the decoding options.
+            case contentJSON(Data, JSONDecodingOptions)
```
dabrahams: Probably, for all tests that are stripping discretionary breaks, the thing to do is make the PR against an empty repo, so all lines are additions and there are no deletions to interrupt our view of the reformatted code.  We don't really care what the old code looked like for this kind of test.

Sources/SwiftProtobuf/AnyMessageStorage.swift
dabrahams: One could argue persuasively for indenting everything inside the `#if` block one level further.  I'm agnostic on this personally, but in principle it seems like showing `#if` block structure is just as important as showing brace structure.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -301,163 +323,173 @@ extension AnyMessageStorage {
...
+            hash = (hash &* i_16777619) ^ _typeURL.hashValue
+          }
+          return hash
+        }
+      #endif
+
+      func isEqualTo(other: AnyMessageStorage) -> Bool {
+            if _typeURL != other._typeURL { return false }
+
+            // Since the library does lazy Any decode, equality is a very hard problem.
```
dabrahams: Don't these comments go past 80 columns?  It seems to me that word wrapping the paragraphs is not unreasonable, but we'd have to be smart about not disturbing ASCII art, etc.  I have some ideas about that if anyone's interested.

Sources/SwiftProtobuf/AnyMessageStorage.swift
dabrahams: Mos' def!

Sources/SwiftProtobuf/AnyMessageStorage.swift
dabrahams: I don't hate it enough to complicate the rules, I think.  With discretionary breaks most people will do the palatable thing and there won't be a problem.  Now the interesting thing is, if we are respecting discretionary breaks but they give us this nasty thing

```swift
if long
    + condition {
    return true }
```

Will we insert the newline before the close brace as follows

```swift
if long
    + condition
{
    return true // <== here
}
```

or would we end up with

```swift
if long
    + condition
{
    return true } // <== yuck
```

?

dabrahams: Obviously(?) I prefer the former.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -301,163 +323,173 @@ extension AnyMessageStorage {
...
+                        serializedData: valueData, partial: true)
+                  return try serializeAnyJSON(
+                        for: m, typeURL: _typeURL, options: options)
+
+            case .message(let msg):
+                  // We should have been initialized with a typeURL, but
+                  // ensure it wasn't cleared.
+                  let url = !_typeURL.isEmpty
+                        ? _typeURL : buildTypeURL(
+                              forMessage: msg,
```
dabrahams: I'm wondering what would happen here if the expression after `:` was a long string of additions instead of a function call (aside from warnings about typechecker performance, I mean ;-) )

Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -18,1477 +18,1505 @@
...
+      }
+
+      private mutating func decrementRecursionDepth() {
+            recursionBudget += 1
+            // This should never happen, if it does, something is probably corrupting memory, and
+            // simply throwing doesn't make much sense.
+            if recursionBudget > options.messageDepthLimit {
+                  fatalError(
+                        "Somehow BinaryDecoding unwound more objects than it started"
+                  )
```
dabrahams: The string exceeds 80 columns; it should be rewritten so it fits.

```swift
print("""
  no \
  newline
  """)
```

Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -18,1477 +18,1505 @@
...
+      /// Return the next field number or nil if there are no more fields.
+      internal mutating func nextFieldNumber() throws -> Int? {
+            // Since this is called for every field, I've taken some pains
+            // to optimize it, including unrolling a tweaked version of
+            // the varint parser.
+            if fieldNumber > 0 {
+                  if let override = unknownOverride {
+                        assert(!options.discardUnknownFields)
+                        assert(
+                              fieldWireFormat != .startGroup && fieldWireFormat
```
dabrahams: This should break before the `&&`

Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -18,1477 +18,1505 @@
...
+                              let c3 = start[3]
+                              fieldNumber |= Int(c3 & 0x7f) << 18
+                              if (c3 & 0x80) == 0 {
+                                    p += 4
+                                    available -= 4
+                              }
+                              else {
+                                    if available < 5 {
+                                          throw
+                                                BinaryDecodingError.malformedProtobuf
```
dabrahams: I'd want it wrapped like this if possible:

```swift
throw BinaryDecodingError
      .malformedProtobuf
```

Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -18,1477 +18,1505 @@
...
+            case WireFormat.fixed32:
+                  let i = try decodeFloat()
+                  value.append(i)
+                  consumed = true
+            case WireFormat.lengthDelimited:
+                  let bodyBytes = try decodeVarint()
+                  if bodyBytes > 0 {
+                        let itemSize = UInt64(MemoryLayout<Float>.size)
+                        let itemCount = bodyBytes / itemSize
+                        if bodyBytes % itemSize != 0 || itemCount > UInt64(
```
dabrahams: break before `||`

Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -18,1477 +18,1505 @@
...
+      }
+
+      internal mutating func decodeSingularSFixed64Field(value: inout Int64)
+            throws
+      {
+            guard fieldWireFormat == WireFormat.fixed64 else { return }
             var i: Int64 = 0
             try decodeEightByteNumber(value: &i)
-            value.append(Int64(littleEndian: i))
+            value = Int64(littleEndian: i)
```
dabrahams: I hope the formatter didn't make this change!

Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -18,1477 +18,1505 @@
...
-                        encoder.putVarInt(value: Int64(v))
-                    }
-                }
-                unknownOverride = field
+      }
+
+      internal mutating func decodeRepeatedEnumField<E: Enum>(value: inout [E])
+            throws where E.RawValue == Int {
+            switch fieldWireFormat {
+            case WireFormat.varint:
```
dabrahams: Lots of people are unhappy with dedented case's.  Personally I appreciate the way they prevent `switch`es from eating up whitespace, but for example, I couldn't convince the people on my current project that they were OK.

Sources/SwiftProtobuf/Enum.swift
dabrahams: Yeah, I believe we got to this point in our analysis and that's when I realized we might need to do something more sophisticated, like reset the nesting->indentation mapping when an unmatched-on-the-same-line close delimiter ends a continuation line, and *that's* when I realized I wanted to stop trying to figure it all out in that PR ;-).  I think the answer fits neatly into your stack-based model, though.

Sources/SwiftProtobuf/JSONScanner.swift
dabrahams: I'm not sure about number 2 above.  It might not be worth it, especially because it will create the expectation that you would do the same thing for multiline, multi-parameter subscripts and/or function argument lists.

dabrahams:
Overall I want to say this is freaking awesome.  There are almost no places where the formatting doesn't work at all and only a few I want to tweak.  Great job, @allevato

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -301,163 +323,173 @@ extension AnyMessageStorage {
...
+            // removes issues like map<> serialization order, some other protocol buffer
+            // implementation details/bugs around serialized form order, etc.; but that
+            // would also greatly slow down equality tests.
+            //
+            // Do our best to compare what is present have...
+
+            // If both have messages, check if they are the same.
+            if case .message(let myMsg) = state,
+                  case .message(let otherMsg) = other.state, type(of: myMsg)
+                  == type(of: otherMsg) {
```
thomasvl: Drive by - Scanning this file, this wrap on the `==` took me a moment to figure out what the line really was.  Not sure if there is a better thing to be done here.

Sources/SwiftProtobuf/BinaryDecoder.swift
thomasvl: 😄 it's the diffs not really lining up, it is different functions with similar bodies confusing things.

Sources/SwiftProtobuf/ExtensionFieldValueSet.swift
```diff
@@ -15,75 +15,67 @@
...
+      #else  // swift(>=4.2)
+        public var hashValue: Int {
+          var hash = 16_777_619
+          for (fieldNumber, v) in values {
+            // Note: This calculation cannot depend on the order of the items.
+            hash = hash &+ fieldNumber &+ v.hashValue
+          }
+          return hash
+        }
+      #endif
```
thomasvl: The comment on this `#endif` seems to have gotten lots, but the on on the `#else` above survived.

Sources/SwiftProtobuf/FieldTag.swift
```diff
@@ -21,49 +21,45 @@
...
+      /// Creates a new tag from its raw numeric representation.
+      ///
+      /// Note that if the raw value given here is not a valid tag (for example, it
+      /// has an invalid wire format), this initializer will fail.
+      init?(rawValue: UInt32) {
+            // Verify that the field number and wire format are valid and fail if they
+            // are not.
+            guard rawValue & ~0x07 != 0,
+                  let _ = WireFormat(rawValue: UInt8(rawValue % 8)) else {
+                  return nil
```
thomasvl: The continuation of the` guard` the the body of the `else` having the same indent doesn't seem so great.

Sources/SwiftProtobuf/api.pb.swift
```diff
@@ -35,17 +35,17 @@
 // THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
 import Foundation
```
thomasvl: Granted this is a generated file, but why lose the blank line between a comment (license) and then the code, this blank seems like it was useful because of the type of content.

Performance/generators/cpp.sh
allevato: Oh, this is just my fork's `master` being more stale than I realized compared to upstream.

I'd rebase/merge and push again, but I'm afraid GitHub will hide all the existing comment threads once the commit is outdated. There haven't been any terribly significant code changes otherwise here that would affect what we want to look at, though.

Sources/SwiftProtobuf/AnyMessageStorage.swift
allevato: They do—non-documentation comments that exceed the column limit are something we don't have a great way to handle yet, so I'm definitely interested in your thoughts on that.

Separately, for documentation comments, one thing I've considered is parsing the Markdown and then rerendering it back out with the desired column limit (which is why the whole `CommonMark` module, wrapping `cmark`, exists, even though I haven't gotten far enough to use it yet). One drawback is that cmark doesn't preserve the original characters used for emphasis, strong text, bulleted lists, etc., so if the user used underscores to emphasize a word they would get converted to asterisks as a result of formatting it. Maybe we'd be OK with that kind of canonicalization? But it also feels a bit heavy-handed.

Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -18,1477 +18,1505 @@
...
+            fieldStartP = p
+            self.extensions = extensions
+            self.options = options
+            recursionBudget = options.messageDepthLimit
+      }
+
+      internal init(
+            forReadingFrom pointer: UnsafePointer<UInt8>, count: Int,
+            parent: BinaryDecoder
+      ) {
```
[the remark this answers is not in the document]
allevato: I can experiment with the groupings to see if we can satisfy all of those at once. I think it should be possible.

Sources/SwiftProtobuf/BinaryDecoder.swift
allevato: We could consider adding a `--swift-version` flag to the formatter that would let us conditionally enable certain transformations only if the target version supports it.

I'm not sure what the right default would be though if it's not specified (the current toolchain version?) and how we would query that cleanly in the context of the formatter.

Sources/SwiftProtobuf/BinaryDecoder.swift
allevato: TBH indenting them would remove another special case, so I wouldn't be heartbroken if folks wanted to do that. I mainly kept it because it's what Xcode does when formatting-while-typing (it initially indent when you type the `{` and move to the next line, but once you type the `:`, the clause jumps back), and I didn't want our developers to constantly be fighting against Xcode. (That's never a winnable battle though, and there are plenty of other places in our style where fighting occurs.)

Sources/SwiftProtobuf/BinaryDecoder.swift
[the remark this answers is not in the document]
thomasvl: Drive by - I can't help but also wonder about https://forums.swift.org/t/trailing-commas-in-all-expression-lists/19527 in this context.  Would some of the people in favor of that rather see these all on their own lines and the closing paren (and return) on the next line.  i.e. - how many of the points in that thread also sorta apply here?  (they also apply to the formatting of the function calls themselves)

Sources/SwiftProtobuf/AnyMessageStorage.swift
allevato: We should put groups around each of the comma-separated clauses to prefer that they wrap as a whole instead of breaking in the middle of one of the expressions unless absolutely necessary.

The curly brace should have been moved to the next line; that's a minor bug I'll have to track down.

Sources/SwiftProtobuf/ExtensionFieldValueSet.swift
allevato: @dabelknap Did I break this, or was this a known issue that we just hadn't fixed yet?

Sources/SwiftProtobuf/FieldTag.swift
allevato: Yeah, the `else {` should have moved down to the next line at the original indentation. I have specific test cases for situations similar to this, so I need to track down why it's not happening here.

Sources/SwiftProtobuf/api.pb.swift
allevato: I think this is a problem with the rule that sorts the imports; it may not be preserving the leading trivia correctly on those lines. I need to overhaul the logic there anyway.

allevato:
> Overall I want to say this is freaking awesome. There are almost no places where the formatting doesn't work at all and only a few I want to tweak. Great job, @allevato

Thanks @dabrahams, but I have to give much of the credit to @dabelknap. He did all the hard work of deciphering Oppen's paper, implementing the algorithm, and configuring most of the rules—I just came along later to tweak the specifics on top of what he had already done 😅

Sources/SwiftProtobuf/AnyMessageStorage.swift
allevato: I agree; a non-scientific and non-exhaustive look at the places where I use `#if` the most is when checking language version or target platform, and they almost always have an `#else` block as well. Indenting those blocks ensures that those two blocks don't blend together into a single harder-to-decipher unit.

Sources/SwiftProtobuf/AnyMessageStorage.swift
allevato: Open and close breaks always guarantee paired consistency†, so if the user inserts a discretionary line break after the open delimiter then we force a line break before the close delimiter as well, and vice versa. So the former, desired outcome is what we'll get.

† The exception to this is for the close break before the right parenthesis/bracket of a function call/subscript expression, which we explicitly allow to remain on the previous line. (And this is controllable using a property on the break that can be adjusted on a case-by-case basis.)

dabrahams:
> On Jan 24, 2019, at 8:25 AM, Tony Allevato <notifications@github.com> wrote:
> 
> Oh, this is just my fork's master being more stale than I realized compared to upstream.
> 
> I'd rebase/merge and push again, but I'm afraid GitHub will hide all the existing comment threads once the commit is outdated.
> 

You don’t need to rebase; just edit the base branch of your PR in the web interface.  You might need to push a branch to use as the base.  GH has gotten a lot better about knowing which comments are actually out of date.
> There haven't been any terribly significant code changes otherwise here that would affect what we want to look at, though.
> 
OK, I’m fine either way.

Sources/SwiftProtobuf/AnyMessageStorage.swift
dabrahams: @allevato: +1.  That said, if it *was* going break before `==` (as will sometimes be unavoidable) it's arguable that the `==` should be further indented

dabrahams:
> On Jan 24, 2019, at 5:31 AM, Thomas Van Lenten <notifications@github.com> wrote:
> 
> 😄 it's the diffs not really lining up, it is different functions with similar bodies confusing things.
> 

…which is why this particular kind of test should be based against a clean repo.

Sources/SwiftProtobuf/BinaryDecoder.swift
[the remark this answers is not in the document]
dabrahams: > I can experiment with 

Cheers! FWIW, when I say “it's arguable” it means I could easily be talked out of it, so if you are going to pursue such ideas I'd like to know that you're +1 and not just doing it “cuz Dave sez”

Sources/SwiftProtobuf/BinaryDecoder.swift
dabrahams: Once we get this thing out there as the official formatter, I think (without evidence—personal guess and not a statement on behalf of Apple) we'll be able to get Xcode to cooperate with it.  I sure as heck hope so.  I'll see what strings can be pulled on this end.

Sources/SwiftProtobuf/ExtensionFieldValueSet.swift
dabelknap: This is a pre-existing issue. The problem is that the entire `#if/#endif` block is an UnknownStmt, and being printed verbatim. I don't think we're currently checking for trailing comments on verbatim tokens. Since the missing comment belongs to the `public` keyword, we're missing it.

Sources/SwiftProtobuf/BinaryDecoder.swift
allevato: Expressions involving multiple binary operators are somewhat tricky to deal with the way that SwiftSyntax exposes them. There's a single `ExprList` node that's just a flat list of alternating terms and operators, so the AST structure doesn't actually reflect the precedence of those operators (I assume this is because of support for custom operators and custom precedence).

We could solve this for the built-in operators by processing that list as a whole and grouping terms based on the precedence of those operators to get the desired wrapping, but it's not obvious what to do if we encounter a custom operator.

Sources/SwiftProtobuf/JSONScanner.swift
dabelknap: Speaking as someone who has had to lay out large matrices in the past, allowing the extra spaces in array declarations is essential for readability.

dabelknap: More specifically, aligning the array elements at the commas.

Sources/SwiftProtobuf/BinaryDecoder.swift
allevato: Just for comments currently.

That being said, there's no good reason that we only do it for comments, and strings aren't a special case either; depending on the indentation, line length, and tokens involved, it's possible to end up with degenerate cases where there's no good way to break a token without it exceeding the limit, so we should unify our handling of these.

Sources/SwiftProtobuf/ExtensionFieldValueSet.swift
dabelknap: This is now fixed with PR https://github.com/google/swift/pull/213. Ordinarily, `#if` blocks are parsed by SwiftSyntax correctly. But, for some reason nesting it inside of a `struct` causes it to be parsed as an `UnknownStmt`.

PR #840, opened 2019-03-09
Accept too-large float values in Text format

tbkka:
Swift CI caught this recently:

The protobuf conformance suite recently added tests for
Text format.  One of their tests asserts that Text format
should accept too-large values for `float` fields
(in contrast with JSON, which rejects such values).

Add a new test assertion for this case and change
the TextFormat scanner to truncate large float
values rather than rejecting them.

More history:

The expected behavior changed in protocolbuffers/protobuf#5839 after it was originally committed in protocolbuffers/protobuf#5765

Specifically, see this edit:  https://github.com/protocolbuffers/protobuf/pull/5839/files#diff-892a679c750b4634959c71ca8930c5fc

### DIFF
--- a/Sources/SwiftProtobuf/TextFormatScanner.swift
+++ b/Sources/SwiftProtobuf/TextFormatScanner.swift
@@ -800,10 +800,7 @@ internal struct TextFormatScanner {
 
     internal mutating func nextFloat() throws -> Float {
         if let d = tryParseFloatString() {
-            let n = Float(d)
-            if n.isFinite {
-                return n
-            }
+            return Float(d)
         }
         if skipOptionalNaN() {
             return Float.nan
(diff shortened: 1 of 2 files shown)


thomasvl:
Should we continue to wait to cut a release until Swift 5 hits GM or just go ahead and cut a release?  The only changes since the last one seem to be edge cases that the conformance test as started checking.

tbkka:
> Should we continue to wait to cut a release until Swift 5 hits GM or just go ahead and cut a release? The only changes since the last one seem to be edge cases that the conformance test as started checking.

If it's working with the current snapshot/beta releases, let's just cut it.  When I was experimenting on Friday with the latest Xcode beta, I did notice some deprecation warnings (Data API churn) that we should figure out how to clean up, but nothing serious.

PR #843, opened 2019-03-16
Swift 5: adopt new Data.withUnsafeBytes API

alanzeino:
Unfortunately I wasn't able to use the `#if compiler(>=5.0)` flag since the project supports older swift versions where this directive is not available.

### DIFF
--- a/Sources/SwiftProtobuf/AnyMessageStorage.swift
+++ b/Sources/SwiftProtobuf/AnyMessageStorage.swift
@@ -66,22 +66,26 @@ fileprivate func unpack(contentJSON: Data,
   }
 
   var value = String()
-  try contentJSON.withUnsafeBytes { (bytes:UnsafePointer<UInt8>) in
-    let buffer = UnsafeBufferPointer(start: bytes, count: contentJSON.count)
-    var scanner = JSONScanner(source: buffer,
-                              messageDepthLimit: options.messageDepthLimit,
-                              ignoreUnknownFields: options.ignoreUnknownFields)
-    let key = try scanner.nextQuotedString()
-    if key != "value" {
-      // The only thing within a WKT should be "value".
-      throw AnyUnpackError.malformedWellKnownTypeJSON
-    }
-    try scanner.skipRequiredColon()  // Can't fail
-    value = try scanner.skip()
-    if !scanner.complete {
-      // If that wasn't the end, then there was another key,
-      // and WKTs should only have the one.
-      throw AnyUnpackError.malformedWellKnownTypeJSON
+  try contentJSON.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
+    if let baseAddress = body.baseAddress, body.count > 0 {
+      let bytes = baseAddress.assumingMemoryBound(to: UInt8.self)
+
+      let buffer = UnsafeBufferPointer(start: bytes, count: body.count)
+      var scanner = JSONScanner(source: buffer,
+                                messageDepthLimit: options.messageDepthLimit,
+                                ignoreUnknownFields: options.ignoreUnknownFields)
+      let key = try scanner.nextQuotedString()
+      if key != "value" {
+        // The only thing within a WKT should be "value".
+        throw AnyUnpackError.malformedWellKnownTypeJSON
+      }
+      try scanner.skipRequiredColon()  // Can't fail
+      value = try scanner.skip()
+      if !scanner.complete {
+        // If that wasn't the end, then there was another key,
+        // and WKTs should only have the one.
+        throw AnyUnpackError.malformedWellKnownTypeJSON
+      }
     }
   }
   return try messageType.init(jsonString: value, options: options)
--- a/Sources/SwiftProtobuf/BinaryDecoder.swift
+++ b/Sources/SwiftProtobuf/BinaryDecoder.swift
@@ -876,13 +876,16 @@ internal struct BinaryDecoder: Decoder {
                 }
                 let fieldSize = Varint.encodedSize(of: fieldTag.rawValue) + Varint.encodedSize(of: Int64(bodySize)) + bodySize
                 var field = Data(count: fieldSize)
-                field.withUnsafeMutableBytes { (pointer: UnsafeMutablePointer<UInt8>) in
+                field.withUnsafeMutableBytes { (body: UnsafeMutableRawBufferPointer) in
+                  if let baseAddress = body.baseAddress, body.count > 0 {
+                    let pointer = baseAddress.assumingMemoryBound(to: UInt8.self)
                     var encoder = BinaryEncoder(forWritingInto: pointer)
                     encoder.startField(tag: fieldTag)
                     encoder.putVarInt(value: Int64(bodySize))
                     for v in extras {
                         encoder.putVarInt(value: Int64(v))
                     }
+                  }
                 }
                 unknownOverride = field
             }
@@ -1204,9 +1207,11 @@ internal struct BinaryDecoder: Decoder {
                 // If there already was fieldData, decode it.
                 if let data = fieldData {
                     var wasDecoded = false
-                    try data.withUnsafeBytes { (pointer: UnsafePointer<UInt8>) in
+                    try data.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
+                      if let baseAddress = body.baseAddress, body.count > 0 {
+                        let pointer = baseAddress.assumingMemoryBound(to: UInt8.self)
                         var extDecoder = BinaryDecoder(forReadingFrom: pointer,
-                                                       count: data.count,
+                                                       count: body.count,
                                                        parent: self)
                         // Prime the decode to be correct.
                         extDecoder.consumed = false
@@ -1216,6 +1221,7 @@ internal struct BinaryDecoder: Decoder {
                                                             fieldNumber: fieldNumber,
                                                             messageExtension: ext)
                         wasDecoded = extDecoder.consumed
+                      }
                     }
                     if !wasDecoded {
                         return .malformed
@@ -1245,9 +1251,12 @@ internal struct BinaryDecoder: Decoder {
                         // Save it as length delimited
                         let payloadSize = Varint.encodedSize(of: Int64(data.count)) + data.count
                         var payload = Data(count: payloadSize)
-                        payload.withUnsafeMutableBytes { (pointer: UnsafeMutablePointer<UInt8>) in
+                        payload.withUnsafeMutableBytes { (body: UnsafeMutableRawBufferPointer) in
+                          if let baseAddress = body.baseAddress, body.count > 0 {
+                            let pointer = baseAddress.assumingMemoryBound(to: UInt8.self)
                             var encoder = BinaryEncoder(forWritingInto: pointer)
                             encoder.putBytesValue(value: data)
+                          }
                         }
                         fieldData = payload
                     } else {
--- a/Sources/SwiftProtobuf/Data+Extensions.swift
+++ b/Sources/SwiftProtobuf/Data+Extensions.swift
@@ -1,0 +1,33 @@
+// Sources/SwiftProtobuf/Data+Extensions.swift - Extension exposing new Data API
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/master/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+///
+/// Extension exposing new Data API to Swift versions < 5.0.
+///
+// -----------------------------------------------------------------------------
+
+import Foundation
+
+#if !swift(>=5.0)
+internal extension Data {
+    func withUnsafeBytes<T>(_ body: (UnsafeRawBufferPointer) throws -> T) rethrows -> T {
+        let c = count
+        return try withUnsafeBytes { (p: UnsafePointer<UInt8>) throws -> T in
+            try body(UnsafeRawBufferPointer(start: p, count: c))
+        }
+    }
+
+    mutating func withUnsafeMutableBytes<T>(_ body: (UnsafeMutableRawBufferPointer) throws -> T) rethrows -> T {
+        let c = count
+        return try withUnsafeMutableBytes { (p: UnsafeMutablePointer<UInt8>) throws -> T in
+            try body(UnsafeMutableRawBufferPointer(start: p, count: c))
+        }
+    }
+}
+#endif
--- a/Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
@@ -66,15 +66,19 @@ extension Google_Protobuf_Any {
     self.init()
     if !textFormatString.isEmpty {
       if let data = textFormatString.data(using: String.Encoding.utf8) {
-        try data.withUnsafeBytes { (bytes: UnsafePointer<UInt8>) in
-          var textDecoder = try TextFormatDecoder(
-            messageType: Google_Protobuf_Any.self,
-            utf8Pointer: bytes,
-            count: data.count,
-            extensions: extensions)
-          try decodeTextFormat(decoder: &textDecoder)
-          if !textDecoder.complete {
-            throw TextFormatDecodingError.trailingGarbage
+        try data.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
+          if let baseAddress = body.baseAddress, body.count > 0 {
+            let bytes = baseAddress.assumingMemoryBound(to: UInt8.self)
+
+            var textDecoder = try TextFormatDecoder(
+              messageType: Google_Protobuf_Any.self,
+              utf8Pointer: bytes,
+              count: body.count,
+              extensions: extensions)
+            try decodeTextFormat(decoder: &textDecoder)
+            if !textDecoder.complete {
+              throw TextFormatDecodingError.trailingGarbage
+            }
           }
         }
       }
--- a/Sources/SwiftProtobuf/JSONEncoder.swift
+++ b/Sources/SwiftProtobuf/JSONEncoder.swift
@@ -333,10 +333,13 @@ internal struct JSONEncoder {
     internal mutating func putBytesValue(value: Data) {
         data.append(asciiDoubleQuote)
         if value.count > 0 {
-            value.withUnsafeBytes { (p: UnsafePointer<UInt8>) in
+            value.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
+              if let baseAddress = body.baseAddress, body.count > 0 {
+                let p = baseAddress.assumingMemoryBound(to: UInt8.self)
+
                 var t: Int = 0
                 var bytesInGroup: Int = 0
-                for i in 0..<value.count {
+                for i in 0..<body.count {
                     if bytesInGroup == 3 {
                         data.append(base64Digits[(t >> 18) & 63])
                         data.append(base64Digits[(t >> 12) & 63])
@@ -369,6 +372,7 @@ internal struct JSONEncoder {
                 default:
                     break
                 }
+              }
             }
         }
         data.append(asciiDoubleQuote)
--- a/Sources/SwiftProtobuf/JSONScanner.swift
+++ b/Sources/SwiftProtobuf/JSONScanner.swift
@@ -190,8 +190,9 @@ private func parseBytes(
     // a closing double quote.
     index = digitsStart
     try value.withUnsafeMutableBytes {
-        (dataPointer: UnsafeMutablePointer<UInt8>) in
-        var p = dataPointer
+        (body: UnsafeMutableRawBufferPointer) in
+      if let baseAddress = body.baseAddress, body.count > 0 {
+        var p = baseAddress.assumingMemoryBound(to: UInt8.self)
         var n = 0
         var chars = 0 // # chars in current group
         var padding = 0 // # padding '=' chars
@@ -201,7 +202,6 @@ private func parseBytes(
             if k < 0 {
                 switch digit {
                 case asciiDoubleQuote:
-                    source.formIndex(after: &index)
                     break digits
                 case asciiBackslash:
                     source.formIndex(after: &index)
@@ -221,7 +221,6 @@ private func parseBytes(
                     while true {
                         switch source[index] {
                         case asciiDoubleQuote:
-                            source.formIndex(after: &index)
                             break digits
                         case asciiSpace:
                             break
@@ -269,7 +268,9 @@ private func parseBytes(
             break
         }
         throw JSONDecodingError.malformedString
+      }
     }
+    source.formIndex(after: &index)
     return value
 }
 
@@ -848,15 +849,18 @@ internal struct JSONScanner {
         let s = try nextQuotedString()
         let raw = s.data(using: String.Encoding.utf8)!
         let n = try raw.withUnsafeBytes {
-          (bytes: UnsafePointer<UInt8>) -> UInt64? in
-          let buffer = UnsafeBufferPointer(start: bytes, count: raw.count)
-          var index = buffer.startIndex
-          let end = buffer.endIndex
-          if let u = try parseBareUInt64(source: buffer,
-                                         index: &index,
-                                         end: end) {
-            if index == end {
-              return u
+          (body: UnsafeRawBufferPointer) -> UInt64? in
+          if let baseAddress = body.baseAddress, body.count > 0 {
+            let bytes = baseAddress.assumingMemoryBound(to: UInt8.self)
+            let buffer = UnsafeBufferPointer(start: bytes, count: body.count)
+            var index = buffer.startIndex
+            let end = buffer.endIndex
+            if let u = try parseBareUInt64(source: buffer,
+                                           index: &index,
+                                           end: end) {
+              if index == end {
+                return u
+              }
             }
           }
           return nil
@@ -907,15 +911,18 @@ internal struct JSONScanner {
         let s = try nextQuotedString()
         let raw = s.data(using: String.Encoding.utf8)!
         let n = try raw.withUnsafeBytes {
-          (bytes: UnsafePointer<UInt8>) -> Int64? in
-          let buffer = UnsafeBufferPointer(start: bytes, count: raw.count)
-          var index = buffer.startIndex
-          let end = buffer.endIndex
-          if let s = try parseBareSInt64(source: buffer,
-                                         index: &index,
-                                         end: end) {
-            if index == end {
-              return s
+          (body: UnsafeRawBufferPointer) -> Int64? in
+          if let baseAddress = body.baseAddress, body.count > 0 {
+            let bytes = baseAddress.assumingMemoryBound(to: UInt8.self)
+            let buffer = UnsafeBufferPointer(start: bytes, count: body.count)
+            var index = buffer.startIndex
+            let end = buffer.endIndex
+            if let s = try parseBareSInt64(source: buffer,
+                                           index: &index,
+                                           end: end) {
+              if index == end {
+                return s
+              }
             }
           }
           return nil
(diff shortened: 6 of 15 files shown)


thomasvl:
> Unfortunately I wasn't able to use the #if compiler(>=5.0) flag since the project supports older swift versions where this directive is not available.

Why?  We already use directive like that with problems.

alanzeino:
> > Unfortunately I wasn't able to use the #if compiler(>=5.0) flag since the project supports older swift versions where this directive is not available.
> 
> Why? We already use directive like that without problems.

Sorry I should've been more clear. `#if swift()` has been supported for a while but `#if compiler()` less so. The project at first glance supports Swift versions right back to 3.0, but `#if compiler()` was added much later than that.

I didn't see any in the codebase when I grep'd for it, but I would prefer to be wrong because I want to use that here :)

tbkka:
> The project at first glance supports Swift versions right back to 3.0,...

Our policy is to support one full major version back.  Since 5.0 is the current release version, that means we still support 4.0 and later.  Once 5.2 is released, we'll be able to use `#if compiler` since we'll only support >= 4.2 at that point.

thomasvl:
I also don't know the `#compiler` would be correct, as some of these seem to be API changes. API changes should be tied to the language version, not the underlying compiler version/release.  If they were tied to the compiler doesn't that mean some code that compiles with Swift 4.x would suddenly stop compiling in 4.x mode with the 5 compiler?

ps - I completely read that initial comment as `if #swift()`, adding to the confusion here. 😄

alanzeino:
> I also don't know the `#compiler` would be correct, as some of these seem to be API changes. API changes should be tied to the language version, not the underlying compiler version/release. If they were tied to the compiler doesn't that mean some code that compiles with Swift 4.x would suddenly stop compiling in 4.x mode with the 5 compiler?
> 
> ps - I completely read that initial comment as `#if swift()`, adding to the confusion here. 😄

Yep you're totally right. These changes are all Swift 5 compiler in 5.0 mode, not backwards compatible with 5 compiler in 4.2 mode.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -51,7 +51,11 @@ fileprivate func emitVerboseTextForm(visitor: inout TextFormatEncodingVisitor, m
 fileprivate func asJSONObject(body: Data) -> Data {
   let asciiOpenCurlyBracket = UInt8(ascii: "{")
   let asciiCloseCurlyBracket = UInt8(ascii: "}")
+  #if swift(>=5.0)
```
tbkka: Is the conditional really needed here?

I just tried the new version `Data([asciiOpenCurlyBracket])` in Swift 4.0 and it seems to work correctly there.

Sources/SwiftProtobuf/Data+Extensions.swift
```diff
@@ -0,0 +1,49 @@
+// Sources/SwiftProtobuf/Data+Extensions.swift - Extension wrapping deprecated Data methods
```
tbkka: Elsewhere, we've done this the other way around.

That is, update the code to use the _new_ syntax, and write an adapter to simulate the new syntax on old Swift versions.  It's a bit more work, but means that when we stop supporting those old versions, we can simply delete the adapter code.

Sources/SwiftProtobuf/JSONEncoder.swift
```diff
@@ -70,7 +70,11 @@ internal struct JSONEncoder {
 
     internal init() {}
 
+    #if swift(>=5.0)
```
tbkka: As above, if the new syntax works in Swift 4, then we can avoid this conditional.

Sources/SwiftProtobuf/Data+Extensions.swift
tbkka: If you could get the adapter right (simulating new syntax on old Swift), and prove that it works by converting just a few places, I'm happy to help get the rest converted.  I've gotten pretty handy at scripting global changes of this sort.

Sources/SwiftProtobuf/JSONEncoder.swift
tbkka: I just tried it; the new syntax does seem to work in Swift 4.

tbkka:
@alanzeino -- I'm testing a branch where I've globally replaced all of the `Data(bytes:)` with `Data(_:)`.  Once I finish running tests (for Swift 4.0 through 5.0), I'll put up a PR for that.  So I think you can focus on just the 24 warnings about `withUnsafeBytes` and `withUnsafeMutableBytes`.  If you can fix those, I think I've got the rest of the 5.0 issues covered.

alanzeino:
> @alanzeino -- I'm testing a branch where I've globally replaced all of the `Data(bytes:)` with `Data(_:)`. Once I finish running tests (for Swift 4.0 through 5.0), I'll put up a PR for that. So I think you can focus on just the 24 warnings about `withUnsafeBytes` and `withUnsafeMutableBytes`. If you can fix those, I think I've got the rest of the 5.0 issues covered.

Yep, I'll rebase.

Sources/SwiftProtobuf/Data+Extensions.swift
alanzeino: So I took a look at it, and I think doing it in this fashion is maybe a bit worse than the current one. 

My opinion is based on this [tweet](https://twitter.com/jckarter/status/1071183859083960320) — basically Apple is saying that we should carefully check that the pointer is both valid and has a length (`count`).

For example, if we try to support the new syntax in the code, this call site (from `BinaryDelimited.swift`): 

```
    data.withUnsafeBytes { (pointer: UnsafePointer<UInt8>) in
      written = stream.write(pointer, maxLength: totalSize)
    }
```

Would have to be converted to this:

```
    data.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
      if let baseAddress = body.baseAddress, body.count > 0 {
        let pointer = baseAddress.assumingMemoryBound(to: UInt8.self)
        written = stream.write(pointer, maxLength: totalSize)
      }
    }
```

But if someone tries to write this again, they might use the old way and forget to do the `count` check and the `baseAddress` check, leading to a crash. With the wrapper, we can make sure that no future additions will crash, and when we want to delete the old code we just remove the `#else` case in the wrapper.

In addition, if we want to convert all the call sites to the new API, we're going to have a problem because the new API in 5.0 is this:

```
func withUnsafeBytes<ResultType>(_ body: (UnsafeRawBufferPointer) throws -> ResultType) rethrows -> ResultType
```

The new function no longer is generic over the content type in the Data like the old API:

```
withUnsafeBytes<ResultType, ContentType>(_ body: (UnsafePointer<ContentType>) throws -> ResultType) rethrows -> ResultType
```

But in the implementation of the closure, you need to know the content type so you can call `assumingMemoryBound(to:)`:

```
let pointer = baseAddress.assumingMemoryBound(to: UInt8.self)
```

I can't find a way to provide that type to `assumingMemoryBound(to:)` at runtime, it needs to be provided at compile time. Either we use the old API approach of genericizing over `ResultType` and `ContentType` (which defeats the purpose of making this change), or we take in the type in the function:

```
internal func withUnsafeBytes<T, ResultType>(innerType: T.Type, _ body: (UnsafeRawBufferPointer) throws -> ResultType) rethrows -> ResultType {
    if count > 0 {
      return try withUnsafeBytes { (pointer: UnsafePointer<T>) -> ResultType in
        return try body(UnsafeRawBufferPointer(start: pointer, count: count))
      }
    }
  }
```

But then we're still not providing the exact same new API to older versions of Swift. So perhaps my approach is the only one that works for now, even though it will require changing the call sites when we remove `>=4.1` support.

But I'll defer to your preference on this, happy to do either approach.

keith:
@thomasvl @tbkka think we could merge this so there can be a new release that supports Swift 5?

tbkka:
The following seems sufficient to emulate the Swift 5 API on Swift 4:
```
#if !swift(>=5.0)
extension Data {
  func withUnsafeBytes<T>(_ body: (UnsafeRawBufferPointer) throws -> T) rethrows -> T {
    let c = count
    return try withUnsafeBytes { (p: UnsafePointer<UInt8>) throws -> T in
      try body(UnsafeRawBufferPointer(start: p, count: c))
    }
  }
  mutating func withUnsafeMutableBytes<T>(_ body: (UnsafeMutableRawBufferPointer) throws -> T) rethrows -> T {
    let c = count
    return try withUnsafeMutableBytes { (p: UnsafeMutablePointer<UInt8>) throws -> T in
      try body(UnsafeMutableRawBufferPointer(start: p, count: c))
    }
  }
}
#endif
```

Note that every place we used the old `withUnsafeBytes` APIs, we always used it with `UInt8`, so this is actually correct for us.

Using the above, I've experimentally switched a couple of places to use the Swift 5 APIs -- the result seems to work correctly on both Swift 4 and Swift 5.  In the process, I did see some spots where the resulting code could probably be simplified, but I think such improvements can wait until after we get a clean build.

alanzeino:
> The following seems sufficient to emulate the Swift 5 API on Swift 4:
> 
> ```
> #if !swift(>=5.0)
> extension Data {
>   func withUnsafeBytes<T>(_ body: (UnsafeRawBufferPointer) throws -> T) rethrows -> T {
>     let c = count
>     return try withUnsafeBytes { (p: UnsafePointer<UInt8>) throws -> T in
>       try body(UnsafeRawBufferPointer(start: p, count: c))
>     }
>   }
>   mutating func withUnsafeMutableBytes<T>(_ body: (UnsafeMutableRawBufferPointer) throws -> T) rethrows -> T {
>     let c = count
>     return try withUnsafeMutableBytes { (p: UnsafeMutablePointer<UInt8>) throws -> T in
>       try body(UnsafeMutableRawBufferPointer(start: p, count: c))
>     }
>   }
> }
> #endif
> ```
> Note that every place we used the old `withUnsafeBytes` APIs, we always used it with `UInt8`, so this is actually correct for us.
> 
> Using the above, I've experimentally switched a couple of places to use the Swift 5 APIs -- the result seems to work correctly on both Swift 4 and Swift 5. In the process, I did see some spots where the resulting code could probably be simplified, but I think such improvements can wait until after we get a clean build.

Ah I didn't think to hardcode the type. Updating now.

alanzeino:
Updated all the cases. One thing to mention, looks like the master branch does not compile. So I did this with a commit that did compile, then once I verified that it all compiled after my changes, I rebased so that the diff didn't have any conflicts.

Any plan to add a CI check to these PRs?

tbkka:
We don't have PR testing, but we do have CI for the master branch:

https://ci.swift.org/job/swift-protobuf-linux-ubuntu-16_04/

According to that, the master branch did build last night.  I'll take a look when I get a chance.

alanzeino:
> We don't have PR testing, but we do have CI for the master branch:
> 
> https://ci.swift.org/job/swift-protobuf-linux-ubuntu-16_04/
> 
> According to that, the master branch did build last night. I'll take a look when I get a chance.

If I were to guess, I'd say that master branch build uses `swift build` and not `xcodebuild build` — and the reason why this doesn't compile is because a new type (`TextFormatEncodingOptions`) was added in a new file, but the `.xcodeproj` wasn't updated to use that new file.

That also raises another question I had — should we add the .xcodeproj to the `.gitignore` and just generate it using Swift PM? Then we can add a CI job that does both `swift build` and then `swift package generate-xcodeproj && xcodebuild build -scheme SwiftProtobuf_iOS `, which would catch these project inconsistencies.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -66,22 +66,26 @@ fileprivate func unpack(contentJSON: Data,
...
-    value = try scanner.skip()
-    if !scanner.complete {
-      // If that wasn't the end, then there was another key,
-      // and WKTs should only have the one.
-      throw AnyUnpackError.malformedWellKnownTypeJSON
+  try contentJSON.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
+    if let baseAddress = body.baseAddress, body.count > 0 {
+      let bytes = baseAddress.assumingMemoryBound(to: UInt8.self)
+
+      let buffer = UnsafeBufferPointer(start: bytes, count: contentJSON.count)
```
tbkka: Please use `body.count` here (the argument given to the closure) instead of `contentJSON.count` (which requires an additional capture)

Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -1204,7 +1207,9 @@ internal struct BinaryDecoder: Decoder {
                 // If there already was fieldData, decode it.
                 if let data = fieldData {
                     var wasDecoded = false
-                    try data.withUnsafeBytes { (pointer: UnsafePointer<UInt8>) in
+                    try data.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
+                      if let baseAddress = body.baseAddress, body.count > 0 {
+                        let pointer = baseAddress.assumingMemoryBound(to: UInt8.self)
                         var extDecoder = BinaryDecoder(forReadingFrom: pointer,
                                                        count: data.count,
```
tbkka: `data.count` ==> `body.count`

Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
```diff
@@ -66,15 +66,19 @@ extension Google_Protobuf_Any {
...
-          if !textDecoder.complete {
-            throw TextFormatDecodingError.trailingGarbage
+        try data.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
+          if let baseAddress = body.baseAddress, body.count > 0 {
+            let bytes = baseAddress.assumingMemoryBound(to: UInt8.self)
+
+            var textDecoder = try TextFormatDecoder(
+              messageType: Google_Protobuf_Any.self,
+              utf8Pointer: bytes,
+              count: data.count,
```
tbkka: `data.count` ==> `body.count`

Sources/SwiftProtobuf/JSONEncoder.swift
```diff
@@ -333,7 +333,10 @@ internal struct JSONEncoder {
...
         data.append(asciiDoubleQuote)
         if value.count > 0 {
-            value.withUnsafeBytes { (p: UnsafePointer<UInt8>) in
+            value.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
+              if let baseAddress = body.baseAddress, body.count > 0 {
+                let p = baseAddress.assumingMemoryBound(to: UInt8.self)
+
                 var t: Int = 0
                 var bytesInGroup: Int = 0
                 for i in 0..<value.count {
```
tbkka: `value.count` ==> `body.count`

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -848,15 +850,18 @@ internal struct JSONScanner {
...
-          let end = buffer.endIndex
-          if let u = try parseBareUInt64(source: buffer,
-                                         index: &index,
-                                         end: end) {
-            if index == end {
-              return u
+          (body: UnsafeRawBufferPointer) -> UInt64? in
+          if let baseAddress = body.baseAddress, body.count > 0 {
+            let bytes = baseAddress.assumingMemoryBound(to: UInt8.self)
+            let buffer = UnsafeBufferPointer(start: bytes, count: raw.count)
```
tbkka: `raw.count` => `body.count`

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -907,15 +912,18 @@ internal struct JSONScanner {
...
-          let end = buffer.endIndex
-          if let s = try parseBareSInt64(source: buffer,
-                                         index: &index,
-                                         end: end) {
-            if index == end {
-              return s
+          (body: UnsafeRawBufferPointer) -> Int64? in
+          if let baseAddress = body.baseAddress, body.count > 0 {
+            let bytes = baseAddress.assumingMemoryBound(to: UInt8.self)
+            let buffer = UnsafeBufferPointer(start: bytes, count: raw.count)
```
tbkka: `raw.count` => `body.count`

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -971,16 +979,19 @@ internal struct JSONScanner {
...
-            if let d = try parseBareDouble(source: buffer,
-                                           index: &index,
-                                           end: end) {
-              let f = Float(d)
-              if index == end && f.isFinite {
-                return f
+            (body: UnsafeRawBufferPointer) -> Float? in
+            if let baseAddress = body.baseAddress, body.count > 0 {
+              let bytes = baseAddress.assumingMemoryBound(to: UInt8.self)
+              let buffer = UnsafeBufferPointer(start: bytes, count: raw.count)
```
tbkka: `raw.count` ==> `body.count`

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -1042,15 +1053,18 @@ internal struct JSONScanner {
...
-            let end = buffer.endIndex
-            if let d = try parseBareDouble(source: buffer,
-                                           index: &index,
-                                           end: end) {
-              if index == end {
-                return d
+            (body: UnsafeRawBufferPointer) -> Double? in
+            if let baseAddress = body.baseAddress, body.count > 0 {
+              let bytes = baseAddress.assumingMemoryBound(to: UInt8.self)
+              let buffer = UnsafeBufferPointer(start: bytes, count: raw.count)
```
tbkka: `raw.count` ==> `body.count`

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -104,12 +108,15 @@ extension Message {
...
-        var decoder = BinaryDecoder(forReadingFrom: pointer,
-                                    count: data.count,
-                                    options: options,
-                                    extensions: extensions)
-        try decoder.decodeFullMessage(message: &self)
+      try data.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
+        if let baseAddress = body.baseAddress, body.count > 0 {
+          let pointer = baseAddress.assumingMemoryBound(to: UInt8.self)
+          var decoder = BinaryDecoder(forReadingFrom: pointer,
+                                      count: data.count,
```
tbkka: `data.count` ==> `body.count`

Sources/SwiftProtobuf/Message+JSONAdditions.swift
```diff
@@ -89,21 +89,25 @@ extension Message {
...
-        if let message = try (Self.self as! _CustomJSONCodable.Type)
-          .decodedFromJSONNull() {
-          self = message as! Self
-        } else {
-          throw JSONDecodingError.illegalNull
+    try jsonUTF8Data.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
+      if let baseAddress = body.baseAddress, body.count > 0 {
+        let bytes = baseAddress.assumingMemoryBound(to: UInt8.self)
+
+        let buffer = UnsafeBufferPointer(start: bytes, count: jsonUTF8Data.count)
```
tbkka: `jsonUTF8Data.count` ==> `body.count`

Sources/SwiftProtobuf/Message+JSONArrayAdditions.swift
```diff
@@ -91,14 +91,19 @@ extension Message {
...
       var array = [Self]()
-      let buffer = UnsafeBufferPointer(start: bytes, count: jsonUTF8Data.count)
-      var decoder = JSONDecoder(source: buffer, options: options)
-      try decoder.decodeRepeatedMessageField(value: &array)
-      if !decoder.scanner.complete {
-        throw JSONDecodingError.trailingGarbage
+
+      if let baseAddress = body.baseAddress, body.count > 0 {
+        let bytes = baseAddress.assumingMemoryBound(to: UInt8.self)
+        let buffer = UnsafeBufferPointer(start: bytes, count: jsonUTF8Data.count)
```
tbkka: `jsonUTF8Data.count` ==> `body.count`

Sources/SwiftProtobuf/Message+TextFormatAdditions.swift
```diff
@@ -70,14 +70,17 @@ extension Message {
...
-                                              extensions: extensions)
-          try decodeMessage(decoder: &decoder)
-          if !decoder.complete {
-            throw TextFormatDecodingError.trailingGarbage
+        try data.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
+          if let baseAddress = body.baseAddress, body.count > 0 {
+            let bytes = baseAddress.assumingMemoryBound(to: UInt8.self)
+            var decoder = try TextFormatDecoder(messageType: Self.self,
+                                                utf8Pointer: bytes,
+                                                count: data.count,
```
tbkka: `data.count` ==> `body.count`

Sources/SwiftProtobuf/TextFormatEncoder.swift
```diff
@@ -263,36 +263,40 @@ internal struct TextFormatEncoder {
 
     mutating func putBytesValue(value: Data) {
         data.append(asciiDoubleQuote)
-        value.withUnsafeBytes { (p: UnsafePointer<UInt8>) in
+        value.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
+          if let baseAddress = body.baseAddress, body.count > 0 {
+            let p = baseAddress.assumingMemoryBound(to: UInt8.self)
+
             for i in 0..<value.count {
```
tbkka: `value.count` ==> `body.count`

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
```diff
@@ -77,11 +77,14 @@ internal struct TextFormatEncodingVisitor: Visitor {
 
   mutating func visitUnknown(bytes: Data) throws {
       if options.printUnknownFields {
-          try bytes.withUnsafeBytes { (p: UnsafePointer<UInt8>) -> () in
+          try bytes.withUnsafeBytes { (body: UnsafeRawBufferPointer) -> () in
+            if let baseAddress = body.baseAddress, body.count > 0 {
+              let p = baseAddress.assumingMemoryBound(to: UInt8.self)
               var decoder = BinaryDecoder(forReadingFrom: p,
                                           count: bytes.count,
```
tbkka: `bytes.count` ==> `body.count`

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
```diff
@@ -107,7 +110,10 @@ internal struct TextFormatEncodingVisitor: Visitor {
               encoder.emitFieldNumber(number: tag.fieldNumber)
               var bytes = Internal.emptyData
               try decoder.decodeSingularBytesField(value: &bytes)
-              bytes.withUnsafeBytes { (p: UnsafePointer<UInt8>) -> () in
+              bytes.withUnsafeBytes { (body: UnsafeRawBufferPointer) -> () in
+                if let baseAddress = body.baseAddress, body.count > 0 {
+                  let p = baseAddress.assumingMemoryBound(to: UInt8.self)
+
                   var testDecoder = BinaryDecoder(forReadingFrom: p,
                                                   count: bytes.count,
```
tbkka: `bytes.count` ==> `body.count`

tbkka:
Does your branch pass the test suite?  (E.g. `make test`?)

I'm seeing failures here in the following three tests:
  `SwiftProtobufTests.Test_JSON testMultipleFields`
  `SwiftProtobufTests.Test_JSON_Array testTwoObjectsWithMultipleFields`
  `SwiftProtobufTests.Test_Map_JSON testMapInt32Bytes`

I find it interesting that the failing cases all relate to JSON and all of them have a `bytes` field which is empty.

SwiftProtobuf.xcodeproj/project.pbxproj
```diff
@@ -288,6 +288,10 @@
 		DB2E0AFF1EB25D1D00F59319 /* Message+JSONArrayAdditions.swift in Sources */ = {isa = PBXBuildFile; fileRef = DB2E0AFD1EB25D1D00F59319 /* Message+JSONArrayAdditions.swift */; };
 		DB2E0B001EB25D1D00F59319 /* Message+JSONArrayAdditions.swift in Sources */ = {isa = PBXBuildFile; fileRef = DB2E0AFD1EB25D1D00F59319 /* Message+JSONArrayAdditions.swift */; };
 		DB2E0B011EB25D1D00F59319 /* Message+JSONArrayAdditions.swift in Sources */ = {isa = PBXBuildFile; fileRef = DB2E0AFD1EB25D1D00F59319 /* Message+JSONArrayAdditions.swift */; };
+		E7038A24224D7F1000B68775 /* Data+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = E7038A23224D7F1000B68775 /* Data+Extensions.swift */; };
```
thomasvl: Fixed this in #859, so when you update just make sure it isn't being added twice.

thomasvl: Oops, fixed TextFormatEncodingOptions.swift in #859, you still need this change.

thomasvl:
> That also raises another question I had — should we add the `.xcodeproj` to the `.gitignore` and just generate it using Swift PM? Then we can add a CI job that does both swift build and then `swift package generate-xcodeproj && xcodebuild build -scheme SwiftProtobuf_iOS` , which would catch these project inconsistencies.

Since SwiftPM doesn't really do iOS, and to better support users wanting Carthage, etc.; I think we end up needing a project file checked in. 🙁 The `Makefile` does have a bunch of Xcode related targets (`test-xcode*`), but I don't think most people bother to check them often.

alanzeino:
> Does your branch pass the test suite? (E.g. `make test`?)
> 
> I'm seeing failures here in the following three tests:
> `SwiftProtobufTests.Test_JSON testMultipleFields`
> `SwiftProtobufTests.Test_JSON_Array testTwoObjectsWithMultipleFields`
> `SwiftProtobufTests.Test_Map_JSON testMapInt32Bytes`
> 
> I find it interesting that the failing cases all relate to JSON and all of them have a `bytes` field which is empty.

I should've taken your advice and just converted a few cases just to be safe. Unfortunately I have a bit of a problem in that I'm starting at a new employer next week that might not want me working on Open Source so how about for now I just land the back–ported extension in `Data+Extensions.swift` and convert one other place for reference, and I'll leave the rest to you since it would be a more careful strategy to do them piecemeal?

tbkka:
If you can fix the style issues I pointed out earlier, I'll see if I can debug that test failure.

alanzeino:
> If you can fix the style issues I pointed out earlier, I'll see if I can debug that test failure.

Okay!

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -190,8 +190,9 @@ private func parseBytes(
     // a closing double quote.
     index = digitsStart
     try value.withUnsafeMutableBytes {
-        (dataPointer: UnsafeMutablePointer<UInt8>) in
-        var p = dataPointer
+        (body: UnsafeMutableRawBufferPointer) in
+      if let baseAddress = body.baseAddress, body.count > 0 {
```
tbkka: The `value` here is actually the output buffer; the code here assumes that we're going to scan the input even if the output is zero-sized.  In particular, the close quote gets consumed deep inside the scan loop.  Since the code now skips that loop for zero-byte output, we end up not consuming the close quote for `""`.

I agree this is a somewhat confusing design.  I think the best fix would change how we handle the close quote:
* Look for two `case asciiDoubleQuote` below and remove `source.formIndex(after: &index)` at those points.
* Add `source.formIndex(after: &index)` just before `return value` at the end of the function.
This moves the close-quote handling outside of the `withUnsafeMutableBytes` which ensures that we skip it even if the output is empty.

alanzeino: Ah, makes total sense. Making those changes now. Thanks!

tbkka:
Looks good.  I'm going to run some more tests before I merge this.

alanzeino:
Cool! Lemme know if there's more changes I should make.

tbkka:
I ran tests on Ubuntu with Swift 4.0 through a 5.0 beta and it all looks good.  Thanks for working through this!

PR #845, opened 2019-03-21
Update some FileIo calls to FileHandle.

mpetrov:
FileHandle.sandardInput and FileHandle.standardOutput are part of
Foundation, and are available in Swift 3.2.

### DIFF
--- a/Sources/protoc-gen-swift/FileIo.swift
+++ b/Sources/protoc-gen-swift/FileIo.swift
@@ -20,9 +20,6 @@ import Foundation
   import Darwin.C
 #endif
 
-// Alias clib's write() so Stdout.write(bytes:) can call it.
-private let _write = write
-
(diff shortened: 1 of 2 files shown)


Sources/protoc-gen-swift/FileIo.swift
```diff
@@ -46,36 +46,8 @@ class Stderr {
 
 class Stdout {
   static func print(_ s: String) { printToFd(s, fd: 1) }
-  static func write(bytes: Data) {
-    bytes.withUnsafeBytes { (p: UnsafePointer<UInt8>) -> () in
-      _ = _write(1, p, bytes.count)
```
tbkka: Can we also delete the `_write` definition above?

mpetrov: 👍 Done.

PR #847, opened 2019-03-21
Drop the support for Swift <4.0

thomasvl:


thomasvl:
This probably shouldn't be merged/rebased until after 5 GMs just incase.

PR #848, opened 2019-03-21
Swift5.0: Change Data(bytes:) to Data(_:)

tbkka:
This is a pretty simple global substitution to stop using the deprecated `Data(byte:)` initializer.

The recommended alternative `Data(_:)` works with Swift 4 and later, so we don't need any compiler conditionals; we can just change the code.

Note:  I've not tested the performance impacts; I believe the `Data(byte:)` initializer was faster at one time.  Hopefully that's been addressed.


tbkka:
With this, the only remaining Swift 5.0 issue is an API change for `Data.withUnsafeBytes` and `Data.withUnsafeMutableBytes`.  That's being investigated in #843.

tbkka:
Testing:  I just verified (on Ubuntu 16.04) that this builds and passes the test suite using Swift 4.0, 4.0.2, 4.0.3, 4.1.2, 4.2.3, and the 5.0 snapshot from March 10.

thomasvl:
> Note: I've not tested the performance impacts; I believe the Data(byte:) initializer was faster at one time. Hopefully that's been addressed.

We likely also need to look at some of the JSON/TextFormat stuff because of the `String` changes.

PR #856, opened 2019-04-01
Adding JSON encoding option: Use proto field names instead of lowerCamelCase names

trad55:
https://developers.google.com/protocol-buffers/docs/proto3#json_options


Sources/SwiftProtobuf/JSONEncodingOptions.swift
```diff
@@ -18,5 +18,8 @@ public struct JSONEncodingOptions {
   /// Always print enums as ints. By default they are printed as strings.
   public var alwaysPrintEnumsAsInts: Bool = false
 
+  /// Always print proto field name instead of JSON(lowerCamelCase) name. By default they are printed as JSON name.
+  public var alwaysPrintProtoFieldNames: Bool = false
```
thomasvl: `alwaysPrint` doesn't seem right here, as it seems like if you weren't print the field names you'd maybe use the field numbers instead?

The C++ seems to call this [`preserve_proto_field_names`](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/util/json_util.h#L81) ([CHANGES.txt entry](preserve_proto_field_names))

So maybe some naming more along those lines so it better indicates it is picking between the two string forms?

Tests/SwiftProtobufTests/Test_JSONEncodingOptions.swift
```diff
@@ -101,4 +101,86 @@ class Test_JSONEncodingOptions: XCTestCase {
 
   }
 
+  func testAlwaysPrintProtoFieldNames() {
+    var jsonNames = JSONEncodingOptions()
+    jsonNames.alwaysPrintProtoFieldNames = false
+    jsonNames.alwaysPrintEnumsAsInts = true
```
thomasvl: Why the enums off the default also?

trad55:
@thomasvl Thank you for your comments.
I fixed problems and resolved conflicts by rebasing.

thomasvl:
@tbkka lgtm, naming ok with you?

thomasvl:
Thanks!

PR #870, opened 2019-05-02
Raise a .illegalNull is the top level JSON is null.

### LINKED ISSUE #869: ci.swift.org: JSON parser should reject top-level `null`
Swift CI caught another issue with a newly-added conformance check.  It appears that we should be rejecting bare JSON `null` as a top-level object, rather than parsing it as an empty object as we do today.

See https://ci.swift.org/job/swift-protobuf-linux-ubuntu-14_04 build 1056 and later.
```
06:59:17 CONFORMANCE TEST BEGIN ====================================
06:59:17 
06:59:17 ERROR, test=Required.Proto3.JsonInput.RejectTopLevelNull: Should have failed to parse, but didn't. request=json_payload: "null" requested_output_format: JSON message_type: "protobuf_test_messages.proto3.TestAllTypesProto3" test_category: JSON_TEST, response=json_payload: "{}"
06:59:17 
06:59:17 These tests failed.  If they can't be fixed right now, you can add them to the failure list so the overall suite can succeed.  Add them to the failure list by running:
06:59:17   ./update_failure_list.py Sources/Conformance/failure_list_swift.txt --add failing_tests.txt
06:59:17 
06:59:17   Required.Proto3.JsonInput.RejectTopLevelNull
06:59:17 
06:59:17 CONFORMANCE SUITE FAILED: 839 successes, 0 skipped, 0 expected failures, 1 unexpected failures.
```

thomasvl:
This still allows it for the things that can custom support decoding from
null, but the test isn't completely clear on what should happen for those.

Fixes #869

### DIFF
--- a/Sources/SwiftProtobuf/Message+JSONAdditions.swift
+++ b/Sources/SwiftProtobuf/Message+JSONAdditions.swift
@@ -104,6 +104,8 @@ extension Message {
           } else {
             throw JSONDecodingError.illegalNull
           }
+        } else {
+          throw JSONDecodingError.illegalNull
         }
         if !decoder.scanner.complete {
           throw JSONDecodingError.trailingGarbage


thomasvl:
> Wow. That was even simpler than I expected.

Yea, I did a lot of searching the code to confirm what calls to that vs. message fields, and the only other thing than developer code that will hit that is decoding an `Any`, and that seems like it should also fail until something officially says otherwise.

Issue #752, opened 2018-05-29
Revisit float/double to string transforms (JSON and TextFormat)


thomasvl:
@tbkka has been landing some code to the Swift StandardLib to speed up the float and double formatting.  `Float.description` should now avoid the heap, but `Double.description` does still use the heap.

We should revisit current releases with our perf harness to see if we can move our code into a conditional block for old versions of Swift, to eventually phase it out.

tbkka:
Resolved via #882.

PR #882, opened 2019-06-04
Perf: Use new float formatting code

tbkka:
This addresses #752 -- it simplifies the code to format float and double numbers in JSON and TextFormat by exploiting new facilities in the Swift standard library.  This is a significant performance boost (up to 4x) for floating point fields in JSON and Text format when using Swift 4.2 or later.

This version also works "correctly enough" in Swift 4.1 and earlier:  it will frequently emit more digits than are strictly necessary, but the result will still round-trip correctly.  Performance-wise, this version is a little slower in Swift 4.1 than the previous code.


Sources/SwiftProtobuf/DoubleParser.swift
```diff
@@ -0,0 +1,63 @@
+// Sources/SwiftProtobuf/DoubleParser.swift - Generally useful mathematical functions
+//
+// Copyright (c) 2014 - 2017 Apple Inc. and the project authors
```
thomasvl: 2019?

Sources/SwiftProtobuf/DoubleParser.swift
```diff
@@ -0,0 +1,63 @@
...
+///
+// -----------------------------------------------------------------------------
+
+import Foundation
+
+/// Support parsing float/double values from UTF-8
+internal class DoubleParser {
+    #if swift(>=4.1)
+      private var work =
+          UnsafeMutableRawBufferPointer.allocate(byteCount: 128,
```
thomasvl: Document how/why 128 was chosen?

Sources/SwiftProtobuf/JSONEncoder.swift
```diff
@@ -179,12 +178,7 @@ internal struct JSONEncoder {
                 append(staticText: "\"Infinity\"")
             }
         } else {
-            if let v = Int64(exactly: Double(value)) {
-                appendInt(value: v)
-            } else {
-                let formatted = doubleFormatter.floatToUtf8(value)
-                data.append(contentsOf: formatted)
-            }
+            data.append(contentsOf: value.debugDescription.utf8)
```
thomasvl: Are there any assurances that `debugDescription` won't change again causing problems?  My guess is this means we get the behavior of the version of the Swift runtime installed on a machine, so this could change behaviors out from under and app in the future.

tbkka: The relevant aspects should never change.  The current behavior is pretty strictly defined (and tested).  For finite values:
 1. Round-trip correct (parsing back yields exactly the same result)
 2. Fewest significant digits that can guarantee the above

NaNs and infinities are handled earlier in this code, so these guarantees suffice.

Fortunately, the first constraint above was true of the previous implementation, but any future implementation should guarantee both of these properties.

thomasvl: Is any of that documented as guaranteed?  Since we're talking about it being different for a single build of the app when deployed to different OS versions, it seems like we need a documented guarantee for developers to trust the library with the library depending on this.

tbkka: Round-trip correctness is absolutely guaranteed -- that was the whole point of the new formatting code, after all.  However, it looks like we didn't get the documentation updated; I'll find someone who can fix that.

Sources/SwiftProtobuf/DoubleParser.swift
tbkka: Good catch.  I updated copyright on all touched files.

Sources/SwiftProtobuf/DoubleParser.swift
tbkka: Done.

tbkka: After looking at this, I added another sanity check to guard against this being called with a zero-length (or negative-length) buffer.

Sources/SwiftProtobuf/JSONEncoder.swift
thomasvl: Any word on the docs update here, I saw https://twitter.com/_inside/status/1138478471590633473 making the rounds today which makes me nervous on relying on these sorta things.

stephentyrone: @thomasvl I don't know what an NSData API has to do with a standard library API; in any event it's absolutely guaranteed that .debugDescription will round trip T -> String -> T, for T any of the standard library floating-point formats. This has been the case since Swift 3 or so (I can look up the exact point at which it landed if you need it). The intention is that any user-defined formats conforming to `FloatingPoint` should also satisfy this requirement, but we (Apple) can't directly enforce that.

There's a bug tracking updating the documentation to make this explicit, but this is entirely uncontroversial.

stephentyrone: That said, there's still lots of room for variation in formatting, even guaranteeing round-tripping. Is your goal to have a single canonical format that never changes? We *could* pin down the exact behavior of [debug]Description for floating-point stdlib types if there's a compelling case for doing so, but that's a decision that hasn't been made yet. I think @tbkka and I are broadly in agreement about what that format would be, and I think the core team would OK it, but it's a step beyond what we've specified so far.

thomasvl: The concern/issue is without some documented proof saying it won't change, it is like folks that got broken be relying on the behavior of `NSData`; while it _worked_ it was never guaranteed to always have that result.  I'd sorta expect anything tagged as `debug` is even less likely to be stable.

tbkka: I don't think we need proof that it won't change _at all_, just that the properties we're relying on here won't change.  Specifically, we need that the output is parseable by JSON decoders and round-trip accurate.  I don't think there's any controversy about those properties; as Steve points out, `debugDescription` provided that much long before my changes.  My recent changes did change the output, but did so by strengthening the existing guarantees and providing better performance.

thomasvl: Sorry, yes, the round tripping is the important part, and that's where I wasn't seeing any docs.

As long as the protobuf spec/conformance tests don't decide to get more specific around significant digits, etc. we should be good as long as the Swift sides are nailed down.

stephentyrone: To me, "suitable for debugging" implicitly requires that it be round-trippable; if a conversion to String discards information, it is not suitable for debugging. That said, yes, we can absolutely document that more clearly.

stephentyrone: (I would also observe that the documentation can say whatever it wants, but tests are what enforce the behavior. And the tests that @tbkka added are pretty restrictive--it would be quite difficult for an implementation that did not guarantee round-trip results to pass them.)

Issue #816, opened 2018-12-10
Investigate byte buffer based apis


thomasvl:
Swift GRPC (via swiftnio?) is having to turn their buffers into `Data` object to then parse the protos.  It might make sense to see if we can extend the parsing apis (binary atleast, but maybe also json and TextFormat), to make use of [SE-0237](https://github.com/apple/swift-evolution/blob/master/proposals/0237-contiguous-collection.m) so other projects can bridge their own storage types into something that swift protobuf can parse without having to do extra copies.

In their, the MutableCollection part of SE-0237 might also work for writing out message, but I'm not completely sure how worth it would be since they likely can extra from the `Data` pretty efficiently and likely have to for other things.

allevato:
From a Foundation session at WWDC: `Data` is always only going to be contiguous. They've introduced the new `DataProtocol` and `MutableDataProtocol` and provide many of the same `Data` operations but don't promise contiguity.

It looks like [NIOFoundationCompat](https://github.com/apple/swift-nio/blob/master/Sources/NIOFoundationCompat/ByteBuffer-foundation.swift#L141-L149) has a `ByteBufferView` that conforms to `DataProtocol & ContiguousBytes`.

So, it may be worth figuring out how to make our serialization methods generic over `DataProtocol` and also provide fast-paths over `DataProtocol & ContiguousBytes`.

ddunbar:
There is also [`init(bytesNoCopy:count:deallocator:)`](https://developer.apple.com/documentation/foundation/data/1780455-init) if it is synchronous.

FranzBusch:
Just going to hijack this issue, we are going to investigate how we can get `swift-protobuf` off of `Foundation`, more specifically off of `Data`. We want to enable easy interop with various bag of bytes types. Right now we are only toying around with some ideas how to achieve this but I wanted to give you all a heads up.

cc @tbkka @thomasvl

thomasvl:
On thing I've debating in the past was to do something like the C++ Coded(Input|Output)Streams to wrap part of the serialization, move the existing plumbing on top of those. That might be a way to provide a _non `Data`_ interface for that part of the api.  There's still the issue/question of what to do for the `bytes` fields, but the two spaces might be worth looking at independently.

FranzBusch:
[...]

PR #883, opened 2019-06-06
Allow messages to be (de/)serialized from Unsafe[Mutable]RawBufferPointer

glbrntt:
This change allows users who already have a buffer allocated to
serialize a message straight into that buffer, avoiding an additional
copy via Data.

The motivation behind this is that in Swift gRPC the currency for
storing bytes is NIOs `ByteBuffer`, and we'd like to avoid having
an additional copy from going via `Data`. As such this change
- makes `serializedDataSize` public and renames it `serializedBinarySize`
- adds a `serializeBinary(into:partial:)` method, and
- adds an `init` to `Message` to read from an `UnsafeRawBufferPointer`.

A couple of errors in the documentation were also fixed.

I also saw issue #816 after implementing this so more than happy to
hear what everyone has to think about this!

### DIFF
--- a/Sources/SwiftProtobuf/Message+BinaryAdditions.swift
+++ b/Sources/SwiftProtobuf/Message+BinaryAdditions.swift
@@ -31,32 +31,58 @@ extension Message {
     if !partial && !isInitialized {
       throw BinaryEncodingError.missingRequiredFields
     }
-    let requiredSize = try serializedDataSize()
+    let requiredSize = try serializedBinarySize()
     var data = Data(count: requiredSize)
     try data.withUnsafeMutableBytes { (body: UnsafeMutableRawBufferPointer) in
-      if let baseAddress = body.baseAddress, body.count > 0 {
-        let pointer = baseAddress.assumingMemoryBound(to: UInt8.self)
+      let bytesWritten = try self.serializeBinary(into: body, partial: partial)
 
-        var visitor = BinaryEncodingVisitor(forWritingInto: pointer)
-        try traverse(visitor: &visitor)
-        // Currently not exposing this from the api because it really would be
-        // an internal error in the library and should never happen.
-        assert(requiredSize == visitor.encoder.distance(pointer: pointer))
-      }
+      // Currently not exposing this from the api because it really would be
+      // an internal error in the library and should never happen.
+      assert(requiredSize == bytesWritten)
     }
     return data
   }
 
   /// Returns the size in bytes required to encode the message in binary format.
-  /// This is used by `serializedData()` to precalculate the size of the buffer
-  /// so that encoding can proceed without bounds checks or reallocation.
-  internal func serializedDataSize() throws -> Int {
+  ///
+  /// This should be used with `serializeBytes(into:partial:)` so that enough space may
+  /// be allocated for the buffer so that encoding can proceed without bounds checks
+  /// or reallocation.
+  public func serializedBinarySize() throws -> Int {
     // Note: since this api is internal, it doesn't currently worry about
     // needing a partial argument to handle proto2 syntax required fields.
     // If this become public, it will need that added.
     var visitor = BinaryEncodingSizeVisitor()
     try traverse(visitor: &visitor)
     return visitor.serializedSize
+  }
+
+  /// Writes the message in the Protocol Buffer binary format into the given buffer.
+  ///
+  /// - Parameters:
+  ///   - buffer: An `UnsafeMutableRawBufferPointer` into which the serialized message
+  ///     will be written. The buffer should be at least `serializedDataSize()` bytes
+  ///     in size.
+  ///   - partial: If `false` (the default), this method will check
+  ///     `Message.isInitialized` before encoding to verify that all required
+  ///     fields are present. If any are missing, this method throws
+  ///     `BinaryEncodingError.missingRequiredFields`.
+  /// - Returns: The number of bytes written into the buffer.
+  /// - Throws: `BinaryEncodingError` if encoding fails.
+  @discardableResult
+  public func serializeBinary(into buffer: UnsafeMutableRawBufferPointer, partial: Bool = false) throws -> Int {
+    if !partial && !isInitialized {
+      throw BinaryEncodingError.missingRequiredFields
+    }
+    guard let baseAddress = buffer.baseAddress, buffer.count > 0 else {
+      return 0
+    }
+
+    let pointer = baseAddress.assumingMemoryBound(to: UInt8.self)
+    var visitor = BinaryEncodingVisitor(forWritingInto: pointer)
+    try traverse(visitor: &visitor)
+
+    return visitor.encoder.distance(pointer: pointer)
   }
 
   /// Creates a new message by decoding the given `Data` value containing a
@@ -96,9 +122,9 @@ extension Message {
   ///     extensions in this message or messages nested within this message's
   ///     fields.
   ///   - partial: If `false` (the default), this method will check
-  ///     `Message.isInitialized` before encoding to verify that all required
+  ///     `Message.isInitialized` after decoding to verify that all required
   ///     fields are present. If any are missing, this method throws
-  ///     `BinaryEncodingError.missingRequiredFields`.
+  ///     `BinaryDecodingError.missingRequiredFields`.
   ///   - options: The BinaryDecodingOptions to use.
   /// - Throws: `BinaryDecodingError` if decoding fails.
   public mutating func merge(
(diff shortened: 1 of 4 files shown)


Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -31,26 +31,24 @@ extension Message {
...
   /// Returns the size in bytes required to encode the message in binary format.
-  /// This is used by `serializedData()` to precalculate the size of the buffer
-  /// so that encoding can proceed without bounds checks or reallocation.
-  internal func serializedDataSize() throws -> Int {
+  ///
+  /// This should be used with `serializeBytes(into:partial:)` so that enough space may
+  /// be allocated for the buffer so that encoding can proceed without bounds checks
+  /// or reallocation.
+  public func serializedBinarySize() throws -> Int {
     // Note: since this api is internal, it doesn't currently worry about
```
glbrntt: I wasn't sure how to handle this comment: it seems like we shouldn't need to check for required fields as that's left to the serialize methods?

thomasvl:
Quick comment:

> * makes serializedDataSize public and renames it serializedBinarySize

We'd sorta been avoiding this because it lets folks fall into the trap of calling it and then fetching the `Data` anyways meaning the do extra passes over the proto graph.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
thomasvl: Probably just promote it to a public comment.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -31,26 +31,24 @@ extension Message {
     if !partial && !isInitialized {
       throw BinaryEncodingError.missingRequiredFields
     }
```
thomasvl: This should get removed, otherwise the validation is done twice because of the chaining of calls.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -59,6 +57,34 @@ extension Message {
...
+  ///     will be written. The buffer should be at least `serializedDataSize()` bytes
+  ///     in size.
+  ///   - partial: If `false` (the default), this method will check
+  ///     `Message.isInitialized` before encoding to verify that all required
+  ///     fields are present. If any are missing, this method throws
+  ///     `BinaryEncodingError.missingRequiredFields`.
+  /// - Returns: The number of bytes written into the buffer.
+  /// - Throws: `BinaryEncodingError` if encoding fails.
+  @discardableResult
+  public func serializeBinary(into buffer: UnsafeMutableRawBufferPointer, partial: Bool = false) throws -> Int {
```
thomasvl: can you wrap this like the others are, we mostly try to keep to 80c

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -109,18 +135,82 @@ extension Message {
...
-          var decoder = BinaryDecoder(forReadingFrom: pointer,
-                                      count: body.count,
-                                      options: options,
-                                      extensions: extensions)
-          try decoder.decodeFullMessage(message: &self)
-        }
+        try self.merge(body: body, extensions: extensions, options: options)
       }
     }
     if !partial && !isInitialized {
```
thomasvl: Aren't we checking twice here also?  I guess here you need it just in an `else` clause for when data was empty.

glbrntt:
Ah, okay, that makes sense. 

I suppose another option would be keeping it internal and changing `serializeBinary(into:partial:)` to something like `serializeBinary(partial: Bool = true, provider: (Int) -> UnsafeMutableRawBufferPointer)` and passing the size into the closure?

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -109,18 +135,82 @@ extension Message {
...
+  ///   - extensions: An `ExtensionMap` used to look up and decode any
+  ///     extensions in this message or messages nested within this message's
+  ///     fields.
+  ///   - partial: If `false` (the default), this method will check
+  ///     `Message.isInitialized` after decoding to verify that all required
+  ///     fields are present. If any are missing, this method throws
+  ///     `BinaryDecodingError.missingRequiredFields`.
+  ///   - options: The BinaryDecodingOptions to use.
+  /// - Throws: `BinaryDecodingError` if decoding fails.
+  public init(
```
thomasvl: Maybe group the two `init` methods together?

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -59,6 +57,34 @@ extension Message {
...
+  ///   - buffer: An `UnsafeMutableRawBufferPointer` into which the serialized message
+  ///     will be written. The buffer should be at least `serializedDataSize()` bytes
+  ///     in size.
+  ///   - partial: If `false` (the default), this method will check
+  ///     `Message.isInitialized` before encoding to verify that all required
+  ///     fields are present. If any are missing, this method throws
+  ///     `BinaryEncodingError.missingRequiredFields`.
+  /// - Returns: The number of bytes written into the buffer.
+  /// - Throws: `BinaryEncodingError` if encoding fails.
+  @discardableResult
```
thomasvl: Seems like we shouldn't add this annotation as callers should always use this value, no?

Tests/SwiftProtobufTests/TestHelpers.swift
```diff
@@ -42,6 +42,22 @@ extension PBTestHelpers where MessageTestType: SwiftProtobuf.Message & Equatable
             } catch {
                 XCTFail("Failed to decode protobuf: \(string(from: encoded))", file: file, line: line)
             }
+
+            let expectedSize = try configured.serializedBinarySize()
```
thomasvl: Please add tests that use `serializeBinary(into:)` with a few more cases:

- a buffer a little too short and a message with fields set to ensure that a proper error is thrown (looks like `BinaryEncodingError` doesn't have something for "too short", so the current code might not support this case (just assert), so it might take a lot of plumbing to get this right.
- a zero length buffer and a message with no fields set (since the binary size will be zero) and ensure it passes
- a zero length buffer and a message with fields set to ensure that does error.
- a buffer larger than needed and a message with fields set to ensure it works and reports the correct size.
- a buffer larger than needed and a message with no fields set to ensure it works and reports zero.

thomasvl:
> I suppose another option would be keeping it internal and changing serializeBinary(into:partial:) to something like serializeBinary(partial: Bool = true, provider: (Int) -> UnsafeMutableRawBufferPointer) and passing the size into the closure?

That was about the only thing I could think of, but I wasn't sure how well that would work for consumers. Would that work for grpc/swiftnio based things?

tbkka:
> `serializeBinary(partial: Bool = true, provider: (Int) -> UnsafeMutableRawBufferPointer)`

I'm not sure that works, actually.  MutableRawBufferPointers are vended only as closure arguments in Swift; they're never returned as results.  I think you'd have to do something like this:
```
   serializeBinary(
        partial: Bool,
        provider: (Int, (UnsafeMutableRawBufferPointer) throws -> Int) throws
   ) throws -> Int
```
That is, the provider is given the size of the message.  It then calls another provided function with the requested buffer pointer.

allevato:
Rather than providing an API that works with unsafe pointers, we should consider using the new data protocols in Foundation that come as part of 5.1. My comment in the [related issue](https://github.com/apple/swift-protobuf/issues/816):

> From a Foundation session at WWDC: `Data` is always only going to be contiguous. They've introduced the new `DataProtocol` and `MutableDataProtocol` and provide many of the same `Data` operations but don't promise contiguity.
> 
> It looks like [NIOFoundationCompat](https://github.com/apple/swift-nio/blob/master/Sources/NIOFoundationCompat/ByteBuffer-foundation.swift#L141-L149) has a `ByteBufferView` that conforms to `DataProtocol & ContiguousBytes`.
> 
> So, it may be worth figuring out how to make our serialization methods generic over `DataProtocol` and also provide fast-paths over `DataProtocol & ContiguousBytes`.

glbrntt:
> > `serializeBinary(partial: Bool = true, provider: (Int) -> UnsafeMutableRawBufferPointer)`
> 
> I'm not sure that works, actually. MutableRawBufferPointers are vended only as closure arguments in Swift; they're never returned as results. I think you'd have to do something like this:
> 
> ```
>    serializeBinary(
>         partial: Bool,
>         provider: (Int, (UnsafeMutableRawBufferPointer) throws -> Int) throws
>    ) throws -> Int
> ```
> 
> That is, the provider is given the size of the message. It then calls another provided function with the requested buffer pointer.

You're totally right, thanks. It feels a little clunky but I don't think that's necessarily a bad thing considering it's "unsafe".

glbrntt:
> Rather than providing an API that works with unsafe pointers, we should consider using the new data protocols in Foundation that come as part of 5.1. My comment in the [related issue](https://github.com/apple/swift-protobuf/issues/816):

Are you opposed to having an unsafe pointer API entirely? Using the new protocols with NIOs `ByteBuffer` is fine for deserialization, but I'm not sure how feasible it is for serialization. The idea of deserializing from non-contiguous `DataProtocol` is interesting though.

glbrntt:
Closing this in favour of using `DataProtocol`.

Issue #481, opened 2017-04-20
JSON parsing issues


fe9lix:
Hi! I've noticed a couple of issues when working with JSON decoding/encoding and existing backend services that work with JSON payloads (I'm aware that this is not the ideal setup for protobuffers). What am I missing...?

- When decoding a JSON model, a type mismatch in a single property causes the entire parsing to fail. Wouldn't it make more sense to just treat such properties as "missing" instead of throwing an error? Although a type mismatch indicates that the contract is violated, in most cases you'd still want to work with partial results and then just ignore incorrect properties or use a default value there. Same for enum cases: When your backend service sends you an enum name that is not listed in the proto file, the UNRECOGNIZED case is not trigged but instead an error is thrown and the entire parsing fails. It'd be nice if there was a less "stricter" version for decoding JSON models.

- When encoding the Swift struct back to JSON representation, the camelCased JSON names are used instead of the case that is used in the proto file. So if your backend works with snake_cased property names, you usually also want to use the snake_cased names when posting data back to the backend. jsonString(), however, can't be used then and constructing the payload properties must be manually done.

- Is there a specific reason why the native JSON parsing classes of iOS (JSONSerialization) aren't internally used for decoding/encoding?

thomasvl:
The formatting of the field names comes from the [protobuf spec](https://developers.google.com/protocol-buffers/docs/proto3#json) for how the JSON encoding should work.  Likewise, matching the spec and other language implementations is what dictates the behaviors for type mismatches, unknown keys, etc.  For the Swift version to properly interop with other protobuf libraries we have to have the same success/failures behaviors.  The protobuf project actually includes some [conformance tests](https://github.com/google/protobuf/tree/master/conformance) to help ensure this, some of it has to be sorted out by looking at their unittests and implementation.

Using JSONSerialization would be likely be very expensive and slower.  To generate the JSON, the protobuf object would have to be converted to a `[String:Any]` (so all new objects), that would be serialized, and then thrown away.  Likewise, for parsing that intermediate object tree would get created and thrown away.

fe9lix:
@thomasvl thanks a lot for your clarifications!

tbkka:
> Re: JSON failure

We're reluctant to stray too far from the behavior of other protobuf implementations.  Among other constraints, it's important to us that we meet Google's "conformance test" which does require JSON decoding to fail completely in certain cases.  We might consider optional behaviors for some cases, but expect spirited discussion.  ;-)

[...]

PR #896, opened 2019-07-28
Add an option to ignore unknown enum cases when decoding from JSON (instead of failing)

kzlekk:
The issue was already described in the past but looks like no PR was uploaded: [https://github.com/apple/swift-protobuf/issues/481#issuecomment-295800326](url)

I know it is unwanted behaviour but in real life scenarios backend teams sometimes coming to the necessity of extending existing enums by adding new cases. This leads to backward incompatibility in the mobile clients using swift-protobuf generated models and when backend encodes enum cases with case names. 

To avoid the issue, an option added to pass unknown enum cases and use first enum case as default case following the best practice in proto programming. Backend teams usually define first enum case as "unknown" or "initial" or any other with the intention that it will be used as a default case, or to indicate the problem.

### DIFF
--- a/Sources/SwiftProtobuf/JSONDecoder.swift
+++ b/Sources/SwiftProtobuf/JSONDecoder.swift
@@ -466,7 +466,7 @@ internal struct JSONDecoder: Decoder {
       value = nil
       return
     }
-    value = try scanner.nextEnumValue() as E
+    value = try scanner.nextEnumValue()
   }
 
   mutating func decodeSingularEnumField<E: Enum>(value: inout E) throws
@@ -475,7 +475,7 @@ internal struct JSONDecoder: Decoder {
       value = E()
       return
     }
-    value = try scanner.nextEnumValue()
+    value = try scanner.nextEnumValue() ?? E()
   }
 
   mutating func decodeRepeatedEnumField<E: Enum>(value: inout [E]) throws
@@ -488,8 +488,9 @@ internal struct JSONDecoder: Decoder {
       return
     }
     while true {
-      let e: E = try scanner.nextEnumValue()
-      value.append(e)
+      if let e: E = try scanner.nextEnumValue() {
+        value.append(e)
+      }
       if scanner.skipOptionalArrayEnd() {
         return
       }
--- a/Tests/SwiftProtobufTests/Test_Enum.swift
+++ b/Tests/SwiftProtobufTests/Test_Enum.swift
@@ -16,6 +16,7 @@
 
 import Foundation
 import XCTest
+import SwiftProtobuf
 
 class Test_Enum: XCTestCase, PBTestHelpers {
     typealias MessageTestType = Proto3Unittest_TestAllTypes
@@ -37,6 +38,23 @@ class Test_Enum: XCTestCase, PBTestHelpers {
     func testJSONrepeated() {
         assertJSONEncode("{\"repeatedNestedEnum\":[\"FOO\",\"BAR\"]}") { (m: inout MessageTestType) in
             m.repeatedNestedEnum = [.foo, .bar]
+        }
+    }
+
+    func testJSONdecodingOptions() {
+
+        var options = JSONDecodingOptions()
+        options.ignoreUnknownFields = true
+
+        assertJSONDecodeSucceeds("{\"optionalNestedEnum\":\"NEW_VALUE\"}", options: options) { (m: MessageTestType) -> Bool in
+            switch m.optionalNestedEnum {
+            case .zero: return true
+            default: return false
+            }
+        }
+
+        assertJSONDecodeSucceeds("{\"repeatedNestedEnum\":[\"FOO\",\"NEW_VALUE\"]}", options: options) { (m: MessageTestType) -> Bool in
+            m.repeatedNestedEnum == [.foo]
         }
     }
 
--- a/Sources/SwiftProtobuf/JSONScanner.swift
+++ b/Sources/SwiftProtobuf/JSONScanner.swift
@@ -1285,7 +1285,14 @@ internal struct JSONScanner {
   /// Parse the next token as a string or numeric enum value.  Throws
   /// unrecognizedEnumValue if the string/number can't initialize the
   /// enum.  Will throw other errors if the JSON is malformed.
-  internal mutating func nextEnumValue<E: Enum>() throws -> E {
+  internal mutating func nextEnumValue<E: Enum>() throws -> E? {
+    func throwOrIgnore() throws -> E? {
+      if ignoreUnknownFields {
+        return nil
+      } else {
+        throw JSONDecodingError.unrecognizedEnumValue
+      }
+    }
     skipWhitespace()
     guard hasMoreContent else {
         throw JSONDecodingError.truncated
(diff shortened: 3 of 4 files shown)


thomasvl:
This still feels sorta wrong to add a specific feature for this just to the Swift implementation.  Normally to get better support for a proto definition to expand with time, I believe the binary form is suggested because of these sorta issues with JSON.

If you do feel strongly about this, It seems like it might be worth bring up the discussion on the main proto repo (or maybe better on their mailing list) as if the functionality is useful, it shouldn't be limited to just one language.

Sources/SwiftProtobuf/JSONDecodingOptions.swift
```diff
@@ -25,5 +25,9 @@ public struct JSONDecodingOptions {
   /// ignored, an error will be raised if one is encountered.
   public var ignoreUnknownFields: Bool = false
 
+  /// If unknown enum cases in the JSON should be defaulted to initial enum case.
+  /// If they aren't ignored, an error will be raised if one is encountered.
+  public var defaultUnknownEnumCases: Bool = false
```
tbkka: I think this should be called `ignoreUnkownEnumValues`

* This is parallel to `ignoreUnknownFields`

* In proto3, ignoring a field is the same as using the default value.

tbkka:
If you can point to other protobuf implementations that support this, that would certainly help.  As you've noticed, we're _very_ conservative about adding features.

kzlekk:
I haven't found any other networking libraries which are "wrapping" SwiftProtobuf. I think the developers should be using Protobufs in conjunction with their own networking layer implementation.

I am agree with you and I hope I understand why it doesn't look right ideologically. But my reasons is practical. I have already the system working and clients are using REST via gRPC bridging to speak with the backend. Some clients can use Swagger, some other handle requests by manually parsing and building JSONs. In that case I can't force other teams to change their already well-working implementation so my decoding will not fail in my iOS client. I don't see real alternative but fix my implementation.

This is an option for those who can't force their backend to change existing encoding for REST clients.

thomasvl:
Sorry, not following the _wrapping_ comment.  I believe we were both asking from the POV of other Protobuf implementations that support something like this.  Hopefully the ones in Google's official distro.

kzlekk:
One of google/protobuf implementations contains already an option called 'ignore_unknown_enum_values' inside  'json_util.cc'.

Actual implementation is inside datapiece.cc inside DataPiece::ToEnum:

`
 // If ignore_unknown_enum_values is true an unknown enum value is ignored.
      if (ignore_unknown_enum_values) {
          *is_unknown_enum_value = true;
          return enum_type->enumvalue(0).number();
      }
 `

thomasvl:
> One of google/protobuf implementations contains already an option called 'ignore_unknown_enum_values' inside 'json_util.cc'.

That appears to be an internal detail for some of the rewriting support between json/binary formats, and isn't directly exposed, if you look at the exposed options, there doesn't seem to be anything: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/util/json_util.h

kzlekk:
@thomasvl, I am not sure I understand your point. 

There is an issue with current SwiftProtobuf implementation so it can't be used in production if your backend produces JSON with enum values encoded as names and not numbers.

thomasvl:
My point is no other proto implementation seems to provide the option you are suggesting and we prefer to stay inline with the other implementation on abilities.

The Swift impl is to spec and in line with the other libraries.  This is one of the downsides of using JSON (vs. binary) and it is usually handled by doing something like  early in your communications having clients identify a "version" they are using and then the server ensure it only sends (and expects) JSON with the constructs known that that time.

The generally problem with the "ignore" options is they are lossy and there is no way for for clients to really know when things aren't known.  The binary format has unknown fields support (and unknown enum value handing with proto3 syntax), so it is more robust and thus better handle these things.

thomasvl:
ps - if you think there is general need/interest, please bring it up on https://github.com/protocolbuffers/protobuf or their discussion group (https://groups.google.com/forum/#!forum/protobuf) to see about getting something like this generally added to multiple languages.  Having the broader feedback is the best way to help figure out if there are likely to be issues/downsides when adding support for something like this and then using it.

Tests/SwiftProtobufTests/Test_Enum.swift
```diff
@@ -40,6 +41,23 @@ class Test_Enum: XCTestCase, PBTestHelpers {
         }
     }
 
+    func testJSONIgnoreUnknownValues() {
```
tbkka, line 44: I think I'd rather have this test case in `Test_JSONDecodingOptions.swift`

Sources/SwiftProtobuf/JSONDecodingOptions.swift
```diff
@@ -25,5 +25,9 @@ public struct JSONDecodingOptions {
   /// ignored, an error will be raised if one is encountered.
   public var ignoreUnknownFields: Bool = false
 
+  /// If unknown enum cases in the JSON should be defaulted to initial enum case.
+  /// If they aren't ignored, an error will be raised if one is encountered.
+  public var ignoreUnknownEnumValues: Bool = false
+
```
tbkka, line 31: According to protocolbuffers/protobuf#7392 the C++ implementation has a single option "ignoreUnknownFields" that also causes unknown enum values to be ignored.  I'd prefer to use that here:  It keeps us aligned with the C++ implementation and also reduces the matrix of options we need to test.

Tests/SwiftProtobufTests/Test_Enum.swift
kzlekk: Done.

Sources/SwiftProtobuf/JSONDecodingOptions.swift
kzlekk: I agree, separated option specific for enums is redundant. I can hardly imagine somebody will prefer to only ignore enums but fall on other fields. Done.

kzlekk:
Good news! This should help us a lot. @tbkka, I've updated the commit as you requested.

Sources/SwiftProtobuf/JSONDecoder.swift
```diff
@@ -466,7 +466,7 @@ internal struct JSONDecoder: Decoder {
       value = nil
       return
     }
-    value = try scanner.nextEnumValue() as E
+    value = try scanner.nextEnumValue()
```
thomasvl, line 469: I think map decoding also comes thru here, so the value will be nil, and the pair won't end up in the map.  We'll need to confirm that is correct vs. using the default value.  (we can correct this on the map handling side since single field might want to say unset in this case).

Tests/SwiftProtobufTests/Test_Enum.swift
```diff
@@ -40,6 +41,23 @@ class Test_Enum: XCTestCase, PBTestHelpers {
...
+            case .zero: return true
+            default: return false
+            }
+        }
+
+        assertJSONDecodeSucceeds("{\"repeatedNestedEnum\":[\"FOO\",\"NEW_VALUE\"]}", options: options) { (m: MessageTestType) -> Bool in
+            m.repeatedNestedEnum == [.foo]
+        }
+    }
+
```
thomasvl, line 60: we'll need to check `map<*, Enum>`, and when the single field is within a `oneof`.

Sources/SwiftProtobuf/JSONDecodingOptions.swift
tbkka: Per other discussion, this has been merged into `ignoreUnknownFields`

tbkka:
@thomasvl has asked for some additional test cases.  Will you be able to implement those?

tbkka:
If you'd like to continue pursuing this, please rebase against the new `main` branch and re-submit.  Thank you!

PR #897, opened 2019-08-12
Move the generated source from descriptor.proto into the runtime libr…

### LINKED ISSUE #727: Proposal:  Move `descriptor` to SwiftProtobuf library for general use
OS: macOS 10.13.3
Xcode version: 9.3 beta 4
Swift version: 4.1
SwiftProtobuf version: 1.0.3

I previously posted this issue in the grpc-swift project ([here](https://github.com/grpc/grpc-swift/issues/179)) but MrMage referred me to the swift-protobuf project since I mistakenly assumed it was a problem with grpc-swift.

In a *.proto file we are defining a custom method option in the following way:
```
extend google.protobuf.MethodOptions {
    bool requiresAuthentication = 50001;
}
```
When generating the Swift code an import for `SwiftProtobufPluginLibrary` is missing in the generated *.pb.swift file. `Google_Protobuf_MethodOptions` is being extended in the generated code but it is being marked as an undeclared type until I manually import `SwiftProtobufPluginLibrary` in addition to the already imported `SwiftProtobuf`. This is fixing the issue but since ideally the generated code is never modified manually this is not a great solution.

Is this a known issue and/or has somebody encountered this issue before?

thomasvl:
…ary.

The runtime doesn't actually use these types/enums, but this should make things
easier for developers that try to generate from .proto files that include
options to annotations of files, fields, messages, etc. as they will no longer
have to generate the descriptor.pb.swift themselves and can just use the one
provided by the runtime.

This does *not* capture the binary descriptor info into the generated code
like some other languages do for protobuf, this just includes the types
defined in descriptor.proto.

Fixes #727


tbkka:
Interesting.  This touched more than I expected.  Was most of this automatic once you tweaked the `Makefile` and moved `descriptor.pb.swift`?  Or did you have to make other manual changes?

thomasvl:
The `Makefile` change was manual as was `Sources/SwiftProtobufPluginLibrary/SwiftProtobufInfo.swift` to mark the file as bundled now.

Then all the changes to `.pb.swift` files were automatic.

From a git pov, I had to manually remove the `descriptor.pb.swift` and add them in the new locations.

Then to get everything building again; all the edits to `Sources/SwiftProtobufPluginLibrary/*.swift` and `Tests/SwiftProtobufPluginLibraryTests/*.swift` to add `import SwiftProtobuf` were manual (since all the types from descriptor moved to the other module).

Oh, and the checked in Xcode project changes were obviously manual.

tbkka:
Thanks for elucidating.  I was just curious whether our build tools seemed to be taking care of everything I expected.

thomasvl:
> Thanks for elucidating. I was just curious whether our build tools seemed to be taking care of everything I expected.

No worries.  I admit, I made the first edits I expected, regenerated and watched things then fail to build and it took a moment to realize all the other edits needed.  Made sense once realized the issue.

basalphenaar:
Thanks for doing this, it was quite frustrating to manually generate _descriptor.pb.swift_ every time!

Issue #735, opened 2018-04-11
Revisit when to use heap based storage for messages.


thomasvl:
Spun out of discussions on #733

The current logic for deciding if a message should use heap storage is:

```swift
    // NOTE: This check for fields.count likely isn't completely correct
    // when the message has one or more oneof{}s. As that will efficively
    // reduce the real number of fields and the message might not need heap
    // storage yet.
    let useHeapStorage = isAnyMessage || descriptor.fields.count > 16 || hasSingleMessageField(descriptor: descriptor)
```

I believe the `hasSingleMessageField` is really there to handle transitive recursion back to this same message (protos making a tree structure).

thomasvl:
I believe the only case where we have to use storage is if there is a non repeated field that is transitively recursive.

Any other limits are about how large we want to let the struct before we push it into the heap to try and cut down on the cost of the struct copying.  A repeated or map field already will be using some heap space, so it probably comes down to how many of those and how many primitive fields we want.

While floating ideas - we could change things, and only put the single fields in the storage.  The repeated fields and map fields already be doing similar things under the hood to avoid copies.

Lukasa:
A relevant size constraint here is that putting a value type into an existential container will allocate if the value type is wider than 3 words (24 bytes). This means any attempt to put a message into a `Message` (e.g. to create `[Message]`) will allocate if those messages are wider than 24 bytes.

The SwiftNIO team tend to use 24 bytes as the maximum size of an inline struct when there is any risk of that structure being placed into an existential container.

A related concern is reference counting: structs with multiple refcounted fields must refcount them all separately whenever their copy operation is invoked. For wide structs with many `string` or `bytes` members (both backed by refcounted structs) this can increase the cost of copying substantially compared to boxing the backing storage, which incurs only one refcount operation per copy.

thomasvl:
Interesting, don't `String`, `Array`, and `Dictionary` end up having >24 bytes per instance because of their multiple storage methods?  i.e. - a message with a single field of one of those types would trip some of these cases?

It likely means we might also need to consider `unknownFields` and `_protobuf_extensionFieldValues` in deciding when to start using Storage.

ydnar:
@Lukasa are the additional refcounts more expensive than the alloc/free operations for boxed storage?

Lukasa:
[...]

Issue #733, opened 2018-04-05
`_StorageClass` makes large arrays of non-trivial messages impractical.


archagon:
I'm trying to use protobufs with fairly large arrays of structs. These arrays receive dozens of changes a second and get serialized and deserialized every few seconds, so they must remain efficient and deterministic at all times: stored in contiguous memory, minimal heap use, structs all the way down. Unfortunately (unless I'm missing something) it seems that the current implementation of swift-protobuf is incapable of handling this case to any degree of complexity, since any nested messages automatically invoke the `_StorageClass` path and cause every element in the array to allocate on the heap. One could splay out the nested messages onto the root array, but you'd still run into the 16 field limit.

Is there any way around this?

archagon:
To clarify, the array of messages will be used as the backing store for a [string CRDT](https://github.com/archagon/crdt-playground). Every millisecond of speed is critical.

thomasvl:
There currently isn't a way to avoid the `_StorageClass` at the moment.

Even if there was, wouldn't `Array` still be likely to do a heap allocation under the hood?  i.e. - it can't put every sized array on the stack, and surely not ones with 100,000+ elements.

tbkka:
Internally, SwiftProtobuf can generate structs with or without a backing StorageClass.  The decision for when to generate a StorageClass is somewhat arbitrary at the moment; we'd appreciate input into how to make this better, especially if that input is based on reproducible performance measurements.

There are two motivations for the StorageClass design:
1. Proto allows circular references and a pure struct-based implementation cannot support that.  At one time, I did experiment with code to analyze the full tree of data structures in order to use StorageClass only when there was in fact a circular reference.  That code didn't ship simply because I wasn't comfortable that I understood all of the implications well enough to justify the complexity, but the approach could be resurrected.
2. Structs sometimes need to be copied.  For the most part, Swift does a good job of avoiding this, but copying a small wrapper struct (plus associated refcounting) is often cheaper than copying a large struct.

Unknown field handling is another complication.  At one time, proto3 did not preserve unknown fields (in Google's C++ implementation), but this was changed.  Some of the implementations have an option to suppress unknown field handling, and we could consider adding such an option.

archagon:
> Even if there was, wouldn't Array still be likely to do a heap allocation under the hood? i.e. - it can't put every sized array on the stack, and surely not ones with 100,000+ elements.

Yes, but only one — not 100,001!

[...]

PR #900, opened 2019-09-17
Use heap-based storage only when a Protobuf has a singular transitive recursion

ydnar:
Hi there—long time reader, first-time PR.

I think this addresses the issues raised in #733 and broken out into #735 by analyzing if a proto has a message field with a transitive recursion.

I broke this out into two major commits: one that [introduces the change to `protoc-gen-swift`](https://github.com/apple/swift-protobuf/pull/900/commits/9aedf6b836aeb928896bf61cb3d3f66d166efb03), and a [second commit](https://github.com/apple/swift-protobuf/pull/900/commits/a9b7ee23f52f70e0679e1d47f396bb888dc1cac1) with the updated generated Swift files. There’s a third commit that undoes a change I needed to build + run the tests on my machine using XCode 11 / Swift 5.1.

Following the instructions in the Makefile, all tests including conformance tests, passed.

Is this the right idea?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,0 @@
-{
-  "swiftformat.enable": false,
-  "files.trimTrailingWhitespace": true
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+  "swiftformat.enable": false,
+  "files.trimTrailingWhitespace": true
+}
--- a/Sources/protoc-gen-swift/MessageGenerator.swift
+++ b/Sources/protoc-gen-swift/MessageGenerator.swift
@@ -53,7 +53,7 @@ class MessageGenerator {
     // when the message has one or more oneof{}s. As that will efficively
     // reduce the real number of fields and the message might not need heap
     // storage yet.
-    let useHeapStorage = isAnyMessage || descriptor.fields.count > 16 || hasSingleMessageField(descriptor: descriptor)
+    let useHeapStorage = isAnyMessage || descriptor.fields.count > 16 || hasRecursiveField(descriptor: descriptor)
 
     oneofs = descriptor.oneofs.map {
       return OneofGenerator(descriptor: $0, generatorOptions: generatorOptions, namer: namer, usesHeapStorage: useHeapStorage)
@@ -504,6 +504,22 @@ fileprivate func hasSingleMessageField(descriptor: Descriptor) -> Bool {
     ($0.type == .message || $0.type == .group) && $0.label != .repeated
   }
   return result
+}
+
+fileprivate func hasRecursiveField(descriptor: Descriptor, visited: [Descriptor] = []) -> Bool {
+  var visited = visited
+  visited.append(descriptor)
+  return descriptor.fields.contains {
+    if $0.type != .message {
+      return false
+    }
+    guard let messageType = $0.messageType else {
+      return false
+    }
+    return visited.contains {
+      $0 === messageType
+    } || hasRecursiveField(descriptor: messageType, visited: visited)
+  }
 }
 
 fileprivate struct MessageFieldFactory {
--- a/Sources/protoc-gen-swift/MessageGenerator.swift
+++ b/Sources/protoc-gen-swift/MessageGenerator.swift
@@ -502,10 +502,16 @@ fileprivate func hasRecursiveSingularField(descriptor: Descriptor, visited: [Des
   var visited = visited
   visited.append(descriptor)
   return descriptor.fields.contains {
-    // Repeated fields use heap storage already
-    if ($0.type != .message && $0.type != .group) || $0.label == .repeated {
+    // Ignore fields that aren’t messages or groups.
+    if $0.type != .message && $0.type != .group {
       return false
     }
+
+    // Repeated fields already use heap storage (for the array).
+    if $0.label == .repeated {
+      return false
+    }
+
     guard let messageType = $0.messageType else {
       return false
     }
--- a/Sources/protoc-gen-swift/MessageGenerator.swift
+++ b/Sources/protoc-gen-swift/MessageGenerator.swift
@@ -498,14 +498,6 @@ class MessageGenerator {
   }
 }
 
-fileprivate func hasSingleMessageField(descriptor: Descriptor) -> Bool {
-  let result = descriptor.fields.contains {
-    // Repeated check also rules out maps.
-    ($0.type == .message || $0.type == .group) && $0.label != .repeated
-  }
-  return result
-}
-
 fileprivate func hasRecursiveSingularField(descriptor: Descriptor, visited: [Descriptor] = []) -> Bool {
   var visited = visited
   visited.append(descriptor)
--- a/Sources/protoc-gen-swift/MessageGenerator.swift
+++ b/Sources/protoc-gen-swift/MessageGenerator.swift
@@ -516,7 +516,7 @@ fileprivate func hasRecursiveSingularField(descriptor: Descriptor, visited: [Des
     }
 
     // Skip other visited fields.
-    if (visited[1...].contains { $0 === messageType }) {
+    if (visited.contains { $0 === messageType }) {
       return false
     }
 
--- a/Sources/protoc-gen-swift/MessageGenerator.swift
+++ b/Sources/protoc-gen-swift/MessageGenerator.swift
@@ -517,9 +517,19 @@ fileprivate func hasRecursiveSingularField(descriptor: Descriptor, visited: [Des
     guard let messageType = $0.messageType else {
       return false
     }
-    return visited.contains {
-      $0 === messageType
-    } || hasRecursiveSingularField(descriptor: messageType, visited: visited)
+
+    // We only care if the message or sub-message recurses to the root message.
+    if messageType === visited[0] {
+      return true
+    }
+
+    // Skip other visited fields.
+    if (visited[1...].contains { $0 === messageType }) {
+      return false
+    }
+
+    // Examine sub-message.
+    return hasRecursiveSingularField(descriptor: messageType, visited: visited)
   }
 }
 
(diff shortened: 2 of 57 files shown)


ydnar:
Subsequent commits (e0c482f and db85e48) restrict the recursive check to singular fields, to mirror the existing behavior which filters out repeated fields and maps.

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -506,6 +506,23 @@ fileprivate func hasSingleMessageField(descriptor: Descriptor) -> Bool {
   return result
 }
 
+fileprivate func hasRecursiveSingularField(descriptor: Descriptor, visited: [Descriptor] = []) -> Bool {
+  var visited = visited
```
thomasvl: If you don't default `visited`, and make it `inout`, can you ensure every node is only checked once?  i.e. - since it is a copy, the deeper nodes can get visited multiple times because nothing is shared between those sub visits?

ydnar: @thomasvl I guess it could be a `Set<Descriptor>`. As it stands, it short-circuits on any recursion, but doesn’t share state between branches. Are you thinking about this as an optimization?

ydnar: Is there a straightforward way to benchmark `protoc` for the test protos and compare against `master`? I guess that would answer the question.

ydnar: @thomasvl tried `inout visited` and got 19 additional heap-allocated messages:

https://github.com/ydnar/swift-protobuf/commit/11698e3cf60f434646e0e3a808e689fcac26fd8a

thomasvl: I'll have to digest this one a little more, I didn't expect there to be any change in what it decided should use the heap.

thomasvl: I think maybe I see the logic issue; `visited` is sorta doing two things.  It is tracking what we've already checked, but also trying to check for recursion, and the second part isn't completely correct.

```protobuf
message A {
  string name = 1;
  B b = 2;
}

message B {
  C c = 1;
}

message C {
  D d = 1;
}

message D {
  B b = 1;
}
```

Shouldn't `A` not need storage because it isn't recursive itself.  But `B`, `C`, and `D` all should use storage because they define a recursive sub loop?

Since `visited` is serving double duty for tracking what's been checked and if there is recursion, it can end up saying that `A` has to use storage?  So the algorithm needs to check purely if the root we are starting from is used again by a sub field, but something like `visitited` is needed to not get stuck in a sub cycle that doesn't directly impact the decision for the root message.

thomasvl: To your other questions - yes I'm concerned about performance here when the message graph gets deep/wide and not doing to much extra work.  As far as benchmarking it, the easiest way is likely to just time making the reference files, but all the files are so impl time times will likely be dwarfed by the other parts of the the overhead.

ydnar: @thomasvl thanks for discovering the logic issue. I think you’re right in your example, `A` won’t need heap storage, but `B` will.

I had originally coded this without a `visited` array, just checking if the root message (in this case `A`) was referenced by any sub-messages. I’ll try that again.

Unrelated: on my machine (2018 MB Air), using Swift 5.1, the package fails to compile because of a missing `default` statement in a generated `switch` (`generated_swift_names_enum_cases.pb.swift:2605`). I have to manually insert it (and accept the inevitable warnings) in order to get the package to build and tests to run.

thomasvl:
Re the compile failure in the generated file, can you open a sep issue for that (we already saw it on linux, but it hadn't hit macOS hosted builds yet).  We can tweak the generated files to dodge it, but the underlying problem can happen in a real proto file if it has enough enum cases, so a complete solution for that will take more thought/work.

ydnar:
Thanks. The exact error I’m seeing:

```sh
swift-protobuf/Tests/SwiftProtobufTests/generated_swift_names_enum_cases.pb.swift:1749:5: error: the compiler is unable to check that this switch is exhaustive in reasonable time
```

Does `swift` have any arguments to extend the timeout for type checking?

Sources/protoc-gen-swift/MessageGenerator.swift
ydnar: @thomasvl I revised the logic based on your findings, and the result is several fewer heap allocations: db85e488

thomasvl: Since it is hard to tell from the diffs, what is left using storage, and does it make sense?

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -53,7 +53,7 @@ class MessageGenerator {
     // when the message has one or more oneof{}s. As that will efficively
     // reduce the real number of fields and the message might not need heap
     // storage yet.
-    let useHeapStorage = isAnyMessage || descriptor.fields.count > 16 || hasSingleMessageField(descriptor: descriptor)
```
thomasvl: looks like hasSingleMessageField didn't get removed (since it isn't called any more)

ydnar: Fixed.

Sources/protoc-gen-swift/MessageGenerator.swift
ydnar: The remaining heap allocations look like they makes sense. I’d love to see it stress-tested by someone hitting performance limits with heap allocs like #733.

thomasvl:
fyi - #904 opened for the general issue with large enum. #906 landed to avoid the problem with our testing file generated in the tree.  So you likely want to sync and rebase these changes off that.

ydnar:
@thomasvl I rebased off current master (post #904), re-ran `make build regenerate test-runtime reference build test`. Regenerated code produced no diffs, and all tests passed.

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -498,12 +498,31 @@ class MessageGenerator {
...
+      return false
+    }
+
+    // We only care if the message or sub-message recurses to the root message.
+    if messageType === visited[0] {
+      return true
+    }
+
+    // Skip other visited fields.
+    if (visited[1...].contains { $0 === messageType }) {
```
thomasvl: Actually, from a performance pov, you don't need this slice.  `visted[0]` was checked in the conditional above, so just check all of `visitited` and avoid making the temp object, etc.

ydnar: Ah, got it. I’m used to @golang where slicing is cheap and on the stack (shared storage with original slice).

allevato: `visited[1...]` returns an `ArraySlice` that shares storage (and indices) with its original array, so it's the same in that regard.

ydnar: @allevato so is it correct to assume that sub slice doesn’t result in any additional heap allocations?

(I’m trying to understand if I should change it per @thomasvl’s suggestion).

thomasvl: My thought was there is a call to make the slice (which may or may not have an alloc to its internal layout).  So that extra call up front (with potential allocs) is all to avoid one extra `===` check within the actually `contains(where:)` loop, so it didn't seem like trying to skip the first one was worth it compare to the cost of that one extra compare.

ydnar: Roger that. Updated in 1dafa90.

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+  "swiftformat.enable": false,
+  "files.trimTrailingWhitespace": true
+}
```
thomasvl: oops.  :)

ydnar: Gah. Oops

ydnar: On that note, why doesn’t this repo use `swiftformat`?

thomasvl: Which _swiftformat_, the new one in the Apple repo?  There isn't yet an general style guide the community has adopted, so we haven't bothered to chase one yet (just like different parts of Swift have different formats at the moment).

ydnar: Great question. We’ve been using the non-Apple `swiftformat`, not the new `swift-format`. A blessed standard format is one of the best aspects of working with Go.

ydnar:
Any thoughts?

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -498,12 +498,31 @@ class MessageGenerator {
...
-    // Repeated check also rules out maps.
-    ($0.type == .message || $0.type == .group) && $0.label != .repeated
+fileprivate func hasRecursiveSingularField(descriptor: Descriptor, visited: [Descriptor] = []) -> Bool {
+  var visited = visited
+  visited.append(descriptor)
+  return descriptor.fields.contains {
+    // Repeated fields use heap storage already
+    if ($0.type != .message && $0.type != .group) || $0.label == .repeated {
+      return false
+    }
```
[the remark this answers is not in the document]
ydnar: @tbkka pushed at 5edbfff

Sources/protoc-gen-swift/MessageGenerator.swift
tbkka: Solving this optimally is tricky:  In @thomasvl's example above, we technically only _need_ storage for one of `B`, `C`, and `D`.  You don't need all three.

Some quick digging suggests that the optimal answer requires solving something called the "maximal acyclic subgraph" problem.  That sounds like more work than I really want to tackle in this PR.

thomasvl:
> My only real concern with this is whether the larger structs we're generating will cause problems for some people. The best way to find out is to merge this and solicit feedback.

I'll go head and merge this now so we can try to get the feedback.

thomasvl:
@ydnar thanks again for the work and patch!

ydnar:
Oh, cool! Thanks @thomasvl @tbkka for the review + feedback!

PR #902, opened 2019-09-25
Fix "documenation" typo in generated header.

glbrntt:
In the header of all generated files the message pointing to the
SwiftProtobuf repo has a typo: "documenation" should be "documentation".

This change fixes the typo in FileGenerator.swift and updates the
generated and reference protobufs.


Sources/protoc-gen-swift/FileGenerator.swift
```diff
@@ -56,7 +56,7 @@ class FileGenerator {
             "// Generated by the Swift generator plugin for the protocol buffer compiler.\n",
             "// Source: \(fileDescriptor.name)\n",
             "//\n",
-            "// For information on using the generated types, please see the documenation:\n",
+            "// For information on using the generated types, please see the documentation:\n",
```
glbrntt: "Real" code change is here:

Issue #904, opened 2019-09-26
Enum generation can make code that causes new Swift 5.1 error message


thomasvl:
A .proto file with an enum with _**a lot**_ of cases ends up needing to generate `switch` statements with a lot of `case`s. As of Swift 5.1 this can cause the compiler to actually error out.

SwiftProtobuf sees this itself with one of the `generated_swift_names_*.swift` files:

```
Tests/SwiftProtobufTests/generated_swift_names_enum_cases.pb.swift:1749:5: error: the compiler is unable to check that this switch is exhaustive in reasonable time
    switch self {
    ^
Tests/SwiftProtobufTests/generated_swift_names_enum_cases.pb.swift:1749:5: note: do you want to add a default clause?
    switch self {
    ^
```

On the near/short term, we need to do something to tweak what the validation generates to work around this, but…

Longer term, we likely need to see if something can be generated differently/annotate the source in some way to avoid this problem.

thomasvl:
https://bugs.swift.org/browse/SR-7907 appears to be the upstream bug for this.  @allevato is likely going to ping that also.

thomasvl:
Capturing it since we've been discussing it locally -

- If we add a `default` all the time, then the compiler does kick out a warning when it is able to safely check the `switch` in time.
- Adding a `default` also means some degree of code bloat because if the compiler can't tell everything was covered, it won't realized whatever code we put into the `default` is never hit and thus won't dead strip it.

thomasvl:
Also, just switching this test file to proto2 syntax doesn't make it _work_, there are enough cases even without the one having an associated value to still cause a the issue.

allevato:
Jordan Rose recommended filing a new bug since the original one was resolved, so let's track https://bugs.swift.org/browse/SR-11533 now.

thomasvl:
So about the only thing we seem to be able to come up with is do something like:

```swift
  var rawValue: Int {
    switch self {
    case .case0: return 0
    case .case1: return 1
    case .case2: return 2
    case .case3: return 3
    case .case4: return 4
    case .case5: return 5
    // ...
    case .caseN: return N
    default: // nothing
  }
    switch self {
    case .caseN_1: return N+1
    case .caseN_2: return N+2
    case .caseN_3: return N+3
    case .caseN_4: return N+4
    case .caseN_5: return N+5
    // ...
    case .case2N: return 2*N
    default: // nothing
  }
  // ...
[...]

PR #906, opened 2019-09-26
Generate one enum per enum case for the testing.

thomasvl:
This is a workaround for #904, the underlying problem is still there, but this
makes the generated source file for testing "work" in that it doesn't hit the
compile failure. A real solution is still needed for large enums.


Makefile
```diff
@@ -461,10 +461,11 @@ Protos/generated_swift_names_enum_cases.proto: Protos/mined_words.txt
 	@echo '// Swift compile errors are probably bugs in protoc-gen-swift' >> $@
 	@echo 'syntax = "proto3";' >> $@
 	@echo 'package protobuf_unittest_generated;' >> $@
-	@echo 'enum GeneratedSwiftReservedEnum {' >> $@
-	@echo '  NONE = 0;' >> $@
-	@cat Protos/mined_words.txt | awk 'BEGIN{n = 1} {print "  " $$1 " = " n ";"; n += 1 }' >> $@
-	@echo '}' >> $@
+	@#echo 'enum GeneratedSwiftReservedEnum {' >> $@
```
tbkka: We should have a really clear comment here explaining why these lines are commented out and what needs to be done to un-comment them.

thomasvl: Oops, meant to do that, thanks, added.

Issue #861, opened 2019-04-03
json string with default value 


seanliu1:
Developing using iOS12, Swift 5. I am about to add an extension which can support to output fields with their default values.  I just want to check whether it is already implemented. 

Based on proto doc, it looks like 
`Emit fields with default values: Fields with default values are omitted by default in proto3 JSON output. An implementation may provide an option to override this behavior and output fields with their default values.`

I wonder does swift version has option to output fileds with their default values. I found python version has it `MessageToJson(message, including_default_value_fields=False)`

https://developers.google.com/protocol-buffers/docs/reference/python/google.protobuf.json_format-module

thomasvl:
The supported JSON encoding options: https://github.com/apple/swift-protobuf/blob/master/Sources/SwiftProtobuf/JSONEncodingOptions.swift

thomasvl:
p.s. - adding this as an option would be tricky since JSON encoding is built on the visitor pattern which is where it is skipping default values.

seanliu1:
@thomasvl thanks for your quick reply. Yes, I realized that default value is skipped in the generated code. 
```
  func traverse<V: SwiftProtobuf.Visitor>(visitor: inout V) throws {
    if !self.cards.isEmpty {
      try visitor.visitRepeatedMessageField(value: self.homeCards, fieldNumber: 1)
    }
    if self.test != false {
      try visitor.visitSingularBoolField(value: self.test, fieldNumber: 2)
    }
    try unknownFields.traverse(visitor: &visitor)
  }

```

tbkka:
Is there a way to do this without breaking the API between generated code and the library?

I see how old generated code could continue to work with an updated library, but I don't see how new generated code (with modified default handling) could work with an old library.

If we can't preserve this compatibility, we would probably want to increment the major version.   It's been a while since 1.0, so this is probably okay.  We should certainly consider whether there are other issues that might require API breakage that should be adopted along with this.

seanliu1:
What I found is that current we do the check in codegen. 
```
  func traverse<V: SwiftProtobuf.Visitor>(visitor: inout V) throws {
    if self.showID != 0 {
      try visitor.visitSingularUInt32Field(value: self.showID, fieldNumber: 1)
    }
    try unknownFields.traverse(visitor: &visitor)
  }

```
 What if we add an option to JSONEncodingOptions, then 

we change 

```
  mutating func visitSingularUInt32Field(value: UInt32, fieldNumber: Int) throws {

[...]

PR #907, opened 2019-09-27
[861] Add default option for JSON decoding  

seanliu1:
https://github.com/apple/swift-protobuf/issues/861

See discussion in the ticket. This PR adds a new method in Visitor Protocol. However this PR did not address backward compatibility issue 

We opened this PR a while ago, we currently use this PR in our codebase.

```
    let shouldIncludeDefault = visitor.shouldIncludeDefault()
    if !self.text.isEmpty || shouldIncludeDefault {
      try visitor.visitSingularStringField(value: self.text, fieldNumber: 1)
    }
    if self.action != .searchCancelEventActionInvalid || shouldIncludeDefault {
      try visitor.visitSingularEnumField(value: self.action, fieldNumber: 2)
    }
    if self.resultCount != 0  || shouldIncludeDefault {
      try visitor.visitSingularUInt32Field(value: self.resultCount, fieldNumber: 3)
    }
  }
```


Sources/SwiftProtobuf/Visitor.swift
```diff
@@ -690,4 +692,8 @@ extension Visitor {
   public mutating func visitExtensionFields(fields: ExtensionFieldValueSet, start: Int, end: Int) throws {
     try fields.traverse(visitor: &self, start: start, end: end)
   }
+    
+    func shouldIncludeDefault() -> Bool {
```
thomasvl: Indent is off compared to the rest of the file, and this likely should get a comment since unlike the rest of these, it isn't remapping to another call, it is providing a default implementation.

Sources/SwiftProtobuf/Visitor.swift
```diff
@@ -443,6 +443,8 @@ public protocol Visitor {
 
   /// Called with the raw bytes that represent any unknown fields.
   mutating func visitUnknown(bytes: Data) throws
+    
+ func shouldIncludeDefault() -> Bool
```
thomasvl: Needs documentation since it is public interface a `Visitor` needs to know about.

Sources/protoc-gen-swift/FieldGenerator.swift
```diff
@@ -51,6 +51,8 @@ protocol FieldGenerator {
   /// Generate any support needed to this field's value is initialized.
   /// The generated code should return false if it isn't set.
   func generateIsInitializedCheck(printer: inout CodePrinter)
+
+  func shouldGenerateIncludeDefault() -> Bool
```
thomasvl: Needs docs.

Sources/SwiftProtobuf/JSONEncodingOptions.swift
```diff
@@ -22,5 +22,8 @@ public struct JSONEncodingOptions {
   /// By default they are converted to JSON(lowerCamelCase) names.
   public var preserveProtoFieldNames: Bool = false
 
+  /// Whether to include proto field with default value
+  public var includeDefaultValue: Bool = false
```
thomasvl: If I'm reading the code right, this only impacts proto3 syntax files, yes?  (does that match the other languages that have this option)? If so, it should be documented as such (and likely renamed to better capture what it means)

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -328,6 +328,13 @@ class MessageGenerator {
 
       var ranges = descriptor.extensionRanges.makeIterator()
       var nextRange = ranges.next()
+      var shouldGenerateInclude = false
```
thomasvl: This can be sped up since you only need to loop until you find one, something like:

```swift
      if fieldsSortedByNumber.contains { $0.shouldGenerateIncludeDefault } {
        p.print("let shouldIncludeDefault = visitor.shouldIncludeDefault()\n")
      }
```

should do it.

thomasvl:
So the approach here doesn't seem to end up with what was suggested on #861.  Specifically, traverse is "polling" the visitor for it if should visit defaults, which like means new generated code will fail to compile if building against the old runtime.

@tbkka do we want to go more in the direction of the original issue or move ahead with this approach instead?

tbkka:
> .. new generated code will fail to compile if building against the old runtime.

I wonder if we need a way to test this?  I think it's important that code generated by 1.x compile against library 1.y regardless of which is newer.  It would be easy to test that new library builds against old generated code (just have a directory with copies of old generated code), but the other way around seems trickier.  Hmmmm...

If we need to break compatibility, then I think we need to discuss a major version bump.

FlamingHairball:
Hello!

We need this functionality for a project we're working on and for obvious reasons we'd prefer to standardize on it rather than re-implement it. I can update the PR to address the comments but we'd still be left with a breaking change -- are we okay with that and if so, what steps should I take to move that forward? Just bumping the version number in the specs and any comments?

tbkka:
I believe it's possible to support this without breaking compatibility.  See https://github.com/apple/swift-protobuf/issues/861#issuecomment-491442253 for details.

Otherwise, this would have to wait until we are ready to release SwiftProtobuf 2.0.  We have no current plans for such a release, so I cannot estimate when that might be.

FlamingHairball:
@tbkka interesting idea, I explored it a bit but it doesn't seem 100% backwards-compatible in the case of new framework + old lib.

You need an updated protoc-gen-swift lib in order to generate the appropriate `traverse` calls, but the new framework would accept a JSONDecodingOptions with `printDefaultFields` set true, even though an old lib wouldn't generate the necessary logic -- so in this scenario it would fail silently.

Am I missing something/is that an acceptable breakage? It seems *mostly* implausible (although e.g if two team members were working on same project with different versions of `protoc-gen-swift` it gets more plausible)

tbkka:
> @tbkka interesting idea, I explored it a bit but it doesn't seem 100% backwards-compatible in the case of new framework + old lib.
>

You are right that someone who doesn't update both the generated code and the library would be unable to use the new feature and that attempts to use it would fail silently.

This is regrettable but I don't think it's a fatal flaw.  Remember that most people are not using your new feature, so it's much more important that the code compiles and all _old_ features work correctly when people are using mis-matched versions.

PR #910, opened 2019-09-27
Version bump for doing a release.

thomasvl:


thomasvl:
Doing a release to get the current changes out and then we will look at landing some of the current PRs which have the potential for more impact and letting them bake on master before the next release.

PR #912, opened 2019-11-07
Fix typo in auto-genned comment

FlamingHairball:


tbkka:
We do not directly edit the auto-generated files.  Those files are routinely rebuilt from their original sources; any direct edits will be lost.

Instead, please submit a PR to correct the master copy of this comment, which is located here:

https://github.com/apple/swift-protobuf/blob/b2baf41e0d62cccda25d12e18e9b1660756a8d4a/Sources/protoc-gen-swift/FileGenerator.swift#L184

Once that is fixed, the comment will automatically get updated in all of the generated files the next time we rebuild them.

FlamingHairball:
whoops! should've looked closer, thank you!

PR #913, opened 2019-11-07
Fix typo

FlamingHairball:


tbkka:
Looks good.  Before we merge this, please go through the following steps.  (Note:  this will require you to have a checkout of Google's protobuf project adjacent to your SwiftProtobuf checkout.)

* `make regenerate` in the top level project to rebuild the generated files used by the library test harness
* `git diff` and verify that the _only_ changes to the generated files are the expected ones
* `make test` should now succeed until it gets to the "Reference" check (which will fail)
* `make reference` to update the reference files
* `git diff Reference` to check that the _only_ changes to the Reference files are the expected ones
* `make test` should now succeed entirely
* Commit the regenerated files as a new commit to your existing `patch-2` branch
* Push the new commit (please don't squash)

The above should take less than an hour (most of which is waiting on protobuf to build).  Once that's done, I'll review the results and merge it.

Thank you!

FlamingHairball:
Thank you for the detailed guide!
I had to `brew install automake` and `brew install libtool` as well in order to get protos to compile. Other than that everything checks out. Let me know if you need anything else!

tbkka:
Looks good!  Thank you for regenerating and going through the extra tests.

PR #915, opened 2019-11-19
Make things generic over ContiguousBytes and @inlinable.

thomasvl:
- Gets better codegen.
- Move some of the internal details of merge into a help so things don't also
  have to be exposed outside the module.
- Also make the versions take Data @inlinable for consistency.


thomasvl:
@Lukasa fyi.

Lukasa:
This LGTM, nice change!

PR #916, opened 2019-11-20
JSON/TextFormat submessage encoding performance

tbkka:
On my local system, this change yields a 5x speedup for JSON and TextFormat encoding with:
```
Performance/perf_runner.sh 100 'repeated message'
```

The test with 100 message fields (not repeated) shows better than 2.5x speedup with this change.

Previously, JSON and TextFormat handled submessages by creating a completely new encoder, encoding the sub message, then appending the result to the outer encoder.  Now, they work the same way as the binary encoder, reusing the same encoder in nested submessages and avoiding the extra data copies.

TODO: The test above also demonstrates problems with decoding array fields in general (due to incremental appends) and with binary encoding of submessages (due to the extra recursive traversal to accumulate size).

### DIFF
--- a/Sources/SwiftProtobuf/JSONEncodingVisitor.swift
+++ b/Sources/SwiftProtobuf/JSONEncodingVisitor.swift
@@ -166,8 +166,19 @@ internal struct JSONEncodingVisitor: Visitor {
 
   mutating func visitSingularMessageField<M: Message>(value: M, fieldNumber: Int) throws {
     try startField(for: fieldNumber)
-    let json = try value.jsonUTF8Data(options: options)
-    encoder.append(utf8Data: json)
+    if let m = value as? _CustomJSONCodable {
+      let json = try m.encodedJSONString(options: options)
+      encoder.append(text: json)
+    } else if let newNameMap = (M.self as? _ProtoNameProviding.Type)?._protobuf_nameMap {
+      encoder.startNestedObject()
+      let oldNameMap = self.nameMap
+      self.nameMap = newNameMap
+      try value.traverse(visitor: &self)
+      self.nameMap = oldNameMap
+      endObject()
+    } else {
+      throw JSONEncodingError.missingFieldNames
+    }
   }
 
   mutating func visitSingularGroupField<G: Message>(value: G, fieldNumber: Int) throws {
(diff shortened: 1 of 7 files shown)


Sources/SwiftProtobuf/JSONEncodingVisitor.swift
```diff
@@ -274,12 +285,35 @@ internal struct JSONEncodingVisitor: Visitor {
...
+      for v in value {
+        if comma {
+          encoder.comma()
+        }
+        comma = true
+        encoder.startNestedObject()
+        try v.traverse(visitor: &self)
+        encoder.endObject()
+      }
+      self.nameMap = oldNameMap
```
thomasvl: Could do this with a `defer` right after swapping it so it is less likely to get missed in any other changes.

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
```diff
@@ -220,15 +237,27 @@ internal struct TextFormatEncodingVisitor: Visitor {
...
       if let any = value as? Google_Protobuf_Any {
-          any.textTraverse(visitor: &visitor)
+          any.textTraverse(visitor: &self)
       } else {
-          try! value.traverse(visitor: &visitor)
+          try! value.traverse(visitor: &self)
       }
-      encoder = visitor.encoder
       encoder.endMessageField()
+      // Restore state
```
thomasvl: defer to keep it close to where you swapped it in?

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
```diff
@@ -63,6 +63,23 @@ internal struct TextFormatEncodingVisitor: Visitor {
...
+          bytes.append(contentsOf: buff)
+      } else if let extensionName = extensions?[fieldNumber]?.protobufExtension.fieldName {
+          bytes.append(UInt8(ascii: "["))
+          bytes.append(contentsOf: extensionName.utf8)
+          bytes.append(UInt8(ascii: "]"))
+      } else {
+          bytes.append(contentsOf: fieldNumber.description.utf8)
+      }
+      return bytes
+  }
```
thomasvl: So this is duplicating logic that is split between the next method and the encoder, which seems like it will let bugs sneak in.  Seems like we should figure out how to these two approaches so the logic isn't duplicated.

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
```diff
@@ -372,18 +401,31 @@ internal struct TextFormatEncodingVisitor: Visitor {
...
+      let oldNameResolver = self.nameResolver
+      let oldExtensions = self.extensions
+      // Update encoding state for new message type
+      self.nameMap = (M.self as? _ProtoNameProviding.Type)?._protobuf_nameMap
+      self.nameResolver = [:]
+      self.extensions = (value as? ExtensibleMessage)?._protobuf_extensionFieldValues
+      // Iterate and encode each message
       for v in value {
-          emitFieldName(lookingUp: fieldNumber)
+          encoder.emitFieldName(name: fieldName)
```
thomasvl: Seems like all repeated handlings could use the speedup of calcing the name once?

Sources/SwiftProtobuf/JSONEncodingVisitor.swift
tbkka: Good catch!  Done.

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
tbkka: I went ahead and did this for all the repeated handlings, though I haven't actually been able to measure any difference.

The real reason I need it for repeated message fields is not really for performance but because of the context switch; the field name has to be looked up against the outer message.  Otherwise, I would have to repeat the context switch for every entry.  (Though honestly, that might not be too bad performance-wise; the context switch is pretty light.)

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
tbkka: Good point.  I don't see a good way to do this right now, but I'll think about it for a day and see what occurs to me.

Given the paltry performance gains from caching the name lookup, another option would be to just ditch the `formatFieldName` entirely.

PR #917, opened 2019-11-20
Add Message.with to the performance suite

Lukasa:
This patch adds `Message.with` to the tasks that are benchmarked in the performance suite.

This work was originally intended as a precursor to a patch that would add `@inlinable` to `Message.with`. However, as #763 exists I'd rather hold off and let the project consider that work more holistically. That said, the extra benchmark does have value in and of itself so I'm proposing it here.

I've added the benchmark without an equivalent in C++-land as there is no natural equivalent.

Below is a screenshot of the result as run on my machine:

Screenshot 2019-11-20 at 09 55 00

### DIFF
--- a/Performance/generators/swift.sh
+++ b/Performance/generators/swift.sh
@@ -98,6 +98,10 @@ extension Harness {
         populateFields(of: &message)
       }
 
+      message = measureSubtask("Populate fields with with") {
+        return populateFieldsWithWith()
+      }
+
       // Exercise binary serialization.
       let data = try measureSubtask("Encode binary") {
         return try message.serializedData()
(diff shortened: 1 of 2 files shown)


thomasvl:
Aside while I think of it - At some point dropping the `with` and just doing a trailing closure would likely be an api more developers expect, that just wasn't the case when this first go added.

allevato:
> Aside while I think of it - At some point dropping the `with` and just doing a trailing closure would likely be an api more developers expect, that just wasn't the case when this first go added.

Before Swift 5.1 I felt this way too, but now I wonder if people would expect something that looks like `TypeName { ... }` to be a function builder, or if there's an expectation that function builders and initializers-with-trailing-closures-that-pass-the-new-value-as-`$0` are equally common patterns.

At the time we first designed the API I don't think there was a prevailing practice. I still don't know if there is—I usually lean on the standard library to provide these patterns.

PR #918, opened 2019-11-21
Use raw pointers instead of typed pointers.

Lukasa:
In general, it is better to use raw pointers instead of typed pointers
whenever those pointers are provided to you by external users. The
reason for this is that Swift requires that each pointer be bound to one
and only one data type at any point in the program execution. If you are
vended a raw pointer (e.g. from a Data), your program does not
unequivocally know what the type of that pointer may be, and so to
assume that it is bound to UInt8 or Int8 is to risk violating Swift's
pointer aliasing rules.

We've learned lessons in the past that violating the pointer aliasing
rules in programming languages can lead to disaster. In our case, we
can avoid the majority of this trouble by electing to use raw pointers
instead of typed pointers. This also helps us by removing tricky pointer
management code.

This patch contains essentially no functional change: wherever possible I
made straightforward, targeted refactors. I think once or twice I turned
a pointer into a buffer pointer, but I generally resisted getting too
frisky with changes.


atrick:
@Lukasa I'm *so* happy to see this. Thanks for doing the work.

I'm guessing you wanted to be explicit anyway, but `fromByteOffset:` defaults to zero.

It looks like you needed to add a subscript to workaround code that still uses standalone pointers instead of buffer pointers. I'm not sure if there's a good reason the code doesn't use buffer pointers.

Lukasa:
There isn’t, but I wanted to make this patch easier to merge by avoiding making larger changes. I plan to follow up with a patch that uses buffer pointers as a next step, when hopefully these shims can be removed.

Sources/SwiftProtobuf/UnsafeRawPointer+Shims.swift
```diff
@@ -0,0 +1,39 @@
...
+// https://github.com/apple/swift-protobuf/blob/master/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+///
+/// Shims for UnsafeRawPointer and friends.
+///
+// -----------------------------------------------------------------------------
+
+
+extension UnsafeRawPointer {
```
tbkka: I've been burned in the past extending standard system types like this.  Inevitably, they end up conflicting with something.

If you plan to remove these in a subsequent PR, please comment them as such.

Lukasa: This extension _should_ be safe, because even if the standard library added these methods these extensions would still take priority over the standard library within the module. As the extended methods do not add _public_ API to the types, nothing bad should happen.

Nonetheless, I do want to remove these, so I've updated them with comments indicating that they are planned to be removed.

Sources/SwiftProtobuf/BinaryEncoder.swift
```diff
@@ -19,30 +19,35 @@ import Foundation
...
+        pointer.storeBytes(of: byte, as: UInt8.self)
+        pointer = pointer.advanced(by: 1)
     }
 
     private mutating func append(contentsOf data: Data) {
-        let count = data.count
-        data.copyBytes(to: pointer, count: count)
-        pointer = pointer.advanced(by: count)
+        data.withUnsafeBytes { dataPointer in
+            if let baseAddress = dataPointer.baseAddress, dataPointer.count > 0 {
```
thomasvl: the next helper used a local for count to avoid fetching it a few times, any reason to do it there and not here?

Sources/SwiftProtobuf/BinaryEncoder.swift
```diff
@@ -123,8 +128,8 @@ internal struct BinaryEncoder {
         let count = value.utf8.count
```
thomasvl: Does it make sense to attempt to use withContiguousStorageIfAvailable to fast path the whole string in here and then fall back to this loop if we can't?

Sources/SwiftProtobuf/DoubleParser.swift
```diff
@@ -21,46 +21,39 @@ internal class DoubleParser {
...
     // necessary, so we size this generously.
-    #if swift(>=4.1)
-      private var work =
-          UnsafeMutableRawBufferPointer.allocate(byteCount: 128,
-                                                 alignment: MemoryLayout<UInt8>.alignment)
-    #else
-      private var work = UnsafeMutableRawBufferPointer.allocate(count: 128)
-    #endif
+    private var work = 
+      UnsafeMutableBufferPointer<Int8>.allocate(capacity: 128)
```
thomasvl: Just checking, no swift conditional needed now?

Sources/SwiftProtobuf/BinaryEncoder.swift
Lukasa: Yeah, `UnsafeRawBufferPointer.count` has [been `inlinable` since Swift 4.0 at least](https://github.com/apple/swift/blob/3416108abd0df3997a533a14c5ad06de20ba8a60/stdlib/public/core/UnsafeRawBufferPointer.swift.gyb#L498-L504), so the compiler should observe that the load is trivial. This was not necessarily true of `Data.count`, and so storing the local to avoid the re-load was potentially worth it.

I can restore the local if you would like, but I wouldn't expect a meaningful performance difference.

Sources/SwiftProtobuf/BinaryEncoder.swift
Lukasa: Yes, this is a good enhancement. For the sake of not boiling the ocean I'd like not to make it here: may I suggest filing an issue for this? It could even be marked contributor friendly and a new contributor could take a swing at it, as it should be easy enough to do.

Sources/SwiftProtobuf/DoubleParser.swift
Lukasa: That's correct, allocating typed pointers has never required an alignment specification as the alignment will naturally be `MemoryLayout<Element>.alignment`.

Sources/SwiftProtobuf/BinaryEncoder.swift
thomasvl: It was more the consistence of the code that was what I noticed.

Sources/SwiftProtobuf/BinaryEncoder.swift
thomasvl: Opened.

Sources/SwiftProtobuf/UnsafeRawPointer+Shims.swift
tbkka: Thanks for adding the comment.

To clarify my concern:  You're right that Swift's scoping rules should prevent this from actually breaking any code (unlike when people try this in Javascript or Obj-C), but it's still an avoidable point of confusion.  People new to this code base would likely assume these methods were the ones from the standard library and be confused by any behavioral differences.

Lukasa: I totally agree. If you’d prefer not to have these shims at all I can replace all the usage sites with `store` and `load`.

Sources/SwiftProtobuf/BinaryEncoder.swift
Lukasa: I’m happy to be guided by your preferences here.

Sources/SwiftProtobuf/UnsafeRawPointer+Shims.swift
tbkka: I'm okay merging this as-is as long as it's clearly commented as a wart to be removed.

thomasvl:
@tbkka did you want to look at this?

thomasvl:
Hm, I tried to fix the conflict in the github ui, but I guess it didn't work. ☹️

@Lukasa can you sync/rebase?

thomasvl:
Oh wait, I just can't rebase to land it, I can squash though.

PR #924, opened 2019-12-11
Avoid collecting unknown fields while scanning for message.

thomasvl:
When generating TextFormat for unknown fields, the attempt to see if a length
delimited field is a submessage was building up the unknown data pointlessly,
instead allow it to be forced to be discarded to skip this overhead.


thomasvl:
Total tunnel vision here, we create the decoder in the calling function, so just set it up from the start.

PR #927, opened 2019-12-12
Guard against TextFormat encoding of unknowns from overflowing the stack

thomasvl:


thomasvl:
> This makes a lot of sense. It looks like the recursion here to dissect potential submessages would yield quadratic performance in the depth of the tree, so the limit here could be a significant performance win (in addition to limiting stack consumption). As a follow-up, would it make sense to add an option to completely suppress dissecting unknown sub messages?

Not sure I follow the perf concern.  For each length delimited, it would walk the tags (not looking for sub messages) and if it works, then expand the message; and in doing the expansion, it would attempt to do that for sub messages.  But at each level, it isn't fully recursive.

Options - There already is a bool to not include unknowns; and TextFormat isn't really a good wire format in the first place because you can't read back in unknowns (you don't know something is a group or message when seen in TextFormat; and I'm not sure all platforms are specific enough on fixed vs. varint output to tell those apart). Anyway, we could add it, just not sure what it really buys folks.

tbkka:
Ah, yes.  You're right, at each level, the length-delimited fields get skipped entirely, so those bytes don't even get scanned, much less parsed.

PR #934, opened 2020-01-03
Use dictionary subscript with default.

thomasvl:
https://developer.apple.com/documentation/swift/dictionary/2894528-subscript
says it goes back to Xcode 9, so at this point, our min Swift version should be
ensuring things always support it.

The only other place that it might make sense to use this appears to be within
the SimpleExtensionMap, but the values looked up get filtered before adding to
them, so it seems like the fast path might for not being in the map might still
be worth while.

### DIFF
--- a/Sources/SwiftProtobufPluginLibrary/SwiftProtobufNamer.swift
+++ b/Sources/SwiftProtobufPluginLibrary/SwiftProtobufNamer.swift
@@ -86,9 +86,9 @@ public final class SwiftProtobufNamer {
   private func computeRelativeNames(enum e: EnumDescriptor) {
     let stripper = NamingUtils.PrefixStripper(prefix: e.name)
 
-    /// Determine the initial canidate name for the name before
+    /// Determine the initial candidate name for the name before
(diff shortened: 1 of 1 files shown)


Sources/SwiftProtobufPluginLibrary/SwiftProtobufNamer.swift
```diff
@@ -103,14 +103,7 @@ public final class SwiftProtobufNamer {
...
       let canidate = canidateName(enumValue)
-
-      if var existing = canidates[canidate] {
-        existing.append(enumValue)
-        canidates[canidate] = existing
-      } else {
-        canidates[canidate] = [enumValue]
-      }
-
+      canidates[canidate, default:[]].append(enumValue)
```
tbkka: This is a nice improvement.

While you're here, I wouldn't mind if you fixed the spelling of `candidate`.  ;-)

thomasvl: Done

PR #935, opened 2020-01-03
Minor code cleanup for JSON additions

tbkka:
Normally, `jsonString()` delegates to `jsonUTF8Data()`, which relies
on our efficient general JSON encoder.  But when the message type
has a custom JSON encoding, that encoding is produced directly as a String,
so the delegation unnecessarily translated string->data->string.

While here, make `init(jsonUTF8Data:options:)` a bit easier to read.

### DIFF
--- a/Sources/SwiftProtobuf/Message+JSONAdditions.swift
+++ b/Sources/SwiftProtobuf/Message+JSONAdditions.swift
@@ -93,6 +93,7 @@ extension Message {
   ) throws {
     self.init()
     try jsonUTF8Data.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
+      // Empty input is valid for binary, but not for JSON.
       guard body.count > 0 else {
         throw JSONDecodingError.truncated
       }
--- a/Sources/SwiftProtobuf/Message+JSONAdditions.swift
+++ b/Sources/SwiftProtobuf/Message+JSONAdditions.swift
@@ -28,6 +28,9 @@ extension Message {
   public func jsonString(
     options: JSONEncodingOptions = JSONEncodingOptions()
   ) throws -> String {
+    if let m = self as? _CustomJSONCodable {
+      return try m.encodedJSONString(options: options)
+    }
     let data = try jsonUTF8Data(options: options)
     return String(data: data, encoding: String.Encoding.utf8)!
   }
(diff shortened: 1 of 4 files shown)


Sources/SwiftProtobuf/Message+JSONAdditions.swift
```diff
@@ -92,17 +95,15 @@ extension Message {
     try jsonUTF8Data.withUnsafeBytes { (body: UnsafeRawBufferPointer) in
       if body.count > 0 {
```
thomasvl: In rereading this, I'm wondering if `guard...else` to have _early outs_ instead of all the nesting might also be a better solution.

tbkka: Good point.  While making this change, I noticed that the guard here was being handled differently for `init(jsonString:)`, which dominoed into some other related changes...

tbkka:
While fixing the handling of empty JSON decoding for the data-based API, I became nervous that there might be other cases where string- and data-based JSON APIs behaved differently.

So I expanded the test helpers to run all JSON tests through both string- and data-based APIs, which should catch any future bugs along these lines.

Of course, that change in turn uncovered a test inconsistency around floating-point signed zeros.  By my reading of the spec, a negative zero should be considered a default value and not encoded in proto3.  It's easy to change this if the protobuf folks disagree.

thomasvl:
Re negative zero - go ahead and open an issue upstream asking about it.  I believe they were having some discussions internally around the floating point edge cases, so filing issues to get this clarified seems like a good idea.

Sources/SwiftProtobuf/Message+JSONAdditions.swift
```diff
@@ -90,23 +93,22 @@ extension Message {
...
-          try decoder.decodeFullObject(message: &self)
-        } else if Self.self is _CustomJSONCodable.Type {
-          if let message = try (Self.self as! _CustomJSONCodable.Type)
-            .decodedFromJSONNull() {
-            self = message as! Self
-          } else {
-            throw JSONDecodingError.illegalNull
-          }
+      guard body.count > 0 else {
+        throw JSONDecodingError.truncated
```
thomasvl: I wonder if this should get a comment since zero length makes sense for binary?

Tests/SwiftProtobufTests/TestHelpers.swift
```diff
@@ -151,6 +151,22 @@ extension PBTestHelpers where MessageTestType: SwiftProtobuf.Message & Equatable
         } catch let e {
             XCTFail("Failed to serialize JSON: \(e)\n    \(configured)", file: file, line: line)
         }
+
+        do {
+            let encodedData = try configured.jsonUTF8Data()
```
thomasvl: indent seems a little odd here.

Tests/SwiftProtobufTests/TestHelpers.swift
```diff
@@ -278,6 +317,14 @@ extension PBTestHelpers where MessageTestType: SwiftProtobuf.Message & Equatable
         } catch {
             // Yay! It failed!
         }
+
+	let jsonData = json.data(using: String.Encoding.utf8)!
```
thomasvl: Indent odd again

Tests/SwiftProtobufTests/TestHelpers.swift
```diff
@@ -213,6 +229,29 @@ extension PBTestHelpers where MessageTestType: SwiftProtobuf.Message & Equatable
...
             return
         }
+
+        do {
+	    let jsonData = json.data(using: String.Encoding.utf8)!
+            let decoded: MessageTestType = try MessageTestType(jsonUTF8Data: jsonData)
+            XCTAssert(check(decoded), "Condition failed for \(decoded) from binary \(json)", file: file, line: line)
+
+            do {
+                let encoded = try decoded.jsonUTF8Data()
```
thomasvl: indent odd

Sources/SwiftProtobuf/Message+JSONAdditions.swift
tbkka: Done.

Issue #919, opened 2019-11-22
`BinaryEncoder` look at using `withContiguousStorageIfAvailable` for `String.utf8`


thomasvl:
Currently the `BinaryEncoder` pulls over string data a byte at a time, but it looks like we could fastpath with `withContiguousStorageIfAvailable` and fall back to the loop only if need be.

RLovelett:
I'm interested in doing this one. I've even come up with with I think is a reasonable first attempt at a patch.

The part that might be unexpected and not great is the modification to `append(contentsOf:)`. I made that function return a value to silence the `Constant 'isAvailable' inferred to have type '()?', which may be unexpected`.

```patch
From 7e6da390148ea6a782c0c4040563069a89f4c84a Mon Sep 17 00:00:00 2001
From: Ryan Lovelett <RLovelett@users.noreply.github.com>
Date: Thu, 6 Feb 2020 11:01:22 -0500
Subject: [PATCH 1/1] Use withContiguousStorageIfAvailable for String encoding

---
 Sources/SwiftProtobuf/BinaryEncoder.swift | 9 ++++++++-
 1 file changed, 8 insertions(+), 1 deletion(-)

diff --git a/Sources/SwiftProtobuf/BinaryEncoder.swift b/Sources/SwiftProtobuf/BinaryEncoder.swift
index b3c73f18..6e3800b1 100644
--- a/Sources/SwiftProtobuf/BinaryEncoder.swift
+++ b/Sources/SwiftProtobuf/BinaryEncoder.swift
@@ -36,10 +36,12 @@ internal struct BinaryEncoder {
         pointer = pointer.advanced(by: count)
     }
 
-    private mutating func append(contentsOf bufferPointer: UnsafeBufferPointer<UInt8>) {
+    @discardableResult
+    private mutating func append(contentsOf bufferPointer: UnsafeBufferPointer<UInt8>) -> Int {
         let count = bufferPointer.count
         pointer.assign(from: bufferPointer.baseAddress!, count: count)
         pointer = pointer.advanced(by: count)
+        return count
     }
 
     func distance(pointer: UnsafeMutablePointer<UInt8>) -> Int {
@@ -120,12 +122,17 @@ internal struct BinaryEncoder {
 
     // Write a string field, including the leading index/tag value.
     mutating func putStringValue(value: String) {
+      // If the collection does not support an internal representation in a form of contiguous storage,
+      // body is not called and nil is returned.
[...]

PR #949, opened 2020-02-06
Use withContiguousStorageIfAvailable for String encoding in BinaryEnc…

RLovelett:
I am proposing the following change for issue #919 which is labeled as "help-wanted".

The part that might be unexpected is the modification to `append(contentsOf:)`. I made that function return a value, the number of bytes appended from the input buffer, to silence the compiler warning: `Constant 'isAvailable' inferred to have type '()?', which may be unexpected`.

Resolves issue #919.

### DIFF
--- a/Sources/SwiftProtobuf/BinaryEncoder.swift
+++ b/Sources/SwiftProtobuf/BinaryEncoder.swift
@@ -39,12 +39,14 @@ internal struct BinaryEncoder {
         }
     }
 
-    private mutating func append(contentsOf bufferPointer: UnsafeRawBufferPointer) {
+    @discardableResult
+    private mutating func append(contentsOf bufferPointer: UnsafeRawBufferPointer) -> Int {
         let count = bufferPointer.count
         if let baseAddress = bufferPointer.baseAddress, count > 0 {
             memcpy(pointer, baseAddress, count)
         }
         pointer = pointer.advanced(by: count)
+        return count
     }
 
     func distance(pointer: UnsafeMutableRawPointer) -> Int {
@@ -125,11 +127,19 @@ internal struct BinaryEncoder {
 
     // Write a string field, including the leading index/tag value.
     mutating func putStringValue(value: String) {
-        let count = value.utf8.count
-        putVarInt(value: count)
-        for b in value.utf8 {
-            pointer.storeBytes(of: b, as: UInt8.self)
-            pointer = pointer.advanced(by: 1)
+        // If the String does not support an internal representation in a form
+        // of contiguous storage, body is not called and nil is returned.
+        let isAvailable = value.utf8.withContiguousStorageIfAvailable { (body: UnsafeBufferPointer<UInt8>) -> Int in
+            putVarInt(value: body.count)
+            return append(contentsOf: UnsafeRawBufferPointer(body))
+        }
+        if isAvailable == nil {
+            let count = value.utf8.count
+            putVarInt(value: count)
+            for b in value.utf8 {
+                pointer.storeBytes(of: b, as: UInt8.self)
+                pointer = pointer.advanced(by: 1)
+            }
         }
     }
 
--- a/Sources/SwiftProtobuf/BinaryEncoder.swift
+++ b/Sources/SwiftProtobuf/BinaryEncoder.swift
@@ -127,20 +127,21 @@ internal struct BinaryEncoder {
 
     // Write a string field, including the leading index/tag value.
     mutating func putStringValue(value: String) {
+        let utf8 = value.utf8
         #if swift(>=5.0)
             // If the String does not support an internal representation in a form
             // of contiguous storage, body is not called and nil is returned.
-            let isAvailable = value.utf8.withContiguousStorageIfAvailable { (body: UnsafeBufferPointer<UInt8>) -> Int in
+            let isAvailable = utf8.withContiguousStorageIfAvailable { (body: UnsafeBufferPointer<UInt8>) -> Int in
                 putVarInt(value: body.count)
                 return append(contentsOf: UnsafeRawBufferPointer(body))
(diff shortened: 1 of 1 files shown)


Sources/SwiftProtobuf/BinaryEncoder.swift
```diff
@@ -125,11 +127,19 @@ internal struct BinaryEncoder {
...
     mutating func putStringValue(value: String) {
-        let count = value.utf8.count
-        putVarInt(value: count)
-        for b in value.utf8 {
-            pointer.storeBytes(of: b, as: UInt8.self)
-            pointer = pointer.advanced(by: 1)
+        // If the String does not support an internal representation in a form
+        // of contiguous storage, body is not called and nil is returned.
+        let isAvailable = value.utf8.withContiguousStorageIfAvailable { (body: UnsafeBufferPointer<UInt8>) -> Int in
+            self.putVarInt(value: body.count)
```
thomasvl: style wise, we don't do `self.`, mixing the style within a single function also really sticks out.

RLovelett: Thank you for the feedback this should now be fixed in 11531df7a63cb0036775610ba14ff593732c8d7a.

Sources/SwiftProtobuf/BinaryEncoder.swift
```diff
@@ -125,11 +127,19 @@ internal struct BinaryEncoder {
...
     // Write a string field, including the leading index/tag value.
     mutating func putStringValue(value: String) {
-        let count = value.utf8.count
-        putVarInt(value: count)
-        for b in value.utf8 {
-            pointer.storeBytes(of: b, as: UInt8.self)
-            pointer = pointer.advanced(by: 1)
+        // If the String does not support an internal representation in a form
+        // of contiguous storage, body is not called and nil is returned.
+        let isAvailable = value.utf8.withContiguousStorageIfAvailable { (body: UnsafeBufferPointer<UInt8>) -> Int in
```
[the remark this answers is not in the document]
RLovelett: According to [SE-0237](https://github.com/apple/swift-evolution/blob/master/proposals/0237-contiguous-collection.md) `withContiguousStorageIfAvailable` was added in Swift 5.0.

I've added `#if swift(>=5.0)` markers around the implementation. After doing this I found that I was repeating the fallback/legacy `for` loop based implementation. To DRY that up I added a new `private` method `loopWrite(value:)` which is the existing loop based write method.

Sources/SwiftProtobuf/BinaryEncoder.swift
```diff
@@ -133,6 +136,23 @@ internal struct BinaryEncoder {
...
+                return append(contentsOf: UnsafeRawBufferPointer(body))
+            }
+            if isAvailable == nil {
+                loopWrite(value: value)
+            }
+        #else
+            loopWrite(value: value)
+        #endif
+    }
+
```
thomasvl: You can probably just do something like this to shrink the code and make the edit when our min lets us remove the conditional simpler:

```swift
    mutating func putStringValue(value: String) {
      #if swift(>=5.0)
        // If the String does not support an internal representation in a form
        // of contiguous storage, body is not called and nil is returned.
        let isAvailable = value.utf8.withContiguousStorageIfAvailable { (body: UnsafeBufferPointer<UInt8>) -> Int in
            putVarInt(value: body.count)
            return append(contentsOf: UnsafeRawBufferPointer(body))
        }
      #else
        let isAvailable: String? = nil 
      #endif
        if isAvailable == nil {
            let count = value.utf8.count
            putVarInt(value: count)
            for b in value.utf8 {
                pointer.storeBytes(of: b, as: UInt8.self)
                pointer = pointer.advanced(by: 1)
            }
        }
    }
```

RLovelett: It's your code base. It is so in 3aa2905bd4221bf9fd7fd481920b1b37aace1362.

thomasvl:
@tbkka you've done more of the text work in the library, can you take a look?

Sources/SwiftProtobuf/BinaryEncoder.swift
```diff
@@ -125,12 +127,24 @@ internal struct BinaryEncoder {
...
-        }
+        #if swift(>=5.0)
+            // If the String does not support an internal representation in a form
+            // of contiguous storage, body is not called and nil is returned.
+            let isAvailable = value.utf8.withContiguousStorageIfAvailable { (body: UnsafeBufferPointer<UInt8>) -> Int in
+                putVarInt(value: body.count)
+                return append(contentsOf: UnsafeRawBufferPointer(body))
+            }
+        #else
+            let isAvailable: String? = nil
```
tbkka: Minor: This should be `Int?` to match the type of the `withContiguousStorage...` call above.

Sources/SwiftProtobuf/BinaryEncoder.swift
```diff
@@ -125,12 +127,24 @@ internal struct BinaryEncoder {
 
     // Write a string field, including the leading index/tag value.
     mutating func putStringValue(value: String) {
```
tbkka: Minor:  I would `let utf8 = value.utf8` to factor out multiple references to the `utf8` view below.

Sources/SwiftProtobuf/BinaryEncoder.swift
thomasvl: That's on me.  I suggested reflowing the conditional and didn't full think about what the type was, so I just went with `String?`, the reality is it doesn't matter since it just needs to define it as some thing that sorta aligns with the other call from the POV of being an optional that can be `if`ed on.

thomasvl:
Thanks for doing this!

PR #953, opened 2020-02-10
Tweak the template to suggest people check the old issues and the new FAQ.

thomasvl:


.github/issue_template.md
```diff
@@ -1,3 +1,9 @@
+Before opening a new issue, please double check the passed issues (both open and
```
allevato: s/passed/past/

.github/issue_template.md
```diff
@@ -1,3 +1,9 @@
...
+closed ones) so see if your problem has been discussed before and already
```
allevato: s/so/to/

.github/issue_template.md
```diff
@@ -1,3 +1,9 @@
...
+contains an answer/solution. Likewise, check the FAQ in the Documentation folder
```
allevato: Even though this will show up as editable text and not a clickable link, I wonder if a URL to the page that the user can quickly copy-and-paste would make this easier/more likely to be read.

.github/issue_template.md
thomasvl: done

.github/issue_template.md
thomasvl: done

.github/issue_template.md
thomasvl: added.

PR #954, opened 2020-02-11
Reflow the camelcase apis so they can support non identifier input.

thomasvl:
The one place Swift identifiers are made from non protobuf identifiers is for the extension registry created for files with extensions. It uses the filename in the creation of the identifier, and based on the file name it could result in and invalid Swift symbol. This changes the camel casing code to handle things outside the protobuf identifier range to ensure the end up as valid Swift symbols also.

While the support for non protobuf identifier characters is new in this code, the reality is it would have resulted in invalid Swift, so it wasn't possible before. i.e. - what didn't work, now works. And the mapping for filenames that were using valid protobuf identifier characters remains unchanged, so no breaking behavior.


Tests/SwiftProtobufPluginLibraryTests/Test_NamingUtils.swift
```diff
@@ -342,6 +342,16 @@ class Test_NamingUtils: XCTestCase {
...
 
       ( "url_foo_http_id", "urlFooHTTPID", "URLFooHTTPID"),
+
+      ( "göß", "göß", "Göß"),
+      ( "göo", "göO", "GöO"),
+      ( "gö_o", "göO", "GöO"),
+      ( "g_🎉_o", "g🎉O", "G🎉O"),
+      ( "g🎉o", "g🎉O", "G🎉O"),
+
+      ( "m\u{AB}n", "m_u171N", "M_u171N"),
```
allevato: Does C++ also use the decimal code point value, not hex?

It probably doesn't matter all that much since the name is going to be mostly unreadable either way, but personally I would "expect" to see hex in a Unicode mangling like this 🤷‍♂ 

Are we worried about collisions? Like `"m\u{6AF}" -> "m_u1711"` and `"m\u{AB}1" -> "m_u1711"`? I guess not, since the examples below also end up having collisions.

thomasvl: > Does C++ also use the decimal code point value, not hex?

C++ uses hex, but is does it character wise, so the test filename gets something like `gro_cc_88_c3_9f`.

> It probably doesn't matter all that much since the name is going to be mostly unreadable either way, but personally I would "expect" to see hex in a Unicode mangling like this 🤷‍♂

I didn't go with hex since there was a chance it could be all letters or atleast end with them, and then any following text could but up against it making the camel casing harder to read.  Numbers seemed like it would avoid that.

I debated adding a trailing underscore, but then two of these in a row also gets odd because we get a double underscore.

> Are we worried about collisions? Like `"m\u{6AF}" -> "m_u1711"` and `"m\u{AB}1" -> "m_u1711"`? I guess not, since the examples below also end up having collisions.

The non Swift identifier case is the least of the collision problem.  The main algorithms handing of valid protobuf identifiers could easily have collisions already (`foo_bar.proto` & `FooBar.proto` both end up the same, just like two fields named that way would).

PR #957, opened 2020-02-13
build: add a CMake based build

compnerd:
This adds a CMake based build for supporting platforms where
swift-package-manager is not yet ready.  This enables building
swift-protobuf on Windows.


compnerd:
@allevato the conditions are reasonable and acceptable :)

tbkka:
I'm okay with this as a stop-gap.  (And honestly somewhat thrilled that people want to use this on Windows.  ;-)

PR #960, opened 2020-02-26
Try to provide some more details on extension usage.

thomasvl:


Documentation/API.md
```diff
@@ -641,36 +644,41 @@ extend CanBeExtended {
...
+* **Extension maps** are collections of extension objects indexed by the target
+  message and field number.  An extension map is generated for every file that
+  defined proto extensions and included as a static global variable.  It is
+  named based on the proto package, filename, and then ends in `_Extensions`,
+  so the above file would be `Sample_Extensions`. These maps are then used
+  the the `Message` apis for parsing/merging binary data to provide the
+  additional data to parse out the extension fields; if a mapping isn't
+  found, the extension field ends up in the `unknownFields` on the message.
+
+  If you need handle extension defined in multiple files, you can build up
```
allevato: s/extension/extensions/

thomasvl: Done.

Documentation/API.md
```diff
@@ -641,36 +644,41 @@ extend CanBeExtended {
...
-  `ExtensibleMessage` and define some additional methods needed by the
-  other components.  You should not use these methods directly.
-* **Extension properties** use Swift's `extension` capability to add new properties to the
-  original message.  In many cases, you can simply use the extension properties
-  without understanding any of the other extension machinery.  The above
-  example creates a Swift extension of `CanBeExtended` that defines a new
-  property `extensionField` of type `Optional<Int32>`.
+  [`ExtensibleMessage`](https://github.com/apple/swift-protobuf/blob/master/Sources/SwiftProtobuf/ExtensibleMessage.swift)
+  and define some additional methods needed by the other components.  You
+  should _not_ need use these methods directly.
```
tbkka: "... need to use ..."

Documentation/API.md
```diff
@@ -641,36 +644,41 @@ extend CanBeExtended {
...
-    public subscript(messageType: Message.Type, protoFieldNumber: Int) -> MessageExtensionBase? {get}
-    public func fieldNumberForProto(messageType: Message.Type, protoFieldName: String) -> Int?
-}
-```
+* **Extension maps** are collections of extension objects indexed by the target
+  message and field number.  An extension map is generated for every file that
+  defined proto extensions and included as a static global variable.  It is
+  named based on the proto package, filename, and then ends in `_Extensions`,
+  so the above file would be `Sample_Extensions`. These maps are then used
+  the the `Message` apis for parsing/merging binary data to provide the
```
tbkka: "... used by the ..." -- omitted word "by"

I suggest simplifying this sentence a bit: "... used by the `Message` apis for parsing/merging extension fields in the binary data; if a mapping ..."

Documentation/API.md
```diff
@@ -641,36 +644,41 @@ extend CanBeExtended {
...
+  additional data to parse out the extension fields; if a mapping isn't
+  found, the extension field ends up in the `unknownFields` on the message.
+
+  If you need handle extensions defined in multiple files, you can build up
```
tbkka: "If you need to handle..."  ---- omitted word "to"

Documentation/API.md
thomasvl: Done in #962

Documentation/API.md
thomasvl: done

Documentation/API.md
thomasvl: done

PR #964, opened 2020-02-27
Mark all generated files to be ignored by swift-format

thomasvl:


allevato:
> At some point, we should probably adjust our formatting to match the emerging consensus style, (rather than following the Swift stdlib style).

The style that swift-format currently applies is based on the stdlib style with a few modifications, so I kind of hope that we can make the consensus style pretty close to the stdlib style 😬

thomasvl:
Agreed on trying to eventually match style, just figured this reduces churn for folks for the time being if they try to run format over their source directories (and they check in their generated source like we are doing).

PR #966, opened 2020-02-28
Make setting a repeated extension field to [] clear it

thomasvl:


thomasvl:
Oh, and just want confirmation that my logic is sound -

I don't believe there is any api breaking change here since it is just an addition.  Nothing generated is changed, it just "fixes" things when people compile against the updated library.

tbkka:
Yeah, I think you're okay.

PR #976, opened 2020-04-14
tweaking the README to expose the packagename

heckj:
With Xcode 11.4 (and swift 5.2), swiftpm won't otherwise recognize the
packages. Took me several hours to hunt this down, so I wanted to share
it so the default value you might copy-pasta into your own Package.swift
would work correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ the plugin above:
 
 ```swift
 dependencies: [
-    .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.6.0"),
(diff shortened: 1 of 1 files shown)


heckj:
I don't have a great answer there - it's definitely a change that's bitten a number of folks with the Swift 5.2 release. I've seen this particular tidbit talked about in forums ([swift forum reference](https://forums.swift.org/t/package-names-in-swift-5-2/34886)) and slack channels. Where it applies most specifically is when the package name doesn't match the repository name (as is the case here), and it took me a bit of time to sort it out - so I wanted to make sure others didn't have the same "WTF happened" experience.

thomasvl:
Thanks, merged!

PR #979, opened 2020-04-27
Drop the 3.11.4 protobuf release from CI

thomasvl:
Since we need newer features to generate, it can't be directly tested without lots of conditionals in the Makefile.


thomasvl:
@toffaletti syncing my fork ended up auto enabling/running the actions, I guess the presence of the file is enough to enable things?

toffaletti:
> @toffaletti syncing my fork ended up auto enabling/running the actions, I guess the presence of the file is enough to enable things?

For your fork, yes but not for the main repo I believe.

PR #984, opened 2020-04-30
Replace Version.moduleName with SwiftProtobufInfo.name.

thomasvl:
The Namer already had to know the module and used the other
constant so generated module references would be correct.


thomasvl:
@tbkka sorry I forgot we already had this constant for the other PR.  if they mismatched WKT references would be correct, so safer to just have one concept in the sources instead of two.

tbkka:
FYI:  @cburrows

PR #999, opened 2020-05-14
Turn off fail-fast.

### LINKED ISSUE #997: CI cancels if one task fails
Looking at the github mails, if one of the linux platforms fails, it cancels the others.  We likely want them all to finish so we know if the failure is specific to a version of swift being tested.

thomasvl:
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

Fixes #997


thomasvl:
@toffaletti fyi

PR #1001, opened 2020-05-14
ci workflow: resolve protobuf_gitref to commit sha

toffaletti:
@thomasvl here is the floating master support. If you look at some old runs you can see an example of having both commit and branch in the matrix.

https://github.com/toffaletti/swift-protobuf/actions/runs/105024487/workflow

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,9 +12,11 @@ jobs:
     strategy:
       matrix:
         swift: ["5.2.1-bionic", "5.1.5-bionic", "4.2.4"]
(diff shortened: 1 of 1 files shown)


toffaletti:
> Thanks! I guess we also wait to see what happens on the other issue about why some times things seem to recompute even when they should have been saved from the previous run?

They got back to me on the issue I filed. It is working as intended; for security reasons the cache from master and PR are not shared. https://github.com/actions/cache/issues/310

toffaletti:
I can think of a possible solution, but it is rather involved. I will open an issue and describe it.

Issue #993, opened 2020-05-14
Conformance failure: proto2 + JSON + extension


tbkka:
This popped up on a CI bot:

CONFORMANCE TEST BEGIN ====================================
 
ERROR, test=Recommended.Proto2.JsonInput.FieldNameExtension.Validator: Expected JSON payload but got type 1. request=json_payload: "{\n        \"[protobuf_test_messages.proto2.extension_int32]\": 1\n      }" requested_output_format: JSON message_type: "protobuf_test_messages.proto2.TestAllTypesProto2" test_category: JSON_TEST, response=parse_error: "JSON failed to parse: unknownField(\"[protobuf_test_messages.proto2.extension_int32]\")"

thomasvl:
I sent some email looking for more info on proto2 extension support in JSON.

tbkka:
Let me know what you find.  I just pushed up a draft PR #1002 which is based on the design for TextFormat and the contents of the new conformance test.  It still needs a little work before I'm happy merging it, but it's enough to at least pass that test.

thomasvl:
So far what I've been able to piece together:

Python has some support and tests that include extensions:
- https://github.com/protocolbuffers/protobuf/blob/master/python/google/protobuf/json_format.py
- https://github.com/protocolbuffers/protobuf/blob/master/python/google/protobuf/internal/json_format_test.py

And Go seems to (didn't hunt down their impl, just found a reference to the test):
- https://github.com/golang/protobuf/blob/e9dc0d72c1a35f349f6d6871842199ddaaaf9edc/jsonpb/json_test.go#L317

tbkka:
Thanks, Thomas!  Those seem to verify what I guessed from the conformance test.  Extensions are handled essentially the same as in TextFormat:  The key name is surrounded by `[]` to mark it as an extension field name, but otherwise the coding is the same as regular JSON.

I think that's enough detail to finish up #1002 and get this landed.

tbkka:
Fixed via #1002

PR #1002, opened 2020-05-14
Support Proto2 extensions in JSON coder/decoder

tbkka:
This is sufficient to pass the conformance check that was recently added, but still needs some additional testing and polish.

Addresses #993


Sources/SwiftProtobuf/Message+JSONAdditions.swift
```diff
@@ -79,6 +80,28 @@ extension Message {
...
+  /// Creates a new message by decoding the given string containing a
+  /// serialized message in JSON format.
+  ///
+  /// - Parameter jsonString: The JSON-formatted string to decode.
+  /// - Parameter extensions: An ExtensionMap
+  /// - Parameter options: The JSONDecodingOptions to use.
+  /// - Throws: `JSONDecodingError` if decoding fails.
+  public init(
+    jsonString: String,
+    extensions: ExtensionMap?,
```
thomasvl, line 92: You could just default this and have one init (still source compat), and it means the decoder doesn't have to make it optional.

Sources/SwiftProtobuf/Message+JSONAdditions.swift
```diff
@@ -113,5 +136,43 @@ extension Message {
...
+  /// text.
+  ///
+  /// - Parameter jsonUTF8Data: The JSON-formatted data to decode, represented
+  ///   as UTF-8 encoded text.
+  /// - Parameter extensions: The extension map to use with this decode
+  /// - Parameter options: The JSONDecodingOptions to use.
+  /// - Throws: `JSONDecodingError` if decoding fails.
+  public init(
+    jsonUTF8Data: Data,
+    extensions: ExtensionMap?,
```
thomasvl, line 151: Same about defaulting.

If keeping them, it might also make sense to make the other (order) inits call to these new ones so there is only one set of code to maintain.

Sources/SwiftProtobuf/JSONDecoder.swift
```diff
@@ -32,10 +34,21 @@ internal struct JSONDecoder: Decoder {
                                ignoreUnknownFields: self.options.ignoreUnknownFields)
   }
 
+  internal init(source: UnsafeRawBufferPointer, options: JSONDecodingOptions,
```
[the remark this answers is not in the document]
tbkka: Removed the extraneous init, since it's internal

Sources/SwiftProtobuf/JSONDecoder.swift
```diff
@@ -32,10 +34,21 @@ internal struct JSONDecoder: Decoder {
...
+                               ignoreUnknownFields: self.options.ignoreUnknownFields)
+    self.messageType = messageType
+    self.extensions = extensions
+  }
+
   private init(decoder: JSONDecoder) {
     // The scanner is copied over along with the options.
     scanner = decoder.scanner
     options = decoder.options
+    extensions = decoder.extensions
```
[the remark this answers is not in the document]
tbkka: Added  messageType argument so we can correctly track across nested messages.

Sources/SwiftProtobuf/Message+JSONAdditions.swift
tbkka: I've changed the older inits to delegate to the new ones to limit code duplication.

I'm still debating whether to actually remove the old inits or not.  I'll leave the other conversation open to track this while I cogitate.  ;-)

Sources/SwiftProtobuf/JSONDecoder.swift
```diff
@@ -697,6 +713,19 @@ internal struct JSONDecoder: Decoder {
     messageType: Message.Type,
     fieldNumber: Int
   ) throws {
-    throw JSONDecodingError.schemaMismatch
+    if let ext = extensions?[messageType, fieldNumber] {
```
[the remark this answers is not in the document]
tbkka: Good idea.  Done.

tbkka:
I'm still filling in tests and tracking down missing bits.  I'll look into Any JSON support next.

Sources/SwiftProtobuf/JSONDecoder.swift
```diff
@@ -713,19 +707,20 @@ internal struct JSONDecoder: Decoder {
...
-        fieldValue = try ext._protobuf_newField(decoder: &self)
-      }
-      if fieldValue != nil {
-        values[fieldNumber] = fieldValue
-      } else {
-        // This most likely indicates a bug in our extension support.
-        throw TextFormatDecodingError.internalExtensionError
-      }
+    guard let ext = extensions?[messageType, fieldNumber] else {
+      return
```
thomasvl, line 711: Is this an internal error since the extension wasn't found?

Or I guess we never get here because we never would have been able to look up the field number in the first place?  So maybe `assert`?

tbkka: I changed this to a simple force-unwrap, which is essentially an `assert`

Sources/SwiftProtobuf/Message+JSONAdditions.swift
tbkka: You're right.  We don't promise binary/ABI compat, so there's no compelling reason to not take advantage of this simplification.

tbkka:
I've filled in support and tests for Any + JSON + extensions, including figuring out a reasonable answer for implicit transcoding (TL;DR: extensions now cause JSON->binary transcoding to throw because none of the other options seemed better).

I've filled in additional tests for message-valued extension fields, arrays, and a few other cases.

Sources/SwiftProtobuf/JSONDecoder.swift
```diff
@@ -45,7 +52,10 @@ internal struct JSONDecoder: Decoder {
     if fieldCount > 0 {
       try scanner.skipRequiredComma()
     }
-    if let fieldNumber = try scanner.nextFieldNumber(names: fieldNameMap!) {
+    let fieldNumber = try scanner.nextFieldNumber(names: fieldNameMap!,
+                                                  messageType: messageType,
+                                                  extensionMap: extensions)
```
thomasvl, line 57: Rather than passing the extensionMap ever call, should we pass it in the initializer?  it can't change during the life of the scanner, right?

Sources/SwiftProtobuf/JSONEncoder.swift
```diff
@@ -128,6 +128,18 @@ internal struct JSONEncoder {
...
 
+    /// Begin a new extension field
+    internal mutating func startExtensionField(name: String) {
+        if let s = separator {
+            data.append(s)
+        }
+        data.append(asciiDoubleQuote)
+        data.append(asciiOpenSquareBracket)
+        data.append(contentsOf: name.utf8)
+        append(staticText: "]\":")
```
thomasvl, line 139: Style wise, we use two `append()`s above for the `"[`, but here we use a single `append(staticText:)`.  Should we be a little more consistent, i.e. - pick one pattern or the other?

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -1250,29 +1250,42 @@ internal struct JSONScanner {
   /// Throws if field name cannot be parsed.
   /// If it encounters an unknown field name, it silently skips
```
thomasvl, line 1251: This comment isn't right, we do throw for unknown fields.

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -1250,29 +1250,42 @@ internal struct JSONScanner {
...
   /// the value and looks at the following field name.
-  internal mutating func nextFieldNumber(names: _NameMap) throws -> Int? {
+  internal mutating func nextFieldNumber(
+    names: _NameMap,
+    messageType: Message.Type?,
```
thomasvl, line 1255: does `messageType` have to be optional, or can we require it?

Sources/SwiftProtobuf/Message+JSONAdditions.swift
```diff
@@ -1,3 +1,4 @@
+
```
thomasvl, line 1: stray newline?

Sources/SwiftProtobuf/Message+JSONAdditions.swift
```diff
@@ -63,17 +64,19 @@ extension Message {
   /// serialized message in JSON format.
   ///
   /// - Parameter jsonString: The JSON-formatted string to decode.
+  /// - Parameter extensions: An ExtensionMap for looking up extensions by name
   /// - Parameter options: The JSONDecodingOptions to use.
   /// - Throws: `JSONDecodingError` if decoding fails.
   public init(
     jsonString: String,
+    extensions: ExtensionMap = SimpleExtensionMap(),
```
thomasvl, line 72: Hm, I know I suggested making these non optional, but looking back at the whole PR, I'm realizing in the binary apis we do use nil to indicate no map.  So now I'm torn.  Oh well, up to you for which makes the code cleaner, etc.  (one option might be to nil it here but default it to a SimpleExtensionMap in that case in the actually Decoder/Visitors so they don't have to check in all case; anyway, up to you)

Tests/SwiftProtobufTests/Test_JSON_Extensions.swift
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    func test_optionalInt32Extension() throws {
+        assertJSONEncode("{\"[protobuf_unittest.optional_int32_extension]\":17}",
+                         extensions: extensions) {
+            (o: inout MessageTestType) in
+            o.ProtobufUnittest_optionalInt32Extension = 17
+        }
+
+        assertJSONDecodeFails("{\"[protobuf_unittest.XXoptional_int32_extensionXX]\":17}",
```
thomasvl, line 44: Maybe a test using the name without the square braces on the end to ensure that doesn't accidentally work.

Maybe a test with only a leading/trailing square brace to also ensure that doesn't accidentally work.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -202,7 +205,18 @@ internal class AnyMessageStorage {
       // never inserted.
       break
 
-    case .contentJSON:
+    case .contentJSON(let contentJSON, let options):
+      do {
+        guard let messageType = Google_Protobuf_Any.messageType(forTypeURL: _typeURL) else {
```
thomasvl, line 210: Shouldn't this go above the `do {...` instead of inside it?  i.e. - why catch and rethrow it?

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -202,7 +205,18 @@ internal class AnyMessageStorage {
...
+            throw BinaryEncodingError.anyTranscodeFailure
+        }
+        _ = try unpack(contentJSON: contentJSON,
+                       extensions: SimpleExtensionMap(),
+                       options: options,
+                       as: messageType)
+      } catch {
+        throw BinaryEncodingError.anyTranscodeFailure
+      }
       // contentJSON requires a good URL and our ability to look up
```
thomasvl, line 220: This seems to be what was lifted above, but lost the comment, so move the comments and don't repeat this check?

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -202,7 +205,18 @@ internal class AnyMessageStorage {
       // never inserted.
       break
 
-    case .contentJSON:
+    case .contentJSON(let contentJSON, let options):
+      do {
+        guard let messageType = Google_Protobuf_Any.messageType(forTypeURL: _typeURL) else {
+            throw BinaryEncodingError.anyTranscodeFailure
+        }
+        _ = try unpack(contentJSON: contentJSON,
```
thomasvl, line 213: should we comment about this doing a full unpack and throwing it away to then have it repeated, but given current flows, there's nothing we can really do about it?

Sources/Conformance/main.swift
```diff
@@ -85,15 +85,18 @@ func buildResponse(serializedData: Data) -> Conformance_ConformanceResponse {
     }
 
     let msgType: SwiftProtobuf.Message.Type
+    let extensions: SwiftProtobuf.SimpleExtensionMap
```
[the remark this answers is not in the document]
tbkka: Done

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: Corrected the comment to explain that it silently skips if `ignoreUnknownFields` is set and throws otherwise.

Sources/SwiftProtobuf/AnyMessageStorage.swift
tbkka: Ooops.  Good catch.

Sources/SwiftProtobuf/AnyMessageStorage.swift
tbkka: Commented.

Sources/SwiftProtobuf/AnyMessageStorage.swift
tbkka: Done.

Sources/SwiftProtobuf/JSONScanner.swift
tbkka: This is internal and all the callers provide it, so yes it can be non-optional.

Sources/SwiftProtobuf/Message+JSONAdditions.swift
tbkka: Fixed.

Sources/SwiftProtobuf/JSONEncoder.swift
tbkka: I changed it so we use the `staticText:` form here for both.

Sources/SwiftProtobuf/JSONDecoder.swift
tbkka: Good idea!  Same argument applies to `options`, so I've moved both of those down into the scanner.  That provoked several minor code simplifications regarding these values as well.

Tests/SwiftProtobufTests/Test_JSON_Extensions.swift
tbkka: Added several additional checks:
* No trailing bracket
* No leading bracket
* No brackets at all
* Missing/invalid package name

Sources/SwiftProtobuf/Message+JSONAdditions.swift
tbkka: Hmmm...  Looks like both binary and textformat use an optional here and at the decoder initializer.  So I've converted the JSON to use the same pattern.  Down at the scanner where the extension map is actually stored, I now fill in an empty SimpleExtensionMap() to avoid nil checks.  (Since `SimpleExtensionMap()` is just a wrapper around a dictionary and empty dictionaries initialize trivially, this is no big deal.)

tbkka:
This feels like it's in pretty good shape now.  Any concerns before I merge it?

PR #1004, opened 2020-05-15
Remove the empty check since it was the only one.

### LINKED ISSUE #988: Empty Packed Enum fails validation
In `BinaryEncodingSizeVisitor.swift` the method `visitPackedEnumField<E: Enum>(value: [E],  fieldNumber: Int)` does not follow the pattern of all other packed field types in that it adds
```
    guard !value.isEmpty else {
      return
    }
```
Before the general pattern of calculating size: 
```
    let tagSize = FieldTag(fieldNumber: fieldNumber,
                           wireFormat: .varint).encodedSize
    serializedSize += tagSize
    var dataSize = 0
    for v in value {
      dataSize += Varint.encodedSize(of: Int32(truncatingIfNeeded: v.rawValue))
    }
    serializedSize += Varint.encodedSize(of: Int64(dataSize)) + dataSize
```

However, when actually writing the binary it will still write the field number (`BinaryEncodingVisitor.swift`)
```
    encoder.startField(fieldNumber: fieldNumber, wireFormat: .lengthDelimited)
    var packedSize = 0
    for v in value {
      packedSize += Varint.encodedSize(of: Int32(truncatingIfNeeded: v.rawValue))
    }
    encoder.putVarInt(value: packedSize)
    for v in value {
      encoder.putVarInt(value: v.rawValue)
    }
```

This causes the calculated size to be different from the actual size when the message is serialized to data.

thomasvl:
The generated traverse methods should never call on a repeated field
that is empty.

Fixes #988

### DIFF
--- a/Sources/SwiftProtobuf/BinaryEncodingSizeVisitor.swift
+++ b/Sources/SwiftProtobuf/BinaryEncodingSizeVisitor.swift
@@ -228,10 +228,6 @@ internal struct BinaryEncodingSizeVisitor: Visitor {
 
   mutating func visitPackedEnumField<E: Enum>(value: [E],
                                      fieldNumber: Int) throws {
-    guard !value.isEmpty else {
-      return
-    }
-
     let tagSize = FieldTag(fieldNumber: fieldNumber,
                            wireFormat: .varint).encodedSize
     serializedSize += tagSize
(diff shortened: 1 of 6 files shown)


thomasvl:
@tbkka we could add `assert(!value.isEmpty)` to every `visitRepeated*`/`visitPacked*` method as added safety so there is no impact in release builds, but I'm not sure if it is worth it.  Thoughts?

thomasvl:
> Thanks! Sorry about the repetitive boilerplate there.

No worries, better safe than sorry.

(although I've been wondering about gyb or something lately for some of this code/tests...)

Issue #42, opened 2016-10-04
Automate/improve the dependency on google/protobuf


thomasvl:
The `Protos` directory contains a bunch of protos taken from [google/protobuf](github.com/google/protobuf/), the reality is they need to stay stay in sync (especially things for the plugin and the Well Known Types).

An ideal setup would likely to be try and pull them via a gitmodule so they always are in sync.  Short of that, something should be done to automate updating them.

Maybe once there is a CI system (#38), it could check/error if they need updating.

tbkka:
The plugin and well known types definitely should be synced somehow.  A gitmodule approach  is certainly worth considering.  At one time, I had a Makefile target to copy the files from google/protobuf.

For the rest, I could go either way.  They could be synced or not.  Syncing them benefits from new cases added upstream, but diminishes their usefulness for generator testing.

thomasvl:
PR #150 added a makefile target that will sync things.  Keeping this open as it would be nice to have things automatically done instead of a manual process.

thomasvl:
Closing this since #1005 adds an easy way to check this.

PR #1005, opened 2020-05-15
Add a check for the upstream protos going stale.

thomasvl:
I'm thinking with this we can count #42 as fixed.

I don't think we want to run this on `make check` because it it would trigger if we updated and someone built without syncing protobuf also.

I'm also not sure if we'd want to wire this into CI because it would only be valid on a build against master, and as soon as we start checking against a release again, it too could mis fire there (unless you can make a step only check in some cases).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -197,6 +197,7 @@ endif
 	build \
 	check \
 	check-for-protobuf-checkout \
+	check-proto-files \
 	check-version-numbers \
 	clean \
 	conformance-host \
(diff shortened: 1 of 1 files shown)


tbkka:
I agree that we can close #42 with this.

As for automating it, I wonder if we could use a completely separate CI pass that just did this comparison (against master) and nothing else?

thomasvl:
> As for automating it, I wonder if we could use a completely separate CI pass that just did this comparison (against master) and nothing else?

That's an idea.  Although would we want it on PRs or just when things land?  i.e. - would we want it to block someone trying to land a PR?  Or just want the signal anytime something landed (or can we have github run the action once a week?)

PR #1007, opened 2020-05-18
Add workflow to check the status of proto files.

thomasvl:


thomasvl:
Here is it working: https://github.com/thomasvl/swift-protobuf/runs/686597675?check_suite_focus=true

Here is it finding a diff: https://github.com/thomasvl/swift-protobuf/runs/686593009?check_suite_focus=true


(looks like my cron trigger isn't fully right, checking that now)

thomasvl:
That seems to have it fixed.  Now I just need to remember to look next week to confirm it worked.  :)

PR #1009, opened 2020-05-19
Add visitRepeated* methods to the BinaryEncodingSizeVisitory.

thomasvl:
What the default impls in Visitor "work", by providing them things should be
more optimized in that it doesn't have to loop for fixed width things and the
tag overhead can be computed and multiplied instead of being added in the loop.

### DIFF
--- a/Sources/SwiftProtobuf/BinaryEncodingSizeVisitor.swift
+++ b/Sources/SwiftProtobuf/BinaryEncodingSizeVisitor.swift
@@ -104,6 +104,136 @@ internal struct BinaryEncodingSizeVisitor: Visitor {
     serializedSize += tagSize + Varint.encodedSize(of: Int64(count)) + count
   }
 
+  // The default impls for visitRepeated*Field would work, but by implementing
+  // these directly, the calculation for the tag overhead can be optimized and
+  // the fixed width fields can be simple multiplication.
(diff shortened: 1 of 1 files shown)


thomasvl:
We could also do this in the actual encoding, but I'm not sure how much of a win it would end up being, the FieldTag is just hopefully something that collapses down to just about constants, and the repeated writing is still needed.

tbkka:
Any measurable perf wins?

tbkka:
I'd also be curious whether `@inlineable` helps any -- it probably doesn't, since there's an unavoidable layer of indirection between the generated code and the visitor, but it might create some additional opportunities for the optimizer, and the implementations here seem sufficiently correct that inlining would be safe.

thomasvl:
> Any measurable perf wins?

I never have luck with the perf_runner.  `Performance/perf_runner.sh -c 3671b6afcbe981d0ebfcfc610a904867d09db399 100 "repeated fixed32"` gives me some data, but it always leaves me in a state where I have staged and unstated changes, so I'm never sure if I can trust what it comes up with between runs.

tbkka:
I've not had much luck using that mode, either.  I generally just run each branch against the C++ version, pencil down the numbers and compare them.

PR #1015, opened 2020-06-24
Don't use StaticString.utf8Start unless there is a pointer rep.

thomasvl:
The docs for StaticString.utf8Start say it should runtime abort for this, so
play it safe and handling when a StaticString doesn't have a ptr rep.

### DIFF
--- a/Sources/SwiftProtobuf/NameMap.swift
+++ b/Sources/SwiftProtobuf/NameMap.swift
@@ -55,6 +55,19 @@ fileprivate class InternPool {
     return immutable
   }
 
+  func intern(utf8Ptr: UnsafeBufferPointer<UInt8>) -> UnsafeRawBufferPointer {
+    #if swift(>=4.1)
+    let mutable = UnsafeMutableRawBufferPointer.allocate(byteCount: utf8Ptr.count,
+                                                         alignment: MemoryLayout<UInt8>.alignment)
+    #else
+    let mutable = UnsafeMutableRawBufferPointer.allocate(count: utf8.count)
+    #endif
+    mutable.copyBytes(from: utf8Ptr)
+    let immutable = UnsafeRawBufferPointer(mutable)
+    interned.append(immutable)
+    return immutable
+  }
+
   deinit {
     for buff in interned {
         #if swift(>=4.1)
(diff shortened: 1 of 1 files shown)


thomasvl:
@tbkka - @allevato noticed this when we were looking at the two issues that came in, it seems like this is the right thing to do to ensure things always work.

No clue if it is related to the two issues…

Sources/SwiftProtobuf/NameMap.swift
```diff
@@ -91,10 +91,19 @@ public struct _NameMap: ExpressibleByDictionaryLiteral {
...
-        self.utf8Buffer = UnsafeRawBufferPointer(start: staticString.utf8Start, count: staticString.utf8CodeUnitCount)
+    // This should not be used outside of this file, as it requires
+    // coordinating the lifecycle with the lifecycle of the pool
+    // where the raw UTF8 gets interned.
+    fileprivate init(staticString: StaticString, pool: InternPool) {
+        if staticString.hasPointerRepresentation {
+            self.nameString = .staticString(staticString)
+            self.utf8Buffer = UnsafeRawBufferPointer(start: staticString.utf8Start, count: staticString.utf8CodeUnitCount)
+        } else {
+            let str = staticString.withUTF8Buffer { String(decoding: $0, as: UTF8.self) }
```
allevato, line 102: This, combined with the interning, is going to end up making two copies: first, a copy will be made to decode the string on this line, and then the `intern` call below will make its own call to `UnsafeMutableRawBufferPointer.allocate` and copy the data from *that* string. The strings themselves will be small (because the pointer rep is only not used for single scalar strings, allegedly), but the extra heap allocations will be slow.

We could instead add a new overload to `intern` that takes a direct UTF8 raw pointer (the one passed to `withUTF8Buffer`) and copies the data out of that, to avoid the extra decoding operation.

thomasvl: But don't we still need the first copy because `self.nameString` ends up using `str` directly also?

thomasvl: Updated to use the `StaticString` for `nameString` still, and directly copy from the other buffer.

tbkka:
I'm fine with the approach proposed here for working around the current concern, but I suspect this code can be radically simplified.  A lot of the complexity here was motivated by issues with the original UTF-16 Swift String implementation, but things are very different now.  For example, we can likely just drop the `utf8Buffer` property entirely (and all of the tedious interning) if we just pre-compute and store the hash value so that we can rapidly compare with identifiers in JSON or TextFormat input.  Careful benchmarking is certainly warranted, but it's entirely possible that a more straightforward approach built on the standard String type would actually be faster than this code.

thomasvl:
Yea, the whole parsing/serialization pipeline likely could get a revalidation due to the `String` changes, there might be a few places things can be streamlined.

PR #1023, opened 2020-07-09
Add back the old init signatures.

### LINKED ISSUE #1022: API-breaking change in #1002
The PR #1002 introduced an API breaking change. Specifically, the diff changed two messages signatures in `Message+JSONAdditions`, from

```
  public init(
    jsonString: String,
    options: JSONDecodingOptions = JSONDecodingOptions()
  ) throws

  public init(
    jsonUTF8Data: Data,
    options: JSONDecodingOptions = JSONDecodingOptions()
  )
```

to

```
  public init(
    jsonString: String,
    options: JSONDecodingOptions = JSONDecodingOptions()
  ) throws

  public init(
    jsonUTF8Data: Data,
    options: JSONDecodingOptions = JSONDecodingOptions()
  )
```

Adding new parameters to Swift methods, even when they are defaulted, is a SemVer major. This is because functions may be referred to by their complete label set: `Message.init(jsonString:options:)`. With this change, the previously-valid fully-qualified method name no longer exists, and so compiles fail.

We need to update the code to restore the old methods, and to instead add new methods that take the `extensions` field. It's fine for the method body of the old methods to simply jump to the new ones, but they must continue to exist.

thomasvl:
Fixes #1022.

### DIFF
--- a/Sources/SwiftProtobuf/Message+JSONAdditions.swift
+++ b/Sources/SwiftProtobuf/Message+JSONAdditions.swift
@@ -63,6 +63,19 @@ extension Message {
   /// serialized message in JSON format.
   ///
   /// - Parameter jsonString: The JSON-formatted string to decode.
+  /// - Parameter options: The JSONDecodingOptions to use.
+  /// - Throws: `JSONDecodingError` if decoding fails.
+    public init(
+    jsonString: String,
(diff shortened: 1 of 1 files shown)


thomasvl:
@Lukasa fyi

PR #1024, opened 2020-07-09
Version bump to do a spot release for the accidental api break.

thomasvl:
Just bumping to x.y.1, the only change so far was docs: https://github.com/apple/swift-protobuf/compare/1.10.0...master


thomasvl:
@Lukasa fyi

PR #1027, opened 2020-07-10
Fixes and version bump to right api break

### LINKED ISSUE #1025: Further API breakages in 1.10.0
Sorry folks, my issue yesterday (#1022) failed to catch the full suite of breakages we accidentally shipped in 1.10.0. I missed this at the time, but when we went to update another project we caught some more.

I finally decided to run SwiftNIO's [`check_no_api_breakages` script](https://github.com/apple/swift-nio/blob/master/scripts/check_no_api_breakages.sh), comparing 1.9.0 and 1.10.1. This is what else slipped through:

```
Checking SwiftProtobuf.json... ERROR
==============================
ERROR: public API change in SwiftProtobuf.json
==============================

/* Generic Signature Changes */

/* RawRepresentable Changes */

/* Removed Decls */

/* Moved Decls */

/* Renamed Decls */
Func Message.array(fromJSONString:options:) has been renamed to Func array(fromJSONString:extensions:options:)
Func Message.array(fromJSONUTF8Data:options:) has been renamed to Func array(fromJSONUTF8Data:extensions:options:)

/* Type Changes */
Func Message.array(fromJSONString:options:) has parameter 1 type change from SwiftProtobuf.JSONDecodingOptions to SwiftProtobuf.ExtensionMap
Func Message.array(fromJSONUTF8Data:options:) has parameter 1 type change from SwiftProtobuf.JSONDecodingOptions to SwiftProtobuf.ExtensionMap

/* Decl Attribute changes */

/* Fixed-layout Type Changes */

/* Protocol Conformance Change */

/* Protocol Requirement Change */

/* Class Inheritance Change */
Checking SwiftProtobufPluginLibrary.json... OK
done
```

This points out that we need the same patch as #1023 but for returning `.array(fromJSONString:options:)` and `.array(fromJSONUTF8Data:options)`. This will require a 1.10.2.

Sorry about not running the full validation earlier. Perhaps we should consider integrating a variant of NIO's script to CI?

thomasvl:
Fixes #1025


thomasvl:
@Lukasa fyi

PR #1028, opened 2020-07-10
Remove empty Data singleton

cburrows:
Added in #466, `Internal.emptyData` improved performance. It seems that Swift has improved since then. See https://github.com/apple/swift-protobuf/pull/1017#discussion_r453014842 for one reason we might not like to generate `emptyData` reference, which is that it requires module name lookup.

Screen Shot 2020-07-10 at 3 49 23 PM

NB `Internal.emptyData` is not removed in SwiftProtobuf because it would break existing generated code. Should we deprecate it?


cburrows:
By appearances, my guess is that this got better in Swift 5.1 due to a refactoring of `Data`'s internal storage in https://github.com/apple/swift/pull/20225.

thomasvl:
> NB `Internal.emptyData` is not removed in SwiftProtobuf because it would break existing generated code. Should we deprecate it?

At the moment, I'd say no just because the files still work, and we haven't ever made old generated files warn before.

thomasvl:
@allevato fyi

PR #1037, opened 2020-08-10
Add 5.0.x build to coverage.

thomasvl:
Let the matrix auto generate and just insert the one special case for 4.2.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        swift: ["5.2.4-bionic", "5.1.5-bionic", "4.2.4"]
+        swift: ["5.2.4-bionic", "5.1.5-bionic", "5.0.3-bionic", "4.2.4"]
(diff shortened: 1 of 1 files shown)


thomasvl:
Hm, was the 5.0 image bad in that it is missing zlib or something?

tbkka:
I suppose we could try adding `zlib-dev` to the `apt-get install` line to ensure it's there?

thomasvl:
Looks like `zlib-dev` doesn't exist.  Looking at the [packages providing the header](https://packages.ubuntu.com/search?searchon=contents&keywords=zlib.h&mode=exactfilename&suite=eoan&arch=any), I'm not sure if one of those is safe to pick without getting extra dependencies by accident.

thomasvl:
ps - I'm trying to figure out how to get 5.0.x added because when 5.3 ships, our min _toolchain_ would be 4.3, which doesn't exist, so having something on 5.0 would get us that older coverage.

tbkka:
You want `zlib1g-dev`

thomasvl:
Thanks, that seems to have worked.

Issue #1034, opened 2020-08-07
EXC_BAD_ACCESS in decodeMessage(decoder:) due to stack overflow


n8gray:
Note: This is likely a duplicate of https://github.com/apple/swift-protobuf/issues/1014 but I didn't want to cram all of this into a comment in a long chain. Feel free to close as duplicate if you want.

Summary/TLDR
-----------
The change to move computed properties backed by `_storage` to properties stored directly on the message structs has caused massive size increases for messages at the top of a deep hierarchy.  The switch code generated for `oneof` parsing is sensitive to these size increases, leading to massive stack frames.  In debug mode without optimization these stack frames can easily exhaust the 512KB stack of a background dispatch thread and cause a crash.

Details
------------
We are experiencing a crash in debug builds after upgrading SwiftProtobuf and `protoc-gen-swift` from 1.7 to 1.10.2. Our crash happens in the `decodeMessage(decoder:)` method of a message containing a `oneof` with lots of cases.  From what I can tell the stack frame generated for the `withExtendedLifetime` closure in that method is huge, about 800KB. Since the decoding is happening on a background thread with a 512KB stack, the app crashes as soon as this closure is invoked.

The message is of the form:
```
message RenderDataDTO {
  SomeMessage1 m1 = 1;
  SomeMessage2 m2 = 2;
  SomeMessage3 m3 = 3;
  oneof render {
    SomeCaseMessage1 case1 = 100;
    SomeCaseMessage2 case2 = 101;
    SomeCaseMessage3 case3 = 102;
    // 20 cases in total
    SomeCaseMessage20 case20 = 120;
  }
```

The code that gets generated for `decodeMessage(decoder:)` has the form:
```swift
  public mutating func decodeMessage<D: SwiftProtobuf.Decoder>(decoder: inout D) throws {
    _ = _uniqueStorage()    // LINE A
    try withExtendedLifetime(_storage) { (_storage: _StorageClass) in    // LINE B
      while let fieldNumber = try decoder.nextFieldNumber() {
        switch fieldNumber {
        case 1: try decoder.decodeSingularMessageField(value: &_storage._m1)
        case 2: try decoder.decodeSingularMessageField(value: &_storage._m2)
        case 3: try decoder.decodeSingularStringField(value: &_storage._m3)
        case 100:
          var v: SomeCaseMessage1?
          if let current = _storage._render {
            try decoder.handleConflictingOneOf()
            if case .case1(let m) = current {v = m}
          }
[...]

PR #1040, opened 2020-08-11
Work around stack usage for non optimize build and `switch` statements

thomasvl:
There's still some generated `switch` code around required fields and oneofs, but reflowing the code is going to be a bit larger, so that will happen in another PR.

Work on #1034


Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -283,14 +283,19 @@ class MessageGenerator {
         p.print("while let \(varName) = try decoder.nextFieldNumber() {\n")
         p.indent()
         if !fields.isEmpty {
-          p.print("switch fieldNumber {\n")
+
+          p.print(
+              "// The use of inline closures is to circumvents an issue where the compiler\n",
```
allevato, line 288: circumvent

thomasvl: Done

PR #1043, opened 2020-08-12
Add .dynamic library type for SPM

hewigovens:
We'd like to progressively adopt Swift Package Manger into our existing app whose dependencies are managed by cocoapods + `use_frameworks!`, without `.dynamic` option, we can't choose dynamic library version of SwiftProtobuf. 

Here is the screenshot after adding `.dynamic` in my fork

Screen Shot 2020-08-12 at 1 29 51 PM

We can still import SwiftProtobuf like below:

```swift
import SwiftProtobuf
```

The final copied framework name is `SwiftProtobufDynamic`

Screen Shot 2020-08-12 at 1 30 36 PM

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,7 +16,9 @@ let package = Package(
   products: [
     .executable(name: "protoc-gen-swift", targets: ["protoc-gen-swift"]),
     .library(name: "SwiftProtobuf", targets: ["SwiftProtobuf"]),
+    .library(name: "SwiftProtobufDynamic", type: .dynamic, targets: ["SwiftProtobuf"]),
     .library(name: "SwiftProtobufPluginLibrary", targets: ["SwiftProtobufPluginLibrary"]),
+    .library(name: "SwiftProtobufPluginLibraryDynamic", type: .dynamic, targets: ["SwiftProtobufPluginLibrary"]),
   ],
   targets: [
     .target(name: "SwiftProtobuf"),


thomasvl:
- Why do you need the SwiftProtobufPluginLibrary that is has to be a dynamic framework?
- We make no promises about library evolution, which building as a dynamic library sorta implies.  Why do you need the SwiftPM target for it to be a library?

hewigovens:
1. we can remove `.dynamic` for `SwiftProtobufPluginLibrary` because we don't actually use it.
2. we have a framework (contains Swift binding code and a few C static libs) that depends on SwiftProtobuf. It's quite slow to build it every time, and Xcode 12 adds SwiftPM binary framework support.

thomasvl:
I'll defer to @tbkka on this, but it feels wrong to me to add the dynamic framework since we don't make the promises around stability that it would seem to imply.

hewigovens:
It's same as Carthage IMHO, not a problem if you distribute binary framework built with `BUILD_LIBRARY_FOR_DISTRIBUTION` or always build it from source (like we do in cocoapods)

Lukasa:
I think another problem is that this _forces_ all downstream consumers to build Protobuf as a dynamic library. This is not a problem for those consumers shipping apps with Xcode, who can use the standard application bundle to distribute the resultant framework. However, it's a huge chore for other applications (e.g. CLI applications or server applications) as they now have to move around a dynamic library as well as a binary.

I think the better option is to provide a feedback request that asks for Xcode to allow SPM packages to be built as frameworks for consumers that want them.

thomasvl:
I don't have any real experience with Carthage.

With cocoapods, they've gone the route of letting the developer integrating decide how to package it, so the package authors aren't the one making any claims about library stability, etc.  The developer choosing to package that way is doing it and thus assuming to know what they are doing.

Having it in the `Package.swift` vended by the project can be interpreted to mean we fully support that path; And we haven't put the thought into what would be required, especially since the library depends on generating code, and ensuring generate code works across framework versions.

I haven't really looked, but I can't say I've heard of other packages listing both targets like this, so it doesn't seem to be a common pattern.

thomasvl:
Following up to @Lukasa comment - if one package picks the shared library and another picks the normal library model, those two packages can never be both used by something else because they'd have dueling dependencies on SwiftProtobuf that couldn't be resolved.

hewigovens:
I understand your concern, this only adds an option to allow developers choosing either static or dynamic linking (developers know what they are doing), and static version is default and not changed.  Currently we have a fork that enables it, just need to regularly merge code from upstream

Package.swift
```diff
@@ -16,7 +16,9 @@ let package = Package(
   products: [
     .executable(name: "protoc-gen-swift", targets: ["protoc-gen-swift"]),
     .library(name: "SwiftProtobuf", targets: ["SwiftProtobuf"]),
+    .library(name: "SwiftProtobufDynamic", type: .dynamic, targets: ["SwiftProtobuf"]),
     .library(name: "SwiftProtobufPluginLibrary", targets: ["SwiftProtobufPluginLibrary"]),
+    .library(name: "SwiftProtobufPluginLibraryDynamic", type: .dynamic, targets: ["SwiftProtobufPluginLibrary"]),
```
tbkka, line 21: The `SwiftProtobufPluginLibrary` does not have a stable API or ABI.  We should certainly not provide a dynamic version of that at this time.

tbkka:
> It's ... not a problem if you distribute binary framework built with `BUILD_LIBRARY_FOR_DISTRIBUTION` ...

`BUILD_LIBRARY_FOR_DISTRIBUTION` is not magic.  It does not make the library invulnerable to API changes, and it has significant performance impacts.  It really just provides an additional layer of information about data layout and inhibits a lot of optimizations.  The result allows data structures in the library to change their layout without needing clients to be recompiled.  It is only needed when you share a framework between multiple programs (for example, when the framework is part of an OS) or update the framework and application separately from each other (for example, when the framework is part of an OS :wink:).  If you always distribute the app and frameworks together as a bundle, it's unnecessary and just serves to make things slower.

It sounds like your only real concern here is build times.  I would certainly urge you to ask the Xcode and SwiftPM folks about that -- I don't believe there is an inherent reason that building frameworks dynamically or statically should have a major impact on build times, and I know that the Swift team has made real progress recently on improving compile times.

hewigovens:
Thanks for all the discussions, let's close this for now

PR #1046, opened 2020-08-17
Revise storage decision

thomasvl:
Revise the way storage calculations are done to take into account the cost of singular message fields since those directly increase the size of the containing message. This should better cap things so Message don't get too big.

This is progress on #735.

### DIFF
--- a/Sources/protoc-gen-swift/MessageStorageDecision.swift
+++ b/Sources/protoc-gen-swift/MessageStorageDecision.swift
@@ -1,0 +1,149 @@
+// Sources/protoc-gen-swift/MessageGenerator.swift - Per-message logic
+//
+// Copyright (c) 2014 - 2016 Apple Inc. and the project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/master/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+
+import SwiftProtobufPluginLibrary
+
+// The file attempts to isolate the decisions around when to use heap based
+// storage vs. inlining it into the value times. At the moment the decicions
+// are entirely based on field counts and language requires, but this could be
+// revised in the future to better account for the actually *real* memory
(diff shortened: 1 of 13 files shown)


Sources/protoc-gen-swift/MessageStorageDecision.swift
```diff
@@ -0,0 +1,190 @@
...
+    }
+  }
+}
+
+/// Maxium computed cost of a Message's fields allow before it uses Storage.
+fileprivate let totalFieldCostRequiringStorage = 17
+
+/// The result of analysis, if the message should use heap storage and the
+/// cost of the message when used as a field in other messages.
+fileprivate struct AnalyseResult {
```
allevato, line 71: Not-strongly-held opinion: `AnalyzeResult` and `analyze(descriptor:)` for consistency, since we probably use American English spelling throughout other APIs 🤷

Sources/protoc-gen-swift/MessageStorageDecision.swift
```diff
@@ -0,0 +1,190 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/master/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+
+import SwiftProtobufPluginLibrary
+
+// The file attempts to isolate the decisions around when to use heap based
```
allevato, line 13: I love this. Having this pulled out into its own self-contained unit will make it a lot easier to reason about changes to the algorithm in the future.

Sources/protoc-gen-swift/MessageStorageDecision.swift
thomasvl: Done

Sources/protoc-gen-swift/MessageStorageDecision.swift
thomasvl: Thanks, I figured if this gets more work in the future it would be clearer this way.

PR #1051, opened 2020-08-21
Fix JSON coding/decoding of NullValue WKT

### LINKED ISSUE #1048: Failing new upstream conformance tests
Upstream has added some new conformance tests that aren't passing.

```
Recommended.Proto3.JsonInput.NullValueInOtherOneofNewFormat.Validator
Recommended.Proto3.JsonInput.NullValueInOtherOneofOldFormat.Validator
```

tbkka:
The google.protobuf.NullValue well-known type is used
to represent JSON `null` values.  As such, it needs custom
encoding/decoding in order to parse itself from a `null`
value or to write itself out in that form.

Prior to this, Message types could customize their JSON
encoding/decoding by conforming to _CustomJSONCodable.
This PR allows Enum types to do the same:
* They can choose to decode themselves from a `null` token
  (the null token is otherwise treated as a missing value)
* They can provide custom encoding (otherwise, the
  enum value is written out as a string or int)

It then uses this to provide the necessary customization for
the NullValue type.  Plus tests, of course.

Thanks to the Google protobuf team's ongoing conformance test
expansion efforts for catching this oversight!

Resolves #1048


Sources/SwiftProtobuf/JSONEncodingVisitor.swift
```diff
@@ -277,10 +280,14 @@ internal struct JSONEncodingVisitor: Visitor {
   }
 
   mutating func visitRepeatedEnumField<E: Enum>(value: [E], fieldNumber: Int) throws {
+    let options = self.options
     let alwaysPrintEnumsAsInts = options.alwaysPrintEnumsAsInts
     try _visitRepeated(value: value, fieldNumber: fieldNumber) {
       (encoder: inout JSONEncoder, v: E) throws in
-      if !alwaysPrintEnumsAsInts, let n = v.name {
+      if let e = v as? _CustomJSONCodable {
```
thomasvl, line 287: Do the class level check so you can consult a boolean inside the loop for the common case and do this check more quickly?

tbkka: That won't really help.  `v` still needs to be cast to `_CustomJSONCodable` to access the protocol method on the instance.  That cast requires an `as?` or `as!` operation and those are implemented the same way in the runtime.  That's why I plan to experiment with changing `encodedJSONString` to a static class method that accepts an instance, as that would allow us to do a single test outside the loop.

thomasvl: You lost me.  You'd completely skip the cast if the class doesn't conform, isn't that a huge win for the common case?

thomasvl: #1058 is what I was thinking.

PR #1058, opened 2020-08-24
Split enum visiting based on custom json support.

thomasvl:
This should only do the check once per repeated enum, which should make the cost
much lower on the common case since enums don't normally support this.

### DIFF
--- a/Sources/SwiftProtobuf/JSONEncodingVisitor.swift
+++ b/Sources/SwiftProtobuf/JSONEncodingVisitor.swift
@@ -280,17 +280,23 @@ internal struct JSONEncodingVisitor: Visitor {
   }
 
   mutating func visitRepeatedEnumField<E: Enum>(value: [E], fieldNumber: Int) throws {
-    let options = self.options
-    let alwaysPrintEnumsAsInts = options.alwaysPrintEnumsAsInts
-    try _visitRepeated(value: value, fieldNumber: fieldNumber) {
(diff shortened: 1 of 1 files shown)


thomasvl:
@tbkka this is what I was thinking to avoid the hit inside the loop.  Just decide up front so when there are multiple, it is much cheaper.

Sources/SwiftProtobuf/JSONEncodingVisitor.swift
```diff
@@ -280,17 +280,23 @@ internal struct JSONEncodingVisitor: Visitor {
   }
 
   mutating func visitRepeatedEnumField<E: Enum>(value: [E], fieldNumber: Int) throws {
-    let options = self.options
-    let alwaysPrintEnumsAsInts = options.alwaysPrintEnumsAsInts
-    try _visitRepeated(value: value, fieldNumber: fieldNumber) {
-      (encoder: inout JSONEncoder, v: E) throws in
-      if let e = v as? _CustomJSONCodable {
+    if let _ = E.self as? _CustomJSONCodable.Type {
```
tbkka, line 283: `if E.self is _CustomJSONCodable.Type {`

thomasvl: I basically had that, but switch to this because it is what we have in `visitRepeatedMessageField` right after it.  For the moment, I'll leave it as is so we're consistent.

PR #1062, opened 2020-08-26
Factor out the common decodeJSON from all the wrappers

tbkka:
While looking at other code here, I realized we had 9 copies of the identical `decodeJSON` implementation.  That seems unnecessary.

While here, simplify the encodedJSONString implementations as well.  In particular, since these all return Swift String instances, we should just use standard library String facilities directly instead of relying on lower-level JSON encoding tools and then converting to String.

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Wrappers+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Wrappers+Extensions.swift
@@ -32,6 +32,14 @@ protocol ProtobufWrapper {
   init(_ value: WrappedType.BaseType)
 }
 
+extension ProtobufWrapper {
+  mutating func decodeJSON(from decoder: inout JSONDecoder) throws {
+    var v: WrappedType.BaseType?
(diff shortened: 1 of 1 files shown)


Sources/SwiftProtobuf/Google_Protobuf_Wrappers+Extensions.swift
```diff
@@ -48,15 +56,16 @@ extension Google_Protobuf_DoubleValue:
...
-    var encoder = JSONEncoder()
-    encoder.putDoubleValue(value: value)
-    return encoder.stringResult
-  }
-
-  mutating func decodeJSON(from decoder: inout JSONDecoder) throws {
-    var v: WrappedType.BaseType?
-    try WrappedType.decodeSingular(value: &v, from: &decoder)
-    value = v ?? WrappedType.proto3DefaultValue
+    if value.isFinite {
```
thomasvl, line 59: The encoding changes give me slight pause, because it means the knowledge about these are in two places.  But since they are pretty simple, I guess it's fine; and I don't immediately see any better ideas to share it without the conversion overhead.

PR #1064, opened 2020-08-31
Reduce foundation usage

tbkka:
This is one very minor refactoring to eliminate a Foundation dependency from the date/time JSON formatting.

### DIFF
--- a/Sources/SwiftProtobuf/StringUtils.swift
+++ b/Sources/SwiftProtobuf/StringUtils.swift
@@ -20,6 +20,41 @@
 
 import Foundation
 
+/*
+  Note: Once our minimum support version is at least Swift 5.3, we
+  should probably recast the following to use
+  String(unsafeUninitializedCapacity:)
+*/
+
+// Note: We're trying to avoid Foundation's String(format:) since that's not
+// universally available.
+
+fileprivate func formatZeroPaddedInt(_ value: Int32, digits: Int) -> String {
+  let s = String(value)
+  if s.count >= digits {
+    return s
+  } else {
+    let pad = String(repeating: "0", count: digits - s.count)
+    return pad + s
+  }
+}
+
+internal func twoDigit(_ value: Int32) -> String {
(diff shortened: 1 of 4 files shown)


tbkka:
If this seems premature, we can wait to merge until we have a critical mass of such changes.  But I thought it might simplify other folks' work on this area to resolve a few such minor issues now.

Sources/SwiftProtobuf/StringUtils.swift
```diff
@@ -20,6 +20,41 @@
...
+/*
+  Note: Once our minimum support version is at least Swift 5.3, we
+  should probably recast the following to use
+  String(unsafeUninitializedCapacity:)
+*/
+
+// Note: We're trying to avoid Foundation's String(format:) since that's not
+// universally available.
+
+fileprivate func formatZeroPaddedInt(_ value: Int32, digits: Int) -> String {
```
allevato, line 32: Do we think we'll need this on other integer types later, where it might be good to avoid churn now and make this and the helpers below generic over `BinaryInteger`?

Should it handle negative values, or have a precondition that fails if the value is negative?

Sources/SwiftProtobuf/StringUtils.swift
```diff
@@ -20,6 +20,41 @@
 
 import Foundation
 
+/*
+  Note: Once our minimum support version is at least Swift 5.3, we
```
thomasvl, line 24: Why wait until the min is that, can't you use a `#if swift(>= 5.3)` block with two code paths?

Sources/SwiftProtobuf/StringUtils.swift
tbkka: There's very little of this sort of formatting, so I don't see much need to generalize this.

Precondition is a good idea.  Done.

Sources/SwiftProtobuf/StringUtils.swift
tbkka: I'll look into this later.

PR #1074, opened 2020-09-23
Fixed minor typo

TheAppMentor:
Fixed minor type in the Quick start section
Changed from : Once you have installed the code generator, **used** it to .. 
to : Once you have installed the code generator, **use** it to ..

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,7 +232,7 @@ Run `carthage update` and drag `SwiftProtobuf.framework` into your Xcode.project
 
 # Quick Start
 
-Once you have installed the code generator, used it to
+Once you have installed the code generator, use it to
 generate Swift code from your `.proto` file, and
 added the SwiftProtobuf library to your project, you can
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -232,7 +232,7 @@ Run `carthage update` and drag `SwiftProtobuf.framework` into your Xcode.project
 
 # Quick Start
 
-Once you have installed the code generator, used it to
+Once you have installed the code generator, use it to
 generate Swift code from your `.proto` file, and
 added the SwiftProtobuf library to your project, you can
```
thomasvl, line 237: I think the intent was for this long sentence to be all past tense "installed", "used", "added".  So if you are shifting "used" to "use", then "added" should likely change to "add"?

TheAppMentor: Yes, makes sense. Thanks for your review.

TheAppMentor:
This change is not required. Sorry for the trouble.

Issue #1075, opened 2020-09-23
map is not sorted by key


miworking3:
* what OS you are developing on (Linux or macOS, including the version)
macOS
* for macOS, what version of Xcode you are using (`xcodebuild -version`),
  for Linux, what version of Swift (`swift --version`)
11.7
* what version of Swift is your code set to compile with (i.e. from project
  settings, etc.)
5.2.4

* what branch/tag of SwiftProtobuf you are using (1.0.0, etc.)
1.12.0 Release

* Problem Description
 the `map` is not implemented as defined by [proto3](https://developers.google.com/protocol-buffers/docs/proto):
```When generating text format for a .proto, maps are sorted by key.```
while the output of `textFormatString()` of a dictionary in SWIFT is randomly sorted.

* Expected Result
If swift-protobuf decides to use `Dictionary` as the backed up data structure for `map` in proto3,  the result of `dictionary.textFormatString()` is expected to be sorted by the key.
e.g.,
```
var map = [String:String]()
map["b"] = "bbb"
map["a"] = "aaa"

print(map) // ["b": "bbb", "a": "aaa"]
print(map.textFormatString()) // [ "a": "aaa", "b": "bbb"]
```
* Proposed fix:
A fix for this line in [TextFormatEncodingVisitor.swift](https://github.com/apple/swift-protobuf/blob/master/Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift#L616)

before
```
for (k,v) in map {
```
after:
```
for (k,v) in map.sorted { $0.0  < $1.0 }.map { $0 } {
       
```

tbkka:
We would appreciate a Pull Request with this change.  Please also include a test case.  There should be some existing tests for map encoding/decoding in TextFormate that could be easily updated.

thomasvl:
There's even a conformance test, but the data set must always pass for Swift.

https://github.com/protocolbuffers/protobuf/blob/e667bf6eaaa2fb1ba2987c6538df81f88500d030/conformance/text_format_conformance_suite.cc#L317-L375

thomasvl:
fyi - I reached out to the protobuf team, it looks like that conformance test parses the TextFormat and compares the binary forms, so it doesn't actually check the raw TextFormat.

thomasvl:
Also, I don't think that proposed fix is drop in.  Since a `Dictionary`'s keys are only `Hashable` it doesn't mean they are `Comparable`, so odds are a few things might need tweaking to get it to work.

[...]

PR #1076, opened 2020-09-24
Sort map fields for TextFormat.

thomasvl:
Since Bool isn't Comparable, add a sorting function to MapKeyType and use that
in TextFormatEncodingVisitor.

### DIFF
--- a/Sources/SwiftProtobuf/FieldTypes.swift
+++ b/Sources/SwiftProtobuf/FieldTypes.swift
@@ -50,6 +50,17 @@ public protocol FieldType {
 /// Marker protocol for types that can be used as map keys
 ///
 public protocol MapKeyType: FieldType {
+    /// A comparision function for where order is needed.  Can't use `Comparable`
+    /// because `Bool` doesn't conform, and since it is `public` there is no way
+    /// to add a conformance internal to SwiftProtobuf.
+    static func _lessThan(lhs: BaseType, rhs: BaseType) -> Bool
+}
+
+// Default impl for anything `Comparable`
+extension MapKeyType where BaseType: Comparable {
+    public static func _lessThan(lhs: BaseType, rhs: BaseType) -> Bool {
+        return lhs < rhs
+    }
 }
 
 ///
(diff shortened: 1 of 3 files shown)


thomasvl:
I'm not sure this is the best way to do this, but this is what occurred to me last night.

I'm not thrilled with it having to make something `public`, but I didn't see a better way to do this that let us restrict the needed definitions to something within the library since the `Visitor` api itself is public.

Fixes #1075

Sources/SwiftProtobuf/FieldTypes.swift
```diff
@@ -50,6 +50,17 @@ public protocol FieldType {
 /// Marker protocol for types that can be used as map keys
 ///
 public protocol MapKeyType: FieldType {
+    /// A comparision function for where order is needed.  Can't use `Comparable`
+    /// because `Bool` doesn't conform, and since it is `public` there is no way
+    /// to add a conformance internal to SwiftProtobuf.
+    static func lessThan(lhs: BaseType, rhs: BaseType) -> Bool
```
allevato, line 56: We could make the name of the method `_lessThan` to make it clear that it's only meant to be used by the library.

(We probably should have done that for the protocol itself, in retrospect.)

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
```diff
@@ -611,9 +611,10 @@ internal struct TextFormatEncodingVisitor: Visitor {
   private mutating func _visitMap<K, V>(
     map: Dictionary<K, V>,
     fieldNumber: Int,
+    lessThanFunc: (K, K) -> Bool,
```
allevato, line 614: Nit: `isOrderedBefore` would be a cleaner name that matches the argument name used by Swift's stdlib for other sorting predicates.

allevato (approved):

Sources/SwiftProtobuf/FieldTypes.swift
thomasvl: Some of these actually leak into `Visitor` interface, so they sorta have to be `public` to allow someone to write a custom `Vistor`.

Prefixed this one.

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
thomasvl: done.

tbkka (approved):
Looks good.

thomasvl:
Talking with @allevato in person, we realized we might be able to do this like the Custom JSON support with a protocol internally to the library.  I'll take a shot at that before moving ahead with this since it would avoid adding to the public interfaces.

thomasvl:
And that isn't gonna work because I'd need an associated type, so I guess we're sorta stuck exposing something else at the moment. :(

thomasvl:
Landing since no one has come up with a better idea.  We can always revisit if something better comes along (especially if we haven't cut a release), but hopefully the underscore symbols gives up a little wiggle room.

PR #1079, opened 2020-10-21
Enable module stability support.

mingchen:
Enable module stability support to allow build as binary release
and used for different version of Xcode, it can save app build time.

see more on https://swift.org/blog/library-evolution/

### DIFF
--- a/SwiftProtobuf.xcodeproj/project.pbxproj
+++ b/SwiftProtobuf.xcodeproj/project.pbxproj
@@ -2161,6 +2161,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 5A82E3462455D26900EB0A70 /* SwiftProtobuf_iOS.xcconfig */;
 			buildSettings = {
+				BUILD_LIBRARY_FOR_DISTRIBUTION = YES;
 			};
 			name = Debug;
 		};
@@ -2168,6 +2169,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 5A82E3462455D26900EB0A70 /* SwiftProtobuf_iOS.xcconfig */;
 			buildSettings = {
+				BUILD_LIBRARY_FOR_DISTRIBUTION = YES;
 			};
 			name = Release;
 		};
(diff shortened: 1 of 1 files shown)


thomasvl:
This seems a lot like #1043 and at the moment this library isn't making most of the promises that would be implied via enabling this, no?

Lukasa:
I think that’s correct. Adding a stable ABI makes this module tricky to evolve in a number of ways. I’d like to see compelling cases of why this is necessary.

mingchen:
Thanks for the info @thomasvl.

One  reason I want to enable module stability is we use carthage to build the our library,  we release our library as binary. our library have a dependence on swiftprotobuf.  So we want can also distributed as binary  along with our library. The goal is those  binaries can  work with different version of xcode. We donot need the  promise of library  evolution.

For example, we released our libfoo 1.2.3 with swiftpotobuf 1.10.0. Then we released libfoo 1.3.0 with swiftprotobuf 1.12.0. When app upgrade from old libfoo 1.2.3 to libfoo 1.3.0 will also  pick up swiftprotobuf 1.12.0 binary. The  app may build with xcode 11, xocde 11.x or xcode 12.x, it is out of control. We want our release can works on all version of  xcode,  so we need module stability setting in library.

tbkka:
What will happen if the application also uses SwiftProtobuf directly?  Including a SwiftProtobuf binary along with your library could cause problems in that case.

mingchen:
@tbkka We donot embedded the binary of SwiftProtobuf in our library directly. When we release we provide both our library and pre build SwiftProtobuf  with module stability, so different version of xcode can just use it. There is only one SwiftProtobuf  in app.

Lukasa:
I believe what @tbkka is driving at is that the application could _start_ to use SwiftProtobuf, at which point there would be conflicts.

tbkka:
If you'd like to continue pursuing this, please rebase against the new `main` branch and re-submit. Thank you!

mingchen:
Thank you @tbkka. The discussion is helpful. I think keep as-is make sense.

PR #1080, opened 2020-10-22
Update some generated files that were missed before

thomasvl:


thomasvl:
@tbkka looks like the checks aren't kicking off for some reason, any ideas?

PR #1094, opened 2020-11-18
Add -j to the protobuf build to try and speed it up on CI.

thomasvl:


thomasvl:
Odd, this works for me locally, not sure what's different here.  Dropping.

PR #1095, opened 2020-11-18
Swift 4.2 break

### LINKED ISSUE #1093: Fix up Swift 4.2 support
#1088 included a change that doesn't seem to build under Swift 4.2.  Need to tweak things so that still builds since it is a supported Swift version still.

thomasvl:
Fixes #1093


thomasvl:
The 4.2 check job has gotten past the point that was failing, so should be in better shape now.

allevato (approved):

PR #1097, opened 2020-11-19
Tune the CI builds of protobuf

thomasvl:
Build the explicit targets need instead of the default target (usually "all")

Add -j CPU count to the builds to build some things in parallel. Make doesn't
have a way to say scale to the CPU count, so look up the CPU count and
explicitly pass it.  The agents on github seem to all be dual core, so this
should help a little.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,7 +59,7 @@ jobs:
       run: make test-runtime
     - name: Cache protobuf
       id: cache-protobuf
-      uses: actions/cache@v1
+      uses: actions/cache@v2
       with:
         path: protobuf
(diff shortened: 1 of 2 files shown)


thomasvl:
Amended to also use `actions/cache@v2`, main thing being it is supposed to have some speed improvements.

thomasvl:
Looking at some of the recent action runs, building protobuf in the different images used to be 10-12 minutes, as the runs for this PR finish, they are now 5-6 min, so the that basically maps to the expected 2 cpu count on the VMs for github actions.

thomasvl:
Also filed protocolbuffers/protobuf#8069 for the fact that the conformance directory has issues building with `-j NUM_CPUS`.

toffaletti (approved):

PR #1109, opened 2021-02-12
Normalize CRLF in comments to avoid double spacing some input.

### LINKED ISSUE #1107: Multiline comments creating extra newline
Please be sure to include:

* macOS 10.15.7

* Xcode 12.3

* Swift 5.3

* what branch/tag of SwiftProtobuf you are using: 1.15.0

* no errors or warnings

Here is .proto file:
```
enum Events {
 /*
   * After attempts to do somethings, there are some things happening
   * that is too long to write in one line.
   */
  THINGS_HAPPENING = 1;
}
```

Here is .swift generated file:
```
enum Events {
  ///
  ///
  /// After attempts to do somethings, there are some things happening
  ///
  /// that is too long to write in one line.
  case thingsHappening // = 1
}
```

Expected result would be:
```
enum Events {
  /// After attempts to do somethings, there are some things happening
  /// that is too long to write in one line.
  case thingsHappening // = 1
}
```

Is it something I'm writing wrong in .proto file? Is there a way to set up comments parameters?

thomasvl:
Fixes #1107

### DIFF
--- a/Sources/SwiftProtobufPluginLibrary/Google_Protobuf_SourceCodeInfo+Extensions.swift
+++ b/Sources/SwiftProtobufPluginLibrary/Google_Protobuf_SourceCodeInfo+Extensions.swift
@@ -34,7 +34,11 @@ extension Google_Protobuf_SourceCodeInfo.Location {
 
     func prefixLines(text: String, prefix: String) -> String {
       var result = String()
-      var lines = text.components(separatedBy: .newlines)
+      // Protoc doesn't normalize newlines in the comments, make sure CRLF
+      // doesn't insert blank lines and the generated file is hopefully then
+      // consistent in using '\n'.
+      var lines =
+        text.replacingOccurrences(of: "\r\n", with: "\n").components(separatedBy: .newlines)
       // Trim any blank lines off the end.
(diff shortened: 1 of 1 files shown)


tbkka (approved):
That looks like a reasonable approach.

allevato:
Another fix would be to avoid Foundation's string methods entirely; in addition to potentially causing unnecessary bridging between `String` and `NSString` (though I haven't traced through where `text` comes from though to see if it would already be a foreign string on Darwin), the Foundation methods won't do any of the grapheme-cluster processing that one expects from Swift string processing, which is why this bug happened; `"\r"` and `"\n"` were treated as individual code units because that's all Foundation does. `CharacterSet`'s name is a lie in Swift because it's really a "`Unicode.Scalar set`".

The following would also work, because `String.split` operates in terms of `Character`s and `"\r\n"` is a single `Character`:

```
var lines = text.split(whereSeparator: \.isNewline)
```

(But we'd need to tweak the lines below a little bit too, since this will return a `[Substring]` instead of a `[String]`.)

thomasvl:
Yea, I did `split` at first thinking about moving to stdlib, but there's no good replacement for `trimmingCharacters(in:)`, so we're already bridging things in here (and other places).

PR #1117, opened 2021-04-13
Make a FuzzTesting directory with package file.

thomasvl:
On macOS this requires using a swift.org toolchain since the swift support in
Xcode doesn't include fuzz support.

Add a second stage to the 'build' CI workflow to ensure these targets stay
building also.


thomasvl:
@catenacyber -

 This should put some testing binaries with this checkout, and provides binaries for all three formats.

My build command lines for linux are slightly different, but these appear to run just fine, are the other args you provided needed in some specific cases?

allevato (approved):

thomasvl:
> My build command lines for linux are slightly different, but these appear to run just fine, are the other args you provided needed in some specific cases?

I've added added a second commit to tweak the build to more closely match what seems to be upstream.

FuzzTesting/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+This subpackage build binaries to be use with Fuzz testing.
+
+The Swift binaries distributed with Xcode do not include the fuzzing support, so
+for macOS, one need to install the swift.org toolchain and use that instead.
+
+To build on macOS:
+
+```
+xcrun --toolchain swift swift build -c debug -Xswiftc -sanitize=fuzzer,address -Xswiftc -parse-as-library
```
catenacyber, line 11: @thomasvl I wonder if `parse-as-library` and `-sanitize=fuzzer` could be part of Package.swift cf https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#target `swiftSettings`

The other settings could vary

catenacyber: I think we can build with release or debug.
But I think we get more readable stack traces with the debug one

thomasvl: I quick attempt to move the flags into the Package file didn't work, a bunch of related symbols didn't seem to be found when linking.

Yea, I switched to debug since it seemed to give better data.

tbkka: I'd love to test both debug and release builds someday.

thomasvl: Switching to debug actually caught an assert which was bad.

Lukasa: We’ve had trouble in the past with fuzzing in debug mode due to the reduced resolution of ASAN (many memory accesses happen in the stdlib, and in debug mode those memory accesses may not be emitted into your binary to be instrumented). I do think it will be necessary to do both: catching bad asserts is useful, but so is getting ASAN to trigger.

catenacyber: > I'd love to test both debug and release builds someday.

Adding this as part of https://github.com/google/oss-fuzz/pull/5612
And it seems to find different bugs

@Lukasa would you know if there is an equivalent of rust `-Z build-std` in swift ?

Lukasa: I am not aware of any such thing. I recommend posting on the Swift forums asking the Swift team for guidance: in principle this could be done by careful use of the OSS toolchains, but we could probably make it easier to do.

PR #1119, opened 2021-04-13
Protect against overallocation on bad input.

thomasvl:
Double/Float packed parse was trying to reserver array space before confirming
that there actually was enough data, valid that first, to better handle
malformed input.


thomasvl:
fyi - all the other types use varint encoding and those codepaths already get this sorta test.

tbkka (approved):
Ouch.  That's a good catch.

PR #1122, opened 2021-04-13
Avoid looping forever on malformed Map TextFormat.

thomasvl:


allevato (approved):

tbkka (approved):
Fuzzing is wonderful!

PR #1124, opened 2021-04-14
Document Debug/Release fuzzing. Fix octal TextFormat decoding failure

thomasvl:


Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -128,7 +128,11 @@ private func decodeString(_ s: String) -> String? {
             if let digit3 = bytes.next(),
               digit3 >= asciiZero && digit3 <= asciiSeven {
               let digit3Value = digit3 - asciiZero
-              let n = digit1Value * 64 + digit2Value * 8 + digit3Value
+              // The max octal digit is actually \377, but looking at the C++
+              // protobuf code in strutil.cc:UnescapeCEscapeSequences(), it
+              // decodes with rollover, so just duplicate that behavior for
+              // consistency between languages.
+              let n = digit1Value &* 64 &+ digit2Value &* 8 &+ digit3Value
```
tbkka, line 135: That's interesting.   Hmmm....  We need to get the protobuf folks to cobble up a conformance tester for TextFormat, I suppose.  We also should have a test case for this.

tbkka: (Or feel free to just commit this and I'll fill in tests later...)

FuzzTesting/do_build.sh
```diff
@@ -6,10 +6,47 @@ readonly FuzzTestingDir=$(dirname "$(echo $0 | sed -e "s,^\([^/]\),$(pwd)/\1,")"
...
 else
-  swift build -c debug -Xswiftc -sanitize=fuzzer,address -Xswiftc -parse-as-library
+  CMD_BASE=(
+    swift build -Xswiftc -sanitize=fuzzer,address -Xswiftc -parse-as-library
+  )
 fi
+
+for CMD_CONFIG in "${CMD_CONFIGS[@]}"; do
+  echo "Building: ${CMD_CONFIG}"
+  echo "${CMD_BASE[@]}" -c "${CMD_CONFIG}"
```
tbkka, line 50: I believe the `[@]` notation is a bash-ism, in case that matters here.

tbkka (approved):

Sources/SwiftProtobuf/TextFormatScanner.swift
thomasvl: I'm looking at doing a follow up to make it easier to make test cases out of stuff the fuzzer finds, so I'll try to push on that next and bring back some of the fixes from them.

FuzzTesting/do_build.sh
thomasvl: It is, but the file has `#!/bin/bash` at the top, so we're good.

PR #1125, opened 2021-04-14
Add unittest based on fuzz testing.

thomasvl:
Should provide the framework to easily debug and then check for regressions in
the future.


Tests/SwiftProtobufTests/Test_FuzzTests.swift
```diff
@@ -0,0 +1,124 @@
...
+//
+// Easily get the hex dumps via `xxd -i < _test_case_file_`
+
+import Foundation
+import XCTest
+
+class Test_FuzzTests: XCTestCase {
+
+  func assertBinaryFails(_ bytes: [UInt8], file: XCTestFileArgType = #file, line: UInt = #line) {
+    do {
```
allevato, line 29: Can these failing helpers be condensed down to the one-liner:

```swift
XCTAssertThrowsError(try Fuzz_Testing_Message(...), file: file, line: line)
```

Tests/SwiftProtobufTests/Test_FuzzTests.swift
```diff
@@ -0,0 +1,124 @@
...
+      }
+      let _ = try Fuzz_Testing_Message(textFormatString: str, extensions: Fuzz_Testing_FuzzTesting_Extensions)
+      XCTFail(file: file, line: line)
+    } catch {
+      // Error while parsing is expected.
+    }
+  }
+
+  func assertTextFormatSucceeds(_ textFormat: String, file: XCTestFileArgType = #file, line: UInt = #line) {
+    do {
```
allevato, line 78: Similarly, I think this could just be

```swift
XCTAssertNoThrow(try Fuzz_Testing_Message(...), file: file, line: line)
```

Tests/SwiftProtobufTests/Test_FuzzTests.swift
```diff
@@ -0,0 +1,124 @@
...
+      let _ = try Fuzz_Testing_Message(textFormatString: textFormat, extensions: Fuzz_Testing_FuzzTesting_Extensions)
+    } catch {
+      XCTFail(file: file, line: line)
+    }
+  }
+
+  func assertTextFormatSucceeds(_ asBytes: [UInt8], file: XCTestFileArgType = #file, line: UInt = #line) {
+    do {
```
allevato, line 86: And this could be

```swift
let str = XCTUnwrap(String(...), file: file, line: line)
XCTAssertNoThrow(try Fuzz_Testing_Message(...), file: file, line: line)
```

(But `XCTUnwrap` is only Xcode 11.0+, in case we're still supporting anything older.)

Tests/SwiftProtobufTests/Test_FuzzTests.swift
thomasvl: done

Tests/SwiftProtobufTests/Test_FuzzTests.swift
thomasvl: done

Tests/SwiftProtobufTests/Test_FuzzTests.swift
thomasvl: we still support/test with Swift 4.2, so…

allevato (approved):

PR #1135, opened 2021-04-16
Change oneof enforcement to allow null

thomasvl:


tbkka (approved):
Nice!  Much simpler than I expected.

thomasvl:
> Nice! Much simpler than I expected.

Yea, I look a peek thinking I was going to need to track some state in the decoder, but realized a small reordering should be able to do it.

PR #1137, opened 2021-04-20
Add modify operation to ExtensionFieldValueSet.

Lukasa:
During decoding, modifications are frequently made to the extension
values in ExtensionFieldValueSet. In particular for repeated extensions,
these modifications include modifications to copy-on-write data types.
Because the _modify accessor hasn't been standardised yet,
ExtensionFieldValueSet implements get/set only, forcing a modification
of existing data to be done by a get/set pair. This will always emit a
copy-on-write operation.

The result of the above is that decoding repeated extensions is a
quadratic operation, making it diabolically slow. oss-fuzz caught this
behaviour.

This patch adds a non-API modify function to ExtensionFieldValueSet and
uses it on all decoding paths. This code simply encapsulates a somewhat
delicate function invocation that manages to convince the Swift compiler
to use Dictionary's _modify accessor for the underlying operation. The
result is that the oss-fuzz sample runtime decreases from 140s to 5s on
my machine.

This patch will likely have miscellaneous performance improvements
across the board. It adds the oss-fuzz case for regression purposes as
well.


thomasvl (approved):

thomasvl:
Going ahead and merging, hopefully oss-fuzz will see this today and update accordingly.

PR #1144, opened 2021-04-29
Add 5.4 testing.

thomasvl:
Move conformance test to 5.4 since it is the latest.

PR #1145, opened 2021-04-30
Don't do characterwise-compares if not needed.

Lukasa:
When attempting to locate a field number for a JSON document, some code
paths will find themselves needing to strip `[` and `]` from a String.
Right now they do this by grabbing `first` and `last` from the String
and then comparing the `Character` to a `String` containing the given
characters.

This is a fairly heavyweight operation: it invokes unicode
normalisation, among other things. This means that with specially
crafted field names it is possible to cause us to spend a surprisingly
long time normalizing strings unnecessarily.

Given that we're looking for UTF-8 characters, and we know these strings
will already be UTF-8, the easier fix is to just do the check on the
UTF-8 view. The result is a vastly shortened runtime, taking the
duration of a clusterfuzz test case (also submitted with this patch)
down from 60s to 186ms: a 300x speedup.

Note that this patch is not terribly _rigorous_: I haven't gone
searching for other places where we do direct String comparisons that
can be done on the UTF-8 view. That would likely be a productive line of
enquiry for performance enhancement in the JSON codepath at the very
least.


tbkka (approved):
Nice!

thomasvl:
#1144 just fixed/tweaked the config for the checks failure.

PR #1147, opened 2021-05-05
Clear previous contents before decoding Any from TextFormat

tbkka:
The standard decoding path requires the previous byte value; when decoded
from TextFormat to a message, obtaining the byte value requires serializing
the message.  This is entirely unnecessary and can be a performance problem
if the previous value is time-consuming to serialize.

Credit: oss-fuzz project

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
@@ -136,7 +136,7 @@ extension Google_Protobuf_Any {
   }
 }
 
-extension Google_Protobuf_Any: _CustomJSONCodable {
+extension Google_Protobuf_Any {
   // Custom text format decoding support for Any objects.
   // (Note: This is not a part of any protocol; it's invoked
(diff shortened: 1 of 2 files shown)


Lukasa (approved):

Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
```diff
@@ -152,10 +152,16 @@ extension Google_Protobuf_Any: _CustomJSONCodable {
     } else {
       // This is not using the specialized encoding, so we can use the
       // standard path to decode the binary value.
+      // First, clear the fields so we don't waste time re-serializing
+      // the previous contents if a field gets repeated.
```
thomasvl, line 156: I'd probably expand this comment a little, this makes it sorta sound like it impacts a repeated Any field.  Instead you are trying to handle the case where the field name/number was repeated but is actually a single field, so the second occurrence will replace the first.

thomasvl: I figured out how to just edit it in place.

thomasvl (approved):

PR #1165, opened 2021-09-07
Handle skipping an object that ends after the open brace.

thomasvl:
- Add a test case.
- Include a case found via local fuzzing.


allevato (approved):

tbkka:
Nice catch!

thomasvl:
> Nice catch!

Fuzzing for the win. As soon as I tried the dictionaries locally this tripped.

PR #1166, opened 2021-09-07
Add dictionaries for help steer the fuzz testing.

thomasvl:


FuzzTesting/FuzzJSON.dict
```diff
@@ -0,0 +1,332 @@
...
+"\"wktTimestamp\":"
+"\"wktType\":"
+"\"wktUint32Value\":"
+"\"wktUint64Value\":"
+
+# Some fragments
+"[]"
+"{}"
+"\"\""
+"''"
```
allevato, line 317: AFAIK JSON doesn't accept single-quoted strings; should these be included? Likewise, `\x` and `\0` aren't valid in JSON, and `\/` isn't covered here.

thomasvl: Removed single quote.

Removed `\x` and `\0`.

Added `\/`.

allevato (approved):

tbkka:
I'm not sure how oss-fuzz works, but other fuzzing tools I've used will randomly remove characters from sample inputs.  So it might help to have some longer well-formed examples here as well.
I also disagree with Tony:  Fuzzing is mostly about testing malformed inputs -- so including things that aren't valid JSON (single quotes, etc) is probably worthwhile.

thomasvl:
I believe it uses them as input to help steer what it generated, so adding/removing using these in addition to random chunks.  I mainly was after helping guide it to feeding the fields with other things.

https://google.github.io/oss-fuzz/getting-started/new-project-guide/#dictionaries

https://llvm.org/docs/LibFuzzer.html#dictionaries

I'm also working on followup for providing some [seed corpus](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#seed-corpus) values also.  But the dictionaries seemed like easy wins to start with (and turned up the "any" issue I just fixed on the first run.

PR #1169, opened 2021-09-09
When skipping a varint for an unknown field, ensure it is valid.

thomasvl:


Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -1288,20 +1288,8 @@ internal struct BinaryDecoder: Decoder {
...
-                available -= 1
-                if available < 1 {
-                    throw BinaryDecodingError.truncated
-                }
-                c = p.load(fromByteOffset: 0, as: UInt8.self)
-            }
-            p += 1
-            available -= 1
+            // Don't need the value, just ensuring it is validly encoded.
+            let _ = try decodeVarint()
```
tbkka, line 1292: It's always nice when the more-correct code is also simpler.  ;-)

tbkka (approved):

Sources/SwiftProtobuf/BinaryDecoder.swift
thomasvl: Yea, I started fixing it up to track length and then realized it wasn't worth a second copy that did everything but collect the value.

PR #1173, opened 2021-09-15
Sort some files correctly

tbkka:
We generally try to keep these in at least roughly sorted order to make it easier to find a particular file.


thomasvl:
Doh, sorry about that.

thomasvl (approved):

PR #1174, opened 2021-09-15
Simple test for recursive Any

tbkka:
Ideally, it would be possible to run this test
in a reasonable time with a nesting of 20 or more.

The recursion is set to only 5 for now since we
have some exponential costs in here.

Note: recursive Any should be vanishingly rare, so the
performance here probably isn't a critical short-term
concern.


thomasvl:
> The recursion is set to only 5 for now since we
> have some exponential costs in here.

I hope it isn't just the fact that we lazy transform, so it have to happen over and over through the recursion?

thomasvl (approved):

PR #1178, opened 2021-09-16
Rework JSON skipping so it's not recursive for nested arrays

tbkka:
The previous implementation was recursive for both objects `{...}`
and for arrays `[...]`.  As a result, we could break the stack for
deep nesting of either type.  Since we only checked nesting for
objects, this left us vulnerable to deeply nested JSON arrays.

Since this is in a "skip" context, we have no semantic information (so
we don't know whether a particular JSON array is a proto repeated
field or a ListValue wrapper), so we cannot accurately match the
recursion depth checks for binary or TextValue, which consider a
ListValue as an object structure that's handled recursively.

Skipping is used for Any bodies and for unknown fields, so this
addresses issues with both of those.

Note: This does allow us to skip things that we cannot parse, but
this is true in general:  For example, a client that does not
know a particular field can correctly skip it as long as it
is well-formed, but a client that does know a particular field
may error because the field doesn't fully meet expectations.

Thanks to oss-fuzz for finding this!


Tests/SwiftProtobufTests/Test_JSONDecodingOptions.swift
```diff
@@ -76,6 +76,9 @@ class Test_JSONDecodingOptions: XCTestCase {
             (true, "{\"unknown\":  7.308e+8}"),
             (true, "{\"unknown\": \"hi!\"}"),
             (true, "{\"unknown\": []}"),
+            (true, "{\"unknown\": [[]]}"),
+            (true, "{\"unknown\": [[[]]]}"),
+            (true, "{\"unknown\": [[],[[]],[[[]],[]],[],7]}"),
```
thomasvl, line 81: Maybe throw in a case with nested values?  `[[[1, 2]]]`

And a cases with values before/after the nested arrays: `[[1,[[]],3]]`

And do you want a test some place to ensure malformed nested is trapped and doesn't crash?  `[[[1,]]]`, `[[[,1]]]`, `[[[],]]`, `[[,[]]]`?

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -1378,6 +1378,12 @@ internal struct JSONScanner {
     return false
   }
 
+  /// If the next non-whitespace character is "]", skip it
```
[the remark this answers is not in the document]
tbkka: Fixed.

Tests/SwiftProtobufTests/Test_JSONDecodingOptions.swift
tbkka: I added a bunch more cases here, including a bunch that fail parsing.

thomasvl (approved):
👍

PR #1179, opened 2021-09-16
Track recursion depth for nested JSON ListValue structures

tbkka:
This prevents stack overrun for inputs like

   [[[[[[....[[[[[[]]]]]]....]]]]]]

Test case exercises direct nesting up to a million
levels.

Thanks to oss-fuzz for finding this!

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_ListValue+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_ListValue+Extensions.swift
@@ -44,6 +44,8 @@ extension Google_Protobuf_ListValue: _CustomJSONCodable {
       return
     }
     try decoder.scanner.skipRequiredArrayStart()
+    // Since we override the JSON decoding, we can't rely
+    // on the default recursion depth tracking.
     try decoder.scanner.incrementRecursionDepth()
     if decoder.scanner.skipOptionalArrayEnd() {
       decoder.scanner.decrementRecursionDepth()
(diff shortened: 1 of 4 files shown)


Sources/SwiftProtobuf/Google_Protobuf_ListValue+Extensions.swift
```diff
@@ -44,14 +44,17 @@ extension Google_Protobuf_ListValue: _CustomJSONCodable {
       return
     }
     try decoder.scanner.skipRequiredArrayStart()
+    try decoder.scanner.incrementRecursionDepth()
```
thomasvl, line 47: Probably worth a comment in this function to call out why manual recursion management is done for anyone looking at the code in the future.

Tests/SwiftProtobufTests/Test_Struct.swift
```diff
@@ -143,6 +143,41 @@ class Test_JSON_ListValue: XCTestCase, PBTestHelpers {
         assertJSONDecodeSucceeds("[true]") {$0.values == [Google_Protobuf_Value(boolValue: true)]}
     }
 
+    func test_JSON_nested_list() throws {
+        let depths = [
```
thomasvl, line 147: Since you avoid hardcoding the limit when do the parse test, maybe something like:

```
let depthLimit = JSONDecodingOptions().messageDepthLimit
let depths = [
            // right around the limit
            depthLimit - 1, depthLimit, depthLimit + 1,
            // Really big, should fail cleanly (not crash)
            1000,10000,100000,1000000,
]
```

thomasvl (approved):

Sources/SwiftProtobuf/Google_Protobuf_ListValue+Extensions.swift
tbkka: Done.

Tests/SwiftProtobufTests/Test_Struct.swift
tbkka: Done.

thomasvl (approved):

thomasvl:
Going ahead and merging this so oss-fuzz can validate things.

PR #1180, opened 2021-09-17
Add a dict to help steer binary fuzzing.

thomasvl:
Since we don't have descriptors, just a simple regex to extract from the proto
file and then generated all the byte runs for the fields.


FuzzTesting/make_FuzzBinary_dict
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+  }
+  print("Error: failed to match: \(fieldName)")
+  exit(1)
+}
+
+print(#"""
+
+# Include of the different values
```
allevato, line 119: s/of/some other word/?

These are supposed to be examples of the different wire-formatted primitives, right? (varints of different length, 32/64-bit ints/floats, strings/bytes?)

allevato (approved):

FuzzTesting/make_FuzzBinary_dict
thomasvl: Yup, updated and commented.

PR #1185, opened 2021-09-20
Reject fieldmasks that have non-ASCII characters for JSON

### LINKED ISSUE #1184: Can't JSON encode something the library JSON Decoded
Another fuzz testing find: https://oss-fuzz.com/testcase-detail/4827369236463616

Given the input `{"wktFieldMask":"{H̱ܻ̻ܻ̻ܶܶAܻD"}`, when the tester tries to generate JSON out of it, it fails and throws.  If we accepted it.

So odds are something in the parse isn't validating the same way and should have failed to decode.

tbkka:
Fieldmasks refer to proto fields which can only have ASCII
names, so any fieldmask with a non-ASCII character is
necessarily ill-formed.

This rejects such field masks when encoding and decoding JSON.
In particular, this fixes a round-trip decode/encode failure
that occurred when certain accented Unicode characters appeared
in JSON fieldmask arguments.

Resolves #1184

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
@@ -20,6 +20,9 @@ private func ProtoToJSON(name: String) -> String? {
   var jsonPath = String()
   var chars = name.makeIterator()
   while let c = chars.next() {
+    if !c.isASCII {
+        return nil // Reject anything with a non-ASCII character
+    }
     switch c {
     case "_":
       if let toupper = chars.next() {
@@ -44,6 +47,9 @@ private func ProtoToJSON(name: String) -> String? {
 private func JSONToProto(name: String) -> String? {
   var path = String()
   for c in name {
+    if !c.isASCII {
+        return nil // Reject anything with a non-ASCII character
+    }
     switch c {
     case "_":
       return nil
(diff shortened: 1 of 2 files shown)


Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -20,6 +20,9 @@ private func ProtoToJSON(name: String) -> String? {
   var jsonPath = String()
   var chars = name.makeIterator()
   while let c = chars.next() {
+    if !c.isASCII {
```
thomasvl, line 23: maybe `guard` since both usage of this is `if !`?

thomasvl: What about `{`, `$`, etc. should those be allowed?  i.e. - should we clamp this down even more?

tbkka: Apparently, the full syntax supported in C++ includes some compressed encodings that I'm not entirely familiar with.  So I'd rather not clamp down too tightly until I better understand that.

`guard` is a good idea.  I need to first find an `isASCII` equivalent that works in Swift 4.2.

tbkka: Reworked the ASCII detection to a form that should be compatible with Swift 4.2

tbkka: I also sketched a next step to filtering the allowed characters more tightly.

thomasvl (approved):

PR #1188, opened 2021-09-21
Switch to the latest image again, as hopefully things are fixed.

thomasvl:


thomasvl:
@Lukasa fyi - looks like it passed, not sure if you want to go back and update that Swift bug also.

PR #1190, opened 2021-09-22
Sync up the scheduled conformance and build workflows on installed th…

thomasvl:
…ings.


thomasvl:
And we still need the one exception for the older swift release.

PR #1192, opened 2021-09-29
Make it compilable using SwiftWasm toolchain

lowkahonn:
[SwiftWasm](https://swiftwasm.org/) allows us to bring Swift code to browser by compiling it to WebAssembly, with some [limitations](https://book.swiftwasm.org/getting-started/foundation.html), e.g. not being able to use Stream and Dispatch.

This PR added compiler directives to the file that uses `Dispatch` and `Stream`. One uncertainty of this PR is the approach of adding compiler flag to `BinaryDelimited.swift`, and this is basically removing the file for WASI target, and not sure what could happen to it


Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
```diff
@@ -47,9 +46,12 @@ internal func typeName(fromURL s: String) -> String {
   return String(s[typeStart..<s.endIndex])
 }
 
+#if !os(WASI)
+import Dispatch
```
thomasvl, line 50: Can you use `canImport` for these gates instead of the `os(WASI)` check?

I think I'd rather keep the import at the top gated also so it is less hidden.

Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
```diff
@@ -105,7 +107,7 @@ extension Google_Protobuf_Any {
     @discardableResult public static func register(messageType: Message.Type) -> Bool {
         let messageTypeName = messageType.protoMessageName
         var result: Bool = false
-        knownTypesQueue.sync(flags: .barrier) {
+        let block = {
```
thomasvl, line 110: I wonder if there's some way to use a file local helper with all the conditionals that takes a closure to keep the main part of the code simple and hide the details.

lowkahonn:
@thomasvl I updated the code as per your comments, can you please review it again?

thomasvl:
Looks pretty reasonable.

@tbkka - ok with you?

PR #1199, opened 2021-11-15
Swift format lint issues

thomasvl:


allevato (approved):
Wow, I can't believe we still had those Swift 2-era enum case names in there.

PR #1202, opened 2021-11-18
Tweak generation for messages with only extension ranges.

thomasvl:
If a message has no fields, `decodeMessage` would generate an `if` instead of a
`swift..case`. If there were a lot of extension ranges, the `if` would be very
complex, and it used raw integers in the source. With enough ranges, this
becomes a problem for compilation. All the the range values could instead use
explicit casts:

  ```
  Int(100) <= fieldNumber && fieldNumer < Int(201)
  ```

But to generally keep the code slightly more readable, just use the
`switch..case` instead of there are more then a few ranges, thus avoiding the
compilation performance issue.

Note: It may be worth revisiting the general fact that extension ranges are
enforced as a gate before doing extension lookup, as removing these gates might
shrink the generated code (and compiled) sizes, but it would have to be
evaluated against the potential costs of extension lookups for what are
otherwise known to be unknown fields.

### DIFF
--- a/Sources/protoc-gen-swift/MessageGenerator.swift
+++ b/Sources/protoc-gen-swift/MessageGenerator.swift
@@ -279,8 +279,12 @@ class MessageGenerator {
       generateWithLifetimeExtension(printer: &p, throws: true) { p in
         p.print("while let \(varName) = try decoder.nextFieldNumber() {\n")
         p.indent()
-        if !fields.isEmpty {
-
+        // If a message only has extensions and there are multiple extension
+        // ranges, get more compact source gen by still using the `switch..case`
+        // code. This also avoids typechecking performance issues if there are
+        // dozens of ranges because we aren't constructing a single large
+        // expression containing untyped integer literals.
+        if !fields.isEmpty || descriptor.extensionRanges.count > 3 {
           p.print(
               "// The use of inline closures is to circumvent an issue where the compiler\n",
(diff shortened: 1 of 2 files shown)


Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -279,8 +279,11 @@ class MessageGenerator {
       generateWithLifetimeExtension(printer: &p, throws: true) { p in
         p.print("while let \(varName) = try decoder.nextFieldNumber() {\n")
         p.indent()
-        if !fields.isEmpty {
-
+        // If a message only has extensions and there are multiple of extension
+        // ranges; get more compact codegen by still using the `switch..case`
+        // code. This also avoids typechecking perforamnce issues if there are
+        // dozens of ranges due to to the comparisions to the literals.
```
allevato, line 285:
```suggestion
        // If a message only has extensions and there are multiple extension
        // ranges, get more compact source gen by still using the `switch..case`
        // code. This also avoids typechecking performance issues if there are
        // dozens of ranges because we aren't constructing a single large
        // expression containing untyped integer literals.
```

allevato (approved):

Sources/protoc-gen-swift/MessageGenerator.swift
thomasvl: done

PR #1206, opened 2022-01-25
Conformance JSON.OneofFieldNullSecond

tbkka:
We previously decoded oneof fields with this template:
    
```
if _storage._oneofField != nil {try decoder.handleConflictingOneOf()}
var v: UInt32?
try decoder.decodeSingularUInt32Field(value: &v)
if let v = v {
  _storage._oneofField = .oneofUint32(v)
}
```

That is, if the field was already set, treat it as a conflict.
For JSON, the conformance suite now expects the following to be
handled identically:
```
   {"oneofString":"test", "oneofUint32": null}
   {"oneofUint32": null, "oneofString":"test"}
```
This makes sense, since JSON decoding should not depend on
order.  But the template above handles these differently; the
first is treated as a conflict and the second is not.

To address this, I've changed the oneof decoding template to
look like this:
```
var v: UInt32?
try decoder.decodeSingularUInt32Field(value: &v)
if let v = v {
  if _storage._oneofField != nil {try decoder.handleConflictingOneOf()}
  _storage._oneofField = .oneofUint32(v)
}
```
With this template, we eagerly decode the field, then
consider it a conflict only if the field was already set
_and_ the field decoded to a non-nil value.

### DIFF
--- a/Reference/conformance/conformance.pb.swift
+++ b/Reference/conformance/conformance.pb.swift
@@ -585,32 +585,40 @@ extension Conformance_ConformanceRequest: SwiftProtobuf.Message, SwiftProtobuf._
       // enabled. https://github.com/apple/swift-protobuf/issues/1034
       switch fieldNumber {
       case 1: try {
-        if self.payload != nil {try decoder.handleConflictingOneOf()}
         var v: Data?
         try decoder.decodeSingularBytesField(value: &v)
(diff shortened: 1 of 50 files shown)


tbkka:
Hmmm...  Seems I was out of date, and Thomas already fixed this one.  ;-)

PR #1208, opened 2022-02-07
Add Sendable conformances

zwaldowski:
- Implicit `Sendable` conformance for all `Message` and `Enum`.
- Generated `Sendable` conformance for `oneof` enums.
- Generated `@unchecked Sendable` for messages with heap storage.
- Explicit `Sendable` conformance for extension-related tooling.

Notes:
* Tested by applying the changes to a medium-sized codebase that has pervasive `oneof` and heap storage, but few uses of advanced features like extensions or RPC.
* The code is formulated to be pre-5.5 compatible, while avoiding putting two copies of the entire type definition inside `#if`. However, I don't have an older compiler handy to check.
* I left types and protocols used purely during serialize/deserialize alone, like `ExtensionMap`, `_NameMap`, `[Blah][De|En|codingOptions`, etc.
* I didn’t bump the compatibility version since it doesn’t technically create API or ABI incompatibilities — just warnings, for now. I don’t know if this was the right call.


tbkka:
I think the compatibility version does need to be bumped: That version flags incompatibilities between the SwiftProtobuf library and the generated code.  Now the generated code requires `_ProtoSendable` which isn't present in old libraries.

tbkka:
@ktoso pointed me to [the Swift Concurrency Adoption Guidelines](https://github.com/swift-server/guides/blob/main/docs/concurrency-adoption-guidelines.md#if-guarding-code-using-concurrency) for the most accurate way to conditionally adopt concurrency.

FuzzTesting/Sources/FuzzCommon/fuzz_testing.pb.swift
```diff
@@ -5392,6 +5392,11 @@ extension Fuzz_Testing_Message: SwiftProtobuf.Message, SwiftProtobuf._MessageImp
   }
 }
 
+#if swift(>=5.5)
```
ktoso, line 5395: You might want to follow the guide from https://github.com/swift-server/guides/blob/main/docs/concurrency-adoption-guidelines.md#if-guarding-code-using-concurrency on these.

I.e. we suggest doing 

```
#if compiler(>=5.5) && canImport(_Concurrency)
```

zwaldowski: The section [Backwards compatibility of declarations and “checked” Swift Concurrency](https://github.com/swift-server/guides/blob/main/docs/concurrency-adoption-guidelines.md#backwards-compatibility-of-declarations-and-checked-swift-concurrency) says to use `swift(>=5.5)` instead. I'm happy to drop in the `canImport`, however.

zwaldowski:
> I think the compatibility version does need to be bumped: That version flags incompatibilities between the SwiftProtobuf library and the generated code. Now the generated code requires _ProtoSendable which isn't present in old libraries.

@tbkka That's what I was missing, thanks. Easy enough.

FuzzTesting/Sources/FuzzCommon/fuzz_testing.pb.swift
zwaldowski: Just out of curiosity, what's the failure case for not having `canImport`, since `Sendable` lives in the stdlib?

tbkka:
I'm surprised that [Sources/SwiftProtobuf/struct.pb.swift](https://github.com/apple/swift-protobuf/pull/1208#diff-693dec7a7530cd67050603766fad4b63ac293ebbfec295bd6029c8d21d4d8b34) doesn't show `Google_Protobuf_Value` and the other types being _ProtoSendable.  Did the regenerate miss something?

zwaldowski:
> Did the regenerate miss something?

It comes from `Message` without code changes.

thomasvl:
> I think the compatibility version does need to be bumped: That version flags incompatibilities between the SwiftProtobuf library and the generated code. Now the generated code requires `_ProtoSendable` which isn't present in old libraries.

Can the `Sendable` conformance just be added in a `#if` gated extension and skip having to define a library specific shim like this?

tbkka:
@thomasvl has a good point:  If `_ProtoSendable` could be made entirely private to the library, then we wouldn't need the compatibility bump.

thomasvl:
Assuming you can add the conformances via extensions, you could add all the conformances in one `#if` block at the end of the file, then the comment about storage doesn't have to be repeated/etc.  That might help keep the impact in the generated files down?

zwaldowski:
Using `@unchecked Sendable` in all the generated code is a fine ask, but one would still want `Sendable` as part of the protocols for use in generics (i.e., `<T: Message>`), and the alias would then still need to be public.

zwaldowski:
> Assuming you can add the conformances via extensions

Yeah, you can in the declaring file. I'll can try that out, but do consider my reply [here](https://github.com/apple/swift-protobuf/pull/1208#issuecomment-1032013283).

Sources/SwiftProtobuf/Data+Extensions.swift
```diff
@@ -32,3 +32,7 @@ internal extension Data {
     }
 }
 #endif
+
+#if swift(>=5.5) && !swift(>=6.0)
+extension Data: @unchecked Sendable {}
```
thomasvl, line 37: Actually, should we really be globally adding to `Foundation`s `Data`?

zwaldowski: I don't think it's global, based on my reading, but let me generate some `swiftinterface` files and poke at them.

FuzzTesting/Sources/FuzzCommon/fuzz_testing.pb.swift
zwaldowski: The question is also made relevant and not a curiosity by @thomasvl's ask to group up conformances in the generated code.

zwaldowski:
> …and the alias would then still need to be public.

But, still, internal to the library in that case, so maybe the version wouldn't need to be bumped. I'll explore that.

thomasvl:
> Using `@unchecked Sendable` in all the generated code is a fine ask, but one would still want `Sendable` as part of the protocols for use in generics (i.e., `<T: Message>`), and the alias would then still need to be public.

The library types could also use extensions so you'd still get this support, no?

zwaldowski:
> The library types could also use extensions so you'd still get this support, no?

Only on concrete types. `extension SomeProtocol: @unchecked Sendable {}` is not valid syntax, that would still need a public alias, but that would avoid bumping the compatibility version as I noted [here](https://github.com/apple/swift-protobuf/pull/1208#issuecomment-1032016063). I added a task to the PR to follow through on this.

Sources/SwiftProtobuf/Data+Extensions.swift
thomasvl: I'd have though this might have needed to be public for a Swift 5.5 build with a Message containing a `bytes` field (`Data`) to be happy since you had to do extra work for the Storage cases.

zwaldowski: You are correct in your reasoning, but because `Sendable` doesn't have an ABI impact I'm not certain it has an observable impact outside of this module.

zwaldowski: Resolved by removal

zwaldowski: I was, though, able to confirm my hypothesis that two retroactive conformances to `Sendable` published by two separate modules doesn't break — _unlike_ normal Swift. We won't make use of that, but it's a neat lesson.

FuzzTesting/Sources/FuzzCommon/fuzz_testing.pb.swift
ktoso: That trick is necessary to survive situations in which one would build this project with a NEW compiler (5.5+) but targeting an SDK that does not have concurrency.

Sources/SwiftProtobuf/UnknownStorage.swift
```diff
@@ -44,3 +44,7 @@ public struct UnknownStorage: Equatable {
     }
   }
 }
+
+#if swift(>=5.5) && canImport(_Concurrency)
+extension UnknownStorage: @unchecked Sendable {}
```
Lukasa, line 49: It may be worth us commenting here that the only reason this is `@unchecked` is that `Data` does not consistently conform to `Sendable`, and when we only support Swift versions where it does we could drop the `@unchecked`.

Lukasa (approved):
Basically LGTM, one minor suggestion around a code comment but I'm otherwise happy.

Sources/SwiftProtobuf/UnknownStorage.swift
thomasvl: Could/Should we tweak the ones generated for the Message types to only include `@unchecked` if they have a `Data` field directly?

FuzzTesting/Sources/FuzzCommon/fuzz_testing.pb.swift
zwaldowski: > That trick is necessary to survive situations in which one would build this project with a NEW compiler (5.5+) but targeting an SDK (mac, ios etc) that does not have concurrency.

Perfect, thanks!

zwaldowski: > I can definitely improve the wording if you found it confusing though

The document wasn’t confusing but it contradicts your suggestion. Your initial comment said to use `compiler(>=5.5.)` (“i.e., we suggest doing”) and linked to the document. The document says to use `compiler(>=5.5)` for concurrency declarations in code, then later says `swift(>=5.5)` for adding `Sendable` conformances. After updates, this PR now follows what the document says: `#if swift(>=5.5) && canImport(_Concurrency)`.

Sources/SwiftProtobuf/UnknownStorage.swift
zwaldowski: The limitation also applies to generated `oneof` enums that have `Data` members and extensible messages. The protocol hierarchy for extensions is otherwise `Sendable`-compatible until you get to `ProtobufBytes`. (Aside that I’m sure has been discussed elsewhere, but I hear `[UInt8]` is really nice this time of year.)

There’s also messages that use heap storage. Like `Array` In the stdlib, we can’t prove that we use COW to the compiler. Generation of `@unchecked Sendable` will likely have to remain for those.

Would you prefer a single comment that addresses all four of the limitations? “One or more of the blah types cannot blah, because blah”?

zwaldowski: Ooops, I see that this comment is on `UnknownStorage` and not the generated code. Trivial add to the docs 😅

Sources/SwiftProtobuf/ProtoSendable.swift
```diff
@@ -0,0 +1,17 @@
...
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+
+#if swift(>=5.5) && canImport(_Concurrency)
+/// SwiftProtobuf Internal: Compatibility alias for indicating a value can be safely used in concurrent code.
+public typealias _ProtoSendable = Swift.Sendable
+#else
+/// SwiftProtobuf Internal: Compatibility alias for indicating a value can be safely used in concurrent code.
```
tbkka, line 15: You should have an additional comment here that `Any` makes this definition essentially a no-op when using a compiler that does not support concurrency.

tbkka:
Do we still need `_ProtoSendable` at all?  It looks like `Enum` and `Message` are the only places still using it.  Would the following suffice?
```
#if swift(>=5.5) && ...
extension Enum: Sendable {}
#endif
```

zwaldowski:
> Do we still need `_ProtoSendable` at all? It looks like `Enum` and `Message` are the only places still using it. Would the following suffice?
> 
> ```
> #if swift(>=5.5) && ...
> extension Enum: Sendable {}
> #endif
> ```

That's not valid syntax. Discussed [here](https://github.com/apple/swift-protobuf/pull/1208#issuecomment-1032018792).

Sources/SwiftProtobuf/UnknownStorage.swift
zwaldowski: > Could/Should we tweak the ones generated for the Message types to only include `@unchecked` if they have a `Data` field directly?

Just to close the loop re: generated types, the `Data` limitation also applies to extensible messages and `oneof` enums. Then, even with that, messages with heap storage will always use `@unchecked`. So that logic would get more and more complex. I'm happy to do that — I'd rather have `Sendable` do what it's supposed to — but I want to call it out as a maintenance cost.

zwaldowski: Cool. Let me add a comment to the generator.

tbkka (approved):

tbkka:
At this point, unless @thomasvl has additional concerns, I think we're ready to go with this.  Nice work!

thomasvl (approved):

thomasvl:
Nice and thank you!

PR #1209, opened 2022-02-08
Fix typo in PLUGIN.md referring to --swift_opt with a dash instead of an underscore

Zedenem:
Noticed this typo while familiarizing myself with `swift-protobuf`.

Hope this helps.


thomasvl (approved):
Thanks!

PR #1211, opened 2022-02-10
Fix typo in PLUGIN.md with reference to --swift_out instead of --swift_opt

Zedenem:
Noticed this _other_ typo while familiarizing myself with `swift-protobuf`.

Hope this helps.


thomasvl (approved):
Thanks again.

(that's likely left over when we had to document putting them on the front of the _out arg because protoc didn't have the _opt support always working for plugins.)

Issue #1212, opened 2022-02-10
Sources generated before #1208 might not compile with newer runtime


thomasvl:
Stumbled into a case where some sources generated before #1208 landed fail to compile when compiled with the updated library. Seems to come down to the Data/Oneof support since we are adding sendable annotations in the new generated sources, the old generated sources don't have them, but the new runtime tries to claim all Message will be sendable, but without those annotations the compiler can't confirm it.

thomasvl:
@zwaldowski fyi

thomasvl:
Repo is easy enough:

1. Checkout current _main_
2. `git revert aaead67360cd4d046aabae0f57bac4b5e15d7628`
3. `swift build`

Get a bunch of issues like:

```
/Users/thomasvl/src/github/swift-protobuf/Sources/SwiftProtobuf/struct.pb.swift:127:14: error: stored property 'kind' of 'Sendable'-conforming struct 'Google_Protobuf_Value' has non-sendable type 'Google_Protobuf_Value.OneOf_Kind?'
  public var kind: Google_Protobuf_Value.OneOf_Kind? = nil
             ^
```

tbkka:
We've bumped the compatibility version before and the world did not actually fall apart.  ;-)

It's nice to avoid bumping the compatibility version when possible, but it doesn't seem possible.

thomasvl:
We might not have bumped it since hitting 1.0.  What are the semver rules for something like this?  No, we aren't changing any apis, but all the generated code won't compile.

Lukasa:
The best fix is probably to add a new protocol (`_ConcurrentMessage`?) that refines `Message` and `Sendable`, and conform message types in the codegen to that as well. That means that if you regen the code you get Sendable everywhere, if you don’t you don’t. You _dont_ automatically get `Message: Sendable` for other generics, but we really can’t avoid the code break there unless we can prevent people updating the runtime.

zwaldowski:
> The best fix is probably to add a new protocol (`_ConcurrentMessage`?) that refines `Message` and `Sendable`, and conform message types in the codegen to that as well. That means that if you regen the code you get Sendable everywhere, if you don’t you don’t. You _dont_ automatically get `Message: Sendable` for other generics, but we really can’t avoid the code break there unless we can prevent people updating the runtime.

[...]

PR #1213, opened 2022-02-11
Add Sendable without library changes

zwaldowski:
This reverts commit 0c727bd7e41384992ca09e9b6686f8bcfcec0356 to address #1212.


Sources/SwiftProtobuf/UnknownStorage.swift
```diff
@@ -44,8 +44,3 @@ public struct UnknownStorage: Equatable {
     }
   }
 }
-
-#if swift(>=5.5) && canImport(_Concurrency)
-// Once our minimum supported version has Data be Sendable, @unchecked could be removed.
-extension UnknownStorage: @unchecked Sendable {}
-#endif
```
tbkka, line 51: Do we need to revert this?

zwaldowski: One of the unit tests has a hand-written conformance to `Message` that isn't code-generated. This was added to allow that type to be `Sendable`, because that had become a requirement of `Message`. The workaround is no longer needed if we're not changing the API of the runtime library for the moment.

tbkka: I was thinking that we know we want all our types to be Sendable to the extent it's possible, even if the `Message` and `Enum` protocols cannot be made so just yet without breaking compatibility.

zwaldowski: > to the extent it's possible

Therein lies the rub. I agree with you that the support library should express `Sendable` in as much public API as it can, and it can't for almost all of its public API because of `Data`. I think leaving just the one conformance is misleading. It implies the support library has been otherwise audited, which it hasn't.

tbkka (approved):

Sources/SwiftProtobuf/UnknownStorage.swift
tbkka: I would like to leave comments in the library near `Message`, `Enum`, and `UnknownStorage`:
```
// TODO: This should be `Sendable` but we cannot do so yet without possibly breaking compatibility.
```
That will reduce the likelihood we miss these when we are able to break compatibility.

zwaldowski: Will do ASAP.

tbkka (approved):

Sources/SwiftProtobuf/UnknownStorage.swift
thomasvl: (without having double checked the raw code), I think the hand written code was actually a copy from one point in time to also ensure things still worked without regenerating.  So UnknownStorage might not need this if we do the version bump because that whole test with the copy of the generated code might be obsolete.

Lukasa (approved):

PR #1221, opened 2022-03-07
Add sendable to core library

thomasvl:


zwaldowski:
Yeah, I suppose the `TODO` on `UnknownStorage` is a duplicate; that's easily resolved.

> what was the intent in the other cases?

The others are all related to `ExtensionFieldValueSet`. `ExtensionFieldValueSet` is a requirement of `ExtensibleMessage`, so moving towards a world of `Sendable` instead of `@unchecked Sendable`, it should be `Sendable`. That family of types can't participate in the `_ProtoSendable` trick because `ProtobufBytes` has an associated type of `Data`, which can't satisfy an associated type requirement for `FieldType.BaseType: _ProtoSendable` for the same reasons involving `Data` elsewhere.

I don't think there's anything to practically be done about those TODOs until `Data` is `Sendable` or it's replaced with a non-Foundation `ContiguousBytes` type that is `Sendable`.

thomasvl:
Thanks, updated to remove the comment on `UnknownStorage`, but leaving the rest.

thomasvl:
@zwaldowski & @Lukasa - if you can also look things over, it would be appreciated.

Lukasa (approved):

Lukasa:
Seems reasonable to me.

Issue #18, opened 2016-09-28
Look at how to factor out JSON/text support


allevato:
JSON/text support is currently provided directly in the same generated message as binary support. In some use cases, a client may not need text/JSON support, so it simply adds bloat:

* A user wants to use protobuf to represent their application's data model on disk and only needs binary coding.
* A user is speaking directly to a service that communicates in binary PB format on the wire for transmission efficiency.

It should be possible to slice JSON/text serialization support out of the standard generated message and move them into extensions, perhaps even in separate files (e.g., foo.pb.swift, foo.pb+json.swift, foo.pb+text.swift; names bike-sheddable).

This would let clients decide which support they need, without adding complexity to the generator with extra switches or options. A user who only needs binary support just takes the .pb.swift files; someone who needs JSON support would also include the .pb+json.swift files, and so forth.

Even in an application that only uses binary encoding for storage/transmission, text format can be useful for debugging. Factoring it out into separate files has the benefit of letting users only include it in their debug/testing builds, and leave production builds slim.

Since the amount of generated code to support these additional formats is not insignificant, this would prevent users from shipping unnecessary bloat (dead stripping can be unpredictable).

tbkka:
If I understand correctly, you're suggesting that we first implement Issue #9, and then:
  * Have the library provide default (empty) fieldNumber → name and name → fieldNumber maps
  * Generate extensions in separate files that override those maps with real information

It's an interesting idea.  I'm curious what the real savings are in code size.

allevato:
Looking more closely, I'm surprised to see that those number/name maps are computed instance properties instead of static properties. That could certainly be a source of performance issues, especially if it's re-creating those arrays every time the property is accessed.

My general thinking is, there's never a scenario where you need the union of *all* the information in the same pass, which is what #9 will solve. You're *either* serializing JSON *or* text, but never both simultaneously. And when you're *not* doing JSON or text serialization, you don't need the names at all—in other words, the core `ProtobufMessage` protocol doesn't have to contain any of that information at all, because the extensions can add conformances for specific visitations that provide that information instead.

[...]

PR #1240, opened 2022-04-07
Split text generation into different files

thomasvl:
This is a draft towards #18.

If we want to go in this directions, there's still a few things needed to be done:
-  decide if we want a generation option to not to control the behavior.
-  update the Package.swift and podspec to expose different targets for the base library support vs. the addition text support, then update the generation accordingly so the runtime parts aren't needed if not using the text support generated files.
-  double check if the Any support code can be further split. This likely is the any storage becoming something with an interface so when it comes in from a _text_ for a different backing could be used vs. when it comes from binary.


thomasvl:
If we split the library code so the text support is optional there also, which should be "SwiftProtobuf"?  The full library with the text support or the main library with only the binary support?  i.e. - do we add something like "SwiftProtobufCore" or "SwiftProtobufTextSupport"?

tbkka:
I'm tempted to split it even finer, but I've not looked into how this actually would work in practice:
 * SwiftProtobufCore - Basic APIs only + binary coder/decoder
 * SwiftProtobufJSON - JSON coder/decoder
 * SwiftProtobufTextFormat - TextFormat coder/decoder
 * SwiftProtobuf - Umbrella that includes all of the above

"TextSupport" is likely to be confusing -- that looks like it only applies to TextFormat, not all text-based formats.

thomasvl:
The naming support for enums (in its current form), has both TextFormat and JSON support.  I believe trying to isolate that likely would make things larger in a bunch of cases when both are desired since it currently is optimized for overlapping cases to carry less data.

thomasvl:
@tbkka any more thoughts on the "TextSupport" vs. trying to split TextFormat and JSON, seems like that split would mean supporting both ends up with more bloat.  Also, do we isolate binary so it is optional?  One of the things I punted on originally was what to do with Any when things are optional, I guess all the conversion apis start failing if support for the needed types wasn't built in (just like if it wasn't registered)?

Issue #194, opened 2016-12-25
how can i use swift-protobuf to generate a swift class, not struct


cgwangding:
I want to use swift class, but i don't know how can i do.

mohpor:
Hi,
This seems to be a valid request.
In my use case, I need to make my `message` exports, conform to a protocol which of NSObjectProtocol. Unfortunately, structs cannot do that!

thomasvl:
So that's actually a huge change because the objects would no longer be value types, so it likely changes just about everything about how the library works.

Out of curiosity, why do you want the to be reference types instead?  What are you after?

mohpor:
I'm trying to use my messages in [IGListKit](https://github.com/Instagram/IGListKit) and for that, I need my messages to be `ListDiffable`.
Since yesterday, I have accomplished what I was trying to do by Boxing my messages inside an object, but it seems kind of interesting to me that such simple requirement could not be met by the decision that the team made and made messages Value Type. I mean, if you are going to use protocols everywhere, why not implementig it in such a way to support both types. and when in doubt, I use `Class` just to make sure it is forwad compatible. I know I'm out of depth here, but this is not the right way to do this.

allevato:
Value types and reference types in Swift have significant enough differences that even with the protocol abstractions, it's highly unlikely that allowing users to swap out `struct` for `class` in the generation would be a simple task—it would require a lot of engineering work.

Even _if_ we wanted to support that, it would spawn a number of other issues:

* Users of Swift protos coming from other frameworks/modules would have to be aware of whether they were generated as `struct`s or `class`es, because that decision has effects on performance and on usage (mutability, for one example).
* How would generated messages across different frameworks/modules interact? If framework A and B both have protos, and A wants to depend on B, does A have to be generated with the same value/reference nature as B? If they could be different, we'd have to ensure that every message `struct` could nest a `class` and vice versa, and behave correctly in all cases.
* Similarly, the runtime library is currently implemented with the assumption that messages are value types and that copying is achieved by simple value assignment. Classes would drop that assumption and creating copies would require explicit cloning support, complicating the message API.

Essentially, supporting both models is a great deal of work for very little gain and it's in direct opposition to one of the main goals of this project, which is to generate natural Swift value types that have all the mutability and safety guarantees that the language provides. Swift in general is modeled to push users toward value types instead of classes when possible, and generating messages as classes would move us in the wrong direction.

[...]

PR #1241, opened 2022-04-07
Simplify the code flow now that 5.0 is the minimum.

thomasvl:
Basically a revisit of #915.

- Remove the version checks.
- Remove `_merge` which was tagged `@usableFromInline` and instead put the guts
  in the leaf call from there other calls to get the same optimization.

### DIFF
--- a/Sources/SwiftProtobuf/Message+BinaryAdditions.swift
+++ b/Sources/SwiftProtobuf/Message+BinaryAdditions.swift
@@ -79,14 +79,9 @@ extension Message {
     options: BinaryDecodingOptions = BinaryDecodingOptions()
   ) throws {
     self.init()
-#if swift(>=5.0)
     try merge(contiguousBytes: data, extensions: extensions, partial: partial, options: options)
-#else
-    try merge(serializedData: data, extensions: extensions, partial: partial, options: options)
-#endif
   }
 
-#if swift(>=5.0)
   /// Creates a new message by decoding the given `ContiguousBytes` value
   /// containing a serialized message in Protocol Buffer binary format.
   ///
@@ -111,7 +106,6 @@ extension Message {
     self.init()
     try merge(contiguousBytes: bytes, extensions: extensions, partial: partial, options: options)
   }
-#endif // #if swift(>=5.0)
 
   /// Updates the message by decoding the given `Data` value containing a
   /// serialized message in Protocol Buffer binary format into the receiver.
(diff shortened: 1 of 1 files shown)


Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -167,38 +154,26 @@ extension Message {
   ///     `BinaryEncodingError.missingRequiredFields`.
   ///   - options: The BinaryDecodingOptions to use.
   /// - Throws: `BinaryDecodingError` if decoding fails.
-  @inlinable
```
cburrows, line 170: I went back to read @Lukasa 's great writeup in the comments of #914 in hopes of answering a question for myself, but I failed to find an answer. So, here goes: what changed such that we no longer need/want this generic method to be `@inlinable`, and are we giving up anything? Like, did generic non-`@inlinable` methods get better in 5.0+, or is the win small enough that we don't care?

I thought the value in `_merge` was not its reuse (which is gone) but rather the way it permitted calling modules to merely give us an `UnsafeRawBufferPointer` on their own terms.

Lukasa: I don't think anything has meaningfully changed to enable this. Swift's cross-module optimisation functionality _should_ be able to make this `@inlinable` unnecessary, but that is not on by default and somewhat painful to enable. So I think the thunk to `_merge` and the `@inlinable` remains useful.

thomasvl: Before there were <5.0 code paths that also called `_merge`, but once I removed all the `#if` checks we ended up with only this one spot calling `_merge`, so that's why I folded it together.  So I thought the only net change here was that the `try bytes.withUnsafeBytes()` frame would be the only net shift, and that likely would matter from a performance POV.  But rereading #194, I guess the issue is if the calling module has custom `ContiguousBytes` with `@inlinable` support for the `withUnsafeBytes` call?

Lukasa: Specifically it's if this is called with any type _other_ than `Data`, in which case this code path will be hit and we can inline the call to `withUnsafeBytes`.

thomasvl: Updated with a code reference to the original thread.

cburrows (approved):

Lukasa (approved):
Nice, LGTM.

PR #1246, opened 2022-05-09
Conform to CustomDebugStringConvertible for Debug only

### LINKED ISSUE #1245: Allow opting-out of CustomDebugStringConvertible to reduce binary size
This observation comes from researching https://github.com/apple/swift-protobuf/issues/1204.

It appears that `CustomDebugStringConvertible`'s contribution to the binary size adds up and is not insignificant.
A quick check shows that removing it from [`Message`](https://github.com/apple/swift-protobuf/blob/1.17.0/Sources/SwiftProtobuf/Message.swift#L34), [`SimpleExtensionMap`](https://github.com/apple/swift-protobuf/blob/1.17.0/Sources/SwiftProtobuf/SimpleExtensionMap.swift#L17), [`AnyExtensionField`](https://github.com/apple/swift-protobuf/blob/1.17.0/Sources/SwiftProtobuf/ExtensionFields.swift#L30) would save us `~800kB`.

For context, our project has around 1000 message definitions.
This was a modified version **1.17.0**,
compiled with `-Osize`,
Xcode 13.2.1 13C100, Swift 5.5.2, Big Sur

Since conforming to `CustomDebugStringConvertible` does not seem essential, perhaps it could could be tied to a compilation flag?
I'd like to hear some initial feedback @thomasvl before getting into details.

dflems:
Resolves #1245

This boxes all references to `CustomDebugStringConvertible` and `debugDescription` in `#if` checks to exclude them from the release binary. The existence of this conformance and default implementation on the `Message` protocol (and some other types) adds up quite considerably. Across our large iOS app, we can attribute 6.7 MB of the uncompressed binary size to generated swift protos. This change cuts that back by 780 KB (11.6%). Other efforts like #1240 will probably also help trim this back even further.

I tried to do this with as little diff as possible. Added some inline comments on some things I'm unsure about.

Special thanks to @bubski for the idea.


Sources/SwiftProtobuf/ExtensionFields.swift
```diff
@@ -39,6 +39,12 @@ public protocol AnyExtensionField: CustomDebugStringConvertible {
   var isInitialized: Bool { get }
 }
 
+#if DEBUG
+public protocol _AnyExtensionField: CustomDebugStringConvertible {}
+#else
+public protocol _AnyExtensionField {}
+#endif
```
dflems, line 46: Not sure if this adds any extra overhead. I couldn't find another good way to conditionally have protocol conformance.

Lukasa (commented):
I'll note here that this change can only be made in a major release, so we can only merge this into the proposed 2.0 release.

thomasvl:
My first thought (and haven't really thought about this) - Is it better to only have the conformance in `DEBUG` builds or is it better that the implementation be different in `DEBUG` builds?  i.e. - how would this impact other code?  Does this force folks to wrap all the places that would end up using this conformance also in guards?

bubski:
The [documentation](https://developer.apple.com/documentation/swift/customdebugstringconvertible) of `CustomDebugStringConvertible` states that:

>Because the `String(reflecting:)` initializer works for instances of any type, returning an instance’s `debugDescription` if the value passed conforms to `CustomDebugStringConvertible`, accessing a type’s `debugDescription` property directly or using `CustomDebugStringConvertible` as a generic constraint is discouraged.

which makes me believe that it should not be a problem, and in cases where engineers do rely on `CustomDebugStringConvertible` or `debugString` they should rather refactor their code to use `String(reflecting:)`/`debugPrint(_:)` (which would compile for both debug and release builds) instead of wrapping their code in `#if` checks.

---

As for having conformance for both configurations, but with different implementations, I suspect it not to change much, if anything.
As far as I understand the cost comes mainly from type metadata, rather than specifics of the implementation.
Each type then also implements `debugDescription` which only forwards to the default implementation shared between all of them.

Lukasa:
Whether engineers _should_ obey the rules above is largely immaterial IMO: this breaks the API. Code that used to work would update to this version and break, and while we can say that the engineers who did that work were holding `CustomDebugStringConvertible` wrong, they were doing it nonetheless.

We can choose to accept the breakage if we want to, but it would be better if we did this as part of a 2.0 release where that kind of breakage is expected.

thomasvl:
Now would be the time, since our next release off _main_ is going to be 2.0 since decided to do some of the breaks now.

bubski:
@Lukasa To be clear, I do agree that this, in its current form, is an API breaking change and should be treated as such. I didn't mean to suggest otherwise.
I only wanted to share some perspective on how reasonable/fitting this approach could be.

Sources/SwiftProtobuf/ExtensionFields.swift
dflems: Right, that should work fine.

Lukasa:
@bubski Fantastic. I think this approach is entirely fine by me, though it might be nice to be able to allow users to override this behaviour if they so choose by replacing the definition with `#if DEBUG || PROTOBUF_ENABLE_DEBUG_STRINGS`. Otherwise I'm +1 on the general shape.

bubski:
@Lukasa in case you haven't seen it, the reason we went with `DEBUG` _alone_ here were concerns around having custom compilation flags brought up in https://github.com/apple/swift-protobuf/issues/1245.
Having more granular control over this does sound pleasing, but I am not sure what would be most ergonomic to engineers.

thomasvl:
@tbkka @Lukasa I'm not sure if @allevato has any strong opinions, but personally, I'll defer to you all on landing something like this.

If we find a solution for #18 (finishing out #1240), it should get but would also raise the question where/how these methods should be implemented.

tbkka:
I think we should take this for 2.0.

Lukasa:
Agree.

allevato:
No concerns here.

thomasvl:
Looks like the sanitizer configs are the only things that catch building the tests in _release_. There's a bunch of tests that will need some work to deal with building in both _debug_ and _release_.

dflems:
@thomasvl RE: tests that access `debugDescription` should I box those out with an `#if DEBUG`? should probably be able to find them easily enough by running the tests in the release config

Sources/SwiftProtobuf/ExtensionFields.swift
dflems: @allevato made this change

Tests/SwiftProtobufTests/TestHelpers.swift
```diff
@@ -368,6 +368,26 @@ extension PBTestHelpers where MessageTestType: SwiftProtobuf.Message & Equatable
...
+}
+
+extension XCTestCase {
+    func assertDebugDescription(_ expected: String, _ m: SwiftProtobuf.Message, fmt: String? = nil, file: XCTestFileArgType = #file, line: UInt = #line) {
+        // `assertDebugDescription` is a no-op in release as `debugDescription` is unavailable.
+        #if DEBUG
+        let actual = m.debugDescription
+        XCTAssertEqual(actual, expected, fmt ?? "debugDescription did not match", file: file, line: line)
+        #endif
+    }
```
dflems, line 390: Added an `assertDebugDescription()` variant where you can pass in an existing `SwiftProtobuf.Message`. Both of the functions no-op for non-debug builds.

dflems:
@thomasvl @tbkka made the requested changes and fixed the tests

tbkka:
Bleargh.  Bug in the upstream conformance test.
```
In file included from conformance_cpp.cc:54:0:
conformance_cpp.cc: In member function 'google::protobuf::util::statusor_internal::StatusOr<bool> google::protobuf::{anonymous}::Harness::ServeConformanceRequest()':
conformance_cpp.cc:235:61: error: invalid conversion from 'const char*' to 'char*' [-fpermissive]
   RETURN_IF_ERROR(ReadFd(STDIN_FILENO, serialized_input.data(), in_len));
                                        ~~~~~~~~~~~~~~~~~~~~~^~
../src/google/protobuf/stubs/status_macros.h:58:10: note: in definition of macro 'RETURN_IF_ERROR'
         (expr);                                                              \
          ^~~~
```

dflems:
@tbkka Is there going to be an alpha branch for 2.0? Or is `main` going to track 2.x and 1.x will branch off from main?

thomasvl:
`main` is already tracking towards 2.0.

Personally, I don't think we'll have another 1.0 release unless there is some issue that comes up.  But that's ultimately up to the Apple folks.

thomasvl:
Did this need anything before landing (since `main` headed to 2.0 and thus accepting breaking changes) aside from approvals by @Lukasa @tbkka ?

tbkka (approved):
Nice work!   I've not been able to review the actual implementation as closely as I'd like, but the idea is good and I'm happy taking this for 2.0.

thomasvl:
@Lukasa this ok to merge?

Lukasa (approved):
LGTM, I'm happy with it. Nice patch!

thomasvl:
Ok, merging…

Thanks for the patch!

acecilia:
👋 Sorry I am a bit confused: seems like this change never made it to the public right? As the latest releases are version 1.X, but this change is marked as [v2.0](https://github.com/apple/swift-protobuf/labels/v2.0)?

By looking at [the commit](https://github.com/apple/swift-protobuf/commit/2a05837723e651495113fd9d7e83c87294c70796), looks to be only present in `main`, and not being part of any of the releases 🤔

thomasvl:
Correct, it wasn't put on the 1.x branch as it could be considered a breaking change since it changes behaviors in a _Release_ build, so the major change seemed safer.

acecilia:
Thanks! In case I would like to build SwiftProtobuf with any similar breaking changes done in the past 2 years, is there a recommended way? Could you confirm: since there hasn't been any 2.X release yet, looks like the only option would be to build from master?

thomasvl:
Yes, `main` would be the only way to get something with those changes, but as there has been no release, you could encounter a break if something else changes on `main` in the future.

As far as how to build off `main`, that depends on your specific project, and I'm not sure what the steps would be for CocoaPods or SwiftPM, can't say I've personally tried it.

Issue #1203, opened 2021-12-14
jsonString lost origin type?


cdoky:
```swift
message User {
    int64 id = 1;
    string name = 2;
}
```
```swift
var user = User()
user.id = 100
user.name = "123"
let json = try? user.jsonString()
// output: {"id": "100", "name": "123"}
// but I need: {"id": 100, "name": "123"}
```

thomasvl:
There are multiple ways to encode/decode an object into JSON.  The Protocol Buffers spec spells how it should be done for messages: https://developers.google.com/protocol-buffers/docs/proto3#json.  That includes the 64bit integer types always being encoded as strings (but accepted as either).

I believe this goes back to ensuring if the payload is received via Javascript it can be received without data loss do to that spec.

cdoky:
@thomasvl OK, Thank you!

tbkka:
If you're working with a non-Protobuf JSON implementation, you should use `int32`, `uint32`, `float`, or `double` types for numbers; Protobuf JSON does not encode those as strings.

Issue #48, opened 2016-10-05
Exporting to json ~> All values are strings


vCrespoP:
Hello all.

When I export a structure to its corresponding json like:

``` 
let id: Int64 = 1
let title = "First Book"
let author = "Who knows"
let cover = Cover.paper
        
let item1 = BookInfo.init(id: id, title: title, author: author, gender: gender)
        
let json = try? item1.serializeJSON() 
````

Then when trying to fetch a value from that dictionary:
```
let idDict = item2["id"] as? Int64
```
idDict results in being nil but
```
let idDict = item2["id"] as? String
```
actually returns "1"
So if I wanted to get the value as I defined it i would always have to parse strings like:
```
let idDict = Int64(item2["id"] as? String ?? "0")
```
Which I find it quite messy, dirty and bound to fail.

All the above also applies for the enum, where I'd get: "paper", the string representation of the case inside the enum for Cover. I assume this is ok.

Options here:
- Build an extension for strings to parse whatever value from the generated dictionary (Int, Enums, ...)
- Wait to have (#46)/implement a JSON parsing method that will contain all this logic.
- Anything else?

Thanks for your time,
Vicente Crespo.

thomasvl:
JSON mapping is a feature of the `proto3` syntax, the definition is on the protobuf docs: https://developers.google.com/protocol-buffers/docs/proto3#json

vCrespoP:
All clear.

Thank you once again, Thomas.

tbkka:
Could you clarify where your `item2` came from?  It sounds like you're taking the String generated by `item1.serializeJSON()` and then parsing it with some other JSON library?

@thomasvl gave you a pointer to the Google protobuf JSON coding spec.  Google defined the protobuf JSON coding to ensure that various JSON parsers would not mangle the result.  Among other things, they decided to store `Int64` as strings so that JSON parsers that don't support full 64-bit integers would not corrupt the values.

Swift Protobuf deals with this for you:  If you parse `json` with `BookInfo(json: json)`, you will see the `id` does get correctly parsed into an `Int64`.

If decoding Protobuf JSON with other JSON parsers is awkward, you should talk to the authors of those libraries about expanding their support.

vCrespoP:
Ok, with the link Thomas provided me I understood the logic behind this.

About 'dict2' that was a mistake I inserted while writing the sample in GitHub. item2 would be json, I edit it right now.

[...]

PR #1249, opened 2022-07-01
Proposal: Introduce a `JSONEncodingOptions` to print `int64` as `number` instead of `string`

Zedenem:
# Proposal

This PR is a proposal to introduce a `JSONEncodingOptions` prop allowing to print `int64` as `number` instead of `string`.

Please let me know if I missed some of the contribution guidelines in that PR (specifically, I am not sure I should have regenerated the protos as part of this PR's diff).
I am looking forward to discussing this proposal.

## Motivation

`int64` being encoded as `string` for messages is a common source of confusion as seen in issues like #48 or #1203.

Moreover, this mapping is only explicitly defined in the `proto3` language guide (see [JSON mapping section](https://developers.google.com/protocol-buffers/docs/proto3#json)) and no clear rule or guideline is stated for `proto2`.

While most Protobuf libraries for specific languages have decided to follow `proto3` json mapping when encoding `proto2` messages, there are some cases when following the language's JSON serialization rules is helpful, e.g. when part of your stack is expecting language-specific JSON serialization and not `proto3` json mapping, especially when you have been using `proto2` on that stack, without JSON mapping rules.

## Proposed solution

By introducing a new `JSONEncodingOptions` and turning it off by default, we give the choice to `SwiftProtobuf` users in need of this specific use case to support it easily, just like the library supports printing enums as ints instead of the default strings (one would argue this use case is different as it is explicitly supported by `proto3` though — see [JSON mapping options](https://developers.google.com/protocol-buffers/docs/proto3#json_options)).

Because it's turned off by default, it doesn't impact any existing code or mechanisms relying on the `proto3` JSON mapping rules.

## Detailed design

The `JSONEncodingOptions` `struct` now has a new boolean prop `alwaysPrintInt64sAsNumbers` which default value is `false`.

`JSONEncodingVisitor` having access to these options can now decide how to print `Int64` and `UInt64` singular or repeated values based on the value of `alwaysPrintInt64sAsNumbers`.

This also works at the wrapper level for `Google_Protobuf_Int64Value` and `Google_Protobuf_UInt64Value`.

## Backwards compatibility

Through unit tests, we can guarantee both encoding and decoding stay backwards compatible by:
- not affecting existing code which is not using the new option (see `Test_JSONEncodingOptions. testAlwaysPrintInt64sAsNumbers `)
- supporting decoding of JSON strings encoded with the new option (see `Test_JSON.testMultipleFields_whenJSONEncodingInt64AsNumbers`)


Zedenem:
To reviewers,

FYI, I am leaving this PR in draft for now as conformance tests are broken due to a change in the main `protobuf` repo.
I have submitted a PR to the repo with a suggested fix for the issue: [protobuf/#10206](https://github.com/protocolbuffers/protobuf/pull/10206)

Sources/SwiftProtobuf/JSONEncodingOptions.swift
```diff
@@ -15,6 +15,10 @@
 /// Options for JSONEncoding.
 public struct JSONEncodingOptions {
 
+  /// Always prints int64s as numbers.
+  /// By default, they are printed as strings as per proto3 JSON mapping rules.
```
tbkka, line 19: We should clarify here that they are still printed as strings when used as Map keys.

Tests/SwiftProtobufTests/Test_JSONEncodingOptions.swift
```diff
@@ -17,9 +17,23 @@ import XCTest
...
+    asStrings.alwaysPrintInt64sAsNumbers = false
+    var asNumbers = JSONEncodingOptions()
+    asNumbers.alwaysPrintInt64sAsNumbers = true
+
+    let msg1 = ProtobufUnittest_Message2.with {
+      $0.optionalInt64 = 1656338459803
+    }
+    XCTAssertEqual(try msg1.jsonString(options: asStrings), "{\"optionalInt64\":\"1656338459803\"}")
+    XCTAssertEqual(try msg1.jsonString(options: asNumbers), "{\"optionalInt64\":1656338459803}")
+  }
```
tbkka, line 34: We should also have a test here to verify that `map<int64,string>` still correctly writes the int64 in quoted form even with this option set.  Ideally, we'd also check `map<uint64,string>` for completeness.

tbkka (commented):
Looks good overall.   Just a minor point to make sure we're correctly handling `map<int64,...>` where the int64 has to be quoted regardless.

Sources/SwiftProtobuf/JSONEncodingOptions.swift
Zedenem: Comment updated. Let me know if this is ok.

Sources/SwiftProtobuf/JSONEncodingVisitor.swift
```diff
@@ -226,9 +233,12 @@ internal struct JSONEncodingVisitor: Visitor {
   }
 
   mutating func visitRepeatedUInt64Field(value: [UInt64], fieldNumber: Int) throws {
+    let options = self.options
     try _visitRepeated(value: value, fieldNumber: fieldNumber) {
       (encoder: inout JSONEncoder, v: UInt64) in
-      encoder.putUInt64(value: v)
+      options.alwaysPrintInt64sAsNumbers
```
thomasvl, line 239: Might make sense to move the condition outside the loop.

Zedenem: Was following the pattern used by `alwaysPrintEnumAsInts` but it's indeed a bit different as it also has a condition on the enum having a value for property `name`.
Will update that.
Thank you.

Zedenem: It does come with some logic duplication though, to a performance benefit I personally consider low:
```swift
if options.alwaysPrintInt64sAsNumbers {
      try _visitRepeated(value: value, fieldNumber: fieldNumber) {
        (encoder: inout JSONEncoder, v: Int64) in
        encoder.putNonQuotedInt64(value: v)
      }
    } else {
      try _visitRepeated(value: value, fieldNumber: fieldNumber) {
        (encoder: inout JSONEncoder, v: Int64) in
        encoder.putInt64(value: v)
      }
    }
```

Or maybe you had a better idea syntax-wise?
Let me know what you think and I'll take care of it.

Zedenem: Hi @thomasvl. Sorry for the ping but any news on this? I'd love for this to be merged and released so I could stop using my fork. Thank you 🙏.

tbkka: As a rule, we generally do split things like this into two loops as @thomasvl suggested.  Although the performance benefit may seem modest to most folks, we have some users who are _very_ passionate about getting every bit of performance improvement they can.

Zedenem: I've made the requested changes.
Thanks for the insights on performance requirements 🙏 .

Sources/Conformance/failure_list_swift.txt
```diff
@@ -1 +1,2 @@
-# No known failures
+Required.Proto3.JsonInput.DurationNegativeNanos.JsonOutput
+Required.Proto3.JsonInput.DurationNegativeNanos.ProtobufOutput
```
Zedenem, line 2: @thomasvl @tbkka here I am following this recent change on google/protobuf where some tests were added to the failure list: https://github.com/protocolbuffers/protobuf/commit/4ea4d1d91a65861a16f4bbb231f1bf2daf795d8a

Let me know if I shouldn't do it this way.

thomasvl: If this is a new test, please spin out a different PR and also open an issue for the tests failing.

thomasvl: I got a few minutes, so I've opened #1257

Zedenem: Awesome! Thank you very much @thomasvl.

tbkka:
This looks good.

It's a little odd that your naming is backwards from how we're already doing things for Int32.  Should we rename `putInt64` to `putQuotedInt64` and `putNonQuotedInt64` to `putInt64` to make them consistent?

It might be even better to use `putQuoted...` and `putNonQuoted...` for Int32 and Int64 (and other number types) consistently throughout.  That would avoid any confusion about whether `putInt64` or `putInt32` was quoted or not.

Zedenem:
> This looks good.
> 
> It's a little odd that your naming is backwards from how we're already doing things for Int32. Should we rename `putInt64` to `putQuotedInt64` and `putNonQuotedInt64` to `putInt64` to make them consistent?
> 
> It might be even better to use `putQuoted...` and `putNonQuoted...` for Int32 and Int64 (and other number types) consistently throughout. That would avoid any confusion about whether `putInt64` or `putInt32` was quoted or not.

@tbkka I am more than happy to do either of those changes. Just let me know what you agree on.

I liked the explicitness of `putNonQuotedInt64` just because it highlighted the fact that this was not the default and that `putInt64` was. But you're right, it makes the reader wonder if `putInt64` will quote or not, and it's also backwards from `Int32` (which I didn't realize before you pointed it out 😕).

tbkka:
I think we should use `putQuoted...` and `putNonQuoted...` throughout this one source file for the types that actually use both forms (which I think is only the integer types).   Re-reading the code, the old naming was a little unclear.

Zedenem:
@tbkka Here is the renaming I have done in my last commit to match with existing `put[Non]Quoted[...]` counterparts:
- `putInt32` > `putNonQuotedInt32`
- `putUInt32` > `putNonQuotedUInt32`
- `putInt64` > `putQuotedInt64`
- `putUInt64` > `putQuotedUInt64`
- `putBoolValue` > `putNonQuotedBoolValue`

tbkka:
Looks great!   I've merged this.  If you have any follow-ups, please put up a new PR.

Issue #1207, opened 2022-01-28
SwiftProtobuf plugin for SwiftPM?


kielgillard:
tl;dr: Is it possible for this repo to offer a canonical SwiftProtobuf plugin for SwiftPM, [as demonstrated here](https://github.com/apple/swift-evolution/blob/main/proposals/0303-swiftpm-extensible-build-tools.md#example-2-swiftprotobuf)?

Now that [Swift 5.6 is out in the Xcode 13.3 betas and later](https://developer.apple.com/documentation/xcode-release-notes/xcode-13_3-release-notes), Swift Packages now support build tool plugins, as defined in [SE-0303](https://github.com/apple/swift-evolution/blob/main/proposals/0303-swiftpm-extensible-build-tools.md) and SE-0325.

However, for the time being, it seems like everyone will have to roll their own protobuf to Swift generator plugin. It doesn't seem like there is a plugin already "out there" and one does not simply copy and paste the example from the proposal. Speaking for myself, I am not that familiar with the tool and I would certainly have no idea how to support a hand rolled plugin going forward.

Is it possible for this repo to offer a canoncial SwiftProtobuf plugin for SwiftPM? All I really want to do is declare a dependency to this repo and add a plug in to my target. Is this straightforward usability naive?

Some details about my use case. I just have some protobuf files organised into folders that I want to generate Swift from. I want to compile that source into a library. This library is a dependency among other targets in different packages.

macOS Monterey 12.2 (21D49)
Xcode 13.3 / Build version 13E5086k
swift-driver version: 1.44.2 Apple Swift version 5.6 (swiftlang-5.6.0.320.8 clang-1316.0.18.8)
Target: x86_64-apple-macosx12.0
https://github.com/apple/swift-protobuf/releases/tag/1.18.0

abertelrud:
There is a prototype in the `eng/add-build-tool-plugin` branch in a fork at https://github.com/abertelrud/swift-protobuf, and a use case in https://github.com/abertelrud/swiftpm-buildtool-plugin-examples that could be used as a start.  It's not robust enough to be real yet but it would be great to work with someone on real support.

Lukasa:
I should note that it will still be required to get hold of `protoc` outside of this repo. However, I'm +1 on upstreaming this support here.

abertelrud:
[...]

PR #1251, opened 2022-07-07
Add SPM plugin

FranzBusch:
# Motivation
SPM is providing new capabilities for tools to expose themselves as plugins which allows them to generate build commands. This allows us to create a plugin for `SwiftProtobuf` which invokes the `protoc` compiler and generates the Swift code. Creating such a plugin is greatly wanted since it improves the usage of the `protoc-gen-swift` plugin dramatically. Fixes https://github.com/apple/swift-protobuf/issues/1207

# Modification
This PR adds a new SPM plugin which generates build commands for generating Swift files from proto files. Since users of the plugin might have complex setups, I introduced a new `swift-protobuf-config.json` file that adopters have to put into the root of their target which wants to use the new plugin. The format of this configuration file is quite simple:

```json
{
    "invocations": [
        {
            "protoFiles": [
                "Foo.proto"
            ],
            "visibility": "internal"
        }
    ]
}

```

It allows you to configure multiple invocations to the `protoc` compiler. For each invocation you have to pass the relative path from the target source to the proto file. Additionally, you have to decide which visibility the generated symbols should have. In my opinion, this configuration files gives us the most flexibility and more complex setups to be covered as well.

# Open topics

## Hosting of the protoc binary
Hosting of the protoc binary is the last open thing to figure out before we can release a plugin for `SwiftProtobuf`. From my point of view, there are three possible routes we can take:

1. Include the `artifactbundle` inside the `SwiftProtobuf` repository
2. Include the `artifactebundle` as an artifact on the GH releases in the protobuf repo
3. Extend the the artifact bundle manifest to allow inclusion of URLs. This would require a Swift evolution pitch most likely.

However, with all three of the above we would still need to release a new version of `SwiftProtobuf` with every new release of `protoc`.

# Future work

## Proto dependencies between modules
With the current shape of the PR one can already use dependencies between proto files inside a single target. However, it might be desirable to be able to build dependency chains of Swift targets where each target includes proto files which depend on protoc files from the dependencies of the Swift target. I punted this from the initial plugin because this requires a bit more work and thinking. Firstly, how would you even spell such an import? Secondly, the current way of doing `ProtoPathModuleMapping` files is not super ideal for this approach. It might make sense to introduce a proto option to set the Swift module name inside the proto files already.

# Result
We now have a SPM plugin that can generate Swift code from proto files. To use it, it provides a configuration file format to declare different invocations to the `protoc` compiler.

TODOs:
- As soon as we agreed on the binary problem, I am going to write documentation how to use the plugin

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -14,29 +14,61 @@ import PackageDescription
 let package = Package(
   name: "SwiftProtobuf",
   products: [
-    .executable(name: "protoc-gen-swift", targets: ["protoc-gen-swift"]),
-    .library(name: "SwiftProtobuf", targets: ["SwiftProtobuf"]),
-    .library(name: "SwiftProtobufPluginLibrary", targets: ["SwiftProtobufPluginLibrary"]),
+    .executable(
+        name: "protoc-gen-swift",
+        targets: ["protoc-gen-swift"]
+    ),
+    .library(
+        name: "SwiftProtobuf",
+        targets: ["SwiftProtobuf"]
+    ),
+    .library(
+        name: "SwiftProtobufPluginLibrary",
+        targets: ["SwiftProtobufPluginLibrary"]
+    ),
+    .plugin(
+        name: "SwiftProtobufPlugin",
+        targets: ["SwiftProtobufPlugin"]
+    ),
   ],
   dependencies: [
     .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
   ],
   targets: [
-    .target(name: "SwiftProtobuf",
-            exclude: ["CMakeLists.txt"]),
-    .target(name: "SwiftProtobufPluginLibrary",
-            dependencies: ["SwiftProtobuf"],
-            exclude: ["CMakeLists.txt"]),
-    .executableTarget(name: "protoc-gen-swift",
-            dependencies: ["SwiftProtobufPluginLibrary", "SwiftProtobuf"],
-            exclude: ["CMakeLists.txt"]),
-    .executableTarget(name: "Conformance",
-            dependencies: ["SwiftProtobuf"],
-            exclude: ["failure_list_swift.txt", "text_format_failure_list_swift.txt"]),
-    .testTarget(name: "SwiftProtobufTests",
-                dependencies: ["SwiftProtobuf"]),
-    .testTarget(name: "SwiftProtobufPluginLibraryTests",
-                dependencies: ["SwiftProtobufPluginLibrary"]),
+    .target(
+        name: "SwiftProtobuf",
+            exclude: ["CMakeLists.txt"]
+    ),
+    .target(
+        name: "SwiftProtobufPluginLibrary",
+        dependencies: ["SwiftProtobuf"],
+        exclude: ["CMakeLists.txt"]
+    ),
+    .executableTarget(
+        name: "protoc-gen-swift",
+        dependencies: ["SwiftProtobufPluginLibrary", "SwiftProtobuf"],
+        exclude: ["CMakeLists.txt"]
+    ),
+    .executableTarget(
+        name: "Conformance",
+        dependencies: ["SwiftProtobuf"],
+        exclude: ["failure_list_swift.txt", "text_format_failure_list_swift.txt"]
+    ),
+    .plugin(
+        name: "SwiftProtobufPlugin",
+        capability: .buildTool(),
+        dependencies: [
+            "protoc-gen-swift"
+        ]
+    ),
+    .testTarget(
+        name: "SwiftProtobufTests",
+        dependencies: ["SwiftProtobuf"]
+    ),
+    .testTarget(
+        name: "SwiftProtobufPluginLibraryTests",
+        dependencies: ["SwiftProtobufPluginLibrary"]
+    ),
   ],
   swiftLanguageVersions: [.v5]
 )
--- a/Plugins/SwiftProtobufPlugin/plugin.swift
+++ b/Plugins/SwiftProtobufPlugin/plugin.swift
@@ -7,8 +7,8 @@ struct SwiftProtobufPlugin: BuildToolPlugin {
     enum PluginError: Error {
         /// Indicates that the target where the plugin was applied to was not `SourceModuleTarget`.
         case invalidTarget
-        /// Indicates that no config file was present in the root of the target directory.
-        case noConfigFile
+        /// Indicates that the file extension of an input file was not `.proto`.
+        case invalidInputFileExtension
     }
 
     /// The configuration of the plugin.
@@ -41,20 +41,17 @@ struct SwiftProtobufPlugin: BuildToolPlugin {
     static let configurationFileName = "swift-protobuf-config.json"
 
     func createBuildCommands(context: PluginContext, target: Target) throws -> [Command] {
-        print("Creating build commands (\(UUID()))")
-
         // Let's check that this is a source target
         guard let target = target as? SourceModuleTarget else {
             throw PluginError.invalidTarget
         }
 
         // We need to find the configuration file at the root of the target
-        guard let configurationFilePath = target.sourceFiles.first(where: { $0.path.lastComponent == Self.configurationFileName }) else {
-            throw PluginError.noConfigFile
-        }
-
-        let data = try Data(contentsOf: URL(fileURLWithPath: "\(configurationFilePath.path)"))
+        let configurationFilePath = target.directory.appending(subpath: Self.configurationFileName)
+        let data = try Data(contentsOf: URL(fileURLWithPath: "\(configurationFilePath)"))
         let configuration = try JSONDecoder().decode(Configuration.self, from: data)
+
+        try self.validateConfiguration(configuration)
 
         // We need to find the path of protoc and protoc-gen-swift
         let protocPath: Path
@@ -112,14 +109,17 @@ struct SwiftProtobufPlugin: BuildToolPlugin {
         var inputFiles = [Path]()
         var outputFiles = [Path]()
 
-        for file in invocation.protoFiles {
+        for var file in invocation.protoFiles {
             // Append the file to the protoc args so that it is used for generating
             protocArgs.append("\(file)")
             inputFiles.append(target.directory.appending(file))
 
             // The name of the output file is based on the name of the input file.
-            let protobufOutputName = file.replacingOccurrences(of: ".proto", with: ".pb.swift")
-            let protobufOutputPath = outputDirectory.appending(protobufOutputName)
+            // We validated in the beginning that every file has the suffix of .proto
+            // This means we can just drop the last 5 elements and append the new suffix
+            file.removeLast(5)
+            file.append("pb.swift")
+            let protobufOutputPath = outputDirectory.appending(file)
 
             // Add the outputPath as an output file
             outputFiles.append(protobufOutputPath)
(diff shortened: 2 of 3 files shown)


FranzBusch:
@thomasvl @tbkka Since we are planning to cut a new SemVer from main at some point I was wondering what our plans are w.r.t. to moving up the supported Swift version?

thomasvl:
A 2.0 will be the next release off `main`, the min is now Swift 5.0.

thomasvl:
> @thomasvl W.r.t. to an SPM plugin this will mean that we need to introduce a new `Package@swift-5.6.swift` to support the plugin capabilities.

This is the one part of SwiftPM that always bugs me.  I guess the other option is move Package.swift to 5.6, and add 5.0, 5.1, 5.2, 5.3, 5.4, 5.5 files instead.  Otherwise, we have to remember to add a 5.7, 5.8, etc.  Since I don't think SwiftPM ever got the support for "hunt" for a version that worked if there wasn't an exact match to major.minor?

tbkka:
> I guess the other option is move Package.swift to 5.6, and add 5.0, 5.1, 5.2, 5.3, 5.4, 5.5 files instead.

Our policy for the Swift code has been to always have the default code paths written against the current Swift version and introduce backwards-compatibility shims to get it to work with older versions.  Then we drop the backwards-compatibility shims as we migrate our minimum language version requirement.

So that would argue for keeping Package.swift up-to-date and using versioned files for backwards compatibility.

FranzBusch:
@tbkka @thomasvl I think this PR is now ready for review. The last change removed the checked-in `protoc` artifactbundle. We are now relying on finding `protoc` through the `context.tool()`. This works without a problem in the CLI if `protoc` is installed in the `$PATH`; however, this does not work in Xcode since Xcode isn't getting the `$PATH`.
To provide a workaround I added a config option that allows you to specify the direct path to the `protoc` binary which allows Xcode builds to still work.

Let me know what you think about this!

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -0,0 +1,139 @@
...
+            /// An array of paths to `.proto` files for this invocation.
+            var protoFiles: [String]
+            /// The visibility of the generated files.
+            var visibility: Visibility?
+        }
+
+        /// The path to the `protoc` binary.
+        ///
+        /// If this is not set, SPM will try to find the tool itself.
+        var protocPath: String?
```
tbkka, line 35: `protocPath` is fundamentally different from the rest of the configuration options here:  It's a property of the build environment, not the project.

I've seen other build tools allow for two configuration files:  One that is checked in and holds configuration that is independent of the particular build environment and a second one that is typically _not_ checked in with configuration for the build environment.  In this case, having a second "swift-protobuf-config-local.json" that will override the protocPath if present might be a good approach.

Package.swift
```diff
@@ -14,22 +14,52 @@ import PackageDescription
...
+    ),
+    .executableTarget(
+        name: "protoc-gen-swift",
+        dependencies: ["SwiftProtobufPluginLibrary", "SwiftProtobuf"]
+    ),
+    .target(
+        name: "Conformance",
+        dependencies: ["SwiftProtobuf"]
+    ),
+    .plugin(
```
tbkka, line 50: Apart from the reformatting, it looks like you've added `.executableTarget` and `.plugin` entries.

Package.swift
```diff
@@ -14,22 +14,52 @@ import PackageDescription
...
+    ),
+    .library(
+        name: "SwiftProtobuf",
+        targets: ["SwiftProtobuf"]
+    ),
+    .library(
+        name: "SwiftProtobufPluginLibrary",
+        targets: ["SwiftProtobufPluginLibrary"]
+    ),
+    .plugin(
```
tbkka, line 29: Apart from the reformatting, you've added one product `.plugin`

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -0,0 +1,139 @@
+import Foundation
+import PackagePlugin
+
+@main
+struct SwiftProtobufPlugin: BuildToolPlugin {
+    /// Errors thrown by the `SwiftProtobufPlugin`
+    enum PluginError: Error {
```
tbkka, line 7: Is this sufficient?  Do we need an error case for failure to find the protoc binary or for failure to run that binary?

Plugins/SwiftProtobufPlugin/plugin.swift
FranzBusch: I really like that idea! Let me change that up and separate it into two files.

tbkka:
You should probably also add a section to `Documentation/PLUGIN.md` describing how to use this.

Package.swift
FranzBusch: Yes we are just vending a new `plugin` product

Plugins/SwiftProtobufPlugin/plugin.swift
tbkka: I would still allow `protocPath` in the checked-in file (for use of CI systems that may not be able to easily manage local configuration files).  But the `-local` file should override that setting.

Package.swift
FranzBusch: I haven't really added a new `executableTarget`. The `protoc-gen-swift` target was always executable but with newer tools versions we introduced `executableTarget`. This change here is just fixing this warning up.

FranzBusch:
@tbkka Yes, if we agree on this approach I will add the documentation!

Plugins/SwiftProtobufPlugin/plugin.swift
FranzBusch: @tbkka Friendly ping. Would love to get your input on this once more.

tbkka: You've convinced me that we don't need local configuration support.

FranzBusch:
@tbkka @thomasvl I added documentation on how to use the new SPM plugin. I only added it to the new DocC documentation for now since the plugin is not fully working in Xcode yet. Let me know what you think!

FranzBusch:
@thomasvl The upstream proto repo removed the `autogen.sh` script and now all the tests are failing. Can we just use `apt-get protobuf-compiler` to install the compiler for CI?

thomasvl:
> @thomasvl The upstream proto repo removed the `autogen.sh` script and now all the tests are failing. Can we just use `apt-get protobuf-compiler` to install the compiler for CI?

We'll have to update to the new build system, we stay on trunk to catch a bunch of things before releases.

thomasvl:
> We'll have to update to the new build system, we stay on trunk to catch a bunch of things before releases.

#1271 should fix things back up.

FranzBusch:
@thomasvl @tbkka I added the docs and the CI got fixed, I think we are ready for a review of this PR. Let me know what you think!

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -0,0 +1,139 @@
...
+        var protocPath: String?
+
+        /// A list of invocations of `protoc` with the `SwiftProtobuf` plugin.
+        var invocations: [Invocation]
+    }
+
+    static let configurationFileName = "swift-protobuf-config.json"
+
+    func createBuildCommands(context: PluginContext, target: Target) throws -> [Command] {
+        print("Creating build commands (\(UUID()))")
```
thomasvl, line 44: Is this useful?  most of the time it seems like having builds be complete silent is what folks prefer.

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -0,0 +1,139 @@
...
+
+        // Let's check that this is a source target
+        guard let target = target as? SourceModuleTarget else {
+            throw PluginError.invalidTarget
+        }
+
+        // We need to find the configuration file at the root of the target
```
thomasvl, line 51: Is sourceFiles recursive?  i.e. - is this really gonna be at the root of the target or could it be found nested in a sub folder?  i.e. - which behavior do you want.

Followup - should we check to see if there are >1 files with the expected name an error since something likely isn't gonna go as the developer might have expected?

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -0,0 +1,139 @@
...
+        var inputFiles = [Path]()
+        var outputFiles = [Path]()
+
+        for file in invocation.protoFiles {
+            // Append the file to the protoc args so that it is used for generating
+            protocArgs.append("\(file)")
+            inputFiles.append(target.directory.appending(file))
+
+            // The name of the output file is based on the name of the input file.
+            let protobufOutputName = file.replacingOccurrences(of: ".proto", with: ".pb.swift")
```
thomasvl, line 121: This could be a little dangerous since it will replaces all locations of ".proto" in the path, I'd suggest doing a more targeted extension swap instead. Someone, somewhere will have "Source/Target/Support.protos/foo.proto"…

And should we validate that the file does end in .proto in the first case?  i.e. - catch another form of user error?

Plugins/SwiftProtobufPlugin/plugin.swift
FranzBusch: I removed this again. This was very useful while developing the plugin because SPM had some hiccups during the 5.7 Betas. Though all of them are fixed up now.

Plugins/SwiftProtobufPlugin/plugin.swift
FranzBusch: You are right. This is actually recursive. I changed the code here to only look at the root now. AFAIK from the discussions in the forums and how our tool is working we should require it to be in the root.

I am also not diagnosing if there are multiple files. If there is no file at the root, the error looks like this
> Error Domain=NSCocoaErrorDomain Code=260 "The file “swift-protobuf-config.json” couldn’t be opened because there is no such file." UserInfo={NSFilePath=/Users/franzbusch/Developer/swift-protobuf-examples/01-same-module-usage/Sources/swift-protobuf-example/swift-protobuf-config.json, NSUnderlyingError=0x600001f3c060 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}

This Error is thrown by the `Data` init and looks sensible to me, so I wouldn't create a custom error for this

Plugins/SwiftProtobufPlugin/plugin.swift
FranzBusch: Good points! I changed it to only swap the suffix and added some validation for the input file extensions.

thomasvl (approved):
Seems good to me.

@tbkka Seem good to you?

tbkka:
I'm happy.

MPLew-is:
This is phenomenal, any thoughts on when a release might be cut with this functionality in it?

thomasvl:
> This is phenomenal, any thoughts on when a release might be cut with this functionality in it?

Things are working towards a 2.0 release, so it likely may be a while since cutting a release stops work on any other breaking changes.

thomasvl:
> @thomasvl i plan on bringing this over to the 1.x branch. Do you see any problems with that?

Nope, but you'll likely have to do a work on the `Package.swift` files over there, and we'll likely have to fix up CI over there (again).

MPLew-is:
Looks like this has been [released in 1.20.0](https://github.com/apple/swift-protobuf/releases/tag/1.20.0), thank you so much for getting this done!

PR #1255, opened 2022-07-15
Move most builds to ubuntu-20

### LINKED ISSUE #1252: Consider moving CI up to ubuntu-20.04
CI is currently using the 18.04 github images, we should probably consider moving up to the 20.04 images.  The 5.0 & 5.1 swift images don't have a matching 20.04 image, so we'd have to keep those on the old image, but we used to be doing that for the old build before raising our min version.

thomasvl:
Fixes #1252

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,16 +11,25 @@ on:
 
 jobs:
   core:
-    runs-on: ubuntu-18.04
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


thomasvl:
I don't really track the linux side of things, so I'm not completely sure when's is the right time, but seems like the 20.04 images have been around for a while, so maybe it's time?  Thoughts?

tbkka (approved):
20.04 is the April 2020 release, so yeah, it's been around for a little over 2 years.  Definitely time.

thomasvl:
Sounds good.  Landing.

Issue #1154, opened 2021-06-17
Of Interest: docc support


thomasvl:
At WWDC2021 docc got announced.  The current `Makefile` has some jazzy support, but might be time to revisit docs flows to instead look at building out docc support instead.

Things that might be needed:
- tweak comments to use code references that link _\`SYMBOL\`_ to _\`\`SYMBOL\`\`_
- Move some of the `Documentation/` pages into things included in the built doc set output instead
- Group/categories pages, etc.  (see the WWDC videos for improving ones docs).

jeroenleenarts:
It would be cool if a manifest is added triggering Swift Package Index DocC generation. More info [here](https://blog.swiftpackageindex.com/posts/auto-generating-auto-hosting-and-auto-updating-docc-documentation/).

jeroenleenarts:
All bits seem to be in place. So I'll ask on the SPI side of things.

thomasvl:
@tbkka is there anything we can do on CI or manually to validate how we'll we've got things documented?  i.e. - missing docs, apis missing params/return result docs/etc.

thomasvl:
@tbkka should this stay open or get moved to more specific issues for what needs followup?

Issue #1032, opened 2020-07-27
README.md links to cocoadocs for the generated API, but it is old


thomasvl:
README.md links to cocoadocs for the generated API, but it is old since cocoadocs is no longer maintained.  It be worth revisit jazzy vs. say https://github.com/SwiftDocOrg/swift-doc and maybe see if we can use github actions to auto generate based on master and/or releases.

thomasvl:
Swift Package Index [announced support](https://blog.swiftpackageindex.com/posts/auto-generating-auto-hosting-and-auto-updating-docc-documentation/) for building/hosting docs.  So adding that to the Package.swift would help developers locally get docs as well as provide a location that auto updates to point too.

thomasvl:
This was effectively fixed by #1270 since it removed the link.

PR #1262, opened 2022-08-02
Enable Swift-DocC plugin for generating docs

tbkka:
I'm going to be going through things and updating our API docs and articles to work cleanly with DocC.  First step is to make it easier to build and view the docs.

With this, you can build and view the docs for a single part of our project with:
```
% swift package --disable-sandbox preview-documentation --target SwiftProtobuf   
```
This should also work for `--target protoc-gen-swift`, `--target SwiftProtobufPluginLibrary`, and `--target Conformance`.


thomasvl (commented):
Should we document how to do any of this?  Is it worth running any sorta CI to catch issue with source comments/etc?

Showing my ignorance, how does this compare to `xcodebuild docbuild`? (https://developer.apple.com/documentation/xcode/distributing-documentation-to-external-developers)

thomasvl:
I guess I'm also sorta think of #1154 and #1032, should we start linking to other things and maybe dropping the jazzy info in the repo?

tbkka:
This uses exactly the same tools as `xcodebuild docbuild` and should produce identical results.  The only difference is that it will work on non-Apple platforms.

I don't think we're ready to drop Jazzy support just yet.  I plan to spend some time over the next few weeks trying to clean up and improve the DocC docs.

thomasvl:
> I plan to spend some time over the next few weeks trying to clean up and improve the DocC docs.

Great!  You might want to self assign those two issues then to let folks know things are being worked on.

thomasvl (approved):

PR #1268, opened 2022-08-10
Add Package.resolved to .gitignore

thomasvl:
Now that there is another package reference (for docc), it causes a
Package.resolved to get created, so ignore it since there probably isn't a
reason to pin that information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ xcbaselines
 /build
 mined_words.txt
 /DescriptorTestData.bin
+/Package.resolved
 
 # Intermediate conformance test outputs
 failing_tests.txt


thomasvl:
@tbkka does it make sense to land this or should we just land the `Package.resolved`?

tbkka:
To be honest, I'm not entirely certain.  The consensus on the Swift forums seems to be that the Package.resolved should be checked in.

* I can see that Package.resolved should be included in any official source bundles we publish, since the Package.resolved is in part an assertion that our software has been tested and works with a particular version of certain other packages.  This argument also suggests that it should be included in release tags.

* But I'm less convinced about including it in the Github main branch:  that branch is supposed to be bleeding edge and should be regularly tested with the most-recent version of other packages we use.  So there I think the best answer is to .gitignore that file.

The above is a really awkward split, though, and I'm not happy about the effort it would take to maintain it.  Fortunately, since the only package affected is the DocC package, I don't think it matters that much either way.  So I'm inclined to go ahead with this PR as-is (.gitignore the Package.resolved).  It's possible we'll have to re-visit that in the future.  🤷

tbkka (approved):

thomasvl:
Yea, to that first point, it also almost feels like those points are more about when you have dependencies for you project to build.  Since the only thing we have is the docs generation, it makes things more fuzzy to me. Anyway, merging this in.

PR #1271, opened 2022-08-15
Move off autotools and use Bazel/bazelisk instead.

thomasvl:


thomasvl:
Bazel doesn't seem to list focal (20.x) - https://bazel.build/install/ubuntu. So the steps are a little more complex, but it doesn't seem worth moving our Swift images back to simplify the CI.

thomasvl:
We can also go down the path of using CMake if we think that might be better.

thomasvl:
> We can also go down the path of using CMake if we think that might be better.
> 
> Talking to some protobuf folks, CMake will be supported long term, so it should be fine if we'd rather that.

Make a run at this in another PR, and it seems to have some of it's own things to sort out, so for now, bazel seems easier.  Maybe after protobuf documents a linux based CMake build a little more we can revisit it.

thomasvl:
Ok, got CMake working in #1273, since it keeps caching, I think I prefer that one. That version also removes support in the `Makefile` for rebuilding the conformance runner, which I sorta prefer since it lets folks build which ever way they would prefer.

PR #1273, opened 2022-08-15
Move off autotools and use CMake instead.

thomasvl:


thomasvl:
So I think I actually like this one a little better because it allows up to keep caching and also removed the knowledge from the `Makefile` about how to build protobuf, so folks can use which every approach the prefer.

FranzBusch:
LGTM! Thanks for fixing this up!

FranzBusch (approved):

Lukasa (approved):

PR #1276, opened 2022-08-23
SPM plugin for 1.x

FranzBusch:
This PR brings over the commit from the plugin merge to the 1x release branch. Furthermore, I cherry-picked all commits that changed the CI recently.


thomasvl (changes requested):
Cherry picking all the changes to Package*.swift and CI changes won't work, as that's removing the 3.x & 4.x support from the branch.

The safest is likely to just do the CMake related changes directly without changing the requirements.

Makefile
```diff
@@ -11,8 +11,6 @@
 # make regenerate
 #   Recompile all the necessary protos
 #   (requires protoc in path)
-# make test-xcode[-NAME]:
```
thomasvl, line 14: I think this all would still be valid on the branch since the Xcode project is still there.

.github/workflows/build.yml
```diff
@@ -11,18 +11,40 @@ on:
 
 jobs:
   core:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        swift: ["5.5-bionic", "5.4-bionic", "5.3.3-bionic", "5.2.5-bionic", "5.1.5-bionic", "5.0.3-bionic", "4.2.4"]
```
thomasvl, line 18: Lost the 4.2.4 setup from in here.

thomasvl (changes requested):

FranzBusch:
@thomasvl I think we are good to go know. Could you give it another look, merge it and create a new release for it?

Makefile
thomasvl: This comment applied to all the other xcode related bits in the Makefile.  Likely should keep those here just incase someone was using them since it is part of the 1.x behaviors/support.

FranzBusch: Got you! Should be fixed now

Makefile
```diff
@@ -653,3 +630,9 @@ test-xcode-watchOS-release:
 		-scheme SwiftProtobuf_watchOS \
 		-configuration Release \
 		build $(XCODEBUILD_EXTRAS)
+test-conformance: build check-for-protobuf-checkout Sources/Conformance/failure_list_swift.txt Sources/Conformance/text_format_failure_list_swift.txt
```
thomasvl, line 633: Make my want a blank line between these two targets

FranzBusch: Yep it wants that!

thomasvl (approved):

PR #1279, opened 2022-08-25
Provide environment variable configuration option

FranzBusch:
# Motivation
We are currently providing a config option to provide a path to the `protoc` binary. This works great if your `protoc` is always the same; however, for a lot of packages you want to build them on various systems and the path to `protoc` is always different. We really can't solve this with any config file.

# Modification
This PR adds the option to set the `PROTOC_PATH` env variable and picks it up in the plugin to find the protoc binary.

# Result
We can now dynamically pass the protoc path to the plugin.

### DIFF
--- a/Sources/protoc-gen-swift/Docs.docc/spm-plugin.md
+++ b/Sources/protoc-gen-swift/Docs.docc/spm-plugin.md
@@ -22,7 +22,7 @@ There are multiple ways to do this. Some of the easiest are:
 1. If you are on MacOS, installing it via `brew install protoc`
 2. Download the binary from [Google's github repository](https://github.com/protocolbuffers/protobuf).
 
-### Adding the proto files to your target
+### Adding the proto files to your target
 
 Next, you need to add the `.proto` files for which you want to generate your Swift types to your target's
 source directory. You should also commit these files to your git repository since the generated types
@@ -56,7 +56,7 @@ let package = Package(
 
 ```
 
-### Configuring the plugin
+### Configuring the plugin
 
 Lastly, after you have added the `.proto` files and modified your `Package.swift` manifest, you can now
 configure the plugin to invoke the `protoc` compiler. This is done by adding a `swift-protobuf-config.json`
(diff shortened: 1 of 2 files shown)


FranzBusch:
@thomasvl @tbkka  Would love to get your opinion on this one. While the config option that I provided in the JSON file is good to have and solves most problems. Sometimes, that didn't really work if you needed your package to be developed cross platform. Furthermore, non-leaf packages couldn't use it at all.
I have been trying a few different things for the last days and was almost going to give up but then I came across an idea of passing an env variable to the plugin. The great thing is this works across command line and Xcode if you start Xcode in the right way.

The real fix is still in SPM however I think it is worth adding this env variable escape hatch.

tbkka:
I'm fine with it.

thomasvl:
Looking at the rendered docs, two of the subsections (`###`) are getting detected.  Can you try an extra blank line before them so we can fix those in the same PR?  You can preview it before we merge it at https://github.com/FranzBusch/swift-protobuf/blob/fb-env-variable-for-protoc/Sources/protoc-gen-swift/Docs.docc/spm-plugin.md.

Sources/protoc-gen-swift/Docs.docc/spm-plugin.md
```diff
@@ -88,13 +88,25 @@ is generating Swift types for the `Bar.proto` file with the `public` visibility.
...
+There are three ways how this can be achieved. First, by default, the package manager looks into
 the `$PATH` to find binaries named `protoc`. This works immediately if you use `swift build` to build
 your package and `protoc` is installed in the `$PATH` (`brew` is adding it to your `$PATH` automatically).
 However, this doesn't work if you want to compile from Xcode since Xcode is not passed the `$PATH`.
-To still make this work from Xcode you can point the plugin to the concrete location of the `protoc`
-compiler by changing the configuration file like this:
+You have to options to set the path of `protoc` that the plugin is going to use. Either you can set
+an environment variable `PROTOC_PATH` that gets picked up by the plugin. Here are two example how you
+can set the variable so that it gets picked up:
+```
```
thomasvl, line 100:
```shell

(might do some highlighting)

File: Sources/protoc-gen-swift/Docs.docc/spm-plugin.md [hunk 2]
```diff
@@ -88,13 +88,26 @@ is generating Swift types for the `Bar.proto` file with the `public` visibility.
...
 your package and `protoc` is installed in the `$PATH` (`brew` is adding it to your `$PATH` automatically).
 However, this doesn't work if you want to compile from Xcode since Xcode is not passed the `$PATH`.
-To still make this work from Xcode you can point the plugin to the concrete location of the `protoc`
-compiler by changing the configuration file like this:
+You have to options to set the path of `protoc` that the plugin is going to use. Either you can set
+an environment variable `PROTOC_PATH` that gets picked up by the plugin. Here are two example how you
+can set the variable so that it gets picked up:
+
+```shell
+// swift build
```
thomasvl, line 102: Shell comments are usually `#` (not `//`)

Sources/protoc-gen-swift/Docs.docc/spm-plugin.md
```diff
@@ -88,13 +88,26 @@ is generating Swift types for the `Bar.proto` file with the `public` visibility.
...
+env PROTOC_PATH=/opt/homebrew/bin/protoc swift build
+
+// To start Xcode (Xcode MUST NOT be running before invoking this)
+env PROTOC_PATH=/opt/homebrew/bin/protoc xed .
+```
```
thomasvl, line 107: Should we mention (or example) anything about using it with `xcodebuild`?  Does the env var (and the `PATH` method) _just work_?

FranzBusch: Good point! `xcodebuild` works with the default `$PATH` as well, but the env var also works. I added the latter part as an example.

thomasvl (approved):

thomasvl:
I guess we need to pull this to `main` also.

thomasvl:
#1283 is a cherry-pick of it over to main.

Issue #1280, opened 2022-08-26
Warning in new version


akhakimzyanov:
Hi, update your library to new version (1.19.1 -> 1.20.0) and see new warning. We are using SPM.
<img width="477" alt="Swift-protobuf" src="https://user-images.githubusercontent.com/5399592/186945231-24c2d9dd-9917-4c99-9888-9ebf381a9113.png">

thomasvl:
What's your host platform, toolchain version, etc.?

akhakimzyanov:
MacOS - 12.5
Xcode - 13.4.1
App only for iOS

<img width="831" alt="Version" src="https://user-images.githubusercontent.com/5399592/186946283-4ca70155-86fb-46e4-ac17-c4ef7f0ef618.png"> - Swift Version default for this Xcode version

FranzBusch:
Ah this is because we adopted a new tools version. Let me open a PR to fix that up.

FranzBusch:
Should be fixed via: https://github.com/apple/swift-protobuf/pull/1281

akhakimzyanov:
Thank you for fast reply and fix!

thomasvl:
1.20.1 should be live for SwiftPM, CocoaPods push happening now.

PR #1281, opened 2022-08-26
Fix warning in new package.swift

FranzBusch:
We have created two new `Package.swift` manifests for landing the SPM plugin on the 1.x branch. While doing this we introduced a new warning, since the manifest now uses `executableTarget` for executable targets.


thomasvl (approved):

thomasvl:
Fixes #1280

PR #1286, opened 2022-09-01
Avoid the warning from `cat`

thomasvl:
Noticed regenerations were getting:

> cat: Sources/SwiftProtobuf/Docs.docc: Is a directory


tbkka (approved):
Good catch!

PR #1287, opened 2022-09-01
Descriptor improvements

thomasvl:
This is all within the plugin library, so it is API breaking (but we're making a 2.0 from main), I've kept the one api that would force swift-grpc to have to update, so they can hopefully not having to version lockstep around this, if something does come up we can look at adding back something to keep things simpler for them.

Generally went for consistency in naming and docs with the C++ descriptor so future things should be easier to maps and it should make new things coming in the future to upstream protobuf easier to model.

### DIFF
--- a/Sources/SwiftProtobufPluginLibrary/Descriptor.swift
+++ b/Sources/SwiftProtobufPluginLibrary/Descriptor.swift
@@ -42,6 +42,7 @@ import SwiftProtobuf
 ///
 /// This is like the `DescriptorPool` class in the C++ protobuf library.
 public final class DescriptorSet {
+  /// The list of `FileDescriptor`s in this set.
   public let files: [FileDescriptor]
   private let registry = Registry()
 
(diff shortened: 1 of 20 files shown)


allevato (approved):
This all seems reasonable from a high level.

Sources/SwiftProtobufPluginLibrary/Descriptor.swift
```diff
@@ -605,35 +878,59 @@ public final class ServiceDescriptor {
   }
 }
 
+/// Describes an individual service method.
+///
+/// SwiftProtobuf does *not* generate anything for these (or services), but
+/// they are here to support things that generate based off RPCs defined in
+/// .proto file (gRPC, etc.).
 public final class MethodDescriptor {
-  public let proto: Google_Protobuf_MethodDescriptorProto
```
glbrntt, line 609: Thanks for keeping grpc-swift in mind when making these changes!

grpc-swift uses `proto` for `clientStreaming` and `serverStreaming` which I see are added below. To maintain the API, could these ] properties be added to `MethodDescriptor` in a 1.x release? grpc-swift can bump its minimum required version such that it does not rely on the existence of `proto`.

We could also do the same for `fileDescriptor(named:)` to avoid carrying over `lookupFileDescriptor(named:)` into the 2.0 release.

thomasvl: Just put up #1291 to bring those two apis to the 1.x branch.

Also just added a commit to this to keep the needed things but taking them as deprecated, we can consider just dropping them before declaring a 2.0 here if we think there was enough time for things to migrate.

thomasvl: ps - There's a little more cleanup I wanted to do, but I'd like to land this first before doing more PRs. Once all that cleanup is done, we could cut a 1.x to get these those changes out, just seems worth waiting to make sure there isn't anything else worth doing to avoid the churn.

glbrntt: Great, thank you! We'll adopt them right after the next 1.x release.

PR #1288, opened 2022-09-01
updating to non-deprecated SwiftPM usage

heckj:
When following the README documentation, the code example for adding the dependencies to a `Package.swift` reported as being deprecated (I was using Swift 5.7 - beta release). The updates were simple and minimal, so I wanted to provide them back to prevent someone else from stumbling over the same.

The change updates the dependencies to use enumeration structures in SwiftPM instead of the strings mechanism which was prevalent in earlier Swift versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,10 +188,13 @@ the plugin above:
 
 ```swift
 dependencies: [
-    .package(name: "SwiftProtobuf", url: "https://github.com/apple/swift-protobuf.git", from: "1.6.0"),
+    .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.6.0"),
 ],
 targets: [
-    .target(name: "MyTarget", dependencies: ["SwiftProtobuf"]),
+    .target(
+      name: "MyTarget", 
+      dependencies: [.product(name: "SwiftProtobuf", package: "swift-protobuf")]
+    ),
 ]
 ```
 


README.md
```diff
@@ -188,10 +188,13 @@ the plugin above:
 
 ```swift
 dependencies: [
-    .package(name: "SwiftProtobuf", url: "https://github.com/apple/swift-protobuf.git", from: "1.6.0"),
+    .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.6.0"),
```
thomasvl, line 191: So this is undoing what you sent in in #976, so I guess different versions of swift want different things?

heckj: Yes, that's it in a nutshell. The manifest format for Package.swift continues to evolve, and this appears to be the "current" (swift 5.6/5.7) iteration. The older format does still work, and Xcode presents you with deprecation warnings in that case - so if you'd prefer to keep it "as is", I'd very much understand. When it was a deprecation warning, I figured it might be worth updating in the README though.

thomasvl: @tbkka should we land this now or wait for 5.7 to GM before updating things?

tbkka:
The beta is already out, so I don't see any compelling reason to wait.

thomasvl (approved):

PR #1292, opened 2022-09-07
Move from `weak` to `unowned`, past issues seem resolved.

thomasvl:


Sources/SwiftProtobufPluginLibrary/Descriptor.swift
```diff
@@ -755,12 +746,12 @@ public final class FieldDescriptor {
...
   /// nil if was declared at a global scope.
-  public private(set) weak var extensionScope: Descriptor?
+  public private(set) unowned var extensionScope: Descriptor?
 
   /// When this is a message field, the message's `Desciptor`.
-  public private(set) weak var messageType: Descriptor!
+  public private(set) unowned var messageType: Descriptor!
   /// When this is a enum field, the enum's `EnumDesciptor`.
-  public private(set) weak var enumType: EnumDescriptor!
+  public private(set) unowned var enumType: EnumDescriptor!
```
tbkka, line 754: I wonder if these should be regular optional instead of IUOs?

Sources/SwiftProtobufPluginLibrary/Descriptor.swift
```diff
@@ -543,9 +534,9 @@ public final class EnumValueDescriptor {
   public let number: Int32
 
   /// The .proto file in which this message type was defined.
-  public weak var file: FileDescriptor! { return enumType.file }
+  public unowned var file: FileDescriptor! { return enumType.file }
```
allevato, line 537: Taking a fresh look at this years later, it turns out that adding `weak`/`unowned` to a computed property [doesn't actually do anything for pure Swift properties](https://forums.swift.org/t/can-we-forbid-weak-keyword-for-computed-properties/19811), but the language has never gotten around to banning/warning against it. I wonder if we should drop it for these. On the one hand, it provides a smidge of documentation to users about the ownership of the returned value; on the other hand, it's easy for that "documentation" to become a lie if the internal implementation changes since nothing is actually enforced here. I could go either way.

Another thought: for these computed properties, we could make the property type just `FileDescriptor` (dropping the IUO) and force-unwrap inside the body.

Part of me thinks that while we're revisiting these it would be great to not leak the IUOs as part of the public getter interfaces, but that would require mirroring every parent-like property with a private stored/public computed property pair, and I don't know if that's worth all the trouble.

tbkka (approved):
Looks good.

Lukasa (approved):

Sources/SwiftProtobufPluginLibrary/Descriptor.swift
thomasvl: yup, on my list of cleanups (regrets) to do as part of 2.0.

Sources/SwiftProtobufPluginLibrary/Descriptor.swift
thomasvl: Yea, I've got some follow up coming here, just trying to do them in smaller chunks.

PR #1296, opened 2022-09-07
Add missing argument for plugin usage

mbrzozowski3:
Building silently fails without adding package to plugin


FranzBusch (approved):
Thanks for spotting this!

Lukasa (approved):

PR #1299, opened 2022-09-08
Enum value alias shuffle

thomasvl:


Sources/SwiftProtobufTestHelpers/Descriptor+TestHelpers.swift
```diff
@@ -9,9 +9,8 @@
 // -----------------------------------------------------------------------------
 
 import SwiftProtobuf
-import SwiftProtobufPluginLibrary
 
-extension Google_Protobuf_FileDescriptorProto {
+public extension Google_Protobuf_FileDescriptorProto {
```
thomasvl, line 13: The other option was to just copy this file into both Tests/* directories, but using a shared target seems slightly better.

Sources/protoc-gen-swift/Descriptor+Extensions.swift
```diff
@@ -298,4 +299,60 @@ extension EnumDescriptor {
...
+    func aliases(_ value: EnumValueDescriptor) -> [EnumValueDescriptor]? {
+      assert(mainValues.first!.enumType === value.enumType)
+      return aliasesMap[value.index]
+    }
+
+    /// Find the original for an alias.
+    ///
+    /// - Parameter value: The value descriptor to look up.
+    /// - Returns The original/main value if this was an alias otherwise `nil`.
+    func original(of: EnumValueDescriptor) -> EnumValueDescriptor? {
```
thomasvl, line 324: I couldn't come up name for this one, open to suggestions.

allevato: This seems fine; `originalValue(of:)` would send the wrong signal (it's returning the descriptor, not the value) and `originalValueDescriptor(of:)` is too verbose IMO.

allevato (approved):
I like the addition of a test helper module; it'll probably come in handy later too.

PR #1300, opened 2022-09-12
Bump to 1.20.2 for a release.

thomasvl:
This way we can cut a release with just the changes to the plugin library to allow grpc swift to look at updating things on their end.


thomasvl:
@tbkka good to land this and cut a release for the Plugin changes to allow grpc to support both 1.x and 2.x more easily?

tbkka:
Yep, that makes sense.

PR #1302, opened 2022-09-14
Changes to SPM Plugin documentation

gjcairo:
This PR introduces some changes to the SPM protobuf codegen plugin documentation.

### DIFF
--- a/Sources/protoc-gen-swift/Docs.docc/spm-plugin.md
+++ b/Sources/protoc-gen-swift/Docs.docc/spm-plugin.md
@@ -19,7 +19,7 @@ configuration file which will be used to customize the invocation of `protoc`.
 First, you must ensure that you have the `protoc` compiler installed.
 There are multiple ways to do this. Some of the easiest are:
 
-1. If you are on MacOS, installing it via `brew install protoc`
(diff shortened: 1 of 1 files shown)


Sources/protoc-gen-swift/Docs.docc/spm-plugin.md
```diff
@@ -109,18 +114,18 @@ env PROTOC_PATH=/opt/homebrew/bin/protoc xed .
...
 }
-
 ```
 
 > Warning: The configuration file option only solves the problem for leaf packages that are using the Swift package manager
 plugin since there you can point the package manager to the right binary. The environment variable
 does solve the problem for transitive packages as well; however, it requires your users to set
 the variable now. In general we advise against adopting the plugin as a non-leaf package!
+
+* You can start Xcode by running `$ xed .` from the command line from the directory your project is located - this should make `$PATH` visible to Xcode.
```
FranzBusch, line 131: I think this should be the first option that we list nowadays, it is the easiest to get started with. I think it is worth calling out though that the `protoc` needs to be in the `$PATH` for this to work.

FranzBusch (approved):

PR #1307, opened 2022-09-21
Align SPM plugin docs with grpc-swift

FranzBusch:
# Motivation
We added some docs to clarify the paths of files in grpc-swift which we should state here as well.


thomasvl (approved):
Should we also do a full example at some point?

PR #1315, opened 2022-10-10
Build protobuf into an output directory.

thomasvl:
https://github.com/protocolbuffers/protobuf/issues/10749 opened for the break, seems like cmake can no longer build in place.


thomasvl:
For the caching, I put the directory still under the `protobuf` directory, we could be it beside things, but then we'd need to tweak the caching to cache both directories.

allevato (approved):

PR #1317, opened 2022-10-10
Be more accepting for generation options.

thomasvl:
Don't require strict casing of the `OutputNaming` and `Visibility` generation options. Also accept underscore splits on `OutputNaming`.

### DIFF
--- a/Sources/protoc-gen-swift/GeneratorOptions.swift
+++ b/Sources/protoc-gen-swift/GeneratorOptions.swift
@@ -11,15 +11,39 @@
 import SwiftProtobufPluginLibrary
 
 class GeneratorOptions {
-  enum OutputNaming : String {
-    case fullPath = "FullPath"
-    case pathToUnderscores = "PathToUnderscores"
-    case dropPath = "DropPath"
+  enum OutputNaming {
+    case fullPath
+    case pathToUnderscores
+    case dropPath
+
+    init?(flag: String) {
+      switch flag.lowercased() {
+      case "fullpath", "full_path":
+        self = .fullPath
+      case "pathtounderscores", "path_to_underscores":
+        self = .pathToUnderscores
+      case "droppath", "drop_path":
+        self = .dropPath
+      default:
+        return nil
+      }
+    }
   }
 
-  enum Visibility : String {
-    case `internal` = "Internal"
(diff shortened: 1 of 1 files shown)


thomasvl:
@FranzBusch @klundberg  fyi

Sources/protoc-gen-swift/GeneratorOptions.swift
```diff
@@ -15,11 +15,35 @@ class GeneratorOptions {
     case fullPath = "FullPath"
     case pathToUnderscores = "PathToUnderscores"
     case dropPath = "DropPath"
+
+    init?(rawValue: String) {
```
allevato, line 19: I would give the argument label here and below a different name, like `init?(flag: String)`, since it's no longer a 1:1 mapping between the actual raw values, accepting that this will also require updating the call-sites.

Going further, do we use the raw values for anything else other than parsing the command line flag? If not, we should just drop them and remove the `String` type from the enum.

thomasvl: Done.

thomasvl:
> @FranzBusch it may make sense to do the same sorta thing to the enum in the plugin.

Hm, makes me also wonder if we should set up anything to share between the plugin and the generator.  Or should we make the plugin just pass the value directly through without the enum and let the generation do the validation/failure instead?

allevato (approved):

Lukasa (approved):

FranzBusch:
> @FranzBusch it may make sense to do the same sorta thing to the enum in the plugin.

Sure let me add it to my PR.

> Hm, makes me also wonder if we should set up anything to share between the plugin and the generator. Or should we make the plugin just pass the value directly through without the enum and let the generation do the validation/failure instead?

Sadly, that is not possible right now. You can't nicely share code with an SPM plugin since the plugin itself can't have any dependencies.

FranzBusch:
> I guess the only option then would be to just make it a plain String (and not Enum) so it passes thru and let the validation happen within the tool running.

Yes basically. However, I think there is real value in doing validation in the SPM plugin since we can surface the errors in a nicer way during the plugin evaluation. Otherwise this would cause a build time error.

PR #1321, opened 2022-10-13
[WiP] Improve the performance when serializing very large objects to TextFormat

tbkka:
The TextFormatEncoder relies on appending a few bytes at a time to an `Array<UInt8>`.  For large objects, this seems to result in a lot more calls to memmove than expected.

To mitigate this, the TextFormateEncoder now uses two arrays: all bytes are first appended to a smaller (4096 byte) fixed-size array; when that fills up, it is appended to the larger array as a block.

This was found by OSS-Fuzz:  It found an input of ~1MB that would decode successfully but took more than 60s to re-encode.  The example it found now encodes in just over 1s.

Resolves https://oss-fuzz.com/testcase?key=6026353006215168

### DIFF
--- a/Sources/SwiftProtobuf/TextFormatEncoder.swift
+++ b/Sources/SwiftProtobuf/TextFormatEncoder.swift
@@ -33,57 +33,91 @@ private let tab = [UInt8](repeating: asciiSpace, count: tabSize)
 
 /// TextFormatEncoder has no public members.
 internal struct TextFormatEncoder {
+    private static let foyerSize = 4096
+    private var foyer = Array<UInt8>(repeating: 0, count: TextFormatEncoder.foyerSize)
+    private var foyerIndex = 0
+
     private var data = [UInt8]()
     private var indentString: [UInt8] = []
-    var stringResult: String {
-        get {
-            return String(bytes: data, encoding: String.Encoding.utf8)!
+
+    internal mutating func constructFinalResult() -> String {
+        if foyerIndex > 0 {
+            let subfoyer = foyer[0..<foyerIndex]
+            data.append(contentsOf: subfoyer)
+            foyerIndex = 0
         }
+        return String(bytes: data, encoding: String.Encoding.utf8)!
+    }
+
+    // Append each byte to the foyer, append and
+    // recycle the foyer when it fills up.
+    private mutating func append(byte: UInt8) {
+        if foyerIndex > TextFormatEncoder.foyerSize - 1 {
+            data.append(contentsOf: foyer)
+            foyerIndex = 0
+        }
+        foyer[foyerIndex] = byte
+        foyerIndex += 1
     }
 
     internal mutating func append(staticText: StaticString) {
         let buff = UnsafeBufferPointer(start: staticText.utf8Start, count: staticText.utf8CodeUnitCount)
-        data.append(contentsOf: buff)
+        for b in buff {
+            append(byte: b)
+        }
     }
 
     internal mutating func append(name: _NameMap.Name) {
-        data.append(contentsOf: name.utf8Buffer)
+        for b in name.utf8Buffer {
+            append(byte: b)
+        }
     }
 
     internal mutating func append(bytes: [UInt8]) {
-        data.append(contentsOf: bytes)
+        for b in bytes {
+            append(byte: b)
+        }
+    }
+
+    private mutating func append(bytes: UnsafeRawBufferPointer) {
+        for b in bytes {
+            append(byte: b)
+        }
     }
 
     private mutating func append(text: String) {
-        data.append(contentsOf: text.utf8)
+        for u in text.utf8 {
+            append(byte: u)
+        }
     }
 
     init() {}
 
     internal mutating func indent() {
-        data.append(contentsOf: indentString)
+        append(bytes: indentString)
     }
 
     mutating func emitFieldName(name: UnsafeRawBufferPointer) {
         indent()
-        data.append(contentsOf: name)
+        append(bytes: name)
     }
 
     mutating func emitFieldName(name: StaticString) {
         let buff = UnsafeRawBufferPointer(start: name.utf8Start, count: name.utf8CodeUnitCount)
-        emitFieldName(name: buff)
+        indent()
+        append(bytes: buff)
     }
 
     mutating func emitFieldName(name: [UInt8]) {
         indent()
-        data.append(contentsOf: name)
+        append(bytes: name)
     }
 
     mutating func emitExtensionFieldName(name: String) {
         indent()
-        data.append(asciiOpenSquareBracket)
+        append(byte: asciiOpenSquareBracket)
         append(text: name)
-        data.append(asciiCloseSquareBracket)
+        append(byte: asciiCloseSquareBracket)
     }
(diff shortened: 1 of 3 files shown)


Sources/SwiftProtobuf/TextFormatEncoder.swift
```diff
@@ -33,57 +33,91 @@ private let tab = [UInt8](repeating: asciiSpace, count: tabSize)
...
     }
 
     mutating func emitFieldName(name: StaticString) {
         let buff = UnsafeRawBufferPointer(start: name.utf8Start, count: name.utf8CodeUnitCount)
-        emitFieldName(name: buff)
+        indent()
+        append(bytes: buff)
     }
 
     mutating func emitFieldName(name: [UInt8]) {
```
thomasvl, line 111: Is this still used?

Sources/SwiftProtobuf/TextFormatEncoder.swift
```diff
@@ -33,57 +33,91 @@ private let tab = [UInt8](repeating: asciiSpace, count: tabSize)
...
+            foyerIndex = 0
         }
+        foyer[foyerIndex] = byte
+        foyerIndex += 1
     }
 
     internal mutating func append(staticText: StaticString) {
         let buff = UnsafeBufferPointer(start: staticText.utf8Start, count: staticText.utf8CodeUnitCount)
-        data.append(contentsOf: buff)
+        for b in buff {
```
thomasvl, line 65: These loops (this and the next few functions) look like potential bottlenecks again, should they flush and directly add to `data` or maybe use some helper that copies chunks that fit into the buffer instead?

Sources/SwiftProtobuf/TextFormatEncoder.swift
```diff
@@ -33,57 +33,91 @@ private let tab = [UInt8](repeating: asciiSpace, count: tabSize)
...
+            append(byte: b)
+        }
     }
 
     internal mutating func append(name: _NameMap.Name) {
-        data.append(contentsOf: name.utf8Buffer)
+        for b in name.utf8Buffer {
+            append(byte: b)
+        }
```
lorentey, line 73: This change looks suspect to me. `append(contentsOf:)` is already supposed to do the right thing and it avoids repeatedly checking capacity.

thomasvl: This is sorta what I was referencing, this would have to atleast do the "flush" of the other buffer before it could directly append.

Sources/SwiftProtobuf/TextFormatEncoder.swift
tbkka: For the particular example found by OSS-Fuzz, almost all of the appends are one or just a couple of bytes.  It appears that in this case, the overhead of appending to data outweighs the cost of doing repeated single-byte stores into the array.

lorentey: I'd expect that repeated calls to `append(_:)` would generally still be slower than a single `append(contentsOf:)` call, _especially_ if the source buffer is just an UnsafeBufferPointer.

`append(_:)` isn't (supposed to be) magically cheaper than chunking...

Sources/SwiftProtobuf/TextFormatEncoder.swift
```diff
@@ -33,57 +33,91 @@ private let tab = [UInt8](repeating: asciiSpace, count: tabSize)
 
 /// TextFormatEncoder has no public members.
 internal struct TextFormatEncoder {
+    private static let foyerSize = 4096
+    private var foyer = Array<UInt8>(repeating: 0, count: TextFormatEncoder.foyerSize)
```
lorentey, line 37: Why not just reserve 4096 bytes worth of capacity directly in `data`?

tbkka: I'm not sure what you mean.  Do you mean starting `data` out with 4096 bytes, or forcing it to extend the allocation by at least 4096 bytes after/before each append?

lorentey: I mean literally adding this to TextFormatEncoder's initializer, and doing nothing else:

```
data.reserveCapacity(4096)
```

lorentey: (The only benefit of this PR that I can see is that it avoids the initial exponential ramp-up from 0 to 4096 bytes in `data`'s backing store. And that is far more easily avoided by simply reserving capacity directly in `data`. I'm pretty sure doing that will be measurably better than the suggested code, too -- e.g., it wouldn't copy data through two buffers for no reason! -- but who knows how good the optimizer is these days...)

Once `data` grows beyond 4096 bytes in size, it doesn't matter if we append 1 byte or 4096, we get the same allocation pattern.

tbkka: I just did that experiment:
* With our current `main` code, this example runs in 44.1s
* With the addition of `data.reserveCapacity`, it takes 43.4s
* With the changes in this PR, it takes 1.5s

I did some profiling of the `main` code.  That showed that over 90% of the total CPU time was being spent in `memmove` triggered by `append(contentsOf:)`.   I'll take another careful look to see if I can narrow down whether that's the `memmove` to reallocate the backing store or the `memmove` that copies data from the string that's being appended.

lorentey: Wow! That is very _very_ unexpected and it definitely needs a closer, stdlib-level look. 

I'd expect copying data from a string to the buffer would be done using `memcpy`, not `memmove` -- unless of course the optimizer is trying to be clever.

I see `putStringValue` is appending data byte by byte (using both `append(_:)` and `append(contentsOf:)`, which is a bit silly -- the stdlib should rather provide an append equivalent of `Array(unsafeUninitializedCapacity:initializingWith:)` to speed that up a bit. However, I really don't see how appending a series of single bytes could ever lead to a drastic increase of Array resizes, _unless there is a critical logic bug within the stdlib_. My own ad-hoc attempts at recreating this did not succeed so far; I'm seeing the expected exponential resizes, not a series of linear ones.

tbkka:
Switched this to a "draft" until I can do some more research into what exactly is going on here.

tbkka:
Thanks to @lorentey for noticing that the Visitor _copies_ the encoder into a sub-Visitor for every child message.  This results in multiple references to the `data` array, which triggers copy-on-write for the entire array for every sub message.  So of course a test case with tens of thousands of nested child messages would take quite a while to serialize.  The easy fix would make the encoder a class, but I'm going to take a little time to tinker with other options before I do that.

thomasvl:
Oh, in the _Any_ message case?  Can we do the thing were we save off the fields and reuse the the visitor instead to avoid that?

tbkka:
Actually, it seems to be maps in this case.  I was planning to change that to reuse the visitor, just as we do for message fields.  I'll look at Any also while I'm here.  (I'll look at JSON after I get TextFormat fixed up.)

tbkka:
Closed in favor of #1322 which solves the actual underlying problem.

PR #1322, opened 2022-10-17
TextFormat Performance:  Avoid quadratic behavior for map and Any fields

tbkka:
Credit: OSS-Fuzz

The TextFormat encoding visitor would create a new child visitor when traversing into map or Any fields.  This resulted in multiple references to the underlying byte array, triggered unnecessary copy-on-write operations, which led to quadratic behavior for large messages with lots of map and/or Any fields.

Regular message fields didn't have this problem because we didn't create a new visitor in that case -- instead we adjusted the configuration of the current visitor object.  This adjusts the handling of map and Any fields to work the same way.

The test case demonstrates > 100x performance improvement for a multi-megabyte text encoding of a message containing 50,000 map fields or Any fields.  The test was tuned so that on a fast machine it used to take >60s but now takes well under 1s.  Hopefully, if anything regresses here, that will be a glaring enough difference to call attention to it.

### DIFF
--- a/Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
+++ b/Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
@@ -33,12 +33,6 @@ internal struct TextFormatEncodingVisitor: Visitor {
   /// Creates a new visitor that serializes the given message to protobuf text
   /// format.
   init(message: Message, options: TextFormatEncodingOptions) {
-    self.init(message: message, encoder: TextFormatEncoder(), options: options)
-  }
-
-  /// Creates a new visitor that serializes the given message to protobuf text
-  /// format, using an existing encoder.
-  private init(message: Message, encoder: TextFormatEncoder, options: TextFormatEncodingOptions) {
     let nameMap: _NameMap?
     if let nameProviding = message as? _ProtoNameProviding {
         nameMap = type(of: nameProviding)._protobuf_nameMap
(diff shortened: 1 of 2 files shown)


tbkka:
This replaces #1321

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
```diff
@@ -308,13 +308,27 @@ internal struct TextFormatEncodingVisitor: Visitor {
   internal mutating func visitAnyVerbose(value: Message, typeURL: String) {
       encoder.emitExtensionFieldName(name: typeURL)
       encoder.startMessageField()
-      var visitor = TextFormatEncodingVisitor(message: value, encoder: encoder, options: options)
```
thomasvl, line 311: Can this old ctor go away to avoid this risk?

Tests/SwiftProtobufTests/Test_TextFormat_Performance.swift
```diff
@@ -0,0 +1,100 @@
...
+    // Each of the following should be under 1s on a reasonably
+    // fast machine (originally developed on an M1 MacBook Pro).
+    // If they are significantly slower than that, then something
+    // may be amiss.
+
+    func testEncoding_manyMapsEncoding_shouldBeUnder1s() {
+        let repeats = 50000
+
+        let child = (
+            "repeated_message {\n"
```
thomasvl, line 30: You probably can use a multi line string with the current min requirements.

tbkka: Personally, I'm often confused about how indentation works for multi line strings, so I prefer to be a bit more explicit.

thomasvl (approved):

Lukasa (approved):

PR #1323, opened 2022-10-18
JSON Performance: Avoid quadratic encoding for maps

tbkka:
OSS-Fuzz found that TextFormat encoding could degenerate into quadratic behavior for certain cases.  This fixes a similar issue with the JSON encoder.

Specifically, when encoding map fields, the JSON encoding visitor would construct a child visitor using the same underlying encoder. This results in multiple references to the backing byte array which triggers copy-on-write for that array.

To avoid that, we modify the JSON encoding visitor to construct a child visitor with a new empty encoder, and then append the contents of that encoder back to the parent.  This avoids the quadratic behavior in this case.

To try to prevent this from regressing, I've added tests that are tuned to take less than 1s on an M1 MacBook Pro with the new code and more than 30s with the old code. If this does regress, that should be a dramatic enough slowdown to attract someone's attention.

### DIFF
--- a/Sources/SwiftProtobuf/JSONEncoder.swift
+++ b/Sources/SwiftProtobuf/JSONEncoder.swift
@@ -77,6 +77,8 @@ internal struct JSONEncoder {
         }
     }
 
+    internal var bytesResult: [UInt8] { return data }
+
     /// Append a `StaticString` to the JSON text.  Because
     /// `StaticString` is already UTF8 internally, this is faster
     /// than appending a regular `String`.
@@ -98,6 +100,10 @@ internal struct JSONEncoder {
     /// Append a `String` to the JSON text.
     internal mutating func append(text: String) {
         data.append(contentsOf: text.utf8)
+    }
+
+    internal mutating func append(bytes: [UInt8]) {
+        data.append(contentsOf: bytes)
     }
 
     /// Append a raw utf8 in a `Data` to the JSON text.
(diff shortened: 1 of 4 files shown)


tbkka:
This is the equivalent of #1322 for JSON.

Lukasa (approved):

tbkka:
@thomasvl   Are you familiar with the [MessageSet support](https://github.com/apple/swift-protobuf/blob/5090df2fc99dc245d8c1aebc56245de9d0943486/Sources/SwiftProtobuf/BinaryEncodingVisitor.swift#L345) ?  It looks like that might have the same issue, and likely could be fixed with the same approach used here.

thomasvl:
> @thomasvl   Are you familiar with the [MessageSet support](https://github.com/apple/swift-protobuf/blob/5090df2fc99dc245d8c1aebc56245de9d0943486/Sources/SwiftProtobuf/BinaryEncodingVisitor.swift#L345) ?  It looks like that might have the same issue, and likely could be fixed with the same approach used here.

I won't be able to look this week, but wouldn't be surprised if it has the same issue. I'll try to look late next week if you don't beat me to it.

thomasvl:
> > @thomasvl   Are you familiar with the [MessageSet support](https://github.com/apple/swift-protobuf/blob/5090df2fc99dc245d8c1aebc56245de9d0943486/Sources/SwiftProtobuf/BinaryEncodingVisitor.swift#L345) ?  It looks like that might have the same issue, and likely could be fixed with the same approach used here.
> 
> I won't be able to look this week, but wouldn't be surprised if it has the same issue. I'll try to look late next week if you don't beat me to it.

Opened up https://github.com/apple/swift-protobuf/pull/1329 from taking a quick look at this.  I believe the existing code was ok because of how simple the types were, but that PR might make it less likely that future changes to the types could make this expensive.

PR #1326, opened 2022-10-26
Use own protocol instead of Data/ContiguousBytes

gjcairo:
## Motivation

This is the first step of several in the path to removing the usage of `Data` (and eventually `Foundation`) from `swift-protobuf`. This is related to https://github.com/apple/swift-protobuf/issues/816.

## Modifications

* Created a new `SwiftProtobufContiguousBytes` protocol and conformed both `[UInt8]` and (temporarily) `Data` to it.
* Changed some of the code in the deserialisation path to use this protocol instead of `Foundation`'s existing `ContiguousBytes` or `Data`.
* This is just a starting point: I couldn't make more changes without modifying the codegen and regenerating all of the protos, so I decided to split this bit of work first. A follow-up PR will contain those changes.

## Result

The deserialisation path in `swift-protobuf` now partially uses the new `SwiftProtobufContiguousBytes` protocol instead of `Foundation`'s `Data`/`ContiguousBytes`.


Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -99,7 +99,7 @@ internal class AnyMessageStorage {
     get {
       switch state {
       case .binary(let value):
-        return value
+        return Data(value)
```
FranzBusch, line 102: Can we track it somewhere that we still need to clean this up? Maybe with a TODO or adding it to an issue.

Sources/SwiftProtobuf/SwiftProtobufContiguousBytes.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+// Copyright (c) 2022 Apple Inc. and the project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+
+public protocol SwiftProtobufContiguousBytes {
```
FranzBusch, line 11: Can we add some docs to this stating why we need it and what owners of other bag of bytes type should do with it?

FranzBusch (commented):
Great start! I think this is an amazing first step to get rid of the Foundation dependency. I left two small inline comments. The only real fear that I have is that we forget some of the intermediate "hacks" that we are doing here like bridging to `Data` or to `Array`. Especially the latter one could be easy to forget later on.

Sources/SwiftProtobuf/AnyMessageStorage.swift
gjcairo: I've actually already removed this in the follow-up PR I'll send. I can still add the TODO if you think it's useful?

FuzzTesting/Sources/FuzzJSON/main.swift
```diff
@@ -8,7 +6,7 @@ public func FuzzJSON(_ start: UnsafeRawPointer, _ count: Int) -> CInt {
   var msg: Fuzz_Testing_Message?
   do {
     msg = try Fuzz_Testing_Message(
-      jsonUTF8Data: Data(bytes),
+      jsonUTF8Data: Array(bytes),
```
thomasvl, line 9: Should this api also get renamed since it is no long a `Data` argument?

Sources/SwiftProtobuf/Data+SwiftProtobufContiguousBytes.swift
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+
+import Foundation
+
+extension Data: SwiftProtobufContiguousBytes {}
```
thomasvl, line 13: This should probably get documented also.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -57,31 +57,6 @@ extension Message {
...
-  ///     fields.
-  ///   - partial: If `false` (the default), this method will check
-  ///     `Message.isInitialized` before encoding to verify that all required
-  ///     fields are present. If any are missing, this method throws
-  ///     `BinaryEncodingError.missingRequiredFields`.
-  ///   - options: The BinaryDecodingOptions to use.
-  /// - Throws: `BinaryDecodingError` if decoding fails.
-  @inlinable
-  public init(
-    serializedData data: Data,
```
thomasvl, line 76: I wonder if we should keep the old names to help folks in migration.

Yes, we're in a 2.0 phase, but this sorta breaking change could be really hard for teams with large code bases and multiple dependencies that all use SwiftProtobuf. Seems like allowing those sub dependencies to migrate one at a time would be worth while.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -107,34 +82,6 @@ extension Message {
...
-  ///     extensions in this message or messages nested within this message's
-  ///     fields.
-  ///   - partial: If `false` (the default), this method will check
-  ///     `Message.isInitialized` before encoding to verify that all required
-  ///     fields are present. If any are missing, this method throws
-  ///     `BinaryEncodingError.missingRequiredFields`.
-  ///   - options: The BinaryDecodingOptions to use.
-  /// - Throws: `BinaryDecodingError` if decoding fails.
-  @inlinable
-  public mutating func merge(
```
thomasvl, line 129: Same migration comment.

Sources/SwiftProtobuf/Message+JSONAdditions.swift
```diff
@@ -103,7 +103,7 @@ extension Message {
   /// - Parameter options: The JSONDecodingOptions to use.
   /// - Throws: `JSONDecodingError` if decoding fails.
   public init(
-    jsonUTF8Data: Data,
```
thomasvl, line 106: As mentioned above, should still still be "Data" in the name?  As is the PR is inconsistent on when it changed names, so some things are breaking and other aren't.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -97,7 +72,7 @@ extension Message {
   ///   - options: The BinaryDecodingOptions to use.
   /// - Throws: `BinaryDecodingError` if decoding fails.
   @inlinable
-  public init<Bytes: ContiguousBytes>(
```
thomasvl, line 100: Can/Should `ContiguousBytes` conform to the custom protocol?  i.e. - if folks were using this api, what is their path forward?

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
FranzBusch: IMO we should provide this method in a new SwiftProtobufFoundationCompat module. This should ease migration

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
FranzBusch: You cannot extend a protocol with a conformance to another protocol. I think the same applies here and we should provide this in the foundation compat module.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
thomasvl: Sounds like there some plan being worked on since I haven't heard "SwiftProtobufFoundationCompat" 😄 , maybe doing a bit of a outline on the expected path forward on the issue would help rather then PRs that are only partial pictures into a plan.

thomasvl:
This might be too much work, but one idea for this PR

1. Add in the new _Core_ and _Foundation_ modules, but don't change the unitests or fuzz tests, this would help show things are working correctly.
2. Then in a followup commit, update the tests to directly depend on the _Core_ module with the needed code updates.

FuzzTesting/Sources/FuzzJSON/main.swift
gjcairo: Will change to `jsonUTF8Bytes`

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
gjcairo: I'll add this this method (i.e. with the old signature) to the FoundationCompat module, and have it call the new method.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
gjcairo: Same response as above.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
gjcairo: I'm not sure what to do here. I've added the `Data` methods (`merge` and `init`) to the `SwiftProtobufFoundationCompat` module, which was easy enough because `Data` conforms to `SwiftProtobufContiguousBytes`. 

However in this case, as Franz said, I cannot extend `ContiguousBytes` to conform to our new protocol. As a consequence, I can't call the new methods in `SwiftProtobufCore` because objects conforming to `ContiguousBytes` may not conform to `SwiftProtobufContiguousBytes`. And I can't simply copy the implementation of the methods because `_merge` is `internal` and so is inaccessible from the FoundationCompat module (and based on the comment left on the method, it was a very conscious decision to take advantage of `@inlinable`). 
Furthermore, `_merge` uses a `BinaryDecoder` which (based on the comments too) we want to keep internal.

Any suggestions to get around this? Or should we just drop support for `ContiguousBytes`?

Sources/SwiftProtobuf/SwiftProtobufContiguousBytes.swift
```diff
@@ -0,0 +1,28 @@
...
+// -----------------------------------------------------------------------------
+
+/// Conformance to this protocol gives users a way to provide their own "bag of bytes" types
+/// to be used for serialization and deserialization of protobufs.
+///
+/// By conforming your own types to this protocol, you will be able to pass instances of said types
+/// directly to `SwiftProtobuf.Message`'s deserialisation methods
+/// (i.e. `init(serializedData:)` for binary format and `init(jsonUTF8Data:)` for JSON).
+// TODO: extend doc to include how this will be used in serialization once that API change has been finalised.
+public protocol SwiftProtobufContiguousBytes {
```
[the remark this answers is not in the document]
gjcairo: That would basically be explaining why we want to remove `Foundation` (since `ContiguousBytes` also lives in `Foundation`). Something like:
>/// This protocol exists because the `ContiguousBytes` protocol, which would have been a viable alternative,
>/// lives inside `Foundation`.
>/// Importing `Foundation` results in very big binaries, which can be a problem for some swift-protobuf usecases.

Do we want to go there in this comment though?

Sources/SwiftProtobuf/Message+JSONAdditions.swift
gjcairo: Yes that's true. I'll rename these, but only in the deserialisation path (i.e. I'll keep the `jsonUTF8Data()` methods for now) since serialisation changes will follow in a later PR.
To keep things in line with the comments on binary decoding, I'll copy this method with the old name (i.e. using `jsonUTF8Data:`) into `SwiftProtobufFoundationCompat` to aid in migration.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
Lukasa: We could expose `_merge` via SPI.

gjcairo: I did think about this, but we support back to Swift 5.0 and I believe `@_spi` was introduced with 5.3 based on what I could find - is that correct?

Lukasa: Yes, probably. We could also do the awkward workaround for older Swifts: make the type public but with a leading underscore.

Sources/SwiftProtobuf/SwiftProtobufContiguousBytes.swift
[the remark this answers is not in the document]
thomasvl: ...provides a general interface for bytes since the Swift Standard Library currently does not provide such a protocol.

Or something like that?

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
thomasvl: Since the protocols match, maybe skip it as all someone has to do is declare conformance for their types themselves?

Sources/Conformance/main.swift
```diff
@@ -24,7 +24,7 @@ import Darwin.C
 import Foundation
```
thomasvl, line 24: Is Foundation still needed for something else?

Sources/SwiftProtobufCore/TextFormatEncodingVisitor.swift
```diff
@@ -248,6 +248,10 @@ internal struct TextFormatEncodingVisitor: Visitor {
   }
 
   mutating func visitSingularBytesField(value: Data, fieldNumber: Int) throws {
+      try visitSingularBytesField(value: Array(value), fieldNumber: fieldNumber)
+  }
+
+  mutating func visitSingularBytesField(value: [UInt8], fieldNumber: Int) throws {
```
thomasvl, line 254: I guess at some point this sorta change need to happen to the base Visitor?  Not sure what this will end up for the Compat module and api in general.

Sources/SwiftProtobufCore/Message+BinaryAdditions.swift
```diff
@@ -97,7 +72,7 @@ extension Message {
   ///   - options: The BinaryDecodingOptions to use.
   /// - Throws: `BinaryDecodingError` if decoding fails.
   @inlinable
-  public init<Bytes: ContiguousBytes>(
+  public init<Bytes: SwiftProtobufContiguousBytes>(
```
thomasvl, line 75: The existing `ContiguousBytes` apis sorta broke the model, but it used to be the apis used "Serialized" for all the "binary" encoding based apis (and "JSON" and "TextFormat" for those).

Since we're dropping all the "Data" apis that are named with "Serialized", should we do a slight api correct and call this: `init(serializedBytes:...)`

@tbkka ?

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
gjcairo: I think it will stop matching when the encoding path stops using `Data` (coming in a future PR), as we'll need to add a couple more things to it. It still shouldn't be wildly different, so I wouldn't mind just dropping it and have users conform to the new one. But if we're worried about migration then I can go for exposing the decoders but with a leading underscore, as @Lukasa suggested.

Sources/SwiftProtobufCore/Message+BinaryAdditions.swift
gjcairo: I'm fine with making that change, I actually prefer the name myself.

Sources/SwiftProtobufCore/TextFormatEncodingVisitor.swift
gjcairo: Yes, the idea is to actually change the base visitor to be generic on a `SwiftProtobufContiguousBytes`-conforming type (i.e. `mutating func visitSingularBytesField<Bytes: SwiftProtobufContiguousBytes>(value: Bytes, fieldNumber: Int) throws`), so the Compat module wouldn't need to redefine the `Visitor` protocol or do anything special, since it will already be conforming `Data`. Was planning on making that change in the next PR to try to split the changes a bit.

Sources/Conformance/main.swift
gjcairo: Yes, there is the `func writeResponse(data: Data)` method. This ends up taking the result of calling `serializedData()` on a proto, so it will be changed once the serialisation path is rid of `Data`.

Sources/SwiftProtobuf/SwiftProtobufContiguousBytes.swift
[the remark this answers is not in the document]
gjcairo: Yeah I think I like that better, I'll change it. Thanks.

Sources/Conformance/main.swift
```diff
@@ -54,7 +54,7 @@ func buildResponse<Bytes: SwiftProtobufContiguousBytes>(serializedData: Bytes) -
 
     let request: Conformance_ConformanceRequest
     do {
-        request = try Conformance_ConformanceRequest(contiguousBytes: serializedData)
+        request = try Conformance_ConformanceRequest(serializedBytes: serializedData)
```
FranzBusch, line 57: Just want to double check here with @thomasvl. Since we talked about keeping the breaking changes limited this is now quite a wide reaching one. Even though we still support `Data` based `init`s we changed the name of the parameter. I am thinking if we should still provide a `convenience init` in `SwiftProtobufCompat` that maintains the old naming with a deprecation so that adopters won't break on this right away?

FranzBusch (approved):
Overall this looks great! I left one comment w.r.t to breaking changes otherwise approved from my side.

Sources/Conformance/main.swift
thomasvl: Part of me wishes we didn't change the tests/conformance tests/fuzz tests, so we can see that the foundation shims were all working (then clean them up in a follow up commit/pr).

thomasvl (commented):
I think I'm generally on board with these, but I'm having a hard time convincing myself how breaking/non breaking this is.  As mentioned in the other comment, if there was a single commit after all other changes that updated Conformance, Tests, FuzzTesting, I might be more convinced FoundationCompat is doing the job we'd hopped, but at the moment, I'm not sure.

Sources/Conformance/main.swift
FranzBusch: Agreed, the previous PR did some changes which we maybe should look into again.

FranzBusch:
> I think I'm generally on board with these, but I'm having a hard time convincing myself how breaking/non breaking this is. As mentioned in the other comment, if there was a single commit after all other changes that updated Conformance, Tests, FuzzTesting, I might be more convinced FoundationCompat is doing the job we'd hopped, but at the moment, I'm not sure.

I think I agree with this but only for Conformance and Fuzz tests. The actual unit tests of the `SwiftProtobufCoreTests` target need to change here but that should be fine IMO.

Sources/SwiftProtobufCore/Message+BinaryAdditions.swift
tbkka: I like that idea, too.

Sources/SwiftProtobufCore/TextFormatEncodingVisitor.swift
tbkka: We should watch the performance impact of this.   Invoking generics across a module boundary is tricky to get right:  If you're not careful, you'll end up either ruining performance (because of meta type lookups on each call) or bloating code size (due to excessive inlining).  It might make sense to keep one or two statically-typed variations (e.g., `[UInt8]`) in addition to the generic version so that suitably-typed users can avoid that.

FuzzTesting/Sources/FuzzBinary/main.swift
```diff
@@ -2,6 +2,8 @@ import Foundation
 
 import FuzzCommon
 
+import SwiftProtobuf
```
thomasvl, line 5: So these could have been FoundationCompat? But no matter which one picks, to maintain Data support folks will have to explicitly import atleast one of them?

gjcairo: They could have, yes. The codegen was changed to import `SwiftProtobufCore` in the autogenerated files instead of `SwiftProtobuf`, so unless `SwiftProtobuf` or `SwiftProtobufFoundationCompat` are also imported, the `Data` methods won't be included.

An alternative would be to re-export `SwiftProtobufFoundationCompat` in `FuzzCommon`, that would avoid the import here.

thomasvl: > An alternative would be to re-export `SwiftProtobufFoundationCompat` in `FuzzCommon`, that would avoid the import here.

Ack.

I was mainly asking to think about the general impacts as these CLs are being done.  If we do go down the path some some generation option for Foundation support, we could change what `import` is generated.

Anyway, not expecting a change because of this, was just asking to make sure I'm reasoning correctly on why the files did have to change.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
thomasvl: This one is edge case enough I'm also ok dropping it completely.

@tbkka ?

Sources/SwiftProtobufCore/TextFormatEncodingVisitor.swift
gjcairo: I agree, I'll take special care and look into it when making that change. You think the changes so far are good enough though to go ahead with this PR?

gjcairo:
@thomasvl once this is approved, should I fixup/squash some of the commits so the history is a bit cleaner? Or do you prefer to do that yourself when merging?

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
tbkka: But yeah, I'm fine just dropping it for now.

tbkka (approved):

thomasvl (approved):

gjcairo:
I've tidied the commits - still left the two distinct commits where I rolled back the changes to the tests and then added them back so there's some history that shows we did test with the FoundationCompat module.

@tbkka / @thomasvl feel free to merge when you can :) thanks!

PR #1327, opened 2022-10-27
Removal of Data from codegen and binary deserialisation

gjcairo:
_Note: only the last three commits are actually related to this change. Previous commits are actually part of https://github.com/apple/swift-protobuf/pull/1326, on which this PR depends._

## Motivation

This is a follow-up PR for the removal of `Data` (and eventually `Foundation`) from `swift-protobuf`. This is related to https://github.com/apple/swift-protobuf/issues/816 and follows the changes made in https://github.com/apple/swift-protobuf/pull/1326

## Modifications

This PR includes the bulk of the required changes for the removal of `Data` from `swift-protobuf`:
* Changes to the codegen, to move away from using `Data` and use `[UInt8]` instead for `byte`-type fields.
* Regenerated protos based on the codegen changes.
* Changes to remove `Data` completely from the binary deserialisation path.
* Other required changes to make things compile with the new generated code.

Other PRs will follow, removing `Data` usages from the binary serialisation path, the JSON (de)serialisation, and any other miscellaneous remaining usages.

## Result

The codegen and binary deserialisation paths are now `Data`-free.


thomasvl (commented):
I'd suggest splitting this in two - do the binary serialization support in one PR, and then do a final PR for the how the `bytes` fields are represented.  That should make the impact of each change easier to follow/digest.

As commented on the base CL, I have some concerns about how folks will be able to accept this when we cut a 2.0 release.  For small code bases, they can update all at once, but for large code bases with dependencies that depend on protobuf, they will never *every* dependency to update, so even though 2.0 will be an api breaking release, I feel we need to consider how to provide a migration path for those larger projects (deprecated apis still in the library, generation options to support old and new code paths, etc.).

I'll also add, that I fully back getting `Data` out of the apis for all the different serializations, but the actually `bytes` fields is what does give me minor pause.  I can see the potential value in now wanting Foundation on linux, but for the iOS/tvOS/matchOS/macOS consumers of this project, I do worry a little that we might actually be making an API that is harder for them to use by not providing the fields as `Data`.  Have you all given that any thought and what the potential impact may be for those clients?

tbkka:
> .... I feel we need to consider how to provide a migration path for those larger projects (deprecated apis still in the library, generation options to support old and new code paths, etc.).
> 
> I'll also add, that I fully back getting `Data` out of the apis for all the different serializations, but the actually `bytes` fields is what does give me minor pause.

I'm a little reluctant to suggest this, but ....  one robust answer here might be to only go part-way for 2.0 and leave the remaining transition for a future 3.0.   So 2.0 would get data out of the serialization and change the bytes fields to provide _both_ Data- and non-Data-based APIs.  Then a future 3.0 could remove the Data-based bytes field support.  That would give people two relatively easy transitions rather than one hard one.  But of course that also means that we wouldn't completely shed the Foundation/Data dependency for at least a full year (I would want at least a full year between major versions).

Though to be honest, I'm not entirely sure this approach is really feasible -- I've not experimented to see what the redundant accessors would look like.

thomasvl:
> > .... I feel we need to consider how to provide a migration path for those larger projects (deprecated apis still in the library, generation options to support old and new code paths, etc.).
> > I'll also add, that I fully back getting `Data` out of the apis for all the different serializations, but the actually `bytes` fields is what does give me minor pause.
> 
> I'm a little reluctant to suggest this, but .... one robust answer here might be to only go part-way for 2.0 and leave the remaining transition for a future 3.0. So 2.0 would get data out of the serialization and change the bytes fields to provide _both_ Data- and non-Data-based APIs. Then a future 3.0 could remove the Data-based bytes field support. That would give people two relatively easy transitions rather than one hard one. But of course that also means that we wouldn't completely shed the Foundation/Data dependency for at least a full year (I would want at least a full year between major versions).
> 
> Though to be honest, I'm not entirely sure this approach is really feasible -- I've not experimented to see what the redundant accessors would look like.

There is potentially some support coming in protobuf next year to allow migration for things like this on a per file (and even per field) basis.  I don't think the team has completely nailed down the details, but it may be something to help with it at some point.

I thought (which likely is wrong), one of the main issues/desires was to reduce the copies when things like grpc are receiving data and then getting a message parsed?  Can we realize some wins in that space with just changes to the serialization apis, and save the actual field migration until a 3.0 time frame?

FranzBusch:
While I understand the concern that migrations should be as smooth as possible and agree that we should try everything that is possible to support that. I don't think we should split this into two major releases. `SwiftProtobuf` is one of the lowest dependencies that other projects have and if that brings in `Foundation` they have no way out. If we do tackle the `Foundation` problem we should remove it holistically in a single major release while providing good migration documentation.

> I thought (which likely is wrong), one of the main issues/desires was to reduce the copies when things like grpc are receiving data and then getting a message parsed? Can we realize some wins in that space with just changes to the serialization apis, and save the actual field migration until a 3.0 time frame?

@gjcairo Is going to provide a larger response on the issue that lays the why's and how's out a bit more detailed. Besides the desire to avoid unnecessary copies when using other bag of bytes types like `ByteBuffer`, the other big goal is to shed the `Foundation` dependency so that sever applications that want to use `SwiftProtobuf` can ship without `Foundation`. `Foundation` increases the binary size drastically and it links a bunch of other things.

gjcairo:
Closing in favour of https://github.com/apple/swift-protobuf/pull/1391

PR #1329, opened 2022-10-27
Ensure BinaryEncoder/BinaryEncoderVistor changes stay lightweight

thomasvl:
Currently, `BinaryEncoder` only has one field, and it is `UnsafeMutableRawPointer`, so the cost of copying the `encoder` in `BinaryEncoderVistor` should be cheap; but to ensure that stats the case, remove the initializer that would do a full copy, and directly expose/update the `pointer` for this one case. Hopefully this will ensure this case says lightweight if the types are revised.

Follow from comments on #1323


Lukasa (approved):
I mean, I'm pretty confident this change is a complete no-op.

PR #1333, opened 2022-11-01
The conformance test isn't needed for the unittests.

thomasvl:
The conformance.proto is just the communication between the conformance runner and the given language binary, to reuse the tests, the messages are all thats needed, and they aren't in conformance.proto.


thomasvl:
Long term we should probably got the tests off the upstream protos files, just so future changes to them don't break us. I did that even upstream for objc (all the test protos are in the objc subdir) and it made this much simpler over the long run.

tbkka (approved):

PR #1337, opened 2022-11-04
Make it easier to compile the SwiftProtobufCore with a different module name.

thomasvl:
This allows one to more simply cope the module with a different name (not `-module-alias` needed) and then it updates the tests so you can run those against against a runtime that was compiled under a different name (`-module-alias` used simple to compile the tests, no edit need to make them all pass).

This is sorta a followup to #1017.


thomasvl:
Build failures are protobuf (their tip just broke)

tbkka:
Oooh.  Nice idea!  Does the Makefile need updating?

tbkka (approved):

thomasvl:
> Oooh. Nice idea! Does the Makefile need updating?

Nope, be default we use the file being generated to figure out which module we're generating for, so it "just works".

Issue #1310, opened 2022-10-08
SwiftPM plugin: solve "filename X.pb.swift used twice"


klundberg:
I'm trying to integrate the swift pm plugin into our setup. We use submodules to easily keep the protobufs up to date with the other teams that use them. When I add the paths to the protobufs in the plugin config and build, I get a "Filename x.pb.swift used twice" error, because some of the proto files in the submodule have the same filenames, even though they're in different directory structures.

Could the plugin either by default or by a config option, use the `--swift_opt=FileNaming=PathToUnderscores`  option when generating code? That would resolve this issue for our use case. I'd prefer to not manually copy/rename files if I can help it as that'll make things more complex than we're willing to allow.

* MacOS 12.6
* Xcode 14.0.1
* Swift 5.7
* Swift Protobuf 1.20.2

FranzBusch:
Just opened https://github.com/apple/swift-protobuf/pull/1312 to provide this option

swarupdonepudi:
thanks a lot for fixing this. is it possible to release a new version to get this fix?

PR #1338, opened 2022-11-07
1_x_branch: Add the file naming option (#1312)

thomasvl:
* Add the file naming option

# Motivation
In some cases, it might be that the proto files in a single target have duplicate file names. This would result in a compilation failure since Swift does not allow duplicate files names in a single module. To resolve this `protoc-gen-swift` has a `FileNaming` option called `PathToUnderscore` to avoid this. We should expose this as an option in the SPM plugin.

# Modification
This PR extends the SPM plugin config to set a `fileNaming` strategy. It defaults to the `FullPath` strategy if none is set.

# Result
Fixes: https://github.com/apple/swift-protobuf/issues/1310


thomasvl:
fyi @FranzBusch

FranzBusch (approved):

PR #1341, opened 2022-11-09
Update CocoaPods support.

### LINKED ISSUE #1334: Get CocoaPods working again
Post #1331, `pod lib lint SwiftProtobuf.podspec` fails because the sources have changed layout.

When we add `products` to the Package.swift, we'll want _subspecs_, but for the time being what we conceptually want is probably:

```
diff --git a/SwiftProtobuf.podspec b/SwiftProtobuf.podspec
index bad55155..ebce3606 100644
--- a/SwiftProtobuf.podspec
+++ b/SwiftProtobuf.podspec
@@ -15,7 +15,7 @@ Pod::Spec.new do |s|
 
-  s.source_files = 'Sources/SwiftProtobuf/**/*.swift'
+  s.source_files = 'Sources/SwiftProtobufCore/*.swift', 'Sources/SwiftProtobufFoundationCompat/*.swift'
 
   s.swift_versions = ['5.0']
 end
```

However that will still fail because the code expects to be compiled into two modules, which isn't what CocoaPods does even if it were `subspecs`

fyi @gjcairo @FranzBusch

thomasvl:
Model what Firebase seems to do, have a podspec for each of the modules and have dependencies between them.

Add a helper to the Makefile for running `pod lib lint`.

Update LibraryVersions.py to handle the multiple podspecs.

Update the docs on what publishing is likely going to take.

Fixes #1334

### DIFF
--- a/DevTools/LibraryVersions.py
+++ b/DevTools/LibraryVersions.py
@@ -18,6 +18,8 @@ import sys
 _VERSION_RE = re.compile(r'^(?P<major>\d+)\.(?P<minor>\d+)(.(?P<revision>\d+))?$')
 
 _PROJECT_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+_CORE_PODSPEC_PATH = os.path.join(_PROJECT_ROOT, 'SwiftProtobufCore.podspec')
+_FOUNDATION_PODSPEC_PATH = os.path.join(_PROJECT_ROOT, 'SwiftProtobufFoundationCompat.podspec')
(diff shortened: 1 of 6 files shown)


thomasvl:
I'm not 100% sure this will work, but looking at [Firebase](https://github.com/firebase/firebase-ios-sdk), they seems to publish a few podspecs all pointing at the same repo and have some dependencies between them, so it seems like it might.  It does pass `pod lib lint` testing, so… 🤞

tbkka (approved):
Nifty!

FranzBusch (approved):
Nice!

PR #1345, opened 2022-11-15
Use JSONDecodingOptions.ignoreUnknownFields to ignore unknown enums.

### LINKED ISSUE #972: Issue parsing unknown enum case
iOS 13.4
Xcode Version 11.4 (11E146)
Swift 5
Swift-Protobuf 1.8.0
proto2

My iOS app is using generated code from outdated proto files. Since I've compiled the `.swift` files from the old protos, we've added some new cases to one of our enums, and started returning the cases to the client from the backend. I'm getting an error when parsing the JSON data into the Swift Proto objects: `unrecognizedEnumValue`.

From my understanding of the proto spec, it should be fine to add new enum cases, and clients who don't have the latest spec map the unknown value to the 0 case. Is there something wrong with how we name our 0th case?

Old Proto
```
message Product {
    enum State {
        State_INVALID = 0;
        ACTIVE = 1;
        DISCONTINUED = 2;
    }
    optional State state = 1;
}
```

New Proto
```
message Product {
    enum State {
        State_INVALID = 0;
        ACTIVE = 1;
        DISCONTINUED = 2;
        FOR_SALE = 3;
        NOT_FOR_SALE = 4;
    }
    optional State state = 1;
}
```

JSON Being Parsed
```
{
    "state": "FOR_SALE"
}
```

Swift JSON Decoding (using Old Proto)
```
do {
    var options = JSONDecodingOptions()
    options.ignoreUnknownFields = true
    return try Product(jsonUTF8Data: data, options: options)
} catch {
    print(error) // error is printed unrecognizedEnumValue
    return nil
}
```

thomasvl:
A new conformance test came in protocolbuffers/protobuf@09f4901 that says ignoreUnknownFields should also result in unknown enums being ignored.

These changes are based on #896.

Fixes #972.


Tests/SwiftProtobufTests/Test_Enum.swift
```diff
@@ -40,6 +41,24 @@ class Test_Enum: XCTestCase, PBTestHelpers {
...
+            }
+        }
+
+        assertJSONDecodeSucceeds("{\"repeatedNestedEnum\":[\"FOO\",\"NEW_VALUE\"]}", options: options) { (m: MessageTestType) -> Bool in
+            m.repeatedNestedEnum == [.foo]
+        }
+
+        // Proto3Unittest_TestAllTypes doesn't have a map<>, Test_Map_JSON covers this case.
+    }
+
```
tbkka, line 61: Do we need to add a test case to verify that we reject unknown enums without this option?

Tests/SwiftProtobufTests/Test_Map_JSON.swift
```diff
@@ -285,6 +285,17 @@ class Test_Map_JSON: XCTestCase, PBTestHelpers {
...
+        var options = JSONDecodingOptions()
+        options.ignoreUnknownFields = true
+
+        // Test_Enum covers the single and repeated field cases.
+
+        assertJSONDecodeSucceeds("{\"mapInt32Enum\": {\"1\": \"NEW_VALUE\"}}", options: options) { (m: MessageTestType) -> Bool in
+          m.mapInt32Enum == [:]
+        }
+    }
+
```
tbkka, line 298: As above, doe need a test case to verify we reject maps containing unknown string enums without this option?

Sources/SwiftProtobufCore/JSONDecoder.swift
```diff
@@ -511,8 +517,9 @@ internal struct JSONDecoder: Decoder {
           throw JSONDecodingError.illegalNull
         }
       } else {
-        let e: E = try scanner.nextEnumValue()
-        value.append(e)
+        if let e: E = try scanner.nextEnumValue() {
+          value.append(e)
+        }
```
tbkka, line 522: This needs some tests to make sure that, e.g.,
```
{ "field": ["enum1", "enum2", "unknown", "enum3"]}
```
gets either rejected (without the option) or decoded as
```
{ "field": ["enum1", "enum2", "enum3"]}
```

tbkka (approved):

tbkka:
Nice work!   I think we might be missing a couple of test cases, but the core logic looks sound.

Tests/SwiftProtobufTests/Test_Enum.swift
thomasvl: Was hoping we had that, but added locally to make it explicit.

Tests/SwiftProtobufTests/Test_Map_JSON.swift
thomasvl: same.

tbkka:
Thank you for adding those.  Looks all ready to go!

PR #1353, opened 2023-01-06
1.x branch: Fail to encode google.protobuf.Value number if not finite.

thomasvl:
From https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Value

> Note that attempting to serialize NaN or Infinity results in error. (We can't
> serialize these as string "NaN" or "Infinity" values like we do for regular
> fields, because they would parse as string_value, not number_value).

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Value+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Value+Extensions.swift
@@ -152,7 +152,11 @@ extension Google_Protobuf_Value {
   ) throws {
     switch kind {
     case .nullValue?: encoder.putNullValue()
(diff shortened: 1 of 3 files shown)


thomasvl:
The new error case is technically a breaking changes if someone had a `switch` on all the cases, but last time we added one, I think we still did a minor version bump and were ok with it.

thomasvl:
The 4.2.4 failure is because protobuf now needs a newer compiler than we have on the image, so not related to this change.

thomasvl:
@tbkka you ok with this change going to the branch so it's ready for a future 1.x release (if we do one)?

tbkka (approved):
Yep.  Small chance of breaking someone with the new error enum case, but I think it's worth it.

PR #1354, opened 2023-01-06
1.x branch: Bump the Swift 5.7.0 testing to 5.7.2.

thomasvl:


thomasvl:
The 4.2.4 failure is because of the protobuf compiler requirements, so unrelated.

tbkka:
This doesn't make any sense at all.  Protobuf build is claiming:
```
-- The C compiler identification is GNU 5.4.0
```
but the Ubuntu 22.04 image we're using should be this one: https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20221212.1
which only lists GCC 9, 10, 11, and 12 as supported compilers according to sbom.json.

GCC 5.4.0 was released in 2016; I can't believe that's really the compiler in this image.

tbkka:
Hmmm...   In the 4.2.4 test run, "Update and Checkout dependencies" shows
```
g++ is already the newest version (4:5.3.1-1ubuntu1).
g++ set to manually installed.
```

thomasvl:
22.04 shouldn't be coming into play, the swift image is the controlling factor: https://github.com/apple/swift-protobuf/blob/1_x_release_branch/.github/workflows/build.yml#L70

tbkka:
Ah.  Then I think I'm inclined to drop the 4.2.4 CI job entirely.

PR #1356, opened 2023-01-16
Only warn about `protoc` version if explicitly set to <3.2.x

rebello95:
There are some cases where it does not make sense to specify the `protoc` version - for example, when compiling with `buf`. (See https://github.com/bufbuild/buf/issues/1244 and https://github.com/bufbuild/buf/issues/1736). Furthermore, the compiler version is not read elsewhere right now.

This change updates the behavior to only warn _if the compiler version is explicitly set and is < 3.2.x_.

### DIFF
--- a/Sources/protoc-gen-swift/main.swift
+++ b/Sources/protoc-gen-swift/main.swift
@@ -153,7 +153,6 @@ struct GeneratorPlugin {
       return 1
     }
 
-    auditProtoCVersion(request: request)
     let response = generate(request: request)
     guard sendReply(response: response) else { return 1 }
     return 0
@@ -237,16 +236,6 @@ struct GeneratorPlugin {
     }
     return Google_Protobuf_Compiler_CodeGeneratorResponse(files: responseFiles,
                                                           supportedFeatures: [.proto3Optional])
-  }
-
-  private func auditProtoCVersion(request: Google_Protobuf_Compiler_CodeGeneratorRequest) {
-    guard request.hasCompilerVersion else {
-      Stderr.print("WARNING: unknown version of protoc, use 3.2.x or later to ensure JSON support is correct.")
-      return
-    }
-    // 3.2.x is what added the compiler_version, so there is no need to
-    // ensure that the version of protoc being used is newer, if the field
-    // is there, the JSON support should be good.
   }
 
   private func sendReply(response: Google_Protobuf_Compiler_CodeGeneratorResponse) -> Bool {
--- a/Sources/protoc-gen-swift/main.swift
+++ b/Sources/protoc-gen-swift/main.swift
@@ -153,6 +153,7 @@ struct GeneratorPlugin {
       return 1
     }
 
+    auditProtoCVersion(request: request)
     let response = generate(request: request)
     guard sendReply(response: response) else { return 1 }
     return 0
@@ -236,6 +237,17 @@ struct GeneratorPlugin {
     }
     return Google_Protobuf_Compiler_CodeGeneratorResponse(files: responseFiles,
                                                           supportedFeatures: [.proto3Optional])
+  }
+
+  private func auditProtoCVersion(request: Google_Protobuf_Compiler_CodeGeneratorRequest) {
+    guard request.hasCompilerVersion else {
+      // Don't warn if the compiler version is not explicitly set.
+      return
+    }
+
+    if request.compilerVersion.major < 3 || (request.compilerVersion.major == 3 && request.compilerVersion.minor < 2) {
+      Stderr.print("WARNING: unknown version of protoc, use 3.2.x or later to ensure JSON support is correct.")
+    }
   }
 
   private func sendReply(response: Google_Protobuf_Compiler_CodeGeneratorResponse) -> Bool {


thomasvl:
So the version is used, to provide the warning that the descriptor set being generated from may not have the correct support to get properly working JSON from the Swift library.

Skimming the links you provided, I think maybe the purpose of the version number might be misunderstood.

The interface between protoc and the plugins is serialized descriptors. The protobuf format provides some degree of api safety because descriptors do (old versions and parse newer version, etc.). Since protoc to a plugin is a cross process api boundary, the version provides the plugin additional knowledge about the calling process, a way to know what other things might have been done to the data being received.

In the case of this warning, the conversion from proto field names to the json strings wasn't formally documented, and for a while protoc was always providing the strings in the descriptors, so the warning is reporting when a developer is likely using some old version so we can't guarantee that what the swift plugin will generate will interoperate correctly since we rely on protoc making those strings.

If that tool really wants to replace the role protoc normally provides, it sorta has to ensure it is providing *all* of the validations/normalization/etc. that protoc does, and if it is stepping up to that, then it seems like it should be setting the protoc version to what it is guaranteeing to match to maintain the plugin api contract.  Future changes to protobuf could well evolve descriptor.proto and introduce new ways plugins might need to use the version to realize when protoc did/didn't do something a plugin relies one.

So if Apple wants to drop this warning, that's fine, but as I mention, that version is providing plugins information about the contract, and so it might be needed in the future to know about other differences.

One footnote here as to why Swift might seem a little different - the Swift plugin is written in Swift, but a lot of other plugins get written in C++. The ones written in C++ get a little extra safety net because they get the official protobuf C++ code that is always used within protoc to re-validate/normalize the descriptor sets as they read them in. Since none of that is officially documented, no plugin not written with in C++ really has a way to ensure they are doing the same things and instead have to rely on the thing build the binary descriptor sets to have done it, and that version number is the only really want to check. Without checking it, developers would be left on their own to try to figure out when something generated doesn't work because the normal checks relied on didn't happen.

tbkka:
We should not remove this check.  I commented on the `buf` thread about this issue to point out that by not setting this field, `buf` is explicitly choosing to proclaim compatibility with pre-3.2 protoc.  And in that case, the plugin has no choice but to assume that the JSON data is incomplete.

bufdev:
We've replied over there too - we'd respectfully disagree that choosing not to set this field is declaring pre-3.2 protoc compatibility, instead it is not making any such claim at all. We'd argue a more faithful implementation of the`CodeGeneratorRequest` API for SwiftProtobuf to make would be "if unset, do not warn, if set and pre-3.2, warn, else do not warn", however reasonable people can disagree here.

rebello95:
I updated the PR with the logic suggested by @bufdev, as I agree that it's probably a more faithful representation of the intent

Sources/protoc-gen-swift/main.swift
```diff
@@ -241,12 +241,13 @@ struct GeneratorPlugin {
 
   private func auditProtoCVersion(request: Google_Protobuf_Compiler_CodeGeneratorRequest) {
     guard request.hasCompilerVersion else {
-      Stderr.print("WARNING: unknown version of protoc, use 3.2.x or later to ensure JSON support is correct.")
+      // Don't warn if the compiler version is not explicitly set.
       return
     }
-    // 3.2.x is what added the compiler_version, so there is no need to
```
thomasvl, line 247: This comment explains exactly why it has to be an unset test vs and explicit version check. Old versions of protoc didn't set the field because it didn't exist and there was no way to detect/report these sorta problems.

tbkka:
If the `buf` developers want to provide some other way to recognize the `buf` version, then I'd consider recognizing that.  Of course, I'd be even happier if you negotiated with the protoc folks to make sure that there was a single uniform way for plugins to recognize the front-end being used.  I think that information would be very useful to include in the generated output sources as well.

bufdev:
Funny and somewhat-unrelated story, one of the original reasons we stopped producing this value was because plugin authors added the version number in their generated sources :-) It ended up causing many diffs with no functional change for many of our users, and we came to the opinion that it was an anti-pattern that was causing developer misery via unnecessary diffs.

Anyways, I think the fundamental disagreement here is that, in our view, plugins should not alter their behavior based on what version of `protoc` or `buf` produces a `CodeGeneratorRequest` for the plugin, and that the behavior of the plugin should remain the same, regardless of if `protoc`, `buf`, some other compiler, or even a handwritten `CodeGeneratorRequest` serialized to a file and piped to the plugin via stdin, is the source of the `CodeGeneratorRequest`. The `compiler_version` field (in our view) should be treated as entirely optional metadata. In that vein, it seems very reasonable to warn if you explicitly detect that the `CodeGeneratorRequest` was produced by a compiler version the plugin author knows will result in certain undesirable behaviors - for example, if the plugin knows that certain (by definition optional) fields on the `FileDescriptorSet` will not be set for `protoc` <3.2, it seems reasonable to warn about this. In the absence of this optional metadata, however, the warning likely should not be printed.

Even better would be for SwiftProtobuf to handle the case where <3.2 versions of protoc do not produce `FileDescriptorSets` that contain whatever they need for proper JSON handling, but this is beyond the scope of my understanding of this specific plugin - I know Protobuf, but I definitely don't know Swift or this plugin implementation, and I'm sure you thought of this heh.

tbkka:
Closing.

PR #1359, opened 2023-01-17
Change the policy for supporting old Swift language versions

tbkka:
"One full major version" used to be approximately the same as "one year."  But Swift 5 changed that and "one full major version" is becoming increasingly intractable.  This is especially problematic for CI; we've based our CI on the Swift-provided Docker images, but the Swift 5.0 Docker images are no longer able to build protobuf for testing, so cannot run our full test suite.

### DIFF
--- a/Documentation/INTERNALS.md
+++ b/Documentation/INTERNALS.md
@@ -17,10 +17,20 @@ behavior are always appreciated.
 
 ## Swift Language Support
 
-The goal is to always support "one full major version”, which basically
-means if the current official release of Swift is `X.Y`, the library will
-support back to `X-1.Y`.  That is, when Swift 4.2 was released, the minimum
-for support got moved up to 3.2.
+We guarantee support for Swift versions that are up to one year old.
+For example, in March 2022 (when Swift 5.6 was released), we still
+guaranteed support for Swift 5.3.3 (released in January 2021, and still
+current in March 2021).
+Support for Swift versions older than one year is at the discretion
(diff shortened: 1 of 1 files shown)


thomasvl (approved):
Good to pull this over to the 1.x branch?

[I'll update CI accordingly]

tbkka:
@Lukasa - Any concerns about this proposed change?

FranzBusch:
This sounds very similar to what we are doing in NIO. We are supporting the last two minor Swift releases. Right now this ends up being almost the same supported version as the "up to one year old" policy here.

Lukasa:
Agreed, this aligns with our policy. I think we technically do a maximum of 18 months (we currently support 5.7, 5.6, and 5.5, where if we assume a six-month-ish cadence of Swift releases 5.5 will end up 18 months old when we eventually drop its support), but that's not a huge delta.

thomasvl:
@tbkka good to commit?

thomasvl:
> Good to pull this over to the 1.x branch?

Just realized we can't pull this to the 1.x branch, as a change to min OS policy would be a breaking change, and thus 2.0, which is trunk. So this will only clear the way for future things, no impact on 1.x.

tbkka:
I don't think it makes much difference either way for 1.x.   That branch is not changing quickly, and I see no need to forcibly remove support for older Swift language versions there.

I suppose the simplest would be to _not_ pull it over and just leave 1.x alone as much as possible.

PR #1360, opened 2023-01-17
Update around where/when protobuf is built.

thomasvl:
- Update build.yml to use the currently release version of protoc and remove the conformance tests.
- Add a new conformance.yml that has all the protobuf building and conformance test running (also run the plugin tests just incase something changes on the plugin interface). This file is much more in sync with regular_conformance.yml.


thomasvl:
And `apt-get protocol-buffer` gets a older than desired version, so that looks like a no go.

PR #1361, opened 2023-01-17
Update CI to match the version policy changes.

thomasvl:
Policy changes from #1359


thomasvl:
Looking at https://www.swift.org/download/, 5.4.3 was Sept 2021, so that and everything before it would be past the 1 year point. 5.2.5 was Aug 2020, so if we want to keep more like two years, that would be the cut off.

thomasvl:
Sorry, 5.3.3 would be as far as we'd have to go back if we wanted two year (assuming we always want the latest point release for any minor version).

thomasvl:
(and I'll add those back if they should be kept until there is an issue, 5.3.3 does mean we need to keep the generate-main support because of SwiftPM differences)

thomasvl:
As a random datapoint, Swift Package Index seems to only keep track back to 5.4 at the moment: https://swiftpackageindex.com/apple/swift-protobuf/builds

thomasvl:
@tbkka - Current PR keep back to 5.4.3 (September 9, 2021), sound good?

(5.5.3 would move up to February 9, 2022, and 5.3.3 would be January 28, 2021)

thomasvl:
If we did move all the way to 5.5.x, then all the conditionals in the source for Concurrency would also be removable.

tbkka (approved):
This looks good.

PR #1365, opened 2023-01-18
Use Ubuntu Jammy for a few places.

thomasvl:
Generally tweak how we express swift versions again to hopefully simplify what has to be expressed in the file.

Document some of the choices for future updates and where to find data.


tbkka (approved):
Nice bit of cleanup!

PR #1320, opened 2022-10-13
fix(ci): replace set-output to GITHUB_OUTPUT

k1rnt:
resolved #1319

PR #1367, opened 2023-01-19
replace set-output to use GITHUB_OUTPUT

thomasvl:
This is #1320 but for the 1.x branch


thomasvl:
You can see the warnings here: https://github.com/apple/swift-protobuf/actions/runs/3957089680

allevato (approved):

PR #1373, opened 2023-02-13
Add import paths to SPM Plugin

samisuteria:
Added an optional parameter to the config file so users can configure import paths for proto files. Useful in a mono repo where the proto is not relative to the target directory and requires imports that are defined local to the proto.

### DIFF
--- a/Plugins/SwiftProtobufPlugin/plugin.swift
+++ b/Plugins/SwiftProtobufPlugin/plugin.swift
@@ -64,6 +64,13 @@ struct SwiftProtobufPlugin: BuildToolPlugin {
             var fileNaming: FileNaming?
         }
 
+        /// Specify the directory in which to search for
+        /// imports.  May be specified multiple times;
+        /// directories will be searched in order.  If not
+        /// given, the current working directory is used.
+        /// Passed via `protoc -I <path>`
+        var importPaths: [String]?
+
         /// The path to the `protoc` binary.
         ///
         /// If this is not set, SPM will try to find the tool itself.
@@ -88,6 +95,14 @@ struct SwiftProtobufPlugin: BuildToolPlugin {
 
         try validateConfiguration(configuration)
 
+        let importPaths: [Path]
+        if let configuredImportPaths = configuration.importPaths {
+            importPaths = configuredImportPaths.map { Path($0) }
+        } else {
+            // We include the target directory as the default
+            importPaths = [target.directory]
+        }
+
         // We need to find the path of protoc and protoc-gen-swift
         let protocPath: Path
         if let configuredProtocPath = configuration.protocPath {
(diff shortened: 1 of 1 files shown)


thomasvl:
@FranzBusch

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -64,6 +64,13 @@ struct SwiftProtobufPlugin: BuildToolPlugin {
             var fileNaming: FileNaming?
         }
 
+        /// Specify the directory in which to search for
+        /// imports.  May be specified multiple times;
+        /// directories will be searched in order.  If not
+        /// given, the current working directory is used.
+        /// Passed via `protoc -I <path>`
```
FranzBusch, line 71:
```suggestion
        /// directories will be searched in order.
        /// The target source directory is always appended
        /// to the import paths.
```

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -88,6 +95,14 @@ struct SwiftProtobufPlugin: BuildToolPlugin {
 
         try validateConfiguration(configuration)
 
+        let importPaths: [Path]
+        if let configuredImportPaths = configuration.importPaths {
+            importPaths = configuredImportPaths.map { Path($0) }
+        } else {
+            // We include the target directory as the default
+            importPaths = [target.directory]
```
FranzBusch, line 103: I think we probably should always include the `target.directory` even when `importPaths` are specified.

samisuteria: sounds fair to me

FranzBusch (approved):
LGTM! Thanks

samisuteria:
@FranzBusch by the way theres a matching PR in https://github.com/grpc/grpc-swift/pull/1568 for the same thing and I made the same change there.

FranzBusch:
@samisuteria I am just coming back to this since I wanted to try this out myself and actually noticed that this behaves differently than I assumed during review. (I will get some tests in place here in another PR).
How are you using this property in your config? Because as far as I can see it only takes absolute paths which makes it very hard to use in general.

samisuteria:
@FranzBusch I actually could not get it to work the way I expected it to. I wanted to reference a proto file that was outside the Sources directory and when running `protoc` manually with the flags it generated code correctly but I have not been able to get it to work with the plugin. I've resorted to soft linking the shared proto folder into each target that needs it.

samisuteria:
@FranzBusch We have a monorepo with multiple iOS/macOS apps and Swift packages so copying files to the target source would mean copying them to many target sources and updating them would be a manual process. We use a git submodule for all proto files and soft link that folder into the targets that need it. 

Correct, I am not using the new option I added. I have not had the time to come back and fix it to the behavior I was expecting.

PR #1380, opened 2023-02-21
Add basic binary delimited fuzz testing.

thomasvl:


tbkka (approved):

thomasvl:
btw - I'm torn on this one, it almost seems like it would be better to have a support for reading the options from the input stream and use an option do delimited instead, since aside from those initial bytes, this is still just a binary parse.

thomasvl:
Eh, I guess having it as a distinct test helps ensure the test harness works.

Issue #1377, opened 2023-02-14
Binary decoding likely has 32bit issues with large data sizes


thomasvl:
Like what #1375 fixed, it looks like we've got a few cases (`getFieldBodyBytes(count:)`, `consume(length:)`, ...) where because the internals of the BinaryDecoder use `Int` we could have failures in a 32bit build.

thomasvl:
fyi - The C++ upstream code seem to use `INT_MAX` in some cased to catch limits, but it seems like that value will change depending on a build being for a 64 or 32 bit platform.  And the new https://protobuf.dev/programming-guides/encoding/#cheat-sheet section calls out some _2GB_ limits on specific things.

thomasvl:
Correction on that comment - `int` (and `INT_MAX`) is still 32bit on most 64bit systems: https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models

tbkka:
If there's a 2GB limit, we should still write it out as `0xffffffff` or similar.

thomasvl:
I'm going to go ahead and close this, I doing another pass, and don't immediately see issues, so at this point I'd suggest we treat them as bugs if found instead of keeping this open longer term without any clear next steps.

PR #1381, opened 2023-02-21
Enforce the 2GB size limits

thomasvl:
This probably takes care of the uses listed in #1377, but it might not hurt to consider updating that to full deal with the mixed use of `Int` for things.

### DIFF
--- a/Sources/SwiftProtobufCore/BinaryDelimited.swift
+++ b/Sources/SwiftProtobufCore/BinaryDelimited.swift
@@ -29,9 +29,9 @@ public enum BinaryDelimited {
     /// read/written.
     case truncated
 
-    /// When decoding (parsing), the length is larger than what will fit in
-    /// an Int for the compiled platform, so the message can't be parsed.
-    case decodeTooLarge
+    /// Messages are limited by the protobuf spec to 2GB; when decoding, if the
+    /// length says the payload is over 2GB, this error is raised.
+    case tooLarge
   }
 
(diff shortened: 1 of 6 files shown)


Sources/SwiftProtobufCore/BinaryDelimited.swift
```diff
@@ -29,8 +29,8 @@ public enum BinaryDelimited {
     /// read/written.
     case truncated
 
-    /// When decoding (parsing), the length is larger than what will fit in
-    /// an Int for the compiled platform, so the message can't be parsed.
+    /// Messages are limited by the protobuf spec to 2GB, if the length says
+    /// the payload is over 2GB, this error is raised.
     case decodeTooLarge
```
allevato, line 34: Should we rename this so that all the error constants are uniformly `tooLarge`?

allevato (approved):

tbkka (approved):

Sources/SwiftProtobufCore/BinaryDelimited.swift
thomasvl: Delimited has on error type, not a split for decode and encode. So that's why I originally stuck decode in there.  Should we split it also?

thomasvl: Actually, we could just use the decoding error one; it didn't exist when this was added the other week.

PR #1387, opened 2023-03-27
Add tests for SPM plugin

FranzBusch:
# Motivation
We have seen a few issues with the SPM plugin over time that we could have caught with tests.

# Modification
This PR adds a new tested `PluginExamples` folder that contains a standalone package that exercises the plugin in various configuration. This does not only help in testing the various configurations but also is a good resource for people that want to adopt the plugin.

# Result
We now have tests in place making sure our plugin works


PluginExamples/Sources/Simple/empty.swift
```diff
@@ -0,0 +1 @@
+
```
glbrntt, line 1: Can we add a comment to all the empty Swift files saying why we need them to avoid them being deleted by mistake.

PluginExamples/Sources/Import/Foo/ImportFoo.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+
+import "Bar/ImportBar.proto";
+
+message ImportFoo {
```
glbrntt, line 5: nit: remove `Import` it makes it sound like the type is importing `Foo` when it's actually importing a Bar

glbrntt (approved):

Makefile
```diff
@@ -274,6 +274,10 @@ test-plugin: build ${PROTOC_GEN_SWIFT}
 	done
 	diff -ru _test Reference
 
+# Test the SPM plugin.
+test-spm-plugin:
```
thomasvl, line 278: I guess we can't add this to `check` since it would then run on the swift versions that don't support plugins.  Guess we just rely on CI to catch things and folks to remember to manually screen things.

thomasvl (approved):

Makefile
FranzBusch: Yeah that was the idea. In a best world we would actually write these as normal tests. However, we need to solve the protoc problem in the ecosystem first.

PR #1388, opened 2023-03-28
Allow target relative path to protoc in SPM Plugin

justintyre:
### Problem
SPM currently requires an absolute path to be passed to the build command.  This is not ideal for situations where `protoc` is in a directory of the repository - e.g. `../tools/protoc`.

### Solution
This adds a check to the user provided `protocPath`.  When the provided path is not absolute, it assumes the provided path is relative to the target directory and adds the target directory as the path's prefix.

### DIFF
--- a/Plugins/SwiftProtobufPlugin/plugin.swift
+++ b/Plugins/SwiftProtobufPlugin/plugin.swift
@@ -104,7 +104,13 @@ struct SwiftProtobufPlugin: BuildToolPlugin {
         let protocPath: Path
         if let configuredProtocPath = configuration.protocPath {
             // The user set the config path in the file. So let's take that
-            protocPath = Path(configuredProtocPath)
+            if configuredProtocPath.hasPrefix("/") {
+                // An absolute path has been provided
+                protocPath = Path(configuredProtocPath)
+            } else {
+                // Not an absolute path, assume it is relative to the target's directory
+                protocPath = target.directory.appending(subpath: configuredProtocPath)
+            }
         } else if let environmentPath = ProcessInfo.processInfo.environment["PROTOC_PATH"] {
             // The user set the env variable. So let's take that
             protocPath = Path(environmentPath)


Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -104,7 +104,13 @@ struct SwiftProtobufPlugin: BuildToolPlugin {
...
+            if configuredProtocPath.hasPrefix("/") {
+                // An absolute path has been provided
+                protocPath = Path(configuredProtocPath)
+            } else {
+                // Not an absolute path, assume it is relative to the target's directory
+                protocPath = target.directory.appending(subpath: configuredProtocPath)
+            }
         } else if let environmentPath = ProcessInfo.processInfo.environment["PROTOC_PATH"] {
             // The user set the env variable. So let's take that
             protocPath = Path(environmentPath)
```
thomasvl, line 116: What about this case, should this assume workspace relative or something?

thomasvl:
@FranzBusch fyi

FranzBusch:
@justintyre Thanks for opening this. I am not quite sure this is the right solution to implement because we are getting into path manipulation territory here and that is deeply rooted to the individual platforms. I don't think the statement holds that `/` is the only indication of an absolute path.

But taking a step back. I agree that the current `protoc` story is not great; however, we are working on an approach that will solve this way better for the ecosystem. We want to solve it similar to how other ecosystem have solves this by providing a canonical `swift-protoc` package that vends the binary. This would allows us to introduce a dependency from `swift-protobuf` onto that package and the plugin will just pull that in.
There is currently a technical restriction on this where we need some enhancements in the package manager to make declaring such a package easier. There is some work going on over here: https://github.com/apple/swift-package-manager/pull/6298

tbkka:
Another approach for handling relative paths is to ensure that the CWD is consistent at the point in the code where those paths are evaluated.  That reduces the need for path surgery.

PR #1391, opened 2023-04-04
Replace Data with SwiftProtobufContiguousBytes in serialization path

gjcairo:
This is a follow-up to https://github.com/apple/swift-protobuf/pull/1326. Usages of `Data` in the serialisation path have been replaced by usages of the new `SwiftProtobufContiguousBytes` protocol so that `[UInt8]` can also be used in the serialisation path, avoiding unnecessary translations between arrays of bytes and `Data`.

Note: we have decided **not** to change the `bytes` field of the protos from `Data` to `[UInt8]`/`SwiftProtobufContiguousBytes`, so this change does not include this.

This PR includes some new `Data` allocations, which are temporary to avoid growing the PR too much. I tried to restrict my changes to just the serialisation path, and plan on sending another one for deserialisation and a third one for JSON encoding/decoding.


Sources/SwiftProtobufCore/JSONEncoder.swift
```diff
@@ -102,18 +102,14 @@ internal struct JSONEncoder {
...
-
     /// Append a raw utf8 in a `Data` to the JSON text.
     internal mutating func append(utf8Data: Data) {
         data.append(contentsOf: utf8Data)
     }
 
     /// Append a raw utf8 in a `[UInt8]` to the JSON text.
-    internal mutating func append(utf8Data: [UInt8]) {
-        data.append(contentsOf: utf8Data)
+    internal mutating func append(bytes: [UInt8]) {
```
thomasvl, line 111: Drive by while you are working on things -- should this keep `utf8` in the name to make it clear it isn't arbitrary bytes, it has to be a UTF8 sequence that is being appended?

Sources/SwiftProtobufCore/BinaryEncodingSizeVisitor.swift
```diff
@@ -98,7 +97,8 @@ internal struct BinaryEncodingSizeVisitor: Visitor {
     serializedSize += tagSize + Varint.encodedSize(of: Int64(count)) + count
   }
 
-  mutating func visitSingularBytesField(value: Data, fieldNumber: Int) throws {
+  @inlinable
+  mutating func visitSingularBytesField<Bytes: SwiftProtobufContiguousBytes>(value: Bytes, fieldNumber: Int) throws {
```
thomasvl, line 101: Now that I think about it, if we're keeping bytes fields as `Data`, what's the value in making this generic like this?  Does it come with any costs?  (generated code size?  performance?)

Sources/SwiftProtobufCore/JSONEncoder.swift
```diff
@@ -69,7 +69,7 @@ internal struct JSONEncoder {
 
     internal init() {}
 
-    internal var dataResult: Data { return Data(data) }
+    internal var dataResult: [UInt8] { return data }
```
thomasvl, line 72: maybe rename from `dataResult` when you do the JSON side of things.

Sources/SwiftProtobufCore/Message+BinaryAdditions.swift
```diff
@@ -60,6 +60,7 @@ extension Message {
   /// Returns the size in bytes required to encode the message in binary format.
   /// This is used by `serializedData()` to precalculate the size of the buffer
   /// so that encoding can proceed without bounds checks or reallocation.
+  @usableFromInline
   internal func serializedDataSize() throws -> Int {
```
thomasvl, line 64: Should we consider renaming this with a shim in Compat?

Sources/SwiftProtobufCore/SwiftProtobufContiguousBytes.swift
```diff
@@ -18,23 +18,34 @@ import Foundation
 /// By conforming your own types to this protocol, you will be able to pass instances of said types
 /// directly to `SwiftProtobuf.Message`'s deserialisation methods
 /// (i.e. `init(serializedBytes:)` for binary format and `init(jsonUTF8Bytes:)` for JSON).
-// TODO: extend doc to include how this will be used in serialization once that API change has been finalised.
-public protocol SwiftProtobufContiguousBytes {
+public protocol SwiftProtobufContiguousBytes: Hashable {
```
thomasvl, line 21: Why was `Hashable` needed?

Sources/protoc-gen-swift/main.swift
```diff
@@ -250,9 +250,11 @@ struct GeneratorPlugin {
   }
 
   private func sendReply(response: Google_Protobuf_Compiler_CodeGeneratorResponse) -> Bool {
+    // NOTE: I can only get rid of Data here if I use FileHandle.standardOutput.write(contentsOf:)
+    // below. This is only available in macOS 10.15.4 - should we increase min OS version?
```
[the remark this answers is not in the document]
thomasvl: We don't currently list supported versions in `Package.swift`, but we do have things defined in the CocoaPods support. But even with that, this is sorta an interesting question, as there's no way in SwiftPM to express that one target has a different min version.  Since the plugin really only comes into play for the developers host platform, it would be nice to use a higher min OS version, but without a way to express that, we'd likely have to generally raise minimums to keep things happy.  Is there any way to _availability_ wrap this to also use the better API on linux when possible?

Sources/SwiftProtobufCore/BinaryEncoder.swift
```diff
@@ -13,9 +13,8 @@
 ///
 // -----------------------------------------------------------------------------
 
-import Foundation
-
 /// Encoder for Binary Protocol Buffer format
+@usableFromInline
```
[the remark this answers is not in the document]
FranzBusch: `@inlinable` and `@usableFromInline` have no API impact but do have ABI impact. What happens with `@inlinable` methods is that their whole body is put into the swift interface files so that dependent modules can specialise their code when calling them.

Sources/SwiftProtobufCore/BinaryEncodingSizeVisitor.swift
FranzBusch: I think this is a fair concern. If we keep bytes fields as `Data` we should already have bridged to `Data` before calling this method.

Sources/SwiftProtobufCore/BinaryEncodingSizeVisitor.swift
```diff
@@ -202,7 +202,8 @@ internal struct BinaryEncodingSizeVisitor: Visitor {
     serializedSize += tagSize * value.count + dataSize
   }
 
-  mutating func visitRepeatedBytesField(value: [Data], fieldNumber: Int) throws {
+  @inlinable
+  mutating func visitRepeatedBytesField<Bytes: SwiftProtobufContiguousBytes>(value: [Bytes], fieldNumber: Int) throws {
```
FranzBusch, line 206: Same here

Sources/SwiftProtobufCore/BinaryEncodingVisitor.swift
```diff
@@ -26,79 +26,96 @@ internal struct BinaryEncodingVisitor: Visitor {
...
 
+  @usableFromInline
   mutating func visitSingularStringField(value: String, fieldNumber: Int) throws {
     encoder.startField(fieldNumber: fieldNumber, wireFormat: .lengthDelimited)
     encoder.putStringValue(value: value)
   }
 
-  mutating func visitSingularBytesField(value: Data, fieldNumber: Int) throws {
+  @inlinable
+  mutating func visitSingularBytesField<Bytes: SwiftProtobufContiguousBytes>(value: Bytes, fieldNumber: Int) throws {
```
FranzBusch, line 106: Same applies here

Sources/protoc-gen-swift/main.swift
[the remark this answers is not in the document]
FranzBusch: I don't think getting rid of `Data` here is needed at all. This is the generator which can freely use `Data`. We only care about the _runtime_ where we want to make the bridging less expensive

Sources/SwiftProtobufCore/BinaryEncoder.swift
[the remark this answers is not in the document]
thomasvl: Sorry, right ABI vs. API, but sorta tied, in that my concern is without this, we could change these things in point release without any concerns, but someone has protos in a different framework, then it would seem like we're more locked in, no?

Sources/SwiftProtobufCore/BinaryEncoder.swift
```diff
@@ -41,20 +41,23 @@ internal struct BinaryEncoder {
     private mutating func append(contentsOf bufferPointer: UnsafeRawBufferPointer) -> Int {
         let count = bufferPointer.count
         if let baseAddress = bufferPointer.baseAddress, count > 0 {
-            memcpy(pointer, baseAddress, count)
+            self.pointer.copyMemory(from: baseAddress, byteCount: count)
```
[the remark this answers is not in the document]
gjcairo: Ah, good catch. This was one of the first files I touched and only realised later on that it wasn't done like this anywhere else - will get rid of them.

Sources/protoc-gen-swift/main.swift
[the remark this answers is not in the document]
gjcairo: Yeah you're right Franz, we can just leave this here.

Sources/SwiftProtobufCore/BinaryEncodingSizeVisitor.swift
gjcairo: You're right, this was unnecessary - visitors will operate on the protos' fields, and bytes fields will remain as `Data`, so no need to change the `Visitor` protocol. I'll roll this back.

Sources/SwiftProtobufCore/SwiftProtobufContiguousBytes.swift
gjcairo: Because of the `HashVisitor`. However, since we're getting rid of the generics in `Visitor` I can get rid of this conformance.

Sources/SwiftProtobufCore/BinaryEncodingSizeVisitor.swift
gjcairo: You're right, this was unnecessary - visitors will operate on the protos' fields, and bytes fields will remain as `Data`, so no need to change the `Visitor` protocol. I'll roll this back.

Sources/SwiftProtobufCore/Message+BinaryAdditions.swift
gjcairo: It's `internal`, so if we want to rename it I don't think the shim in Compat is necessary?

Sources/SwiftProtobufCore/BinaryEncoder.swift
```diff
@@ -28,8 +27,8 @@ internal struct BinaryEncoder {
         pointer = pointer.advanced(by: 1)
     }
 
-    private mutating func append(contentsOf data: Data) {
-        data.withUnsafeBytes { dataPointer in
+    private mutating func append<Bytes: SwiftProtobufContiguousBytes>(contentsOf bytes: Bytes) {
```
thomasvl, line 30: Does this still need to be generic?

thomasvl (commented):
Have you gotten a change to look at changes in binary sizes do to all the inlinable/usableFromInline additions?

Sources/SwiftProtobufCore/BinaryEncoder.swift
gjcairo: Yes, it's used by `BinaryEncoder/putBytesValue(value:)`, which is used by `BinaryDelimited/serialize(message:to stream:partial:)` and uses `[UInt8]`. However other uses of `append(contentsOf:)` use `Data`, so we're using both

thomasvl:
I think a FoundationCompat `serializedData(...)` addition and the question about the overall size of things now are my last concerns.

@FranzBusch @tbkka @Lukasa - have you gotten a chance to look?

FranzBusch:
@thomasvl I think providing a `serializedData()` overload is totally reasonable and will avoid breaking a bunch of clients when they want `Data` in the end anyhow.

W.r.t. to binary size @gjcairo can do a quick release compile of the `protoc-gen-swift` tool on `main` and this branch and we should see the size impact. I don't expect it too be too much though.

gjcairo:
@thomasvl - building `protoc-gen-swift` for release both with these changes and with the latest `main` yields binaries of 4,992,301 bytes and 5,065,117 bytes, respectively. This is about a ~73KB size increase for the `@inlinable` version. This is a very small change, so I wouldn't be worried about binary size.

I've also just added a `serializedData()` method to FoundationCompat.

Sources/SwiftProtobufCore/BinaryEncoder.swift
gjcairo: Actually, scratch all that - I'll revert this to being non-generic. I can just use Data in BinaryDelimited and the TextFormatEncoding visitor. We'll be reading Data fields from the protobufs themselves anyways, and will have to deserialise them into Data when converting back to protobuf, so it doesn't make sense for this to be generic.

FranzBusch:
@thomasvl We haven't run any benchmark to this particular code but we know from experience in NIO that generic method call across module boundaries that are not marked as `@inlinable` are one of the biggest perf cost. The problem is that the compiler is not able to see through the method so it fails to specialise anything.

We don't have any benchmarking infrastructure in this repo as far as I know but that might be a good addition in the future.

thomasvl:
> We don't have any benchmarking infrastructure in this repo as far as I know but that might be a good addition in the future.

https://github.com/apple/swift-protobuf/tree/main/Performance is what @tbkka was using last time performance work was done.

Sources/SwiftProtobufCore/BinaryEncoder.swift
gjcairo: Looking at this again - I've reverted some of the generics in other methods, but did put it back here. I decided to still use `SwiftProtobufContiguousBytes` instead of `Data` in `BinaryDelimited/serialize` to give the user more flexibility, since the method serialises multiple `Message`s into an `OutputStream`. I had missed that here we're not encoding a field, but rather reusing the `putBytesValue` method of the `BinaryEncoder` to encode the **whole** `Message`.

gjcairo:
@thomasvl doesn't this test the performance of the generator? I think Franz meant we don't have anything benchmarking runtime/application performance.

tbkka:
No, that performance suite benchmarks the runtime speed of the generated code.  Tony put it together originally and it works pretty well.

gjcairo:
Gotcha - I can run this and post the results.

tbkka:
Definitely test the performance before sprinkling too much inlining around -- inlining does help perf, but it also limits the ability to change the implementation later on.

FranzBusch:
@tbkka inlining doesn't limit anything here since we are not shipping this in a library evolution mode. In packages inlining is just there to make the perf better and we can remove it any point since it only affects ABI and not API.

We are doing this in NIO and other server libraries.

gjcairo:
@thomasvl @tbkka I've spent the last day or so trying to run the performance script (`Performance/perf_runner.sh`) but, after some changes to get it to build and run again, I end up hitting some apparently asbeil-related issues which might have to do with linking on Apple Silicon. The following if one of the (many) errors I get:
```
Undefined symbols for architecture arm64:
  "_AbslInternalSpinLockWake_lts_20230125", referenced from:
      absl::lts_20230125::base_internal::SpinLockWake(std::__1::atomic<unsigned int>*, bool) in libprotobuf.a(text_format.cc.o)
      absl::lts_20230125::base_internal::SpinLockWake(std::__1::atomic<unsigned int>*, bool) in libprotobuf.a(descriptor.cc.o)
      absl::lts_20230125::base_internal::SpinLockWake(std::__1::atomic<unsigned int>*, bool) in libprotobuf.a(message_differencer.cc.o)
      absl::lts_20230125::base_internal::SpinLockWake(std::__1::atomic<unsigned int>*, bool) in libprotobuf.a(type_resolver_util.cc.o)
      absl::lts_20230125::base_internal::SpinLockWake(std::__1::atomic<unsigned int>*, bool) in libprotobuf.a(message.cc.o)
      absl::lts_20230125::base_internal::SpinLockWake(std::__1::atomic<unsigned int>*, bool) in libprotobuf.a(wire_format.cc.o)
      absl::lts_20230125::base_internal::SpinLockWake(std::__1::atomic<unsigned int>*, bool) in libprotobuf.a(generated_message_reflection.cc.o)
```
I found this (still open) issue on google/protobuf https://github.com/protocolbuffers/protobuf/issues/10908 that seems to be the same.

I think the best way forward is probably to do the following:
- I can remove the `@inlinable`s from this PR and land it;
- We can fix the perf_runner separately (although this seems more like a protobuf issue in general, not just Swift proto related);
- Once we run the perf benchmarks, we can add inlining back again and decide what to do.

Does this sound like a plan? Any objections?

thomasvl:
> Does this sound like a plan? Any objections?

Makes sense to me.  Keeping the commits to one task at a time generally helps from a history pov if something comes up.

thomasvl:
LGTM.

@tbkka @Lukasa ?

gjcairo:
[This commit](https://github.com/protocolbuffers/protobuf/commit/b9e056f779f1c6cbf07a7a11be6f5ee6a4d433b8) in protobuf seems to be breaking the build for Swift 5.4, 5.5 and 5.6

thomasvl:
> [This commit](https://github.com/protocolbuffers/protobuf/commit/b9e056f779f1c6cbf07a7a11be6f5ee6a4d433b8) in protobuf seems to be breaking the build for Swift 5.4, 5.5 and 5.6

Sending out https://github.com/protocolbuffers/protobuf/pull/12492.

thomasvl:
> > [This commit](https://github.com/protocolbuffers/protobuf/commit/b9e056f779f1c6cbf07a7a11be6f5ee6a4d433b8) in protobuf seems to be breaking the build for Swift 5.4, 5.5 and 5.6
> 
> Sending out [protocolbuffers/protobuf#12492](https://github.com/protocolbuffers/protobuf/pull/12492).

Rerun tests passed.

Sources/SwiftProtobufCore/BinaryDelimited.swift
```diff
@@ -89,6 +91,31 @@ public enum BinaryDelimited {
     }
   }
 
+  /// Serialize a single size-delimited message from the given stream. Delimited
```
tbkka, line 94: s/from/to/

Sources/SwiftProtobufCore/BinaryDelimited.swift
```diff
@@ -44,31 +44,33 @@ public enum BinaryDelimited {
...
-  public static func serialize(
+  public static func serialize<Bytes: SwiftProtobufContiguousBytes>(
     message: Message,
     to stream: OutputStream,
+    as: Bytes.Type,
     partial: Bool = false
   ) throws {
     // TODO: Revisit to avoid the extra buffering when encoding is streamed in general.
-    let serialized = try message.serializedData(partial: partial)
+    let serialized: Bytes = try message.serializedBytes(partial: partial)
```
tbkka, line 62: This is the only use of the `Bytes` generic type.  Making this method generic seems to serve no real purpose, so I would just use `[UInt8]` here and drop the generic type parameter entirely.

Sources/SwiftProtobufCore/Message+BinaryAdditions.swift
```diff
@@ -16,18 +16,19 @@ import Foundation
...
   ///     fields are present. If any are missing, this method throws
   ///     `BinaryEncodingError.missingRequiredFields`.
-  /// - Returns: A `Data` value containing the binary serialization of the
-  ///   message.
+  /// - Returns: A `SwiftProtobufContiguousBytes` instance containing the binary serialization
+  /// of the message.
+  ///
   /// - Throws: `BinaryEncodingError` if encoding fails.
-  public func serializedData(partial: Bool = false) throws -> Data {
+  public func serializedBytes<Bytes: SwiftProtobufContiguousBytes>(partial: Bool = false) throws -> Bytes {
```
tbkka, line 31: It may not be appropriate today, but I wonder if we should consider someday rephrasing this to _append_ the serialized message to a provided inout buffer object.   That would require a slightly different protocol with a way to obtain an initializing view.  (CC: @atrick)

Sources/SwiftProtobufCore/SwiftProtobufContiguousBytes.swift
```diff
@@ -18,23 +18,34 @@ import Foundation
...
     ///            outside of the lifetime of the call to the closure.
     func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R
+
+    /// Calls the given closure with the contents of underlying storage.
+    ///
+    /// - note: Calling `withUnsafeBytes` multiple times does not guarantee that
+    ///         the same buffer pointer will be passed in every time.
+    /// - warning: The buffer argument to the body should not be stored or used
+    ///            outside of the lifetime of the call to the closure.
+    mutating func withUnsafeMutableBytes<R>(_ body: (UnsafeMutableRawBufferPointer) throws -> R) rethrows -> R
```
tbkka, line 46: The "initializing view" I mentioned above would be functionally equivalent to the following (but with compiler-enforced safety guarantees that we're just starting to plan out).
```
// Allocate `size` bytes after the current data and call the specified closure to initialize those bytes
mutating func withUnsafeAppendedBytes<R>(size: Int, initializer: (UnsafeMutableRawBufferPointer) throws -> R) rethrows -> R
```

tbkka (approved):
This looks quite good.  I have a couple of small nits and an idea for how this might evolve as we continue to improve the language safety model in the future.

Sources/SwiftProtobufCore/Message+BinaryAdditions.swift
atrick: The client could do an append operation which provides an "initializing view" of the new storage, passing that to a closure. The closure would need to promise to initialize all the bytes. While you're holding onto a "view" of any kind, you won't be able to append or do anything that modifies the storage structure.

atrick: But serializedDataSize() is internal... Having the message allocate the storage seems safer and more encapsulated.

atrick: The only concern I might have with this approach is that it forces `SwiftProtobufContiguousBytes` to be able to dynamically allocate and manage storage. So this is more than just a "contiguous bytes" protocol now.
When we expose a public "contiguous bytes" protocol (a.k.a. BufferView), we can't make that guarantee.

thomasvl (approved):

thomasvl:
@tbkka any things else you were looking for here, or should we go ahead and squash/merge?

tbkka:
Go for it!

PR #1392, opened 2023-04-21
Clean up deserialisation path and regenerate protos

gjcairo:
This is a follow up of https://github.com/apple/swift-protobuf/pull/1391.

After the changes to remove `Data` from the (de)serialisation paths, some functions were actually left in that aren't necessary anymore, since we decided **not** to change the type of the `bytes` field in protos (i.e. it will remain as `Data`). This PR gets rid of the unused `[UInt8]` overloads.

I've also regenerated protos with the latest changes.


thomasvl (changes requested):
looks like you might also need to do a `make reference` for the plugin tests to pass.

thomasvl (approved):

PR #1395, opened 2023-04-25
Retire SwiftProtobufFoundationCompat.

thomasvl:
Since Foundation is still needed within SwiftProtobufCore, move the few files back and rename them to call out that have the `Data` bridging.


thomasvl:
fyi @FranzBusch @gjcairo

thomasvl:
> Do we also want to move `SwiftProtobufCore` back to being just `SwiftProtobuf`? Or do we still see a need for having the separation?

I was going to do that in a followup PR.  I was waiting to make sure there wasn't any other `Data` related work about to land, but also because if we ever pick up the idea of splitting off TextFormat/JSON so it is optional, I figured have the _Core_ shuffle might be useful for someone to revert/look at.

allevato (approved):

PR #1400, opened 2023-05-07
Add ExtraModuleImports option

pgmitche:
Adds `ExtraModuleImports` generator option, just [like that present in the gRPC plugin](https://github.com/grpc/grpc-swift/blob/75b390e901c7d70af9b4c5ca2677e035f00cd1ab/Sources/protoc-gen-grpc-swift/options.swift#L64)
 & [connect-swift plugin](https://github.com/bufbuild/connect-swift/blob/ed165049ecd6e111e8bb2834a14292719556d0f8/Plugins/ConnectPluginUtilities/GeneratorOptions.swift#L109)

### DIFF
--- a/Sources/protoc-gen-swift/GeneratorOptions.swift
+++ b/Sources/protoc-gen-swift/GeneratorOptions.swift
@@ -49,7 +49,7 @@ class GeneratorOptions {
   let outputNaming: OutputNaming
   let protoToModuleMappings: ProtoFileToModuleMappings
   let visibility: Visibility
-  let extraModuleImports: [String]?
+  let extraModuleImports: [String]
 
   /// A string snippet to insert for the visibility
   let visibilitySourceSnippet: String
@@ -91,14 +91,14 @@ class GeneratorOptions {
                                                       value: pair.value)
         }
       case "ExtraModuleImports":
-      if !pair.value.isEmpty {
-          externalModuleImports.append(pair.value)
-      } else {
-        throw GenerationError.invalidParameterValue(name: pair.key, value: pair.value)
-      }
-      default:
-        throw GenerationError.unknownParameter(name: pair.key)
-      }
+        if !pair.value.isEmpty {
+            externalModuleImports.append(pair.value)
+        } else {
+          throw GenerationError.invalidParameterValue(name: pair.key, value: pair.value)
+        }
+        default:
+          throw GenerationError.unknownParameter(name: pair.key)
+        }
     }
 
     if let moduleMapPath = moduleMapPath {
--- a/Sources/protoc-gen-swift/GeneratorOptions.swift
+++ b/Sources/protoc-gen-swift/GeneratorOptions.swift
@@ -49,6 +49,7 @@ class GeneratorOptions {
   let outputNaming: OutputNaming
   let protoToModuleMappings: ProtoFileToModuleMappings
   let visibility: Visibility
+  let extraModuleImports: [String]?
 
   /// A string snippet to insert for the visibility
   let visibilitySourceSnippet: String
@@ -58,6 +59,7 @@ class GeneratorOptions {
     var moduleMapPath: String?
     var visibility: Visibility = .internal
     var swiftProtobufModuleName: String? = nil
+    var externalModuleImports: [String] = []
 
     for pair in parseParameter(string:parameter) {
       switch pair.key {
@@ -88,6 +90,12 @@ class GeneratorOptions {
           throw GenerationError.invalidParameterValue(name: pair.key,
                                                       value: pair.value)
         }
+      case "ExtraModuleImports":
+      if !pair.value.isEmpty {
+          externalModuleImports.append(pair.value)
+      } else {
+        throw GenerationError.invalidParameterValue(name: pair.key, value: pair.value)
+      }
       default:
         throw GenerationError.unknownParameter(name: pair.key)
       }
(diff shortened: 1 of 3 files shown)


Sources/protoc-gen-swift/GeneratorOptions.swift
```diff
@@ -88,6 +90,12 @@ class GeneratorOptions {
           throw GenerationError.invalidParameterValue(name: pair.key,
                                                       value: pair.value)
         }
+      case "ExtraModuleImports":
+      if !pair.value.isEmpty {
```
thomasvl, line 94: Indent seems off here.

Sources/protoc-gen-swift/GeneratorOptions.swift
```diff
@@ -49,6 +49,7 @@ class GeneratorOptions {
   let outputNaming: OutputNaming
   let protoToModuleMappings: ProtoFileToModuleMappings
   let visibility: Visibility
+  let extraModuleImports: [String]?
```
thomasvl, line 52: If I'm reading the code correctly, this is always getting set, so the optional part doesn't seem worth while.  Does the generation code then also not need to be an `if let` but instead a check to see if it is empty?

thomasvl (commented):
Looking back at gRPC, I don't really follow what this might be needed for, do you have an example of why this would be needed?

glbrntt:
> Looking back at gRPC, I don't really follow what this might be needed for, do you have an example of why this would be needed?

Its presence in gRPC predates our involvement with it. The PR (https://github.com/grpc/grpc-swift/pull/402) which added it has some rationale behind it but as far as I'm aware it's not used in gRPC at the moment and wouldn't make sense to add to SwiftProtobuf.

pgmitche:
> Looking back at gRPC, I don't really follow what this might be needed for, do you have an example of why this would be needed?

Hi @thomasvl, thanks for reviewing!

I'm trying to inject import statements for modules containing SwiftProtobuf outputs that aren't generated to the same directory, nor to the same Swift project.

For example, if I have two proto packages that belong to different projects within a Swift registry, and I want to generate code for both of them separately, where PaymentProject declares a Payment message that depends on a message from the OrderProject:

```proto
message Payment {
    ...
    order.v1alpha.Order order = 2; 
}
```

I want to be able to generate both projects independently, but inject an import statement for the dependency so that the package will resolve when the registry dependency is provided in the manifest file like:
```swift
import OrderV1Alpha1
```

thomasvl:
That should already be covered by https://github.com/apple/swift-protobuf/blob/main/Documentation/PLUGIN.md#generation-option-protopathmodulemappings---swift-module-names-for-proto-paths.

You just provide the mapping of what files end up in which module, and the generator will generate the correct imports as well use module references for symbols to ensure there is no ambiguity if some other module happens to have the same symbol name.

thomasvl:
@pgmitche grpc/swift does support the [module mappings](https://github.com/grpc/grpc-swift/blob/main/docs/plugin.md#protopathmodulemappings), and it would appear that connect-swift [does also](https://github.com/bufbuild/connect-swift/blob/main/Plugins/ConnectPluginUtilities/GeneratorOptions.swift#L143). So there may be something universally supported that you could use.

> Please feel free to close this out if you don't want to add it, and thanks for the feedback!

I'll defer to @tbkka for it Apple wants to add/maintain this option.

tbkka:
I don't find this PR that interesting as-is.  I mostly agree with Thomas that the module mappings file makes more sense.  In particular, it is more accurate since it inserts import statements only and exactly when they are needed.  So I'm not eager to have a redundant option for the protoc plugin.

I think there would be value in having a way to express the import information entirely in the SwiftPM package file (as configuration for the SwiftPM plugin) without requiring a separate mapping file.  But the SwiftPM integration seems to be missing from this PR entirely.

pgmitche:
> @pgmitche grpc/swift does support the [module mappings](https://github.com/grpc/grpc-swift/blob/main/docs/plugin.md#protopathmodulemappings), and it would appear that connect-swift [does also](https://github.com/bufbuild/connect-swift/blob/main/Plugins/ConnectPluginUtilities/GeneratorOptions.swift#L143). So there may be something universally supported that you could use.
> 
> > Please feel free to close this out if you don't want to add it, and thanks for the feedback!
> 
> I'll defer to @tbkka for it Apple wants to add/maintain this option.

One key difference between the two that preclude use of `ProtoPathModuleMappings` for us is that for users who develop the protobuf package, building out a mapping file beforehand is simple, but for package-agnostic systems, building out a mapping file at runtime is more complex/involved than simply providing the extra import.

tbkka:
> ... for package-agnostic systems, building out a mapping file at runtime ...

Could you please spell that out in detail?  I think I'm missing some key point here...

FranzBusch:
> @tbkka I think we'd have to duplicate some info in the config json vs. what's in the Package.swift files, as I don't think the SwiftPM plugin interface would allow the plugin to auto generate the module mappings as just not enough info is expose (and propagated between targets).

When developing the SPM plugin I have played around with this some. I wanted to get rid of an explicit `ModuleMapping` file completely by just using the dependency information that SPM gives a plugin.
I was imagining this to work like this:
- TargetA: Contains `foo.proto`
- TargetB: Contains `bar.proto` which imports `TargetA/foo.proto`

In the plugin we are going automatically create a ModuleMapping file for any module that has a `proto` file. In the end, we are going to merge them together so that any target gets all proto module mappings from downstream. There were a bunch of open TODOs such as extending SPM to expose the necessary information but IMO we should explore something here which leverages the SPM dependency tree.

PR #1402, opened 2023-05-22
Fix compilation for web assembly

mwilsoncricut:
When compiling for web assembly on linux it currently gives this error. `swift-protobuf/Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift:20:5: error: cannot find 'DispatchQueue' in scope`. Adding `#if !os(WASI)` should fix this issue. Note it is also being used this way in [foundation]( https://github.com/apple/swift-corelibs-foundation/blob/818de4858f3c3f72f75d25fbe94d2388ca653f18/Sources/Foundation/NSIndexSet.swift#L10)

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
@@ -14,11 +14,13 @@
 // -----------------------------------------------------------------------------
 
 import Foundation
+#if !os(WASI)
 #if canImport(Dispatch)
 import Dispatch
 fileprivate var knownTypesQueue =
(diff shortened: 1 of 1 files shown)


tbkka (approved):

Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
```diff
@@ -14,12 +14,14 @@
 // -----------------------------------------------------------------------------
 
 import Foundation
+#if !os(WASI)
 #if canImport(Dispatch)
```
thomasvl, line 18: Any idea why this is true if Dispatch symbols aren't provided?

PR #1407, opened 2023-06-14
Add XcodeBuildToolPlugin and Directories support to SPM Plugin

denil-ct:
## Overview
This PR adds support for XcodeBuildToolPlugin and directory support to the SPM plugin. 

## Motivation

The motivation for adding XcodeBuildToolPlugin support is to enable the SPM plugin to be added to the Xcode `Run Build Tool Plugins` build phase. This will allow users to generate Swift code from proto files automatically when building their projects. The XcodeBuildToolPlugin protocol defines the interface and behavior for plugins that can be invoked by Xcode during the build process.

The motivation for adding directory support is to simplify the usage of the SPM plugin when there are multiple proto files in a directory or sub-directories. Instead of listing each proto file individually in the JSON configuration, users can specify a directory path and the plugin will recursively find all the proto files in that directory and its sub-directories. This will reduce the verbosity and complexity of the JSON configuration.

## Example JSON usage

To use the directory support, users need to add a new key called `protoDirectories` in their JSON configuration. This can be added along with the existing `protoFiles` key. The value of `protoDirectories` should be an array of strings, each representing a directory path that contains proto files. Also, take care that the directories don't have a trailing `/`.
For example:

```json
{
    "invocations": [
        {
            "protoDirectories": [
                "ProtoFiles/Feature1"
            ],
            "protoFiles": [
                "ProtoFiles/Feature2/feature.proto"
            ],
            "visibility": "internal"
        }
    ]
}
```

The plugin will find all the proto files in `ProtoFiles/Feature1` and their sub-directories, and add them to the list of input files along with `ProtoFiles/Feature2/feature.proto`. Users can also give an empty string to the directories key which will be equivalent to getting all the proto files in the target root directory and its sub-directories.

There are also three additional errors added
```swift
/// Indicates that the input directory had a trailing `/`.
case invalidDirectoryFormat
/// Indicates that there was no configuration file at the required location.
case noConfigFound
/// Indicates that there was no file or directory mentioned in the json.
case noProtoSourceFound
```

### DIFF
--- a/Plugins/SwiftProtobufPlugin/plugin.swift
+++ b/Plugins/SwiftProtobufPlugin/plugin.swift
@@ -2,13 +2,19 @@ import Foundation
 import PackagePlugin
 
 @main
-struct SwiftProtobufPlugin: BuildToolPlugin {
+struct SwiftProtobufPlugin {
     /// Errors thrown by the `SwiftProtobufPlugin`
     enum PluginError: Error {
         /// Indicates that the target where the plugin was applied to was not `SourceModuleTarget`.
         case invalidTarget
         /// Indicates that the file extension of an input file was not `.proto`.
         case invalidInputFileExtension
+        /// Indicates that the input directory had a trailing `/`.
+        case invalidDirectoryFormat
+        /// Indicates that there was no configuration file at the required location.
+        case noConfigFound
+        /// Indicates that there was no file or directory mentioned in the json.
+        case noProtoSourceFound
     }
 
     /// The configuration of the plugin.
@@ -57,7 +63,9 @@ struct SwiftProtobufPlugin: BuildToolPlugin {
             }
 
             /// An array of paths to `.proto` files for this invocation.
-            var protoFiles: [String]
+            var protoFiles: [String]?
+            /// An array of directories that contains `.proto` files for this invocation.
+            var protoDirectories: [String]?
             /// The visibility of the generated files.
             var visibility: Visibility?
             /// The file naming strategy to use.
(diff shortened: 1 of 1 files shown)


denil-ct:
@FranzBusch could you take a look at this, please?

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -189,15 +228,101 @@ struct SwiftProtobufPlugin: BuildToolPlugin {
...
+    ///   - parent: The parent directory of the path.
+    /// - Returns: An array of proto files from the directory and all its subdirectories.
+    private func getProtoFiles(
+        from path: String,
+        in parent: String
+    ) throws -> [String] {
+        let url = URL(fileURLWithPath: parent.appending("/\(path)"))
+        var files = [String]()
+        if url.hasDirectoryPath {
+            if let enumerator = FileManager.default.enumerator(
```
FranzBusch, line 245: This part here is where it gets a bit tricky. SPM does have an internal path and file management implementation that works across different OS's. I would rather we add globbing support to SPM than to reinvent the wheel here.

denil-ct:
Ah I see, makes sense. I was also conflicted on adding it in the plugin, but since we had many proto files adding them individually to the JSON is a pain, hence helplessly added it.

Will split it up and continue on their individual threads.

FranzBusch:
If you want to add the directory support, I would encourage you to look into adding globbing support on the SPM side. That would be a great contribution and help any SPM plugin.

denil-ct:
Sure, that sounds interesting. However as a stopgap solution can we add something similar to this?
Initially I used a bash find process to get globbing support, but then switched over to using FileManager and added a new property on the json.

denil-ct:
Cool, I'll close this PR and open a new one for the `XcodeBuildToolPlugin` alone.

Issue #1410, opened 2023-06-25
Errors thrown and performance degradation when reading binary delimited data directly from a file


willtemperley:
When reading a 26MB binary delimited file with an InputStream created directly from a file (i.e. ```let inputStream = InputStream(url: fileUrl)```, the following test always throws a ```SwiftProtobuf.BinaryDelimited.Error.truncated``` error with a measured [Time, seconds] average: 1.893

```
    func testReadArchiveError() throws {
        measure {
            var nMessages = 0
            let inputStream = InputStream(url: fileUrl)
            if let inputStream = inputStream {
                inputStream.open()
                while (inputStream.hasBytesAvailable) {
                    do {
                        _ = try BinaryDelimited.parse(messageType: Io_Temperley_Archive_BaseMessage.self, from: inputStream)
                        nMessages += 1
                    } catch {
                        print(String(reflecting: error))
                    }
                }
                
            }
            print("n messages: \(nMessages)")
        }
    }
```

however, reading the same file into memory first never throws, with a measured [Time, seconds] average: 1.236

```
    func testReadArchive_No_Error() throws {
        measure {
            var nMessages = 0
            let data = try? Data(contentsOf: fileUrl)
            if let data = data {
                let inputStream = InputStream(data: data)
                inputStream.open()
                while (inputStream.hasBytesAvailable) {
                    do {
                        _ = try BinaryDelimited.parse(messageType: Io_Temperley_Archive_BaseMessage.self, from: inputStream)
                        nMessages += 1
                    } catch {
                        print(String(reflecting: error))
                    }
                }
            }
            print("n messages: \(nMessages)")
        }
    }
```

This feels like a bug - performance is significantly worse when reading directly from the file, plus it's clear that the stream isn't truncated. Maybe InputStream.hasBytesAvailable is just unreliable, but it seems surprising. The Java version doesn't have this issue. (I'm porting a Java app to Swift - the input file was created in Java). The following runs in ~1.18 seconds.

```
    @Test
    public void messageCount() throws IOException {
        FileInputStream inputStream = new FileInputStream(fileName);

        int nMessages = 0;
        while (inputStream.available() > 0) {
            ArchiveSchema.BaseMessage baseMessage = ArchiveSchema.BaseMessage.parseDelimitedFrom(inputStream);
            if (baseMessage != null) {
                nMessages++;
            }
        }

        System.out.println("i = " + nMessages);
    }
```

* what OS you are developing on (Linux or macOS, including the version)
[...]

PR #1414, opened 2023-06-30
Added StreamDecodingIterator extension to Message

willtemperley:
As discussed in https://github.com/apple/swift-protobuf/issues/1410


Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
```diff
@@ -0,0 +1,131 @@
...
+      let messageLength = try decodeVarint()
+      if (messageLength == 0) {
+        return nil
+      }
+      var message = M()
+      let data = try buffer.read(for: Int(messageLength))
+      try message.merge(serializedData: data)
+      return message
+    } catch {
+      print(error)
```
thomasvl, line 46: Probably don't want to `print` here. I guess there's no way to expose to the iterator when the stream was corrupted vs. just ended?

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
```diff
@@ -0,0 +1,131 @@
...
+  fileprivate let buffer: ResizingBuffer
+  
+  public init(inputStream: InputStream, bufferLength: Int = 32768) {
+    buffer = ResizingBuffer(inputStream: inputStream, bufferLength: bufferLength)
+  }
+  
+  public func next() -> M? {
+    do {
+      let messageLength = try decodeVarint()
+      if (messageLength == 0) {
```
thomasvl, line 38: So this means a message that has all default values can never be read back on the stream, all default values encodes to zero length.

You'll likely need to let decodeVarint return errors, no data, and/or and explicit zero to that message case can be handled.  Maybe just returning a optional would do it, so if it return nil you know something went wrong?

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
```diff
@@ -0,0 +1,131 @@
...
+        return nil
+      }
+      var message = M()
```
thomasvl, line 41: You should also handle the .tooLarge case from delimited data. If the message is over the max size, it means you've got a bad stream.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
```diff
@@ -0,0 +1,131 @@
...
+
+private class ResizingBuffer {
+  
+  let inputStream: InputStream
+  let bufferLength: Int
+  var pointer: Int = 0
+  var buffer: [UInt8]
+  var data: Data
+  
+  init(inputStream: InputStream, bufferLength: Int = 32768) {
```
thomasvl, line 86: No need to default bufferLength here, have all the callers pass it (and they can default it).

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
```diff
@@ -0,0 +1,131 @@
...
+///
+/// Extensions to `Message` to provide binary coding and decoding using ``Foundation/Data``.
+///
+// -----------------------------------------------------------------------------
+
+import Foundation
+
+extension Message {
+  
+  public static func streamDecodingIterator(inputStream: InputStream, bufferLength: Int = 32768) -> StreamDecodingIterator<Self> {
```
thomasvl, line 19: Maybe do a file scoped constant for the two public apis that default the buffer size?

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
```diff
@@ -0,0 +1,131 @@
...
+  }
+}
+
+private class ResizingBuffer {
+  
+  let inputStream: InputStream
+  let bufferLength: Int
+  var pointer: Int = 0
+  var buffer: [UInt8]
+  var data: Data
```
thomasvl, line 84: Does this means we're double buffering between these two?

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
```diff
@@ -0,0 +1,131 @@
...
+    } else {
+      //more data may be required
+      let prefix = data[pointer..<data.count]
+      let remainingByteCount = nBytes - prefix.count
+      
+      var toLoad = bufferLength
+      if (remainingByteCount > bufferLength) {
+        let nPages = Int(ceil(Float(remainingByteCount) / Float(bufferLength)))
+        toLoad = nPages * bufferLength
+        self.buffer = [UInt8](repeating: 0, count: toLoad)
```
thomasvl, line 115: At this point, shouldn't bufferLength get updated to the new value? (toLoad), otherwise it might resize it again when it already is big enough.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
```diff
@@ -0,0 +1,131 @@
...
+      }
+      
+      let result = inputStream.read(&buffer, maxLength: buffer.count)
```
thomasvl, line 118: I think this might need to be a loop that keeps reading until an error, end of stream, or the desired total bytes is read.

InputStream can return less than the requested length, but that doesn't mean the end of the stream has been reached, it could be the process got a signal, etc. and the IO got stopped at that point.

Tests/SwiftProtobufTests/Test_StreamDecodingIterator.swift
```diff
@@ -0,0 +1,58 @@
...
+        $0.optionalInt64 = 123456789
+        $0.optionalGroup.a = 456
+        $0.optionalNestedEnum = .baz
+        $0.repeatedString.append("wee")
+        $0.repeatedFloat.append(1.23)
+      }
+      testInt += 1
+      XCTAssertEqual(msg1, message)
+    }
+  }
```
thomasvl, line 57: Probably want to assert that testInt got to 10.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
```diff
@@ -0,0 +1,131 @@
...
+  public func next() -> M? {
+    do {
+      let messageLength = try decodeVarint()
+      if (messageLength == 0) {
+        return nil
+      }
+      var message = M()
+      let data = try buffer.read(for: Int(messageLength))
+      try message.merge(serializedData: data)
+      return message
```
tbkka, line 44: Or more simply:
```
let data = try buffer.read(for: Int(messageLength))
return try M(serializedData: data)
```

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
```diff
@@ -0,0 +1,131 @@
...
+  let bufferLength: Int
+  var pointer: Int = 0
+  var buffer: [UInt8]
+  var data: Data
+  
+  init(inputStream: InputStream, bufferLength: Int = 32768) {
+    self.inputStream = inputStream
+    self.bufferLength = bufferLength
+    self.buffer = [UInt8](repeating: 0, count: bufferLength)
+    let result = inputStream.read(&buffer, maxLength: buffer.count)
```
thomasvl, line 90: This initial `read` means you can't use the iterator to just pull of a few messages, it could read more data than what is actually iterated through.  So the public apis for making the iterator probably should document that they read in advance, and so a developer should not make any other calls on the stream after creating the Iterator referencing the stream.

Tests/SwiftProtobufTests/Test_StreamDecodingIterator.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+// -----------------------------------------------------------------------------
+
+import Foundation
+import XCTest
+import SwiftProtobuf
+
+class Test_StreamDecodingIterator: XCTestCase {
+  
+  func testEverything() {
```
tbkka, line 17: You should also have a test for the case of a zero-sized message (as Thomas pointed out above, zero-size messages are perfectly valid in protobuf).

Also tests for various truncation cases.  I would suggest have a suite of tests where you include short binary streams with known content (and problems).  See Test_AllTypes.swift for an example of this sort of thinking.

Among others:
* Empty stream with no bytes
* Stream with truncated varint
* Stream with zero varint and nothing else
* A stream with legal non-zero varint but no message
* Stream with valid varint, message, but the following varint is truncated

tbkka (commented):
Overall structure looks good.  I agree with Thomas' issues, and I'd also like to see a number of additional tests.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: Good catch, I think it makes sense to make ```decodeVarint``` return nil on a zero-length read. Attempting a read seems to be the only way to know we're at the end an ```InputStream``` and in normal execution that will happen when attempting to read the varint data.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: Yes I wasn't sure what to do here as ```IteratorProtocol.next``` can't throw. Maybe setting a status field on the iterator would do the trick?

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: No. ```Data``` is just a view on the start of of the buffer - at the end of the stream the buffer will have extraneous data from the previous read.

willtemperley: Incidentally I tried using ArraySlice<Int8> instead of Data and got a ~20% performance hit, no idea why.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: Yep!

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
```diff
@@ -0,0 +1,131 @@
...
+}
+
+/**
+ Iterates over a binary-delimited protobuf input stream
+ */
+public struct StreamDecodingIterator<M: Message>: IteratorProtocol {
+  
+  fileprivate let buffer: ResizingBuffer
+  
+  public init(inputStream: InputStream, bufferLength: Int = 32768) {
```
tbkka, line 31: I don't think this init needs to be public.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: Yes

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: yes much nicer

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: To deal with error-handling I added an optional StreamErrorDelegate to the iterator initializer - demonstrated in the test suite.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: Done

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: Done

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: It wasn't a double buffer, it was a view on the valid section of the buffer (the prefix based on how many bytes the read managed) however you're right the ```Data``` var was untidy - refactored to remove this.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: True. I removed all the buffer resizing.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: Good call. Refactored to ensure the read result and stream status is always checked.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: True, added a comment and removed the initial read.

Tests/SwiftProtobufTests/Test_StreamDecodingIterator.swift
willtemperley: Those tests have been added, plus one using a small buffer to test multiple-buffer message situation.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: Done

Tests/SwiftProtobufTests/Test_StreamDecodingIterator.swift
willtemperley: Done

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
```diff
@@ -0,0 +1,150 @@
...
+        messageData += buffer.prefix(lastReadResult)
+        requiredBytes -= lastReadResult
+        consumedBytes = 0
+      }
+    }
+    return messageData
+  }
+}
+
+public protocol StreamErrorDelegate {
```
thomasvl, line 148: Document since it is public?

Tests/SwiftProtobufTests/Test_StreamDecodingIterator.swift
```diff
@@ -0,0 +1,232 @@
...
+    
+    let inputStream = InputStream(data: data)
+    inputStream.open()
+    let iterator = ProtobufUnittest_TestAllTypes.streamDecodingIterator(inputStream: inputStream)
+    var count = 0
+    while let message = iterator.next() {
+      XCTAssertEqual(message, ProtobufUnittest_TestAllTypes())
+      count += 1
+    }
+    XCTAssert(count == 10)
```
thomasvl, line 44: XCTAssertEqual so it reports the values when they don't match

Tests/SwiftProtobufTests/Test_StreamDecodingIterator.swift
```diff
@@ -0,0 +1,232 @@
...
+    inputStream.close()
+  }
+  
+  //Empty stream with no bytes
+  func testEmptyStream() {
+    let inputStream = InputStream(data: Data(count: 0))
+    inputStream.open()
+    let iterator = ProtobufUnittest_TestAllTypes.streamDecodingIterator(inputStream: inputStream)
+    while let _ = iterator.next() {
+      XCTFail("Stream was empty")
```
thomasvl, line 54: Probably want to clarify this to be something like - Shouldn't have returned a value for an empty stream.

Tests/SwiftProtobufTests/Test_StreamDecodingIterator.swift
```diff
@@ -0,0 +1,232 @@
...
+    errorExpectation = expectation(description: "Should encounter a BinaryDecodingError.truncated")
+    
+    let inputStream = InputStream(data: Data([192]))
+    inputStream.open()
+    let iterator = ProtobufUnittest_TestAllTypes.streamDecodingIterator(inputStream: inputStream, errorDelegate: self)
+    while let _ = iterator.next() {
+      XCTFail("Zero messages should be encountered.")
+    }
+    waitForExpectations(timeout: 1)
+    XCTAssert(error as! BinaryDecodingError == BinaryDecodingError.truncated)
```
thomasvl, line 70: AssertEqual?

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
willtemperley: Done

Tests/SwiftProtobufTests/Test_StreamDecodingIterator.swift
willtemperley: Done

Tests/SwiftProtobufTests/Test_StreamDecodingIterator.swift
willtemperley: Done

Tests/SwiftProtobufTests/Test_StreamDecodingIterator.swift
willtemperley: Done

Tests/SwiftProtobufTests/Test_StreamDecodingIterator.swift
```diff
@@ -0,0 +1,232 @@
...
+import XCTest
+import SwiftProtobuf
+
+class Test_StreamDecodingIterator: XCTestCase, StreamErrorDelegate {
+  
+  var error: Error?
+  var errorExpectation: XCTestExpectation?
+  
+  func onError(error: Error) {
+    self.error = error
```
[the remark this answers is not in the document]
willtemperley: Ah nice, done.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
```diff
@@ -0,0 +1,158 @@
...
+
+fileprivate let defaultBufferLength = 32768
+
+extension Message {
+  
+  /**
+   Creates an iterator for a binary-delimited protobuf input stream. Errors can be caught by providing a
+   delegate. The returned iterator uses a read-ahead buffer, therefore no assumptions about the position of
+   the stream should be made after any messages have been read.
+   */
```
FranzBusch, line 25: Can we use `///` for doc comments please and document all the parameters of the methods. This follows what we use in other places in this repository.

FranzBusch (commented):
Before diving into a deep review here, I am a bit wary of the usage of `InputStream`. This type is pretty old and the new [swift-foundation](https://github.com/apple/swift-foundation) repository indicates that they don't intend to bring it forward.

I would like to take a step back and revisit what use-case we want to solve here. Are we trying to stream data from a file and decode `Message`s on the fly? Or similarly are we trying to decode messages in a networking stack?

How are these messages delimited concretely and is this format something standardised?

FranzBusch:
@willtemperley Thanks for expanding again on the goals and laying out the current restrictions. Very much appreciated! I think looking at the `AsyncSequence` based approach would be great and we can compare how they look differently. In the end, the `AsyncSequence` based approach could potentially be a type that is itself an `AsyncSequence` and just takes a base async sequence of bytes that it iterates and then transforms it into an async sequence of `Message`s.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
tbkka: That seems like a reasonable approach.

FranzBusch:
> I think the real question here is whether we want to eventually have both versions (an `InputStream`-based one and an `AsyncStream`/`AsyncSequence`-based one) or not. We already have a plan for segregating the various Foundation-based portions of the library so end users will be able to opt into using Foundation or not. That would seem to open a door for accepting this now with a plan to include it as part of the Foundation extension library once we are ready to make that split.

Personally, I think we should move over to the `AsyncSequence` based approach. `InputStream` is a pretty old type and from the looks of it not going to be part of the new `swift-foundation` package. This would mean we would only support it for Darwin based platforms. Additionally, all new APIs are using `AsyncSequence` for their interfaces. In general, I am not opposed to create new `InputStream` APIs if there is a real need; however, in this particular case it looked like the `AsyncSequence` based approach was a better fit.

> That raises a similar question for async/await-based code. Do we want to segregate that also? That's less compelling to me, but I'm curious what others think?

I don't think we should separate the async/await code out. The motivation behind a separate foundation support target was driven by the dependencies corelibs-foundation required on non-Darwin platforms and the resulting binary bloat. Async/await is already bundled with the language and runtime so I don't see a big need for separating it out.

Sources/SwiftProtobuf/Message+StreamDecodingIterator.swift
```diff
@@ -0,0 +1,158 @@
...
+    guard let messageLength = decodeVarint() else {
+      return nil
+    }
+    guard messageLength <= 0x7fffffff else {
+      errorDelegate?.onError(error: BinaryDecodingError.tooLarge)
+      return nil
+    }
+    if messageLength == 0 {
+      return M()
+    }
```
tbkka, line 58: @thomasvl Is this right for proto2 messages with required fields?  Seems that such messages could fail to decode with a zero-length input.

thomasvl: Good catch, it is an edge case, but yea, if they are required, then this case should fail.  Looking back at BinaryDelimited.swift, we did add the `partial` arguments in there, so we could work those in here also to still get the initialization checks.

The catch is still the error flow - we don't want the iterator to fail and return `nil` because that indicates the end of the iteration, so how would that sorta failure get exposed?  Would it just skip on to the next message after reporting the error?

thomasvl:
> @FranzBusch Here's a [basic async decoder](https://github.com/willtemperley/swift-protobuf/commit/791e404f4aab1feb872ca92e9cb2ebddba095e34)
> 
> It's about three times slower on my own data (large protobuf files) than the `InputStream` approach - not sure why. It's much more compact though.

Does this mean we maybe look at landing both on the hopes the async one getter faster in the future?

willtemperley:
Maybe be the AsyncSequence approach is faster over a network? I was only testing with files.

My suggestion with the IteratorProtocol implementation is a bring-your-own-buffer approach. Then the caller could decide to use an InputStream if they wanted, or maybe use SwiftNIO.

Then SwiftProtobuf could maybe provide something like:

```
public protocol DecodingBuffer {
  func read(for nBytes: Int) -> [UInt8]
}
```

There'd be zero Foundation types and it'd tidy everything up a lot, e.g. no need to document buffer behaviour.

Another thing is as making ```IteratorProtocol``` conform to ```Sequence``` comes for free this might as well be done?

FranzBusch:
I am personally a bit wary of introducing new APIs that work on top of `InputStream`. The type is basically unused in the Swift on Server ecosystem and pretty old. But if others here think that we should land this, I am happy with that!

On the `AsyncSequence` front, just looking at the implementation that is definitely slower than needed. The usage of `AsyncStream` is not necessary. What we should do instead is the following:

- Make this generic over an `AsyncSequence` of `UInt8`
- Call `next` to get the length
- Then call `next` until we have the full message
- Then decode the message
- Rinse and repeat until the upstream returns `nil` or `throws`

At this point it is probably not as performant as we want it to be, but I am quite sure some sprinkling of `@inlinable` will get us quite close. `AsyncBytes` has been optimised to very fast and if it is not we should raise it in the Swift repository with a performance trace.

willtemperley:
@FranzBusch Is this roughly what you meant?  https://gist.github.com/willtemperley/92becd3c96d5e965b6bdf7998cd95dab

It it significantly faster (2.7 seconds down from 3.9 on my test data) than the previous async version but a long way off the buffering input-streams approach.

I think the slowdown is from a lot of unnecessary array writes and reads - every byte gets written to an array and is then read  back again when decoding the message, when the message could instead be constructed directly from the stream.

FranzBusch:
@willtemperley Yeah that is what I meant. Out of interested could you produce a quick repo that has a binary serialized messages in file and some executable that runs your code (basically whatever you currently use to benchmark). I just want to quickly run this locally and profile it to see where we actually spent time. The problem is most likely that we are just grabbing byte by byte which is bad but I wanna make sure we are not doing any executor hopping here.

willtemperley:
@FranzBusch sure: https://github.com/willtemperley/SwiftProtobufAsync

There's a single test ```testAsyncRead``` which reads an uncompressed binary protobuf file from the Resources folder. It's a dummy version of production data - nested binary messages have been replaced with random bytes but effectively equivalent to what I was testing on.

Thinking of compression, I think any API dealing with binary protobuf stream would need to support compression.

I imagine it's somehow possible to use hardware-based decompression with AsyncBytes. That coupled with a stream-based decoder, I think things would get very fast indeed.

FranzBusch:
@willtemperley Sorry for delay here. I just took a quick look at your repo. Are you running this test in release mode? I would recommend you to write an executable target instead of a test when doing the perf profiling and make sure that it is run in release mode.

willtemperley:
@FranzBusch 
I had no idea the results would be so wildly different in release mode: AsyncBytes is 16x faster.
I also decided to compare the implementation in this PR - the results were interesting:

Using AsyncMessages:
0.1692420244216919
0.16054797172546387
0.1598290205001831
0.15847301483154297
0.1629500389099121

Using StreamDecodingIterator:
0.19241297245025635
0.19839191436767578
0.19552505016326904
0.19856798648834229

Parsing directly from inputstream:
0.835036039352417
0.8314739465713501
0.8335990905761719
0.8354449272155762
0.8284729719161987

thomasvl:
What's next steps here?  Is this being closed out in favor of a new PR with the other approach?

willtemperley:
That would seem the way forward to me.

FranzBusch:
That is exactly what I expected performance wise. You can never trust performance in debug mode.

I think having an async sequence based approach would be a great value add for this library.

willtemperley:
Definitely a lesson learned for me there. Closing and I'll open another pull request for the async version.

PR #1421, opened 2023-07-17
Restructure the Protos directory.

thomasvl:
"upstream/" is now a copy of the files protocolbuffers/protobuf, this is basically moving the "google/protobuf" and "conformance" directories just down a level, but it will hopefully make things a little more clear.

All the _local_ .proto files for other purposes are now moved into sub directories that match the name of their directories in "Sources/" or "Tests/".

The places were SwiftProtobuf was using upstream files for tests, those files have now been forked. So they exist within "Protos/SwiftProtobufTests/" but they will *no* longer be updated to match upstream. This is a *good* thing as in the past there have been build breaks with upstream changed and it broke the test. Now the files are specific to this project and any edits desired can be made. NOTE: At some point the files should get updated/reduced down to only what is needed because they have always had some extra message/enums/fields that the local tests didn't need, and so the simplification would likely be good.

Lastly, a few upstream .proto files are now mirrored from "Protos/upstream/" in to the specific project level directory needed (SwiftProtobuf, Conformance, etc.). This gets the files clearly in a target directory where they are used, but does mean they have to be updated in this second location. Symlinks might have worked for this, but to avoid issues building the package on any platform (Windows?), the copies were used instead.

The "Reference/" directory was then update to mirror this new layout.


thomasvl:
Upstream is starting to land the Editions work, this will set things up so the local .proto files for tests, etc. are directly tied to the upstream state generally making things safer for us.  It will also allow us to sync back up the upstream state without fully updating some of the things we depend on until we've gotten the change to start the support needed for Editions within the plugin library target.

tbkka (approved):

tbkka:
Hmmm....  I hadn't been aware of the Editions work.  That looks like it will be fun for us to implement.  ;-)

thomasvl:
> Hmmm.... I hadn't been aware of the Editions work. That looks like it will be fun for us to implement. ;-)

Yea, I've been tracking it, I've got in on my plans for the quarter.

Issue #1066, opened 2020-09-03
Feature Request: Add dynamic as library target option


matthesoundman:
It seems to properly embed swift-protobuf inside a framework the```type: .dynamic``` needs to be set in the package manifest for the particular library style. 

It would be great if there were two library options.

thomasvl:
This sounds sorta like #1043.  It seemed like this should be a FR against Xcode because *every* `Package.swift` shouldn't have to export two products.

PR #1426, opened 2023-07-25
Add explicit Dynamic and Static product libraries

GrzegorzSpryszynski-TomTom:
SwiftPM promises it can always correctly detect what library type (dynamic or static) should be use in the project. It is the case when all dependencies are given as a source code.
This is not always possible if some of the dependencies are binary targets. In that case SPM is not able to figure out what libraries are compiled (or linked) inside the binary. In that case SPM cannot properly resolve dependencies and set proper linking type for used libraries.
Currently there is also no easy way to list those dependencies for binary targets.

In this PR we propose to add explicit targets that provide required linking:
- SwiftProtobufDynamic for dynamic linking
- SwiftProtobufStatic for static linking

Those new targets will not break current setups and allow to manually manage dependencies if SPM cannot handle them properly by itself.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -23,6 +23,16 @@ let package = Package(
         targets: ["SwiftProtobuf"]
     ),
     .library(
+        name: "SwiftProtobufDynamic",
+        type: .dynamic,
+        targets: ["SwiftProtobuf"]
+    ),
+    .library(
+        name: "SwiftProtobufStatic",
+        type: .static,
+        targets: ["SwiftProtobuf"]
+    ),
+    .library(
         name: "SwiftProtobufPluginLibrary",
         targets: ["SwiftProtobufPluginLibrary"]
     ),


GrzegorzSpryszynski-TomTom:
> What is the reason for this? Normally, SPM should decide when a target should be linked statically or dynamically.

As I mentioned in the description. When working with binary targets SPM is not able to detect it correctly.
Let say we have a library which uses SwiftProtobuf. We don't publish the library code but instead publish compiled binaries on a public repository.
If you use the library in your own project SPM is not aware that the SwiftProtobuf was used. So if you also need SwiftProtobuf in your project you will end up with two same libraries. And depending what type of linking you choose (and what type was chosen for the library), you may or may not face some issues related with that.
In that case explicit specifying dynamic linking in both places solves the problem.

FranzBusch:
This is expected behaviour and SPM intentionally does not support this right now. If you bring in a binary dependency which requires specific dependency versions to be linked into the final executable then you are in all kinds of trouble. At this point SPM cannot help you anymore.

The current recommended practice is that either:
1. Your binary framework drops all its dependencies
2. Your binary framework starts to internalise all of its dependencies via something like git submodules

We shouldn't go around and add `dynamic` and `static` targets to all packages just because someone tried to ship a binary framework that depends on SPM packages. The real solution here would be make SPM aware of dependencies for binary targets and how they should be linked. This is something worth raising over in the SPM repository.

GrzegorzSpryszynski-TomTom:
You can specify what dependencies (and in what versions) are needed to use your product. That's not a problem.

The main issue here is that linking type (static by default) is inbuilt into library and client cannot change it.
I assume that soon or later such functionality will be available in SPM. But now we have to play with what we have.

thomasvl:
https://github.com/apple/swift-protobuf/issues/1066 and https://github.com/apple/swift-protobuf/pull/1043 also provide some context for this from the past.

@tbkka fyi

GrzegorzSpryszynski-TomTom:
> ps - I also worry that some other package will explicit pick to use the static or dynamic one, and thus end up with duplicate symbols because some other package picks different, so when both get included into a single binary, you get the code twice via these different targets.

That why separate target is proposed so you have to explicit chose it. By default SPM does its magic.

In this PR the `dynamic` target is most important. `Static` one is added only for completeness. We can remove it to not encourage others to use them.

thomasvl:
> > ps - I also worry that some other package will explicit pick to use the static or dynamic one, and thus end up with duplicate symbols because some other package picks different, so when both get included into a single binary, you get the code twice via these different targets.
> 
> That why separate target is proposed so you have to explicit chose it. By default SPM does its magic.

I'm not sure I follow.  As soon as there is the option to pick, there is the risk different packages will pick different values, and then when those two packages end up in a single binary there are duplicate definitions.

GrzegorzSpryszynski-TomTom:
> I'm not sure I follow. As soon as there is the option to pick, there is the risk different packages will pick different values, and then when those two packages end up in a single binary there are duplicate definitions.

Right, there is such a risk. But without it we are forced to use static linking, ending up with increasing binary size and warnings that same method is available is multiple places and system will pick any of them randomly

FranzBusch:
> Right, there is such a risk. But without it we are forced to use static linking, ending up with increasing binary size and warnings that same method is available is multiple places and system will pick any of them randomly

I think this is just a problem with how your dependencies are currently delivered to you. We cannot influence the choice that you have a framework which depends on swift-protobuf being linked into the final executable and there is sadly not much we can do here to help you work around this. As @thomasvl said we cannot offer two products with different linkage because this is going to diverge the ecosystem.

For your immediate help what you can do is just maintain a fork of swift-protobuf where you do that and ask whoever delivered you that framework to remove the protobuf dependency or internalise it.

GrzegorzSpryszynski-TomTom:
Thanks for your input. I understand your concerns.
One more thing worth to mention. Please have a look at the official code snippets for SPM package [template](https://github.com/apple/swift-package-manager/blob/main/Documentation/PackageDescription.md#product). Looks we are not the first ones that came with such idea.

FranzBusch:
The document you linked is merely documenting the APIs that SPM offers but doesn't recommend to do this at all. I still think the problem you are having is worth solving it just not up to individual packages to solve but rather for SPM to solve this. I would open an issue over in the SPM repository for dependencies for binary targets.

GrzegorzSpryszynski-TomTom:
Thanks all of you for contribution. In this case we will try to solve the issue different way.
For now we can close the PR

Issue #1415, opened 2023-07-06
Mark SwiftProtobuf as safe for use in application extensions (set Require Only App-Extension-Safe API to YES)


tkafka:
I am using SwiftProtobuf in a custom framework (for sharing app code among app targets) that is marked with `Require Only App-Extension-Safe API = YES`, so that I can use it in WidgetKit and Notification app extensions.

During build, I am getting this warning about SwiftProtobuf not being marked as `Require Only App-Extension-Safe API = YES` (which forces compile time checks to make sure the library isn't touching eg. `UIApplication` and other objects which aren't available in app extensions).

Can this flag be added to SwiftProtobuf? I believe it isn't using any of the app objects anyway.

Thanks for consideration!

Please be sure to include:
* what OS you are developing on (Linux or macOS, including the version): 13.4.1
* for macOS, what version of Xcode you are using (`xcodebuild -version`): 15.0 beta 2
* what version of Swift is your code set to compile with (i.e. from project
  settings, etc.): 5.9
* what branch/tag of SwiftProtobuf you are using (1.0.0, etc.): 1.22.0 release
* if you are getting compile errors, please be sure include all errors/warnings,
  sometimes the error before the one you are stuck on is important.:
```
Linking against a dylib which is not safe for use in application extensions: (path hidden)/Library/Developer/Xcode/DerivedData/AppName-eksxezdcznqhaibhzaurdwsgkvku/Build/Products/Debug-iphonesimulator/PackageFrameworks/SwiftProtobuf_7CEE9AB92C12BC6B_PackageProduct.framework/SwiftProtobuf_7CEE9AB92C12BC6B_PackageProduct
```

Lukasa:
I do not believe that SwiftPM enables setting this flag.

PR #1433, opened 2023-08-02
Mark the  library safe for use in app extensions

tkafka:
According to https://stackoverflow.com/a/75928528/38729 - should fix https://github.com/apple/swift-protobuf/issues/1415

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -37,12 +37,28 @@ let package = Package(
   targets: [
     .target(
         name: "SwiftProtobuf",
-        exclude: ["CMakeLists.txt"]
-    ),
+        exclude: ["CMakeLists.txt"],
+				swiftSettings: [
+					/// giving you a nice red compile error in Xcode on the spot if you try to use some unsafe API
+					.unsafeFlags(["-Xfrontend", "-application-extension"])
+				],
+				linkerSettings: [
+					/// tells the linker 'this is an extension safe package' so you won't get warnings if you link the package to an extension target or another framework that is extension safe
+					.unsafeFlags(["-Xlinker", "-application_extension"])
+				]
+		),
     .target(
         name: "SwiftProtobufPluginLibrary",
         dependencies: ["SwiftProtobuf"],
-        exclude: ["CMakeLists.txt"]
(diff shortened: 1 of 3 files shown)


thomasvl:
Driveby 2 - Would linux (or Windows) be ok with these flags?

Package.swift
```diff
@@ -37,12 +37,28 @@ let package = Package(
   targets: [
     .target(
         name: "SwiftProtobuf",
-        exclude: ["CMakeLists.txt"]
-    ),
+        exclude: ["CMakeLists.txt"],
+				swiftSettings: [
+					/// giving you a nice red compile error in Xcode on the spot if you try to use some unsafe API
+					.unsafeFlags(["-Xfrontend", "-application-extension"])
```
FranzBusch, line 43: We can't add those flags sadly because the use of `unsafeFlags` makes packages not consumable by upstream packages/applications anymore. There are ongoing discussions to make some of the unsafe flags actually safe to use by elevating them to concrete options in the package description. cc @neonichu

FranzBusch (changes requested):

Package.swift
thomasvl: > We can't add those flags sadly because the use of `unsafeFlags` makes packages not consumable by upstream packages/applications anymore.

Sounds like this means we should go ahead and close this since we'd immediately hit problems with Swift gRPC using this and then folks trying to use that?

FranzBusch: Yes, I think we can't go forward with this PR. I would recommend @tkafka that you file a Feedback in the Apple feedback system.

tkafka: @thomasvl What kind of problems, please? It seems to me that 
1. we just tell xcode, that `swift-protobuf` isn't using some APIs (while the library users, like `Swift gRPC`, can still call them just like before)
2. and we extend the potential uses of `swift-protobuf` to application extensions (again, not changing anything for the existing users)

I am not sure what should I file the feedback to Apple about - that they added flags to let package users know that it is using a safe subset of APIs, but that we refuse to use this flag (even though it seems we safely use that subset of APIs)?

Issue #1210, opened 2022-02-08
Consider moving up to Swift 5.0 as the minimum compiler and language versions


thomasvl:
We currently require a 4.2 or later compiler and support the language version of 4.2 or later.

According to https://www.swift.org/download/, 4.2 came out with Xcode 10.0 on September 17, 2018; 4.2.1 was Xcode 10.1 on October 30, 2018.  4.2.2 & 4.2.3 were linux only, with the later being on February 28, 2019.  4.2.4 was on March 29, 2019 and is Ubuntu 16.04 only.

There is specific support for 4.2 in the library and generator, so it would simply things.  The AppStore won't accept things off that older Xcode version, but in theory , someone could be using a newer compiler but still compiling the generated code in the 4.2 mode.

[Swift Package Index](https://swiftpackageindex.com/) compatibility reports these days only mention the 5.x version, so that might be a signal that the 4.x support really can be dropped.

Thoughts?

thomasvl:
We should also revisit https://github.com/apple/swift-protobuf/labels/API%20change for things to do as part of a major version bump.

thomasvl:
Calling this fixed/done as the only conditionals left in the library and generated code are now around the 5.5 needs for `Sendable`.

PR #1434, opened 2023-08-07
Added an async sequence over binary delimited protobuf.

willtemperley:
As discussed in https://github.com/apple/swift-protobuf/pull/1414


Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,102 @@
...
+///
+// -----------------------------------------------------------------------------
+
+import Foundation
+
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+extension Message {
+  
+  /// Creates an asynchronous sequence of messages decoded from resource bytes
+  public static func asyncSequence(asyncBytes: URL.AsyncBytes) -> AsyncMessageSequence<Self> {
```
thomasvl, line 22: To fully support users, you'll need the same optional arguments as on the delimited version: https://github.com/apple/swift-protobuf/blob/main/Sources/SwiftProtobuf/BinaryDelimited.swift#L117-L119

And then they will have to be applied to where you actually create the messages.

willtemperley: Yes. This was a bit tricky because neither ExtensionMap nor BinaryDecodingOptions are Sendable. I implemented this using a factory to 1). avoid the double declaration of all the optional arguments and 2). so it could be marked @unchecked Sendable to avoid the warnings.

It seemed BinaryDecodingOptions can be made Sendable easily but then I ran into this:
https://github.com/apple/swift-protobuf/blob/main/Sources/SwiftProtobuf/MessageExtension.swift#L18

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,135 @@
...
+      asyncBytesIterator: asyncBytes.makeAsyncIterator(),
+      messageFactory: messageFactory
+    )
+  }
+}
+
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+public struct AsyncMessageSequence<M: Message> : AsyncSequence, Sendable {
+  
+  let asyncBytesIterator: URL.AsyncBytes.AsyncIterator
```
FranzBusch, line 43: This is a bit to specific for `AsyncBytes` what we want to have here is the `AsyncMessageSequence` to be generic over two things like this `AsyncMessageSequence<Base: AsyncSequence, M: Message>: AsyncSequence where Base.Element = UInt8`

Relatedly, we shouldn't store an iterator here but rather the base async sequence.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,135 @@
...
+    }
+    
+    /// Asynchronously advances to the next message and returns it, or ends the
+    /// sequence if there is no next message.
+    ///
+    /// - Returns: The next message, if it exists, or `nil` to signal the end of
+    ///   the sequence.
+    public mutating func next() async throws -> M? {
+      guard let messageSize = try await nextVarInt() else {
+        return nil
```
FranzBusch, line 84: Whenever we return early in this `next()` method we should make sure to `nil` out the `iterator` to make sure we never call `next()` on it again.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,135 @@
...
+///
+// -----------------------------------------------------------------------------
+
+import Foundation
+
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+extension Message {
+  
+  /// Creates an asynchronous sequence of messages decoded from resource bytes
+  public static func asyncSequence(
```
FranzBusch, line 22: This extension doesn't make too much sense to me. What you rather want is an extension on `AsyncSequence` it self that provides a non-static method like `decodedMessages()` (name can be bike shed)

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,135 @@
...
+    )
+  }
+}
+
+// Currently neither ExtensionMap nor BinaryDecodingOptions are Sendable
+// hence the @unchecked to suppress warnings
+fileprivate struct MessageFactory<M: Message>: @unchecked Sendable {
+  let extensions: ExtensionMap?
+  let partial: Bool
+  let options: BinaryDecodingOptions
```
FranzBusch, line 126: Yeah we really can't do this. As long as they are not `Sendable` we have should exclude them from the API. I would suggest getting rid of this type altogether for now.

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
```diff
@@ -0,0 +1,241 @@
...
+/// Tests the decoding of binary-delimited message streams, ensuring various invalid stream scenarios are
+/// handled gracefully.
+///
+// -----------------------------------------------------------------------------
+
+import Foundation
+import XCTest
+import SwiftProtobuf
+
+class Test_AsyncMessageSequence: XCTestCase {
```
FranzBusch, line 20: NIT
```suggestion
final class Test_AsyncMessageSequence: XCTestCase {
```

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
```diff
@@ -0,0 +1,241 @@
...
+    let expected: [Int32] = Array(1...5)
+    var messages = [SwiftProtoTesting_TestAllTypes]()
+    for messageNumber in expected {
+      let message = SwiftProtoTesting_TestAllTypes.with {
+        $0.optionalInt32 = messageNumber
+      }
+      messages.append(message)
+    }
+    let url = temporaryFileURL()
+    try writeMessagesToFile(url, messages: messages)
```
FranzBusch, line 33: Once your type becomes generic over the base async sequence you don't have to write to files at all in tests but can just construct the byte stream in process via something like `AsyncStream` which should work way better.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
thomasvl: It seems like adding `Sendable` to those generally would be a good thing since since just like you'd need it here, it seems like folks could need it in other places.

@Lukasa @FranzBusch @tbkka - thoughts?

thomasvl: Sorry, I thought the issue was for `ExtensionMap` and `BinaryDecodingOptions`?

FranzBusch: Feel free to add Sendable to `ExtensionMap` `AnyMessageExtension` and `BinaryDecodingOptions` in this PR. These all are fine to require `Sendable` really.

willtemperley: @thomasvl yes they're not ```Sendable``` but the warnings disappeared when I removed ```Sendable``` from  ```AsyncMessageSequence```

FranzBusch: The `AsyncMessageSequence` should really be `Sendable` though otherwise it looses a bunch of its functionality

willtemperley: I haven't been able to make `AsyncMessageSequence` sendable without a compiler warning. Storing the `Base` sequence on this, as suggested earlier, gives a sendable warning.  I made all the optional param types `Sendable` anyway.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
willtemperley: Yes. Done.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,135 @@
...
+  public typealias Element = M
+  
+  /// An asynchronous iterator that produces the messages of this asynchronous sequence
+  public struct AsyncIterator : AsyncIteratorProtocol, Sendable {
+    
+    public var iter: URL.AsyncBytes.AsyncIterator
+    fileprivate let messageFactory: MessageFactory<M>
+    
+    /// Aysnchronously reads the next varint
+    @inlinable public mutating func nextVarInt() async throws -> UInt64? {
```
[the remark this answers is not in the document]
willtemperley: Done.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
willtemperley: Done.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,135 @@
...
+  let asyncBytesIterator: URL.AsyncBytes.AsyncIterator
+  fileprivate let messageFactory: MessageFactory<M>
+  
+  /// The message type in this asynchronous sequence.
+  public typealias Element = M
+  
+  /// An asynchronous iterator that produces the messages of this asynchronous sequence
+  public struct AsyncIterator : AsyncIteratorProtocol, Sendable {
+    
+    public var iter: URL.AsyncBytes.AsyncIterator
```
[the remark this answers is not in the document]
willtemperley: Done.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
willtemperley: I removed it altogether. TBH I didn't understand what having an extension on AsyncSequence would achieve.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
willtemperley: Removed the factory in favour of a flattened version (no nested iterator) which avoids the declaration duplication.

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
willtemperley: Done

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
willtemperley: Agreed in principle, but the current binary delimited serialization is limited to writing to `OutputStream`.  AFAIK it's necessary to do something horrible like:
```
    // See https://bugs.swift.org/browse/SR-5404
    let nsData = memoryOutputStream.property(forKey: .dataWrittenToMemoryStreamKey) as! NSData
    return Data(referencing: nsData)
```
which is worse IMO. Something to revisit maybe?

Sources/SwiftProtobuf/AsyncMessageSequence.swift
```diff
@@ -0,0 +1,109 @@
...
+//
+// -----------------------------------------------------------------------------
+///
+/// An async sequence of messages decoded from a binary delimited protobuf stream.
+///
+// -----------------------------------------------------------------------------
+
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+public struct AsyncMessageSequence<Base: AsyncSequence, M: Message>:
+  AsyncSequence, AsyncIteratorProtocol where Base.Element == UInt8 {
```
FranzBusch, line 18: Please make the iterator a separate struct instead of conforming the sequence to iterator as well

Sources/SwiftProtobuf/AsyncMessageSequence.swift
```diff
@@ -0,0 +1,109 @@
...
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+///
+/// An async sequence of messages decoded from a binary delimited protobuf stream.
+///
+// -----------------------------------------------------------------------------
+
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+public struct AsyncMessageSequence<Base: AsyncSequence, M: Message>:
```
FranzBusch, line 17: We need to add a conditional conformance on `Sendable` to this otherwise it becomes hardly usable

```
extension AsyncMessageSequence: Sendable where Base: Sendable
```

Sources/SwiftProtobuf/AsyncMessageSequence.swift
```diff
@@ -0,0 +1,109 @@
...
+  AsyncSequence, AsyncIteratorProtocol where Base.Element == UInt8 {
+  
+  /// The message type in this asynchronous sequence.
+  public typealias Element = M
+  
+  var iterator: Base.AsyncIterator?
```
FranzBusch, line 23: This should be inside the `Iterator` struct

Sources/SwiftProtobuf/AsyncMessageSequence.swift
```diff
@@ -0,0 +1,109 @@
...
+  let extensions: ExtensionMap?
+  let partial: Bool
+  let options: BinaryDecodingOptions
```
FranzBusch, line 26: Please make those things `private`

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
FranzBusch: Ah interesting I thought we can write binary delimited data into `Data` as well.

FranzBusch (changes requested):

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
FranzBusch: Please add this back. The extension is super helpful because it allows you to chain algorithms together, e.g.

```
let someByteSequence = AsyncStream<UInt8>()
someByteSequence.decodeBinaryDelimitedMessages().filter { ...}.map {}
```

Sources/SwiftProtobuf/BinaryDecodingOptions.swift
```diff
@@ -13,7 +13,7 @@
 // -----------------------------------------------------------------------------
 
 /// Options for binary decoding.
-public struct BinaryDecodingOptions {
+public struct BinaryDecodingOptions: Sendable {
```
thomasvl, line 16: I believe this needed to be [`_ProtoSendable`](https://github.com/apple/swift-protobuf/blob/main/Sources/SwiftProtobuf/ProtoSendable.swift) and not `Sendable`, because we still support Swift < 5.5, so `Sendable` doesn't always exist.

Sources/SwiftProtobuf/ExtensionMap.swift
```diff
@@ -22,7 +22,7 @@
 /// This is a protocol so that developers can build their own
 /// extension handling if they need something more complex than the
 /// standard `SimpleExtensionMap` implementation.
-public protocol ExtensionMap {
+public protocol ExtensionMap: Sendable {
```
thomasvl, line 25: `_ProtoSendable`

Sources/SwiftProtobuf/MessageExtension.swift
```diff
@@ -18,7 +18,7 @@
 // TODO: `AnyMessageExtension` should require `Sendable` but we cannot do so yet without possibly breaking compatibility.
 
 /// Type-erased MessageExtension field implementation.
-public protocol AnyMessageExtension {
+public protocol AnyMessageExtension: Sendable {
```
thomasvl, line 21: `_ProtoSendable`

Sources/SwiftProtobuf/MessageExtension.swift
```diff
@@ -18,7 +18,7 @@
 // TODO: `AnyMessageExtension` should require `Sendable` but we cannot do so yet without possibly breaking compatibility.
```
thomasvl, line 18: This comment can go away, it's left from the 1.x branch.

Sources/SwiftProtobuf/MessageExtension.swift
```diff
@@ -28,7 +28,7 @@ public protocol AnyMessageExtension {
 /// A "Message Extension" relates a particular extension field to
 /// a particular message.  The generic constraints allow
 /// compile-time compatibility checks.
-public class MessageExtension<FieldType: ExtensionField, MessageType: Message>: AnyMessageExtension {
+public final class MessageExtension<FieldType: ExtensionField, MessageType: Message>: AnyMessageExtension, Sendable {
```
thomasvl, line 31: `_ProtoSendable` ?

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
```diff
@@ -0,0 +1,241 @@
...
+///
+/// Tests the decoding of binary-delimited message streams, ensuring various invalid stream scenarios are
+/// handled gracefully.
+///
+// -----------------------------------------------------------------------------
+
+import Foundation
+import XCTest
+import SwiftProtobuf
+
```
thomasvl, line 19: This files is going to need some swift version guards so the build on older Swift versions still works.

thomasvl (changes requested):

Sources/SwiftProtobuf/BinaryDecodingOptions.swift
FranzBusch: This is on `main`. What is our Swift policy for releasing the next `main` version?

tbkka: Hmmm...  I thought we had discussed changing the policy to "1 year old Swift releases" rather than the previous "1 major version".  (The two used to be nearly synonymous, but no longer are.)

FranzBusch: FWIW I would recommend the NIO way which is the last three Swift releases. However, if we do a new major we would also consider only supporting the latest minor and build up from there if there are features that we could otherwise not use.

thomasvl: We updated `main` to 5.0 - https://github.com/apple/swift-protobuf/blob/main/README.md#system-requirements

Doesn't look like we had much discussion in https://github.com/apple/swift-protobuf/issues/1210, what I can find in email didn't really talk about linking it to any date thing.

Swift Package Index seems to be currently showing the computed data for 4 version (5.9 and back), so that would also sorta slight with the last three minor Swift releases.

I'm not opposed to doing that, but with 6.x on the horizon I do wonder if by the time we see 6.2 would it be ok to drop 5.9 at that point, since we don't know how much change 6.x will be; but maybe that just means we just move the min language version up to 5.6 for now, and not officially state it will always be 3 releases?

Also, do we want to move the 1.x branch up?

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
willtemperley: @FranzBusch I may be missing something here but I feel like the extension on `Message` makes far more sense than an extension on `AsyncStream` .

With the extension on `Message`, originally suggested by @tbkka , the message type is elegantly inferred and is very clear at the call site, e.g: 

```
    let messageExtMethod = SwiftProtoTesting_TestAllTypes
      .asyncSequence(baseSequence: url.resourceBytes).filter { m in
      m.optionalInt32 > 2
    }.map { m in
      m.optionalInt32
    }
    for try await optionalInt in messageExtMethod {
      print(optionalInt)
    }
```

With an extension on AsyncSequence it gets cumbersome and actually makes method chaining more difficult. I haven't found a way to specify the message type other than:

```
    let asyncSeqExtMethod: AsyncMessageSequence<URL.AsyncBytes, SwiftProtoTesting_TestAllTypes>
    = url.resourceBytes.decodedMessages()
    let filteredInt32s = asyncSeqExtMethod.filter { m in
      m.optionalInt32 > 2
    }.map { m in
      m.optionalInt32
    }
    for try await msg in filteredInt32s {
      print(msg)
    }
 ```

(It doesn't seem possible to pass the message type as a parameter).

The `AsyncSequence` extension idea as I understood it:
```
extension AsyncSequence {
  
  public func decodedMessages<M: Message>(
    partial: Bool = false,
    options: BinaryDecodingOptions = BinaryDecodingOptions()
  ) -> AsyncMessageSequence<Self, M> {
    AsyncMessageSequence<Self, M>(
      baseSequence: self,
      extensions: extensions,
      partial: partial,
      options: options
    )
  }
}
```

File: Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift (hunk 1 above)
Comment (willtemperley) [reply Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift]: Everything is now `Sendable` or `_ProtoSendable` as applicable.

File: Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift (hunk 5 above)
Comment (willtemperley) [reply Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift]: All params are `Sendable` now.

File: Sources/SwiftProtobuf/AsyncMessageSequence.swift (hunk 10 above)
Comment (willtemperley) [reply Sources/SwiftProtobuf/AsyncMessageSequence.swift]: Done

File: Sources/SwiftProtobuf/AsyncMessageSequence.swift (hunk 11 above)
Comment (willtemperley) [reply Sources/SwiftProtobuf/AsyncMessageSequence.swift]: Done

File: Sources/SwiftProtobuf/AsyncMessageSequence.swift (hunk 13 above)
Comment (willtemperley) [reply Sources/SwiftProtobuf/AsyncMessageSequence.swift]: Done

File: Sources/SwiftProtobuf/ExtensionMap.swift (hunk 15 above)
Comment (willtemperley) [reply Sources/SwiftProtobuf/ExtensionMap.swift]: Done

File: Sources/SwiftProtobuf/MessageExtension.swift (hunk 16 above)
Comment (willtemperley) [reply Sources/SwiftProtobuf/MessageExtension.swift]: Done

File: Sources/SwiftProtobuf/MessageExtension.swift (hunk 17 above)
Comment (willtemperley) [reply Sources/SwiftProtobuf/MessageExtension.swift]: Done

File: Sources/SwiftProtobuf/MessageExtension.swift (hunk 18 above)
Comment (willtemperley) [reply Sources/SwiftProtobuf/MessageExtension.swift]: Done

File: Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift (hunk 19 above)
Comment (willtemperley) [reply Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift:19]: Done

File: Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift (hunk 4 above)
Comment (willtemperley) [reply Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift]: I've put both extensions in there, I guess we can decide which to keep.

**Comment by willtemperley** [comment]
I just noticed in a few places, e.g. https://github.com/apple/swift-protobuf/blob/main/Sources/SwiftProtobuf/BinaryDelimited.swift#L105C1-L108C55

The docs for `partial` say:
> this method will check Message.isInitialized before encoding 

but it should read "after decoding" I think.

**Comment by thomasvl** [comment]
> I just noticed in a few places, e.g. https://github.com/apple/swift-protobuf/blob/main/Sources/SwiftProtobuf/BinaryDelimited.swift#L105C1-L108C55
> 
> The docs for `partial` say:
> > this method will check Message.isInitialized before encoding 
> 
> but it should read "after decoding" I think.

You are correct, sending out https://github.com/apple/swift-protobuf/pull/1437.  Thanks!

File: Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift [hunk 20]
```diff
@@ -0,0 +1,232 @@
...
+    options: BinaryDecodingOptions = BinaryDecodingOptions()
+  ) -> AsyncMessageSequence<Base, Self> {
+    AsyncMessageSequence<Base, Self>(
+      baseSequence: baseSequence,
+      extensions: extensions,
+      partial: partial,
+      options: options
+    )
+  }
+}
```
FranzBusch, line 52: Let's remove this extension. The usability here is really questionable in my opinion. Your normally start from an async sequence and then apply transformations and that is how the whole ecosystem has been shaped up.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,232 @@
...
+  ///   - options: The BinaryDecodingOptions to use.
+  /// - Returns: An asynchronous sequence of messages read from the `AsyncSequence` of bytes.
+  /// - Throws: `BinaryDecodingError` if decoding fails, throws
+  ///           `BinaryDelimited.Error` for some reading errors,
+  ///           `BinaryDecodingError.truncated` if the stream ends before fully decoding a
+  ///           message or a delimiter,
+  ///           `BinaryDecodingError.malformedProtobuf`if a delimiter could not be read and
+  ///           `BinaryDecodingError.tooLarge` if a size delimiter of 2GB or greater is found.
+  public func decodedBinaryDelimitedMessages<M: Message>(
+    messageType: M.Type,
```
FranzBusch, line 77:
```suggestion
    of messageType: M.Type = M.self,
```

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,232 @@
...
+      options: options
+    )
+  }
+}
+
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+/// An asynchronous sequence of messages decoded from an asynchronous sequence of bytes.
+public struct AsyncMessageSequence<Base: AsyncSequence, M: Message>:
+  AsyncSequence, Sendable
+where Base: Sendable, Base.Element == UInt8 {
```
FranzBusch, line 95:
```suggestion
public struct AsyncMessageSequence<
    Base: AsyncSequence,
    M: Message
>: AsyncSequence Base.Element == UInt8 {
```

FranzBusch: Can we please conditionalise the extension on Sendable at the bottom with a

```
extension AsyncMessageSequence: Sendable where Base: Sendable
```

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,232 @@
...
+  ///   - options: The BinaryDecodingOptions to use.
+  /// - Returns: An asynchronous sequence of messages read from the `AsyncSequence` of bytes.
+  /// - Throws: `BinaryDecodingError` if decoding fails, throws
+  ///           `BinaryDelimited.Error` for some reading errors,
+  ///           `BinaryDecodingError.truncated` if the stream ends before fully decoding a
+  ///           message or a delimiter,
+  ///           `BinaryDecodingError.malformedProtobuf`if a delimiter could not be read and
+  ///           `BinaryDecodingError.tooLarge` if a size delimiter of 2GB or greater is found.
+  public init(
+    baseSequence: Base,
```
FranzBusch, line 125:
```suggestion
    base: Base,
```

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,232 @@
...
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+/// An asynchronous sequence of messages decoded from an asynchronous sequence of bytes.
+public struct AsyncMessageSequence<Base: AsyncSequence, M: Message>:
+  AsyncSequence, Sendable
+where Base: Sendable, Base.Element == UInt8 {
+  
+  /// The message type in this asynchronous sequence.
+  public typealias Element = M
+  
+  private let baseSequence: Base
```
FranzBusch, line 100:
```suggestion
  private let base: Base
```

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,232 @@
...
+  ) {
+    self.baseSequence = baseSequence
+    self.extensions = extensions
+    self.partial = partial
+    self.options = options
+  }
+  
+  /// An asynchronous iterator that produces the messages of this asynchronous sequence
+  @frozen public struct AsyncIterator: AsyncIteratorProtocol, Sendable
+  where Base.AsyncIterator: Sendable {
```
FranzBusch, line 138: This is important. We should not use `@frozen` here and async iterators must never be `Sendable`
```suggestion
  public struct AsyncIterator: AsyncIteratorProtocol {
```

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,232 @@
...
+      }
+      if messageSize == 0 {
+        return try M(
+          serializedBytes: [],
+          extensions: extensions,
+          partial: partial,
+          options: options
+        )
+      } else if messageSize > 0x7fffffff {
+        throw BinaryDecodingError.tooLarge
```
FranzBusch, line 197: We should `nil` out the iterator here

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
FranzBusch: Do we still need those guards with the new version support @thomasvl ?

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,232 @@
...
+  /// messages in the asynchronous sequence.
+  public func makeAsyncIterator() -> AsyncMessageSequence.AsyncIterator {
+    AsyncIterator(
+      iterator: baseSequence.makeAsyncIterator(),
+      extensions: extensions,
+      partial: partial,
+      options: options
+    )
+  }
+}
```
FranzBusch, line 232: Can we add an explicit `Sendable` unavailability to the iterator here please

```
@available(*, unavailable)
extension AsyncMessageSequence.Iterator: Sendable { }
```

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
```diff
@@ -0,0 +1,232 @@
...
+  ///     `BinaryEncodingError.missingRequiredFields` will be thrown.
+  ///   - options: The BinaryDecodingOptions to use.
+  /// - Returns: An asynchronous sequence of messages read from the `AsyncSequence` of bytes.
+  /// - Throws: `BinaryDecodingError` if decoding fails, throws
+  ///           `BinaryDelimited.Error` for some reading errors,
+  ///           `BinaryDecodingError.truncated` if the stream ends before fully decoding a
+  ///           message or a delimiter,
+  ///           `BinaryDecodingError.malformedProtobuf`if a delimiter could not be read and
+  ///           `BinaryDecodingError.tooLarge` if a size delimiter of 2GB or greater is found.
+  public func decodedBinaryDelimitedMessages<M: Message>(
```
FranzBusch, line 76: @thomasvl @glbrntt Do you think that name makes sense? I think it does but just want to get a few more opinions on it.

```suggestion
  public func decodedBinaryDelimitedMessages<M: Message>(
```

FranzBusch (commented):
Looks very good now. I think we are almost there! Left some last comments

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
willtemperley: Yes done. Renamed file to AsyncMessageSeqeunce.swift because Message+AsyncMessageSequence.swift didn't make sense anymore.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
willtemperley: Done.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
willtemperley: Done.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
willtemperley: Done.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
willtemperley: Done.

Sources/SwiftProtobuf/AsyncMessageSequence.swift
```diff
@@ -0,0 +1,202 @@
...
+//
+// -----------------------------------------------------------------------------
+///
+/// An async sequence of messages decoded from a binary delimited protobuf stream.
+///
+// -----------------------------------------------------------------------------
+
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+extension AsyncSequence {
+
```
FranzBusch, line 18: NIT
```suggestion
```

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
willtemperley: Done.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
willtemperley: Done. I also `nil`ed the iterator in the early returns in the `nextVarInt` function.

Sources/SwiftProtobuf/AsyncMessageSequence.swift
```diff
@@ -0,0 +1,202 @@
...
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+///
+/// An async sequence of messages decoded from a binary delimited protobuf stream.
+///
+// -----------------------------------------------------------------------------
+
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+extension AsyncSequence {
```
FranzBusch, line 17:
```suggestion
extension AsyncSequence where Element == UInt8 {
```

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
willtemperley: Done.

FranzBusch (approved):
This LGTM from my side now except two minor nits. @thomasvl can you give this another check.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
thomasvl: The only nit I can see is we usually say _parse_ not _decode_ on the public apis, so maybe `parseBinaryDelimitedMessages`?  But both names also fee wrong since it sounds a little to _active_ the api is created a new sequence to iterate on, but I'm not sure what would be the best naming for that.

FranzBusch: `binaryDelimitedMessages` works for me!

willtemperley: You're probably right it's parsing not decoding we're doing here. 

Both the [C#](https://cloud.google.com/dotnet/docs/reference/Google.Protobuf/latest/Google.Protobuf.MessageParser
) and [Java](https://github.com/protocolbuffers/protobuf/blob/023b43dacf7b0a1f6ded403a46965d277e8cbf3c/java/core/src/main/java/com/google/protobuf/Parser.java#L225) APIs use `parseDelimitedFrom(Stream)`

But that doesn't make sense here because the extension is defined on the byte sequence, not the message.

maybe
`bytes.parseDelimitedSequence(of: MessageType.self)
`
or
`bytes.parseDelimitedMessages(of: MessageType.self)
`
?

There is a [discussion](https://github.com/protocolbuffers/protobuf/issues/10229) in the protobuf repo about harmonizing binary delimited APIs.

allevato: These are Swift APIs though, so we should aim to be harmonious with [Swift's API design guidelines](https://www.swift.org/documentation/api-design-guidelines/) around naming, where in general value-returning functions return noun phrases describing the value being returned. `parse...` is an active verb phrase. (The existing `BinaryDelimited` functions in SwiftProtobuf don't obey those, unfortunately, but we should also revisit _that_ in the future.)

So a more Swift-appropriate name that better matches the C#/Java terms would be `parsedDelimitedMessages`, which reads oddly to me. I also don't think we should prefer matching other languages' naming here; the goal of protobuf is to provide—as much as one can—natural interfaces in the target language for the generated messages.

`binaryDelimitedMessages` seems like a nice compromise around not being too verbose, explaining what the function does, and fitting well with other async stream methods/properties that use the noun-phrase naming rule.

willtemperley: Is binary even necessary given the context? Protobuf is primarily a binary serialization format.

`sizeDelimitedMessages` would fit the guidelines and describe the format more specifically. The format is described as [length delimited](https://protobuf.dev/programming-guides/encoding/#length-types) in the protobuf docs but size seems more appropriate for Swift.

willtemperley:
Hmm looks like `AsyncBytes` is [in progress](https://github.com/apple/swift-corelibs-foundation/pull/3036/commits/4860800c24e16ca19c1fa54df5cfcbb8eb08a132) - I can remove this from the tests by writing to memory instead and use AsyncStream like @FranzBusch suggested earlier.

Sources/SwiftProtobuf/Message+AsyncMessageSequence.swift
tbkka: Sticking with the protobuf terminology is generally appropriate:  We're optimizing for people who either see this in SwiftProtobuf and are searching for more information or know it from some other language and are trying to use it in SwiftProtobuf.

thomasvl: > Is binary even necessary given the context? Protobuf is primarily a binary serialization format.

Protobuf (and the Swift impl included), have three possible formats: serialized binary, JSON, and TextFormat. When you look at extensions each format adds to `Message`, I think there's benefit it always including the format in the name - it helps auto complete, consistency between apis using formats, etc.

I fear someone that has seen _JSON Streaming_ apis might think `parseDelimitedMessages` follows that _standard_ of a newline between messages, and a JSON payload on each line. So calling out the _serialized_ or _binary_ for protobuf naming might be good to steer folks expectations a little better.

Having said, that, _naming_ is always one of the classic programming problems, so it comes down to balance between all the competing things.

allevato: Right, I'm not suggesting we should unnecessarily diverge from protobuf terminology, but we shouldn't do it at the expense of Swift naming conventions either. It's definitely possible to marry the two.

I'm fine with `sizeDelimitedMessages`, if nobody else objects.

tbkka: And what would we use for a hypothetical streaming JSON decoder?  (Which I can imagine us wanting to implement someday.)

willtemperley: Ok thanks for the input everyone, I think I'm going to go with `binaryDelimitedMessages` as that seems the least controversial for now.

willtemperley:
That should fix the Linux build issues

Sources/SwiftProtobuf/AsyncMessageSequence.swift
```diff
@@ -13,7 +13,8 @@
 ///
 // -----------------------------------------------------------------------------
 
-@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+#if swift(>=5.5) && canImport(_Concurrency)
+@available(macOS 10.15, *)
```
FranzBusch, line 17: This needs the full range of availability annotations i.e. iOS, tvOS and watchOS as well

willtemperley: Yes, I'll use the same annotation as on `AsyncSequence`. Can we get rid of the Swift version cards @thomasvl given what you just committed?

thomasvl: > Yes, I'll use the same annotation as on `AsyncSequence`. Can we get rid of the Swift version cards @thomasvl given what you just committed? 

Yup, I'm working on the changes now to drop the gates in the rest of the library, so you can also drop the sendable wrapper also and directly use `Sendable`.  (I was going to keep the wrapper in my PR so we weren't racing and clean it up afterwards if need be)

Sources/SwiftProtobuf/AsyncMessageSequence.swift
```diff
@@ -0,0 +1,202 @@
...
+      }
+      return nil // End of stream reached.
+    }
+    
+    /// Asynchronously advances to the next message and returns it, or ends the
+    /// sequence if there is no next message.
+    ///
+    /// - Returns: The next message, if it exists, or `nil` to signal the end of
+    ///   the sequence.
+    public mutating func next() async throws -> M? {
```
FranzBusch, line 147: We want this method to be `@inlinable` which means we have to mark a bunch of other things here as `@usableFromInline`

willtemperley: I went ahead and made `nextVarInt` `@inlinable`  too

FranzBusch (approved):
LGTM! @thomasvl @tbkka @allevato You want to give this another look?

FranzBusch:
> I'm a little unsure about nextVarInt() and next() being @inlinable vs just @usableFromInine as it seems like that could be a lot of code getting duplicated into each use case; but I'll defer to others for that call.

This should be totally fine.

thomasvl:
I'll leave things for @tbkka's signoff on the naming before we merge.

tbkka (approved):
This looks great!  Clean implementation, good ergonomics.

Looking at it all together, I'm still unsure about the naming.  If we're going to extend `AsyncSequence` (a practice I'm generally dubious about by itself), then we need to be very careful to be certain we won't interact badly with any other libraries that might be in use in the same project.

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
```diff
@@ -0,0 +1,223 @@
...
+      let message = SwiftProtoTesting_TestAllTypes.with {
+        $0.optionalInt32 = messageNumber
+      }
+      messages.append(message)
+    }
+    let serialized = try serializedMessageData(messages: messages)
+    let asyncBytes = asyncByteStream(bytes: serialized)
+    
+    // Recreate the original array
+    let decoded = asyncBytes.binaryDelimitedMessages(of: SwiftProtoTesting_TestAllTypes.self)
```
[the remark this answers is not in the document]
FranzBusch: > I'm not thrilled about extending other people's types in this way, though I agree this reads pretty nicely.

This is totally normal with `AsyncSequence`s and a pattern that `swift-async-algorithms` and the standard library follows e.g. `.map {}` on `AsyncSequence comes from the standard library.

Sources/SwiftProtobuf/AsyncMessageSequence.swift
```diff
@@ -0,0 +1,208 @@
...
+  ///   - options: The BinaryDecodingOptions to use.
+  /// - Returns: An asynchronous sequence of messages read from the `AsyncSequence` of bytes.
+  /// - Throws: `BinaryDecodingError` if decoding fails, throws
+  ///           `BinaryDelimited.Error` for some reading errors,
+  ///           `BinaryDecodingError.truncated` if the stream ends before fully decoding a
+  ///           message or a delimiter,
+  ///           `BinaryDecodingError.malformedProtobuf`if a delimiter could not be read and
+  ///           `BinaryDecodingError.tooLarge` if a size delimiter of 2GB or greater is found.
+  @inlinable
+  public func binaryDelimitedMessages<M: Message>(
```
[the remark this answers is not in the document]
FranzBusch: See my reply above. We really should put this on `AsyncSequence` but I am happy if we name this `binaryProtobufDelimitedMessages` instead.

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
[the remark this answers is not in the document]
FranzBusch: Adding to this. If this is not an extension the chainability gets completely broken. You want to be able to write this

```
rootAsyncSequence.map { ... }.binaryProtobufDelimitedMessages().map { ... }.filter { ... }
```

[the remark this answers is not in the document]
cburrows: I am with @tbkka on this, I don't think that publicly extending such a fundamental type is appropriate. Independent of how this example reads.

Sources/SwiftProtobuf/AsyncMessageSequence.swift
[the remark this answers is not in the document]
tbkka: Ok.  Let's use the longer name and leave this where it is.

It's not my personal style, but if that's what's conventional in the ecosystem, then we can stick with it.

Sources/SwiftProtobuf/AsyncMessageSequence.swift
cburrows: Also @FranzBusch would you be so kind as to elaborate on why we "want" this method to be inlinable? I ask because at first blush, I am seeing this code run a byte buffer through the underlying `iterator` byte-by-byte. So I wonder if there is any data to suggest that `@inlinable` is actually a performance win, in this case. With a baseline opinion of "inlinable is a complicated promise to make and should only be done for a well understood benefit."

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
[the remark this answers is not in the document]
FranzBusch: > I am with @tbkka on this, I don't think that publicly extending such a fundamental type is appropriate. 

This is how `AsyncSequence` chaining has been envisioned and how we have spelled this in both the standard library and `swift-async-algorithms`. This is very much the reactive chaining approach and if we would not add that extension the usability of this addition will degrade significantly.

Sources/SwiftProtobuf/AsyncMessageSequence.swift
FranzBusch: This has been a learning in the swift on server ecosystem. Basically any type that has a generic parameter benefits heavily from `@inlinable` on its method since the compiler is going to specialise the generic code across module boundaries. Of course, this often should be proven with benchmarking code but we have learned over the years that exactly this pattern is benefiting heavily from `@inlinable` annotations.

The good thing is that we can always remove the `@inlinable` annotation again without breaking any API.

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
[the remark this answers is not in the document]
cburrows: In the case of the standard library (which I believe introduces `AsyncSequence` to begin with) and the async algorithms package, the _primary subject_ under consideration is the sequence and the way it composes. The methods that those modules introduce provide abstract and generic functionality over the sequence.

In our case, the subject of this module is not the sequence, we do not define or own the type, and neither is this extensions abstract and generic.

I fully appreciate the syntactic value of a method provided by an extension, and I love introducing nice composability. But I don't think that it is fair to say polluting the namespace of the members of a public type is what anyone has envisioned. I also don't think usability will degrade significantly--at most it will degrade slightly. And I also think it's not our job here to fix this particular defect in the ecosystem. It would be more proper for a new abstract chunking combinator (that could handle the varint delimiters and reduce the bytes) or even something like the F# pipe operator, which is specifically amazing at fixing the ordering of these things in syntax.

As a thought exercise, if we were to later add this functionality for decoding JSON messages, do you think anyone would object to putting a method on `String` called `jsonProtobufDelimitedMessages()`?

Sources/SwiftProtobuf/AsyncMessageSequence.swift
tbkka: I would love to see some measurements to justify whether the increased code size is warranted in this particular case.  It's not obvious to me either way.

But I don't think this should block landing this particular PR.  So I'm going to mark this "resolved" for now.

Sources/SwiftProtobuf/AsyncMessageSequence.swift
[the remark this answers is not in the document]
tbkka: @willtemperley Please lengthen the name some more, resolve the one outstanding conflict that seems to have popped up, and then I think we're ready to merge this.  Thanks so much for your hard work on this!

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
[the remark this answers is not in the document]
tbkka: I'm okay with this as long as we adopt the more specific name.  Resolving this for now so we can land this PR.

tbkka (approved):
Looks great!  I think we should merge this, and debate the remaining issues as follow-up PRs.

thomasvl:
Thanks again for the PR!  Landing this now.

PR #1439, opened 2023-08-15
Start removing support for Swift < 5.6.

thomasvl:
NOTE: Swift 5.7 seems to be when `Data` got marked as `Sendable`, as a result
things still have to be marked as `@unchecked Sendable` until 5.7 (or later) is
the minimum supported version.

- Remove older Package@*.swift files.
- Update core library code to remove conditionals.


thomasvl:
I'll do the OneOf changes generator changes after this, figured I'd make sure I had everything right with the `@unchecked` first in something a little smaller.

FranzBusch (commented):
Can we enable strict concurrency checking in the CI as well with this

```
            swiftSettings: [
                .enableExperimentalFeature("StrictConcurrency=complete"),
            ]
```

We should also make sure `warnings-as-error` is enabled (probably is just double checking)

FranzBusch (approved):

thomasvl:
> Can we enable strict concurrency checking in the CI as well with this
>
> ```
>             swiftSettings: [
>                 .enableExperimentalFeature("StrictConcurrency=complete"),
>             ]
> ```

Done.

@FranzBusch please take another look.

thomasvl:
> We should pass it to swift test in our CI invocation. We do this for all our NIO repos

Ok, I think I've got this wired up now.

FranzBusch (approved):
Nice! Thanks

PR #1440, opened 2023-08-16
Add `Sendable` to more places that folks might want it.

thomasvl:
The covers all places with the TODO comments from the first Sendable work.

Also ensure all the Decoding/Encoding Options types are marked Sendable.


thomasvl:
I've dropped some of the attempts at making some of the extension related parts `Sendable` as the `Data` issue wasn't making it hard to sort out some of the issues.

allevato (approved):

PR #1443, opened 2023-08-18
Fix docs errors (1.x branch)

thomasvl:
There were noticed in https://github.com/apple/swift-protobuf/pull/1434#issuecomment-1674608422.

### DIFF
--- a/Sources/SwiftProtobuf/BinaryDelimited.swift
+++ b/Sources/SwiftProtobuf/BinaryDelimited.swift
@@ -141,7 +141,7 @@ public enum BinaryDelimited {
   ///     extensions in this message or messages nested within this message's
   ///     fields.
   ///   - partial: If `false` (the default), this method will check
-  ///     `Message.isInitialized` before decoding to verify that all required
+  ///     `Message.isInitialized` after decoding to verify that all required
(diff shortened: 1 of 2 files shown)


Sources/SwiftProtobuf/BinaryDelimited.swift
```diff
@@ -99,9 +99,9 @@ public enum BinaryDelimited {
   ///     extensions in this message or messages nested within this message's
   ///     fields.
   ///   - partial: If `false` (the default), this method will check
-  ///     `Message.isInitialized` before encoding to verify that all required
+  ///     `Message.isInitialized` after decoding to verify that all required
```
allevato, line 102: You changed this one to "after decoding", but the others below just changed "decoding" so they're "before decoding". Should those also be "after"?

thomasvl: Ugh, just created #1445 to hopefully get more things correct on `main`, once that lands I'll cherry-pick that into this.

thomasvl: Second cherry-pick added.

allevato (approved):

PR #1444, opened 2023-08-18
Make the Extension related types Sendable.

thomasvl:


thomasvl:
@FranzBusch @allevato @tbkka @Lukasa do you all see any way to get this to work for Swift 5.6 since `Data` wasn't `Sendable`?  I'm coming up short on how do this.  The only other idea I have to to add the `Sendable` conformance only for Swift ≥5.7 and dodge the whole issue.

FranzBusch:
For 5.6 you can conditionally mark it as `@preconcurrency`. You just gotta figure out the right dance of `#if swift` and `#if canImport(Darwin)`.

thomasvl:
Now that the min is 5.7, this should be good to go.

FranzBusch (approved):

allevato (approved):

PR #1447, opened 2023-08-18
Add FuzzAsyncMessageSequence to test the new parsing api.

thomasvl:


thomasvl:
@FranzBusch @Lukasa @tbkka –

1. I'm open to a better way to write the test then the current way.  (I have done much `async`/`await` work.
2. The failure sorta looks like we what address on the Delimited `InputStream` approach in #1382, but I don't immediately follow what is causing it to not pick off the large size or if we're getting way more allocations in small parts that expected.

thomasvl:
@willtemperley fyi

thomasvl:
The chunking was added in #1382 if folks want something smaller to compare too.

FranzBusch:
> Should we move that also to a chunked approach instead?

I think it makes sense to read in chunks with a limit of the size of the chunks so that we avoid a truncation attack here. For the small message case we should still only need a single allocation.

thomasvl:
@FranzBusch added the chunking, can you take another look?

FranzBusch (approved):

Issue #1266, opened 2022-08-10
Concurrency warning about allCases in the generated file 


leuski-ict:
* what OS you are developing on (Linux or macOS, including the version)

macOS 12.5

* for macOS, what version of Xcode you are using (`xcodebuild -version`),
  for Linux, what version of Swift (`swift --version`)

XCode 14.0 beta 4

* what version of Swift is your code set to compile with (i.e. from project
  settings, etc.)

Swift 5

* what branch/tag of SwiftProtobuf you are using (1.0.0, etc.)

1.19.0

* if you are getting compile errors, please be sure include all errors/warnings,
  sometimes the error before the one you are stuck on is important.

warning: reference to static property 'allCases' is not concurrency-safe because it involves shared mutable state

* lastly, if it all possible, provide a snippet of `.proto` and/or source
  that shows the problem.

```
#if swift(>=4.2)

extension Module_MyEnum: CaseIterable {
  // The compiler won't synthesize support with the UNRECOGNIZED case.
  public static var allCases: [Module_MyEnum] = [
    .case1,
    .case2,
  ]
}

#endif  // swift(>=4.2)
```

Lukasa:
Yeah, this seems like something the codegen should fix up: `allCases` has no need to be `var`, it can be `let`.

Lukasa:
(Note that changing this in the codegen is _technically_ breaking, so the other option is to protect it with a lock. That said, if anyone is mutating this something horrible is happening anyway.)

thomasvl:
For `main`, the _break_ is fine since we're working toward 2.0. (I've got a PR about to go up).

Should we pull this back to 1.x also?

PR #1451, opened 2023-08-31
Fix allCases to generate a let

gjcairo:
## Description

This commit cherry-picks [Fix allCases to generate a let](https://github.com/apple/swift-protobuf/commit/920e28b3191cf8f5706fafa89cef03bec6738d1a) onto the 1.x release branch, which fixed https://github.com/apple/swift-protobuf/issues/1266 

## Result

Enums `allCases` property is now a `let` instead of a `var`.


thomasvl:
I believe you also need to regenerate the reference directory.

gjcairo:
Oops, fixed

Issue #151, opened 2016-12-01
Support deprecated annotation on messages, fields, enums, enum values.


thomasvl:
descriptor.proto defines a deprecated bool on message, enum, enum value, field; the generated things should get the Swift annotations also.

note: Extensions should honor it since they are a field.

thomasvl:
Ok, this one is turning into a complex problem.  ☹️ 

The annotations can go on the file, messages, enum, enum values, and fields.  Spitting out the `@availablity()` line is easy enough.  The hard part is then the rest of the code generated and what happens when it compiles.

All the other code we generate (for `isEqual`, `hash`, etc. can run into trying to access things that are now tagged as deprecated and the compiler spits out warnings.  For C++/ObjC/etc. we work around this by using `#pragma` to push/pop the warnings on/off in the code we generate so the code will compiled silently, but the usage by developers will get the warnings.  But I don't see a way to do something like this.

For the fields, I can cheat and always ensure a deprecated field gets a _\_name_ private field that all the code uses and only the public _var name` gets tagged as deprecated.

But when it comes to tagging enums and structs for messages, I don't see any way to cheat, they can be deprecated but still used in a deprecated field of another message also; and so that seems like it will always generate warnings when the code is compiled.

It doesn't seem right to say the generated code won't compile warning free, so I'm not sure how we support deprecated…

thomasvl:
Un-assigning this at the moment.  I think we need to support this, but if also seems like Swift doesn't have the support to do this correctly for Message/Enum/Enum Values, and if we generate code that doesn't compile cleanly, it will probably result in bugs against this project.

allevato:
I've filed [SR-3357](https://bugs.swift.org/browse/SR-3357) to start a discussion about easing up on deprecation warnings originating from references within the same compilation unit or module, to see if the language itself can help here.

thomasvl:
Moving this to an enhancement, given the current support in Swift, it does't seem like we can do this and have the generate code compile (cleanly), so until that support is improved, we likely have to put this off.

SebastianThiebaud:
Any update on this or the limitations remain?

thomasvl:
[SR-3357](https://bugs.swift.org/browse/SR-3357) doesn't have any updates, so without it, we wouldn't be able to generate something that would actually compile cleanly in the first place.

tbkka:
Any updates?    Since #861 is opening the door for a possible major version bump, we should consider whether it's time to do this.

[...]

PR #1455, opened 2023-09-08
Add comments for deprecation information

thomasvl:
From @DanF-Go's comment on #151, relate the deprecated information into docstring comments so developers have a chance to notice the information in dies.


thomasvl:
@tbkka thoughts?  (see the referenced issue for some more context)

tbkka:
No thoughts, to be honest.  Comment seems the best we can do here, so it's probably worth doing until the language can clean up the deprecation handling.

allevato (approved):

PR #1459, opened 2023-09-20
Add Swift 5.9 to CI.

thomasvl:


thomasvl:
@FranzBusch @Lukasa @tbkka - Looks like swift-docc-plugin doesn't compile cleanly with 5.9, did you all hit this with any other projects?  Know if they have plans to address it?

thomasvl:
And I guess we need a annotation for our dispatch queue usage as the compiler considers that unsafe.

thomasvl:
> @FranzBusch @Lukasa @tbkka - Looks like swift-docc-plugin doesn't compile cleanly with 5.9, did you all hit this with any other projects? Know if they have plans to address it?

Opened https://github.com/apple/swift-docc-plugin/issues/65

thomasvl:
#1462 should avoid the last of the issues, will sync this PR up after that lands.

PR #1465, opened 2023-09-22
Remove some '@unchecked' usages.

thomasvl:
There were needed because of `Data` and Swift 5.6, but with 5.7 as a minimum they can go away.


thomasvl:
Was it not 5.7 that made `Data` sendable?  https://developer.apple.com/documentation/foundation/data says it conforms, but I don't see anything to indicate when it was added, I had thought it was 5.7 give some other things we'd see in the past.

thomasvl:
I went back to Xcode 14.3.1 (swiftlang-5.8.0.124.5), and I can't get this failure.

Was `Data` not correctly marked `Sendable` in the Linux version of Foundation for both 5.8 an d 5.7?

PR #1469, opened 2023-09-25
Block more things

thomasvl:
So if one tries to move say `Sendable` off the generated `extension [MessageName]` and declare it when the `struct` is declare, then things go comically wrong as the compiler decides we're trying to say all the other `struct`s try to use the local version of _Sendable_ which is another `struct`. And trying to scope it as `Swift.Sendable` also fails, because the locally defined `Swift` gets in the way.

I guess no one has ever really hit this in the field, but it seems better to play it safe and force any attempt to use these as type should get them remapped to avoid the potential confusion they will create.


tbkka (approved):
Good catch!  FWIW, I think this is fine for 1.x.

PR #1471, opened 2023-09-25
Sendable updates

thomasvl:


tbkka (approved):

allevato (approved):
I'm still trying to improve my understanding of Swift concurrency, so seeing the `@unchecked` vs. regular `Sendable` generation got me thinking. Ignoring the unfortunate situation with `Data` on older versions of non-Darwin Foundation, should we be declaring the storage class as `Sendable` as well to avoid using `@unchecked` there? I'm guessing that the primitives underlying `uniqueStorage()` would do the right thing for multi-threaded mutation, since they're the same primitives used by the stdlib types?

FranzBusch:
> I'm guessing that the primitives underlying `uniqueStorage()` would do the right thing for multi-threaded mutation, since they're the same primitives used by the stdlib types?

The only that we get the compiler to understand that a CoW type is actually `Sendable` is if the compiler would generate the code for us. Or if we would re-wrap the underlying storage with a type that does the `uniqueStorage` and we only need to make that type `@unchecked Sendable`.

In general, I am a big opponent of adding `@unchecked Sendable` to types and really only want to see it with a conditional dependency but here in this PR it makes sense since our generator can be make sure that those types really are `Sendable`.

FranzBusch (approved):

PR #1472, opened 2023-09-26
Add support for `package` visibility modifier

gjcairo:
This PR adds support for the new `package` visibility modifier, [introduced with Swift 5.9](https://github.com/apple/swift-package-manager/blob/main/CHANGELOG.md#swift-59).
It also adds support for Swift 5.9, by updating the Package.swift's `swift-tools-version`t o 5.9 (required to use `package`) and moving the existing Package.swift to Package@swift-5.8.swift.


Sources/protoc-gen-swift/GeneratorOptions.swift
```diff
@@ -33,13 +33,20 @@ class GeneratorOptions {
   enum Visibility {
     case `internal`
     case `public`
+    #if swift(>=5.9)
```
FranzBusch, line 36: Do we really need the `#if` here? We can just leave them out and if the user generates them their code will fail to compile.

FranzBusch (changes requested):
We also need to extend the `Plugin.swift` to support this.

Sources/protoc-gen-swift/GeneratorOptions.swift
gjcairo: Yeah okay, good point, this doesn't really safeguard the user from anything - I got rid of them.

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.8
+// swift-tools-version:5.9
```
thomasvl, line 1: Is this needed?

Package@swift-5.7.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.6
+// swift-tools-version:5.7
```
thomasvl, line 1: Same question.

Package.swift
gjcairo: Yes - without this, required compiler flags for understanding the `package` visibility modifier aren't set, and the build will fail.

Package@swift-5.7.swift
gjcairo: This one isn't needed, I just thought I'd align the Swift version with the swift-tools version. Is there a reason why you think this shouldn't change?

Package.swift
thomasvl: But this package doesn't use _package_ visibility, all we're doing is optionally changing part of what we generate.

thomasvl: Or does this value ripple in some way because of the plugin?  i.e. - could someone not use the plugin generating _package_ visibility, because this value come into play in their Package.swift instead of their locally specified version?

gjcairo: Oh yes you're 100% right, we're not adding the `package` modifier to anything in this package. We should be fine then, I don't think anything will ripple as you suggest - I'll revert all this.

FranzBusch (approved):
LGTM!

thomasvl (approved):

Issue #1477, opened 2023-10-13
Deterministic JSON serialization for `map` types


rebello95:
## Problem

It's not currently possible to deterministically produce JSON data for types that include Protobuf `map` fields since the keys in those maps are not sorted. For example:

```swift
for _ in 0..<10 {
    var message = Some_Message.with { message in
        message.other = [
            "aaa": "abc",
            "111": "abc",
            "eee": "baz",
            "bbb": "baz",
            "ccc": "baz",
            "ddd": "baz",
        ]
    }
    message.other["0new"] = "abc"
    print(String(data: try message.jsonUTF8Data(), encoding: .utf8)!)
}
```

Outputs:

```
{"other":{"ddd":"baz","ccc":"baz","eee":"baz","111":"abc","0new":"abc","aaa":"abc","bbb":"baz"}}
{"other":{"111":"abc","bbb":"baz","ddd":"baz","0new":"abc","aaa":"abc","ccc":"baz","eee":"baz"}}
{"other":{"ddd":"baz","ccc":"baz","eee":"baz","111":"abc","0new":"abc","aaa":"abc","bbb":"baz"}}
{"other":{"111":"abc","bbb":"baz","ddd":"baz","0new":"abc","aaa":"abc","ccc":"baz","eee":"baz"}}
{"other":{"ddd":"baz","ccc":"baz","eee":"baz","111":"abc","0new":"abc","aaa":"abc","bbb":"baz"}}
{"other":{"111":"abc","bbb":"baz","ddd":"baz","0new":"abc","aaa":"abc","ccc":"baz","eee":"baz"}}
{"other":{"ddd":"baz","ccc":"baz","eee":"baz","111":"abc","0new":"abc","aaa":"abc","bbb":"baz"}}
{"other":{"111":"abc","bbb":"baz","ddd":"baz","0new":"abc","aaa":"abc","ccc":"baz","eee":"baz"}}
{"other":{"ddd":"baz","ccc":"baz","eee":"baz","111":"abc","0new":"abc","aaa":"abc","bbb":"baz"}}
{"other":{"111":"abc","bbb":"baz","ddd":"baz","0new":"abc","aaa":"abc","ccc":"baz","eee":"baz"}}
```

## Use case

[...]

PR #1478, opened 2023-10-13
Add `useDeterministicOrdering` to `JSONEncodingOptions`

rebello95:
Adds an option to ensure that JSON serialization is deterministic when serializing Protobuf `map` fields. This should be the only type that needs to be sorted, since individual fields are serialized in order by the generated code that invokes `try visitor.visit(...)` for each field.

In order to avoid the overhead of sorting for the default case, there are no changes to the existing serialization strategy when this setting is disabled.

This new option is similar to what [Protobuf provides in Go](https://pkg.go.dev/github.com/golang/protobuf/proto#Buffer.SetDeterministic), and the docstring added in this PR matches that.

Resolves https://github.com/apple/swift-protobuf/issues/1477


Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
```diff
@@ -590,26 +590,26 @@ internal struct TextFormatEncodingVisitor: Visitor {
...
 
   /// Helper to encapsulate the common structure of iterating over a map
   /// and encoding the keys and values.
-  private mutating func _visitMap<K, V>(
+  private mutating func iterateAndEncode<K, V>(
     map: Dictionary<K, V>,
     fieldNumber: Int,
     isOrderedBefore: (K, K) -> Bool,
-    coder: (inout TextFormatEncodingVisitor, K, V) throws -> ()
+    encode: (inout TextFormatEncodingVisitor, K, V) throws -> ()
```
rebello95, line 612: Renamed to match the `encode` parameter in the function in this file above

Sources/SwiftProtobuf/TextFormatEncodingVisitor.swift
```diff
@@ -590,26 +590,26 @@ internal struct TextFormatEncodingVisitor: Visitor {
...
+    try iterateAndEncode(value: value, fieldNumber: fieldNumber) {
       (v: E, encoder: inout TextFormatEncoder) in
       encoder.putEnumValue(value: v)
     }
   }
 
   /// Helper to encapsulate the common structure of iterating over a map
   /// and encoding the keys and values.
-  private mutating func _visitMap<K, V>(
+  private mutating func iterateAndEncode<K, V>(
```
rebello95, line 608: Updated this and the function above to remove the underscore which doesn't not seem to be commonly used for `private` functions in the library

rebello95:
@thomasvl would you mind taking a look at this? 🙏🏽

thomasvl:
> @thomasvl would you mind taking a look at this? 🙏🏽

Sorry, been OOO for a bit, seems like these PR and issue got some discussion.  Will catch up and hopefully some of the Apple folks will take a look.

If these PRs do land, there's also the question of it they can/should go back to the 1.x branch/release or just sit awaiting a 2.0 release.

Sources/SwiftProtobuf/JSONEncodingOptions.swift
```diff
@@ -27,5 +27,19 @@ public struct JSONEncodingOptions: Sendable {
   /// By default they are converted to JSON(lowerCamelCase) names.
   public var preserveProtoFieldNames: Bool = false
 
+  /// Whether to use deterministic ordering when serializing.
+  ///
+  /// Note that the deterministic serialization is NOT canonical across languages. 
+  /// It is not guaranteed to remain stable over time. It is unstable across
```
thomasvl, line 33: s/not/NOT/ (just to emphasis things the same as the previous sentence)

rebello95:
> Sorry, been OOO for a bit, seems like these PR and issue got some discussion. Will catch up and hopefully some of the Apple folks will take a look.

Thanks @thomasvl.

> If these PRs do land, there's also the question of it they can/should go back to the 1.x branch/release or just sit awaiting a 2.0 release.

These are all backward-compatible and opt-in, so I think they should be okay to include in 1.x (not sure what the release plan for 2.0 is).

thomasvl:
> > If these PRs do land, there's also the question of it they can/should go back to the 1.x branch/release or just sit awaiting a 2.0 release.
> 
> These are all backward-compatible and opt-in, so I think they should be okay to include in 1.x (not sure what the release plan for 2.0 is).

Since it is a new Option, it is sorta an ABI change, but I think the past stance has been these sorta things are ok.  That's where Apple (as repo owner) gets the final call.

Sources/SwiftProtobuf/JSONEncodingOptions.swift
rebello95: Updated

FranzBusch:
> Since it is a new Option, it is sorta an ABI change, but I think the past stance has been these sorta things are ok. That's where Apple (as repo owner) gets the final call.

IMO this is fine to add from just a pure API/ABI compatibility perspective. We are just adding a new field to a non-frozen struct. 

On the overall feature, I am +1. If other languages are providing a similar option I don't see why we shouldn't and the documentation calls out that this is unstable and break over versions.

tbkka:
For the record, I'm happy with the intent here.

Adding a new field to a non-frozen struct is forward-compatible API and ABI.  (Old client source/binaries work correctly with new library.) So back-porting to 1.x should be fine.

  I'll do another skim over the source to see if I have any last-minute nits...

tbkka (approved):

thomasvl (approved):

thomasvl:
Thank you!

rebello95:
Thank you both!

PR #1480, opened 2023-10-19
Add `BinaryEncodingOptions` with option for deterministic ordering

rebello95:
Implements the same setting added for JSON in https://github.com/apple/swift-protobuf/pull/1478, but for binary serialization.

Related to https://github.com/apple/swift-protobuf/issues/1477.

### DIFF
--- a/Sources/SwiftProtobuf/BinaryEncodingOptions.swift
+++ b/Sources/SwiftProtobuf/BinaryEncodingOptions.swift
@@ -1,4 +1,4 @@
-// Sources/SwiftProtobuf/BinaryEncodingOptions.swift - Binary decoding options
+// Sources/SwiftProtobuf/BinaryEncodingOptions.swift - Binary encoding options
 //
 // Copyright (c) 2014 - 2017 Apple Inc. and the project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
@@ -8,7 +8,7 @@
 //
 // -----------------------------------------------------------------------------
 ///
-/// Binary decoding options
+/// Binary encoding options
 ///
 // -----------------------------------------------------------------------------
 
--- a/Sources/SwiftProtobuf/BinaryEncodingOptions.swift
+++ b/Sources/SwiftProtobuf/BinaryEncodingOptions.swift
@@ -1,4 +1,4 @@
-// Sources/SwiftProtobuf/BinaryDecodingOptions.swift - Binary decoding options
+// Sources/SwiftProtobuf/BinaryEncodingOptions.swift - Binary decoding options
 //
 // Copyright (c) 2014 - 2017 Apple Inc. and the project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
--- a/Sources/SwiftProtobuf/BinaryEncodingOptions.swift
+++ b/Sources/SwiftProtobuf/BinaryEncodingOptions.swift
@@ -1,0 +1,32 @@
+// Sources/SwiftProtobuf/BinaryDecodingOptions.swift - Binary decoding options
+//
+// Copyright (c) 2014 - 2017 Apple Inc. and the project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+///
+/// Binary decoding options
+///
+// -----------------------------------------------------------------------------
+
+/// Options for binary encoding.
+public struct BinaryEncodingOptions: Sendable {
+    /// Whether to use deterministic ordering when serializing.
+    ///
+    /// Note that the deterministic serialization is NOT canonical across languages.
+    /// It is not guaranteed to remain stable over time. It is unstable across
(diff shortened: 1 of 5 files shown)


Sources/SwiftProtobuf/BinaryEncodingOptions.swift
```diff
@@ -0,0 +1,32 @@
+// Sources/SwiftProtobuf/BinaryEncodingOptions.swift - Binary encoding options
+//
+// Copyright (c) 2014 - 2017 Apple Inc. and the project authors
```
thomasvl, line 3: This should probably be 2014-2023 to get the current year?

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -22,13 +22,17 @@ extension Message {
...
   ///     `BinaryEncodingError.missingRequiredFields`.
+  ///   - options: The `BinaryEncodingOptions` to use.
   /// - Returns: A `SwiftProtobufContiguousBytes` instance containing the binary serialization
   /// of the message.
   ///
   /// - Throws: `BinaryEncodingError` if encoding fails.
-  public func serializedBytes<Bytes: SwiftProtobufContiguousBytes>(partial: Bool = false) throws -> Bytes {
+  public func serializedBytes<Bytes: SwiftProtobufContiguousBytes>(
+    partial: Bool = false,
+    options: BinaryEncodingOptions = BinaryEncodingOptions()
```
thomasvl, line 34: I think for 1.x (if we pull this back), we'd have to have a method with and without this option rather than defaulting it.  Atleast I think that's why we have in some other things on the 1.x branch.

Sources/SwiftProtobuf/BinaryEncodingVisitor.swift
```diff
@@ -338,7 +370,9 @@ extension BinaryEncodingVisitor {
       let length = try value.serializedDataSize()
       encoder.putVarInt(value: length)
       // Create the sub encoder after writing the length.
-      var subVisitor = BinaryEncodingVisitor(forWritingInto: encoder.pointer)
+      var subVisitor = BinaryEncodingVisitor(
+        forWritingInto: encoder.pointer, options: BinaryEncodingOptions()
```
thomasvl, line 374: This seems wrong, don't you need the correct options or else the submessage here won't be able to honor the requested options?

Sources/SwiftProtobuf/BinaryEncodingOptions.swift
rebello95: 👍🏽 copied from another file (files don't seem to be consistent), but will update

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
rebello95: Not opposed to this, but I'm curious as to what this would break considering the function is not overrideable or part of a protocol?

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -48,7 +52,7 @@ extension Message {
     var data = Bytes(repeating: 0, count: requiredSize)
```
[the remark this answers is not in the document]
rebello95: I can add an explicit note

Sources/SwiftProtobuf/BinaryEncodingVisitor.swift
rebello95: Yep, not sure why I missed this originally - updated.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
FranzBusch: This is a public method so changing its signature is an API breaking change. Users might be holding a partially applied call to this method and their signature will change and potentially break their code.

Tests/SwiftProtobufTests/Test_BinaryEncodingOptions.swift
```diff
@@ -0,0 +1,74 @@
...
+///
+/// Test for the use of BinaryEncodingOptions
+///
+// -----------------------------------------------------------------------------
+
+import Foundation
+import XCTest
+import SwiftProtobuf
+
+class Test_BinaryEncodingOptions: XCTestCase {
```
FranzBusch, line 19:
```suggestion
final class Test_BinaryEncodingOptions: XCTestCase {
```

rebello95: Agreed, and updated. I had left it without `final` before because it looks like the rest of the test case classes are missing the keyword, so I figured that was intentional

thomasvl (commented):
Seems good to me, pulling in some Apple folks to get signoff.

tbkka (approved):

thomasvl (approved):

thomasvl:
Thank you!

Tests/SwiftProtobufTests/Test_BinaryEncodingOptions.swift
thomasvl: Just put up #1486 for this.

rebello95:
Thank you both!

PR #1483, opened 2023-10-25
Better plugin support

thomasvl:
This sequence builds more baseline functionallity into the plugin support library so when someone does author a plugin there is less common info they have to do. The hope is this provides something easy for swift-grpc (or anything else) to migrate to so when the support for Protobuf Editions is rolled in they don't have to duplicate anything and it will be more "automatic" like what the C++ plugin interfaces provide.


thomasvl:
PS - there are some other things (like output path calcs, etc.) that happy to be copied between swift-grpc and swift-protobuf, so maybe those things should also move into the shared location, but since they are strictly things all plugins would need, I resisted doing things in that space in this sequence.

allevato (approved):
🥳

PR #1485, opened 2023-10-26
1.x branch Better Plugin Support

thomasvl:
This is #1483, but for the 1.x release branch

There are two minor changes:
1. It leaves out the move to `@main`
2. Small code edit around `serializeBytes` vs. `serializedData`.


thomasvl:
The build failures on upstream protobuf are known and I think they are working to resolve.

allevato (approved):

thomasvl:
The 5.0.3 and 4.2.4 breaks confuse me, I didn't even change those files.  I wonder is something just happened to the swift docker images?

thomasvl:
> The 5.0.3 and 4.2.4 breaks confuse me, I didn't even change those files. I wonder is something just happened to the swift docker images?

Oh, those are just _notes_, they must always be there. But I am using another api that I guess is an issue.

thomasvl:
CI passes now with the fix for FileHandle on Swift <5.1 (and upstream protobuf having been fixed)

PR #1487, opened 2023-10-26
Add `useDeterministicOrdering` opt-in option to JSON + Binary encoding

rebello95:
Cherry-picks #1478 and #1480 from `main` onto the 1.x branch. Includes various fixes because these were not clean cherry picks.

### DIFF
--- a/Sources/SwiftProtobuf/JSONEncodingVisitor.swift
+++ b/Sources/SwiftProtobuf/JSONEncodingVisitor.swift
@@ -374,7 +374,7 @@ internal struct JSONEncodingVisitor: Visitor {
   ) throws {
     try startField(for: fieldNumber)
     encoder.append(text: "{")
-    var mapVisitor = JSONMapEncodingVisitor(encoder: JSONEncoder(), options: options)
+    var mapVisitor = JSONMapEncodingVisitor(encoder: encoder, options: options)
     if options.useDeterministicOrdering {
       for (k,v) in map.sorted(by: { isOrderedBefore( $0.0, $1.0) }) {
         try encode(&mapVisitor, k, v)
--- a/Sources/SwiftProtobuf/JSONEncodingVisitor.swift
+++ b/Sources/SwiftProtobuf/JSONEncodingVisitor.swift
@@ -340,39 +340,49 @@ internal struct JSONEncodingVisitor: Visitor {
   // Packed fields are handled the same as non-packed fields, so JSON just
   // relies on the default implementations in Visitor.swift
 
-
-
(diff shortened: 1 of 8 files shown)


rebello95:
I don't understand the test failure - looks like `make generate-linux-main` isn't including one of the new test cases? @thomasvl

Sources/SwiftProtobuf/JSONEncodingVisitor.swift
```diff
@@ -374,7 +374,7 @@ internal struct JSONEncodingVisitor: Visitor {
   ) throws {
     try startField(for: fieldNumber)
     encoder.append(text: "{")
-    var mapVisitor = JSONMapEncodingVisitor(encoder: JSONEncoder(), options: options)
```
thomasvl: Is this different than what was on `main`?

thomasvl:
> I don't understand the test failure - looks like `make generate-linux-main` isn't including one of the new test cases? @thomasvl

@tbkka wrote the test case collector we used back when SwiftPM didn't have native support to do the work for Linux and can hopefully provide some info.

My guess is [this line](https://github.com/apple/swift-protobuf/blob/1_x_release_branch/DevTools/CollectTests.awk#L46) doesn't deal with the `final class` usage, try switching to just `class`.

rebello95:
> My guess is [this line](https://github.com/apple/swift-protobuf/blob/1_x_release_branch/DevTools/CollectTests.awk#L46) doesn't deal with the final class usage, try switching to just class.

Ah yea that makes sense. I'll remove `final` for now

Sources/SwiftProtobuf/JSONEncodingVisitor.swift
rebello95: `main` uses a new `JSONEncoder()` and this branch seems to require `self.encoder`

tbkka:
#1488 fixes the problem with having `final` test classes in the 1.x branch.  (The main branch doesn't need to use a script for test discovery, so doesn't need this fix.)

rebello95:
Thanks @tbkka! If we land this one, it could be a good test of #1488 to add `final` to all the test classes to validate it as well

thomasvl (approved):

thomasvl:
Thank you for doing the work to pull it to 1.x!

rebello95:
> Thank you for doing the work to pull it to 1.x!

🙌🏽 do you know when a new release is expected for 1.x?

thomasvl:
@tbkka make sense to cut a release?  [Change since last release](https://github.com/apple/swift-protobuf/compare/1.24.0...1_x_release_branch):
- The plugin refactoring
- These two new encoding options
- CI for Swift 5.9

tbkka:
I think it's reasonable.

rebello95:
Thank you!

PR #1488, opened 2023-10-26
Consistently use the last word before colon as the class name

tbkka:
This allows the test collection script to work correctly for test classes defined in either of the following ways:
```
class Test_Foo: XCTestCase
final class Test_Foo: XCTestCase
```

Note:  The CollectTests.awk is only still needed for compatibility with old Swift systems that did not automatically discover tests on Linux.

### DIFF
--- a/DevTools/CollectTests.awk
+++ b/DevTools/CollectTests.awk
@@ -49,8 +49,8 @@ BEGIN {
 	printf("}\n")
     }
     split($0, a, ":")
-    split(a[1], words, " ")
-    CLASS=words[2]
+    n=split(a[1], words, " ")
+    CLASS=words[n]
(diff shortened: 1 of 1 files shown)


tbkka:
I don't think this script is needed at all in `main`, so I'm putting up this PR only for `1_x_release_branch`.  We should probably just delete the script entirely from `main`.

tbkka:
@rebello95  I think this is sufficient to address the problem you had with adding `final` to test classes in the 1.x branch.

tbkka:
Hmmm.... `Illegal reference to array 'words'`.    I guess GNU awk doesn't support length() of an array?

tbkka:
No.  GNU awk and others do support length() of an array, but apparently it's not in POSIX and not universally supported by all awk implementations.  So here's a different approach.

thomasvl (approved):

PR #1490, opened 2023-10-27
Update the generated_swift_names_* files.

thomasvl:


Protos/SwiftProtobufTests/generated_swift_names_enums.proto
```diff
@@ -48,6 +48,7 @@ message GeneratedSwiftReservedEnums {
   enum BinaryEncodingError { NONE_BinaryEncodingError = 0; }
   enum BinaryEncodingMessageSetSizeVisitor { NONE_BinaryEncodingMessageSetSizeVisitor = 0; }
   enum BinaryEncodingMessageSetVisitor { NONE_BinaryEncodingMessageSetVisitor = 0; }
+  enum BinaryEncodingOptions { NONE_BinaryEncodingOptions = 0; }
```
tbkka, line 51: The enum name diff is so nice and small.  I wonder if we can rework the enum cases generator as well?  Maybe combine enum name and enum cases to generate one enum with a single case each?   That would make these easier to review.

tbkka: Hmmm....   that wouldn't test for accidental collisions from renaming rules, though.  Hmmm...

tbkka (approved):

Protos/SwiftProtobufTests/generated_swift_names_enums.proto
thomasvl: Yea, I was just thinking we could do an enum per case to reduce the diffs if we wanted.  Might also help some of the compile times for these files.  But anything else I come up with (100 cases per enum), still would have the ripple effects as things are added/removed.

tbkka: Group by first letter, maybe?  That would limit the ripple effect, and seems like it would still keep things together that might be prone to collide.

PR #1498, opened 2023-11-09
Convert _MessageBase to a typealias

mjburghard:
Remove the `_MessageBase` protocol in favor of a type alias with different definitions for debug and release builds.
This should decrease the code size without requiring an underscored attribute (see #1497).

### DIFF
--- a/Sources/SwiftProtobuf/Message.swift
+++ b/Sources/SwiftProtobuf/Message.swift
@@ -31,7 +31,7 @@
 ///
 /// The actual functionality is implemented either in the generated code or in
 /// default implementations of the below methods and properties.
-public protocol Message: _MessageBase {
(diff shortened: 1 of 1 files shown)


allevato:
This seems like an even better option than the maker protocol.

Since `_MessageBase` is really just "a list of common protocols that messages conform to", I wonder if it's worth renaming it to hammer that point home. Something like `_CommonMessageConformances`, so people don't get the idea that they should try to constrain generics based on it or add functionality to it (which fortunately they won't be able to if it's just a typealias).

mjburghard:
I like the new name.I will rename the type alias.

tbkka (approved):
Nice!

allevato (approved):

thomasvl:
Thanks for the patch!

PR #1499, opened 2023-11-09
A performance-oriented test decoding 1 million messages from a stream

tbkka:
This is inspired by a slow case discovered by the oss-fuzz service, this attempts to decode a stream consisting of a million zero bytes.  Each of these zero bytes represents a single message with no non-default fields.

I'm checking this in to help aid in future performance optimization for AsyncMessageSequence.


thomasvl (approved):
I guess the only down side is these will slow down all test runs.  But having it to eventual do the performance improvements makes some sense.

tbkka:
I cut it back to 100,000 bytes instead of a million.  That should reduce the overall impact on test time.  I've factored out the number of iterations so future tinkerers can easily experiment.

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
```diff
@@ -199,6 +199,26 @@ final class Test_AsyncMessageSequence: XCTestCase {
...
+
+  // Slow test case found by oss-fuzz: 1 million zero-sized messages
+  // A similar test with BinaryDelimited is about 4x faster, showing
+  // that we have some room for improvement here.
+  // (Note this currently only tests 100,000 zero-sized messages,
+  // but the constant below is easy to edit if you want to experiment.)
+  func testLargeExample() async throws {
+    let messageCount = 100_000
+    let bytes = [UInt8](repeating: 0, count: messageCount)
+    let byteStream = asyncByteStream(bytes: bytes)
```
FranzBusch, line 211: I can see why this might be slow. We are using `AsyncStream` here for the underlying `AsyncSequence` which does a bunch of internal locking and is not `@inlinable`. We should probably just copy over the [AsyncSyncSequence](https://github.com/apple/swift-async-algorithms/blob/main/Sources/AsyncAlgorithms/AsyncSyncSequence.swift) for our test here which should perform significantly better.

tbkka: I tried that in the Fuzz-testing harness and it seems to make very little difference (at least, not for the debug build, which is where we're having problems).

FranzBusch: Oh I didn't know we run the fuzzer in debug mode. That's a bit strange and most likely to be the reason for the time outs. `AsyncSequence` heavily relies on the compiler to do some optimisations. Why do we run the fuzzers in debug mode?

thomasvl: Just completeness.  There are a few added checks/validations within the library, so it just helps make sure none of those have potential issues.

PR #1505, opened 2023-11-18
Add FieldMask utilities to Message types

pouyayarandi:
# Summary

This PR adds a bunch of utilities that enable message types to modify some of their fields (by Google well-known type, FieldMask). Here are added APIs:

```swift
extension Message {

  /// Checks whether the given path is valid for Message type.
  public static func isPathValid(_ path: String) -> Bool

  /// Merges fields specified in a FieldMask into another message.
  public mutating func merge(with source: Self, fieldMask: Google_Protobuf_FieldMask) throws

  /// Removes from 'message' any field that is not represented in the given FieldMask.
  public mutating func trim(fieldMask: Google_Protobuf_FieldMask) -> Bool
}

extension Google_Protobuf_FieldMask {

  /// Initiates a field mask with all fields of the message type.
  public init<M>(allFieldsOf messageType: M.Type)

  /// Initiates a field mask from some particular field numbers of a message.
  public init<M>(fieldNumbers: [Int], of messageType: M.Type) throws

  /// Adds a path to FieldMask after checking whether the given path is valid.
  public mutating func addPath<M>(_ path: String, of messageType: M.Type) throws

  /// Converts a FieldMask to the canonical form.
  public var canonical: Google_Protobuf_FieldMask { get }

  /// Creates a union of two FieldMasks.
  public func union(_ mask: Google_Protobuf_FieldMask) -> Google_Protobuf_FieldMask

  /// Creates an intersection of two FieldMasks.
  public func intersect(_ mask: Google_Protobuf_FieldMask) -> Google_Protobuf_FieldMask

  /// Creates a FieldMasks with paths of the original FieldMask that does not included in mask.
  public func subtract(_ mask: Google_Protobuf_FieldMask) -> Google_Protobuf_FieldMask

  /// Returns true if the given path is covered by the FieldMask.
  public func contains(_ path: String) -> Bool

  /// Checks whether the given FieldMask is valid for type M.
  public func isValid<M>(for messageType: M.Type) -> Bool
}
```

# Implementation

This feature is empowered by `_ProtoNameProviding` to map proto field numbers to names. This enables us to have the aforementioned inits of FieldMask. Also, for masking a message two Decoders have been implemented: `GetPathDecoder` and `SetPathDecoder`. The first one gets the value of a field by its path, the latter sets a value to a field of the message by the path. In `override` function, value will be captured by GetPathDecoder and then, it will be set by the SetPathDecoder.

# Use case

Assuming we have a caching system which prevent some fields of a response from recalculation. Backend gives us a proto message and tells that some of its fields should be cached (i.e. field number 1). Next time, when we are going to make a request, we evaluate our client-side's cache and send the cached field numbers in the header of the request (i.e. we send back 1 as a field number that its value is already cached). As a result, backend response with leaving those fields empty (or filled with default values). And in client-side, we need to modify those field numbers with cached values.

For this purpose, first of all, we can store the original backend response somewhere, and in the rest of requests we can override the responses with the cached proto message:

```swift
// Defines a mask field which includes particular paths based on field numbers
let mask = Google_Protobuf_FieldMask(fieldNumbers: cachedFieldNumbers, of: MyMessage.self)

// Get response by telling backend not to recalculate those fields and backend leave those fields unset
var response: MyMessage = await network.getResponse(exclude: cachedFieldNumbers)

// Get cached response from storage
let cachedResponse: MyMessage = cache.getCache()

// Merge those fields of response with the cached one.
try response.merge(to: cachedResponse, fieldMask: mask)
```

# Testing

Unit tests have been added for the new feature, and all tests have passed.

# Considerations

- Merge options and trim options have not been implemented as they require some generated code.
- Dot pattern is supported in paths. Path `a.b` refers to field `b` of the field `a` in the message.
- Merges will occur in order of FieldMask's paths.


thomasvl:
At quick glance you seem to have two distinct proposals in here (they don't seem connected in implementation), if that's the case, I'd suggest splitting this into two PRs so there can be easier discussion on them independently. It likely would help to also provide some examples of what usecases you are trying to enable.

pouyayarandi:
Hi @thomasvl, the protobuf mirroring feature has been deleted from this PR, so the current one is a proposal just for `applying`. Also the use case section is added to the description.

thomasvl:
Handing off to Apple folks.

Idle thought - Since `Decoder` is public, could this entirely implemented via a custom decoder that feeds field numbers and then the values via the decoding apis?  Then there wouldn't have to be any more codegen to support it?  The library code itself would likely be bigger.  But if that is possible, this could be entirely done as a layer on top of SwiftProtobuf and not actually need library additions.

pouyayarandi:
@tbkka Thanks for your response.

> * How would you do this in the C++ protobuf implementation?   Java?  Python?  Go?  C#?

The concept is the same in all the languages you've mentioned but when it comes to implementation, we would see different APIs based on the nature of each language. As I found in most of them, they use Descriptor and Reflection. Here are some examples:

- [C++](https://stackoverflow.com/a/36696706/6547133)
- [Python](https://stackoverflow.com/a/60877463/6547133)
- [Java](https://stackoverflow.com/a/19460650/6547133)

> * Can the FieldMask concept that's in the protobuf specification be used for this kind of application? (We have never fully implemented FieldMask in SwiftProtobuf; if that could provide this support, then it might be worth implementing that instead.)

Maybe, but AFAIK, it just helps us to tell the backend which fields should be excluded, when the client receives the response we have to replace somehow a bunch of values with those ones we've cached before. Here we need an API to set values by field numbers.

> Related: Your ProtobufMirror idea should be implementable as a new encoder, again without requiring any new code generation.

The `ProtobufMirror` feature had been implemented by visitors and it had no generated code.

thomasvl:
I believe FieldMask usually also has helpers that work with copying things from one message to another message - "Copy just these fields".  And by having it within the library it would have access to the field names data that currently isn't exposed.

To really make these Apply methods useful would likely be figuring out how to include the field numbers in the generated files for folks.  My initial thought would be an `enum` doesn't really help because you want `Int` values and you don't want to keep having to do `rawValue`.  So likely just `Int` constants scoped in some way.  But unlike a C based languages where an `enum` of just using `#define` is zero runtime cost, I'm not sure what would work for Swift to make the field number constant not add to binary size since would be `public` and thus if things get linked into a framework they would all become exported symbols.

tbkka:
> Here we need an API to set values by field numbers.

There are a number of inter-related features here that I have in mind.  Since many of these can be implemented in terms of the others, it's worth thinking about them together to figure out how we want to provide them:

* Descriptors.  SwiftProtobuf does not currently embed descriptors into generated messages, mostly due to concerns about binary size.
* Setting values by field number and/or name.  You've shown how we could implement each of these directly with the current decoder APIs.
* FieldMask utilities.  We currently only have support for constructing FieldMask messages.  We also want someday to have utilities that can apply a FieldMask to a message in memory to remove all masked fields or keep only masked fields.
* Merge utilities.  We don't currently have a way to set fields of one message from another.  A nice way to handle this would be to set fields of a message from another message + FieldMask, where the FieldMask indicates what fields need to be overridden.

If we had APIs to query, set, and clear fields by name and by number, those could be used to implement the FieldMask utilities.  They could also be used as building blocks for Descriptor-based utilities.   I think those APIs could all be built using the techniques you've been exploring.

(And coincidentally, in #1504, Max is proposing a different way to organize the traversal logic that might make the above more efficient.)

pouyayarandi:
To sum up, we'd better implement some utilities as extensions on the FieldMask and Messages. Here are some of those utils I found helpful:

```swift
extension Message {

  /// Returns a field mask with all fields of the message type.
  static var fieldMask: Google_Protobuf_FieldMask

  /// Builds a field mask from some particular field numebers of a message
  /// - Parameter fieldNumbers: Field numbers of paths
  /// - Returns: Field mask that include paths of corresponding field numbers
  static func fieldMask(fieldNumbers: [Int]) -> Google_Protobuf_FieldMask

  /// Builds a field mask by excluding some paths
  /// - Parameter paths: Paths to be excluded
  /// - Returns: Field mask that does not include the paths
  static func fieldMask(excludedPaths paths: [String]) -> Google_Protobuf_FieldMask

  /// Builds a field mask by reversing another mask
  /// - Parameter mask: Mask to be reversed
  /// - Returns: Field mask that does not include the `mask` paths
  static func fieldMask(reverse mask: Google_Protobuf_FieldMask) -> Google_Protobuf_FieldMask

  /// Clears masked fields and keep the other fields unchanged
  /// - Parameter mask: Field mask which determines what fields shoud be cleared
  mutating func mask(by mask: Google_Protobuf_FieldMask)

  /// Overrides value of masked fields in original message with the input message
  /// - Parameters:
  ///   - message: Message which overrides some fields of the original message
  ///   - mask: Field mask which determines what fields shoud be overriden
  mutating func override(with message: Self, by mask: Google_Protobuf_FieldMask)
 
  func masked(by mask: Google_Protobuf_FieldMask) -> Self
  func overriden(with message: Self, by mask: Google_Protobuf_FieldMask) -> Self
}

extension Google_Protobuf_FieldMask {
  func union(_ mask: Google_Protobuf_FieldMask) -> Self
  func intersection(_ mask: Google_Protobuf_FieldMask) -> Self
}
```

If you agree with these APIs, I could implement it. Please let me know if you have any other idea. @tbkka @thomasvl

thomasvl:
I haven't had a chance to really think about these or compare to other apis; but one general question that does come to mind:

Would the `static` functions make more sense to be in an `extension` on `Google_Protobuf_FieldMask` and have them take a `Message.Type` (instead of being on `Message`)?

I'm not sure which makes for a more *swifty* api, but maybe something to think about.  They may actually make more sense to developers as convenience initializers?

tbkka:
Can these all be implemented with the information we generate today?  It seems like the name map might have everything you need for this?   Or does this require something like #1504 first, in order to expose the field information in a more easily-digestible form?

pouyayarandi:
> Can these all be implemented with the information we generate today?

Yes it could be implemented already. But obviously something like #1504 makes it much easier

pouyayarandi:
@tbkka @thomasvl New implementation is added for some field mask utils. Note that some other utils like (intersecting, union, etc.) would better to get implemented separately.

thomasvl:
Haven't looked in detail, just taking a quick skim, did you compare the proposed apis with what are available in other langaues?  https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/util/field_mask_util.h for example seems to have a few _path_ related things that would should also provide.

Historically, we've also found that by looking at the [tests in other languages](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/util/field_mask_util_test.cc), it can provide some tests to directly bring over to ensure our implementations are correct.  Usually makes sense to provide comments on the test methods pointing back to what been brought over.

pouyayarandi:
@thomasvl Thanks for sharing. All APIs updated based on the file you sent. Just merge and trim options aren't implemented as they require some kind of data to be generated like whether field required or not, and is a field set or not yet.

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -184,3 +179,201 @@ extension Google_Protobuf_FieldMask: _CustomJSONCodable {
...
+  /// Converts a FieldMask to the canonical form. It will:
+  ///   1. Remove paths that are covered by another path. For example,
+  ///      "foo.bar" is covered by "foo" and will be removed if "foo"
+  ///      is also in the FieldMask.
+  ///   2. Sort all paths in alphabetical order.
+  public var canonical: Google_Protobuf_FieldMask {
+    var mask = Google_Protobuf_FieldMask()
+    let sortedPaths = self.paths.sorted()
+    var set = Set<String>()
+    for path in sortedPaths {
```
thomasvl, line 246: Is there a more performant version of this?  Once the list is sorted, can you simply check to see if entry N-1 is the prefix of the next entry (with a dot added)?

i.e. -

Once sorted, we'd have something like:

[ "foo", "foo.bar", "foobaz" ]

So then for index 1, you can see "foo." (entry zero with a dot appended) is the prefix of index 1, so index 1 gets dropped.  Then for index 2, "foo." is not a prefix of "foobaz", so index 2 gets kept.

This might reduce things to a single sort and then a linear walk of the list pulling things over.

thomasvl: Oh, does this also have to dedup things?  That would be a tweak to just check for the previous also being an exact match.

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -184,3 +179,201 @@ extension Google_Protobuf_FieldMask: _CustomJSONCodable {
...
+  /// Creates an intersection of two FieldMasks.
+  ///
+  /// - Parameter mask: FieldMask to intersect with.
+  /// - Returns: FieldMask with intersection of two path sets.
+  public func intersect(
+    _ mask: Google_Protobuf_FieldMask
+  ) -> Google_Protobuf_FieldMask {
+    let set = mask.pathsSet
+    var _paths: [String] = []
+    for path in paths where set.contains(path) {
```
thomasvl, line 283: If the current mask had duplicates, i believe this continues to have the duplicates, is that desired?

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -184,3 +179,201 @@ extension Google_Protobuf_FieldMask: _CustomJSONCodable {
...
+  /// that does not included in mask.
+  ///
+  /// - Parameter mask: FieldMask with paths should be substracted.
+  /// - Returns: FieldMask with all paths does not included in mask.
+  public func subtract(
+    _ mask: Google_Protobuf_FieldMask
+  ) -> Google_Protobuf_FieldMask {
+    let set = mask.pathsSet
+    var _paths: [String] = []
+    for path in paths where !set.contains(path) {
```
thomasvl, line 301: Similar concerns about duplicates

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -184,3 +179,201 @@ extension Google_Protobuf_FieldMask: _CustomJSONCodable {
...
+    return (0..<comps.count).map {
+      comps[0...$0].joined(separator: ".")
+    }
+  }
+
+  private func contains(
+    _ path: String,
+    in set: Set<String>
+  ) -> Bool {
+    for level in levels(path: path) {
```
thomasvl, line 336: Also a little worried about memory usage/performance here, might make sense to use string ops to keep things in check.

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -184,3 +179,201 @@ extension Google_Protobuf_FieldMask: _CustomJSONCodable {
...
+  ) -> Bool {
+    let message = M()
+    return paths.allSatisfy { path in
+      message.isPathValid(path)
+    }
+  }
+}
+
+/// Describes errors could happen during FieldMask utilities.
+public enum FieldMaskUtilsError: Error {
```
thomasvl, line 362: Maybe just FieldMaskError?

Sources/SwiftProtobuf/Message+FieldMask.swift
```diff
@@ -0,0 +1,86 @@
...
+  /// Merges fields specified in a FieldMask into another message.
+  ///
+  /// - Parameters:
+  ///   - source: Message should be merged to the original one.
+  ///   - fieldMask: FieldMask specifies which fields should be merged.
+  public mutating func merge(
+    to source: Self,
+    fieldMask: Google_Protobuf_FieldMask
+  ) throws {
+    var copy = self
```
thomasvl, line 47: Do we need to make the copy, or can we just directly update self?

Sources/SwiftProtobuf/Message+FieldMask.swift
```diff
@@ -0,0 +1,86 @@
...
+      return false
+    }
+    if fieldMask.paths.isEmpty {
+      return false
+    }
+    var tmp = Self.init()
+    do {
+      try tmp.merge(to: self, fieldMask: fieldMask)
+      let changed = tmp != self
+      self = tmp
```
thomasvl, line 80: Doesn't this result in loss of unknown fields and extensions (if supported on the message)?

Sources/SwiftProtobuf/Message+FieldMask.swift
```diff
@@ -0,0 +1,86 @@
...
+  ) throws {
+    var copy = self
+    var pathToValueMap: [String: Any?] = [:]
+    for path in fieldMask.paths {
+      pathToValueMap[path] = try source.get(path: path)
+    }
+    for (path, value) in pathToValueMap {
+      try copy.set(path: path, value: value)
+    }
+    self = copy
```
thomasvl, line 55: Doesn't this result in loss of unknown fields and extensions (if supported on the message)?

Sources/SwiftProtobuf/SetPathDecoder.swift
```diff
@@ -0,0 +1,343 @@
...
+
+  init(path: String, value: Any?) {
+    self.path = path
+    self.value = value
+    if let firstPathComponent {
+      self.number = T.number(for: firstPathComponent)
+    }
+  }
+
+  var firstPathComponent: String? {
```
thomasvl, line 50: Similar concerns about the paths, might be worth splitting once and tracking positions, etc.

Sources/SwiftProtobuf/SetPathDecoder.swift
```diff
@@ -0,0 +1,343 @@
...
+  mutating func decodeSingularFloatField(value: inout Float) throws {
+    value = try _value(as: Float.self)
+  }
+
+  mutating func decodeSingularFloatField(value: inout Float?) throws {
+    value = try _value(as: Float?.self)
+  }
+
+  mutating func decodeRepeatedFloatField(value: inout [Float]) throws {
+    value = try _value(as: [Float].self)
```
thomasvl, line 86: The repeated handling appears to always being doing a full "set" vs. an "append", and some of the upstream docs seems to imply repeated fields should be appends for some operations.  Do we need a way to do either operation?

Sources/SwiftProtobuf/GetPathDecoder.swift
```diff
@@ -0,0 +1,325 @@
...
+
+import Foundation
+
+struct GetPathDecoder<T: Message>: Decoder {
+
+  private let path: String
+  private var number: Int?
+
+  private var _value: Any?
+  private var _hasPath: Bool = false
```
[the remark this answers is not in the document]
pouyayarandi: A value could be optional (and possibly nil) but its path exists.

Sources/SwiftProtobuf/GetPathDecoder.swift
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  mutating func decodeMapField<KeyType, ValueType>(
+    fieldType: _ProtobufMap<KeyType, ValueType>.Type,
+    value: inout _ProtobufMap<KeyType, ValueType>.BaseType
+  ) throws {
+    captureValue(value)
+  }
+
+  mutating func decodeMapField<KeyType, ValueType>(
```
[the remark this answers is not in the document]
pouyayarandi: There might be. I would investigate.

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
pouyayarandi: With the algorithm you've mentioned above, we won't need it anymore

Sources/SwiftProtobuf/Message+FieldMask.swift
pouyayarandi: I guess it's needed since we shouldn't update a message unless we make sure our merge won't be thrown. i.e. assume we merge a mask field with 2 paths. The first one performs correctly and when it comes to the second one, it throws an error, so merging happens partially.

Sources/SwiftProtobuf/Message+FieldMask.swift
pouyayarandi: Since we create our copy from `self`, I guess it keeps everything.

Sources/SwiftProtobuf/Message+FieldMask.swift
pouyayarandi: Yeah, it will clear everything but the paths in the field mask. Not desired?

Sources/SwiftProtobuf/SetPathDecoder.swift
pouyayarandi: It was a merge option in the file you sent and I didn't implement merge options as I said earlier. But this particular option could be implemented regardless of the other ones. I will go for it.

Sources/SwiftProtobuf/GetPathDecoder.swift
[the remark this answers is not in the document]
pouyayarandi: Based on cpp implementation [here](https://github.com/protocolbuffers/protobuf/blob/7e7037db27aeaa3b9a3e7623e525da6b67150c38/src/google/protobuf/util/field_mask_util.cc#L444) it seems that subpath is only valid for singular message fields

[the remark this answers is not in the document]
pouyayarandi: Oh and about map fields we may have a merge option (let's call it `replaceMapFields` like the similar one for repeated fields) that decides whether a map field should be merged partially or replaced completely. what do you think? Note that cpp doesn't have this merge option.

Sources/SwiftProtobuf/SetPathDecoder.swift
pouyayarandi: I added `MergeOption` which has a case `replaceRepeatedFields` to determine it

Sources/SwiftProtobuf/GetPathDecoder.swift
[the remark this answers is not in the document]
thomasvl: Oh right, since this is a `Decoder` we're feeding path segments to walk to the right place?  Probably should be a comment on the on generally how this decoder works for someone looking in the future.

Sources/SwiftProtobuf/GetPathDecoder.swift
```diff
@@ -44,211 +46,214 @@ struct GetPathDecoder<T: Message>: Decoder {
     return number
   }
 
-  private mutating func captureValue(_ value: Any?) {
+  private mutating func captureValue(_ value: Any?) throws {
+    if !nextPath.isEmpty {
```
thomasvl, line 50: This might read a little better as a `guard`.

```swift
guard nextPath.isEmpty else {
  throw PathDecodingError.pathNotFound
}
```

pouyayarandi: Done, and also previous comments were fixed. I appreciate if you review them and mark them as resolved if they are fixed.

thomasvl:
> @thomasvl Thanks for sharing. All APIs were updated based on the file you sent and tried to make them a little more swifty. Just merge and trim options aren't implemented as they require some kind of data to be generated like whether field required or not, and is a field set or not yet. BTW, I added comments for tests.

Sorry, I haven't gotten a chance to go back through this in a more complete train of thought.  With my current schedule, it might be a little while until I have that time. 😞 

Your right we don't capture anything reusable about required, we conditionally generate the validation/partial support, but that probably doesn't help. Can you provide some pointers as to where this information would been needed and how it impacts things?

Within the Message/Storage we do have if things are set or not, but for the Visitor/Decoder we don't. The proposed change to the Visitors could get traversal for the cases where things aren't set also, so if that lands first, that might provide the information to fill in the gaps.

Sources/SwiftProtobuf/Message+FieldMask.swift
pouyayarandi: I checked cpp implementation, it uses reflection to clear fields which has not been implemented in swift protobuf. How about copying unknown fields and extensions?

Sources/SwiftProtobuf/GetPathDecoder.swift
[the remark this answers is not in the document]
thomasvl: Can you add that mentioned comment?

Sources/SwiftProtobuf/Message+FieldMask.swift
```diff
@@ -0,0 +1,105 @@
...
+  /// - Parameter path: Path to be checked
+  /// - Returns: Boolean determines path is valid.
+  public static func isPathValid(
+    _ path: String
+  ) -> Bool {
+    var message = Self()
+    return message.hasPath(path: path)
+  }
+
+  internal mutating func isPathValid(
```
thomasvl, line 30: I guess this has to be mutating just because it used decoding to do the work?

Sources/SwiftProtobuf/GetPathDecoder.swift
```diff
@@ -0,0 +1,324 @@
...
+//
+// -----------------------------------------------------------------------------
+///
+/// Decoder which captures value of a field by its path.
+///
+// -----------------------------------------------------------------------------
+
+import Foundation
+
+struct GetPathDecoder<T: Message>: Decoder {
```
thomasvl, line 17: Generally could use a comment to say what its used for, etc.

Sources/SwiftProtobuf/GetPathDecoder.swift
```diff
@@ -0,0 +1,324 @@
...
+    value: inout _ProtobufMessageMap<KeyType, ValueType>.BaseType
+  ) throws {
+    try captureValue(value)
+  }
+
+  mutating func decodeExtensionField(
+    values: inout ExtensionFieldValueSet,
+    messageType: Message.Type,
+    fieldNumber: Int
+  ) throws {}
```
thomasvl, line 304: a precondition here might not hurt since this should never get called.

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -184,3 +179,202 @@ extension Google_Protobuf_FieldMask: _CustomJSONCodable {
...
+  /// - Parameter mask: FieldMask to intersect with.
+  /// - Returns: FieldMask with intersection of two path sets.
+  public func intersect(
+    _ mask: Google_Protobuf_FieldMask
+  ) -> Google_Protobuf_FieldMask {
+    let set = mask.pathsSet
+    var _paths: [String] = []
+    var _buffer = Set<String>()
+    for path in paths where set.contains(path)
+    && !_buffer.contains(path) {
```
thomasvl, line 287: indent is a little odd/hard to read here.

Sources/SwiftProtobuf/Message+FieldMask.swift
```diff
@@ -0,0 +1,105 @@
...
+    _ path: String
+  ) -> Bool {
+    hasPath(path: path)
+  }
+}
+
+/// Defines available options for merging two messages.
+public enum MergeOption: Equatable {
```
thomasvl, line 38: I think the better choice might be to model this as a struct, that way you don't end up having a parameter else where that is an array of these. It also brings things in line with how most of the other *Options types in this library have worked. For the one option you have right now, it could then become a boolean to control replace vs. append for repeated and map fields.

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -184,3 +179,202 @@ extension Google_Protobuf_FieldMask: _CustomJSONCodable {
...
+  /// Creates a FieldMasks with paths of the original FieldMask
+  /// that does not included in mask.
+  ///
+  /// - Parameter mask: FieldMask with paths should be substracted.
+  /// - Returns: FieldMask with all paths does not included in mask.
+  public func subtract(
+    _ mask: Google_Protobuf_FieldMask
+  ) -> Google_Protobuf_FieldMask {
+    let set = mask.pathsSet
+    var _paths: [String] = []
```
thomasvl, line 305: Generally, this library seem to only use a leading underscore for properties, not locals. So to make the code fit a bit better, I'd suggest using local names without the leading underscore.  (I think there were other places that did this, not just here)

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -184,3 +179,202 @@ extension Google_Protobuf_FieldMask: _CustomJSONCodable {
...
+    var _buffer = Set<String>()
+    for path in paths where !set.contains(path)
+    && !_buffer.contains(path) {
```
thomasvl, line 308: indent here also

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -184,3 +179,202 @@ extension Google_Protobuf_FieldMask: _CustomJSONCodable {
...
+    }
+    return false
+  }
+
+  // Set containing paths of FieldMask
+  private var pathsSet: Set<String> {
+    .init(paths)
+  }
+
+  private func levels(path: String) -> [String] {
```
thomasvl, line 338: Is this still needed, doesn't seem like anything calls it.

Sources/SwiftProtobuf/Message+FieldMask.swift
thomasvl: That does mean this could use a lot of memory depending on the graph size.

The binary format merging doesn't bother to do the buffering here (for memory/performance reasons).  Might be the better choice here as if someone wants to preflight it, they can take the memory hit otherwise, we can error on the side of less memory hit.

@tbkka any thoughts?

Sources/SwiftProtobuf/Message+FieldMask.swift
thomasvl: Ah right.

Sources/SwiftProtobuf/Message+FieldMask.swift
thomasvl: That works.  If you want, you could add a "todo" comment at the to revisit this if we ever have a way to revisit clearing fields instead.

Sources/SwiftProtobuf/SetPathDecoder.swift
```diff
@@ -0,0 +1,364 @@
...
+  private var number: Int?
+  private let nextPath: [String]
+  private let replaceRepeatedFields: Bool
+
+  init(
+    path: [String],
+    value: Any?,
+    replaceRepeatedFields: Bool
+  ) {
+    if let firstComponent = path.first {
```
thomasvl, line 54: Does this need to do any validation on the path?  If not, maybe add a comment about why it can assume the paths are always valid?

Sources/SwiftProtobuf/GetPathDecoder.swift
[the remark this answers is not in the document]
thomasvl: `map<>` is just shorthand for a repeated field of a message with a key & value field, so maybe the C++ is reusing the same repeated field option for maps?

thomasvl:
> Your right we don't capture anything reusable about required, we conditionally generate the validation/partial support, but that probably doesn't help. Can you provide some pointers as to where this information would been needed and how it impacts things?

I saw the comment about being able to generically clear fields, but what were the other places we current don't provide needed information?

Sources/SwiftProtobuf/GetPathDecoder.swift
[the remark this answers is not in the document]
pouyayarandi: Done. now `replaceRepeatedFields` is being applied for both repeated fields and maps

Sources/SwiftProtobuf/Message+FieldMask.swift
pouyayarandi: I changed it and there is no copy anymore. Now we just validate the captured values and throw before any set. Note that our source is `Self` so it should never throw an error (i.e. type mismatch) because merging message fields and source fields are identical.

Sources/SwiftProtobuf/Message+FieldMask.swift
pouyayarandi: Todo added. Additionally, add some lines of code to copy unknown fields and extensions (if it's extensible)

Sources/SwiftProtobuf/SetPathDecoder.swift
pouyayarandi: @thomasvl Did you check it? It's done

Sources/SwiftProtobuf/Message+FieldMask.swift
pouyayarandi: Exactly. To make it non-mutating, we had to make a copy which brings those memory concerns, etc.

Sources/SwiftProtobuf/GetPathDecoder.swift
pouyayarandi: Done

Sources/SwiftProtobuf/SetPathDecoder.swift
pouyayarandi: I changed the code in a more understandable way, after checking for first path component, we check whether this component refers to a valid field number of message or not. If it doesn't, the number property remains nil and decoder does nothing. (As it has no `nextFieldNumber` to pass)

pouyayarandi:
> but what were the other places we current don't provide needed information

There are some merge/trim options in cpp headers which need more info and couldn't be implemented at the moment. So we can make a todo comment for them.

- replaceMessageFields in merge options: to merge a sub-message partially we need to know which fields are unset to prevent them from merging.
- keepRequiredFields in trim options: we need to know what fields of a message are required to add them to the field mask so that those fields remain unchanged after trimming.

pouyayarandi:
Any updates? @thomasvl @tbkka

Sources/SwiftProtobuf/Message+FieldMask.swift
```diff
@@ -0,0 +1,126 @@
...
+    _ path: String
+  ) -> Bool {
+    hasPath(path: path)
+  }
+}
+
+/// Defines available options for merging two messages.
+public struct MergeOption {
+
+  public init(replaceRepeatedFields: Bool = false) {
```
thomasvl, line 40: I think the other review is debating if we want to list all the options on `init`, or have folks just manually set them after making an instance. In anycase, we should be consistent.

Sources/SwiftProtobuf/Message+FieldMask.swift
```diff
@@ -0,0 +1,126 @@
...
+
+  internal mutating func isPathValid(
+    _ path: String
+  ) -> Bool {
+    hasPath(path: path)
+  }
+}
+
+/// Defines available options for merging two messages.
+public struct MergeOption {
```
thomasvl, line 38: This ends up a top level type in the module, I'm wondering if it should be scoped or named in such a way as to scope it.  It seems like "MergeOptions" is pretty generic and someone might expect it apple to other apis?

The two options that jump to mind are either name it with _FieldMask_ in some way, or declare it scoped to the FieldMask message.

@tbkka any thoughts on better approaches?

Sources/SwiftProtobuf/Message+FieldMask.swift
```diff
@@ -0,0 +1,126 @@
...
+
+extension Message {
+
+  /// Merges fields specified in a FieldMask into another message.
+  ///
+  /// - Parameters:
+  ///   - source: Message should be merged to the original one.
+  ///   - fieldMask: FieldMask specifies which fields should be merged.
+  public mutating func merge(
+    to source: Self,
```
thomasvl, line 59: "to" seems wrong, it's this "from", i.e. - the `source` is where the fields are copied _from_?  Then are copied "to" `self`?

Sources/SwiftProtobuf/SetPathDecoder.swift
```diff
@@ -0,0 +1,396 @@
...
+  init(
+    path: [String],
+    value: Any?,
+    mergeOption: MergeOption
+  ) {
+    if let firstComponent = path.first,
+       let number = T.number(for: firstComponent) {
+      self.number = number
+    }
+    self.nextPath = .init(path.dropFirst())
```
thomasvl, line 70: Maybe also set `nextPath` to an empty array if the current component lookup fails?

thomasvl:
> Any updates? @thomasvl @tbkka

Sorry for the delay, got swamped in a few other projects with the start of the new year also.

thomasvl:
@tbkka can you also take a look?  I have a small concern about the amount of code here for overall binary size.  We've not actually generating anything new, so it is less of an issue, but it seems like it could be seen by some users as a larger increase for things they may not end up using.

tbkka:
> I have a small concern about the amount of code here for overall binary size. We've not actually generating anything new, ...

The size of the library has always been a secondary concern for me.  So I'm really only worried about inadvertently increasing the size of compiled messages.  I don't _think_ this should add any code to the compiled generated messages unless someone actually uses this feature.  But it would be good to check.   Hmmm....  I think I see a way to verify that assumption.

pouyayarandi:
> but it seems like it could be seen by some users as a larger increase for things they may not end up using.

We could encapsulate all these features inside an entity (let's say `FieldMaskUtilites`) to recude added APIs. It's more like cpp's implementation but maybe less swifty. Also `MergeOptions` could be an internal struct of the class so we'll have it inside FieldMaskUtilities scope like `FieldMaskUtilities.MergeOptions`

Sources/SwiftProtobuf/Message+FieldMask.swift
thomasvl: My initial suggestion would just be:

```swift

extension Google_Protobuf_FieldMask {
  public struct MergeOptions {
    // ...
  }
}
```

That way it is (hopefully) clear that it is specific to FieldMask.

Sources/SwiftProtobuf/Message+FieldMask.swift
thomasvl: My initial suggestion would just be:

```swift

extension Google_Protobuf_FieldMask {
  public struct MergeOptions {
    // ...
  }
}
```

That way it is (hopefully) clear that it is specific to FieldMask.

tbkka:
> but it seems like it could be seen by some users as a larger increase for things they may not end up using.

I built the entire test suite in Release mode with and without @pouyayarandi's  changes and looked through the results to see what impact it had on size.  The results were quite interesting:

* The SwiftProtobuf library became 150k (6%) bigger.  Looking through the `.o` files, this is almost all from the new source files, though a few of the existing object files also grew a tiny bit.  For example, AnyMessageStorage.swift.o was a little over 1k larger in the version with Pouya's additions.

* The statically-linked protoc-gen-swift binary also became bigger by about 100k.  None of the existing `.o` files for that program changed size, so this seems to be entirely due to the larger library.

* I deliberately omitted the tests Pouya added, so the test suite source code was the same, only the library itself was changed.  I saw that the test suite did not change at all.  None of the existing object files became larger as a result of Pouya's changes, and the final dynamically-linked test suite was the same size.

So it looks like the only impact on existing adopters that don't use these new features is that the SwiftProtobuf library itself is bigger.  This does regrettably impact statically-linked consumers of the library.  But in particular, generated code is not any bigger, even when compiled against the new library, so the impact should be a fixed size, even for statically-linked executables.  (My fear was that there would be growth proportional to the number of protos, but I saw no evidence of that.)

I think this is within reason.  I'd obviously be happier if the library growth were smaller, but I think that's something we can optimize for later on.  The only real disappointment is Swift's failure to properly dead-strip static library code.  But that's something that Swift itself will need to tackle someday, and I don't think that's an entirely convincing reason to stop adding useful features.

@Lukasa -- Any thoughts from your side?

Sources/SwiftProtobuf/Message+FieldMask.swift
pouyayarandi: Done

Sources/SwiftProtobuf/Message+FieldMask.swift
pouyayarandi: Done. Also other comments are resolved. could you please check all again?

pouyayarandi:
Thank @tbkka for the thorough analysis. Could we do anything to reduce the binary size?

pouyayarandi:
Just for follow-up, I fixed the comments. @tbkka @thomasvl

pouyayarandi:
Just for follow up, build issues with swift 5.8 and later should be fixed now. @thomasvl

thomasvl:
> Just for follow up, build issues with swift 5.8 and later should be fixed now. @thomasvl

Thanks, triggered the CI again to run on things.

Since this is some new functionality, I'd like @tbkka (or some of the other Apple folks) to sign off on it before it's merged.  Given the time of year, the Apple folks might be a little busy with WWDC prep, so it make take some more time before they get a chance to look.

pouyayarandi:
Thank @thomasvl, BTW I also found an issue in decoding `oneOf` fields. First for a better vision, when a non one of field is getting decoded it passes its current value to the decoder as an inout parameter. Like this:
```swift
case 11: try { try decoder.decodeSingularFloatField(value: &_storage._singularFloat) }()
```
But in one of fields it happens completely different. It's a sample of code that generated for decoding a one-of field:
```swift
case 71: try {
  var v: Float?
  try decoder.decodeSingularFloatField(value: &v)
  if let v = v {
    if _storage._o != nil {try decoder.handleConflictingOneOf()}
    _storage._o = .oneofFloat(v)
  }
}()
```
In the above code the inout parameter that is passing to decode method is always nil (v variable is never set before passing to decoder method). My suggestion would be to add a line of code setting the current value to variable if it's possible:
```swift
case 71: try {
  var v: Float?
  if case .oneofFloat(let _v)? = _storage._o {v = _v}    // add this line
  try decoder.decodeSingularFloatField(value: &v)
  if let v = v {
    if _storage._o != nil {try decoder.handleConflictingOneOf()}
    _storage._o = .oneofFloat(v)
  }
}()
```
This line of code can be added with this PR (as it's directly affecting field masks) or could be added during another PR. If you prefer second one, we'd better drop supporting for one-of fields in this PR. What's your thought?

thomasvl:
@pouyayarandi – Can you open a new issue for that so we can track it and the fix?  It should be pretty easy to make a test case using the Merging api to show the problem and also do a solution, but rather than flutter this PR up with that process, a new issue will make it easier to discuss (and provide an easy to follow history if need be).

pouyayarandi:
> Can you open a new issue for that so we can track it and the fix?
I opened the issue and will fix it in a new PR, ASAP.

pouyayarandi:
I've made some changes which fixes the problem with decoding oneof fields. I used visitor for getting values of a message instead of `GetPathDecoder`. @thomasvl can you please check it again?

thomasvl:
Sorry, been swapped with a few things of late, will try to look again soon.

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -184,3 +179,193 @@ extension Google_Protobuf_FieldMask: _CustomJSONCodable {
...
+      mask.paths = M.allProtoNames
+    }
+  }
+
+  /// Initiates a field mask from some particular field numbers of a message
+  ///
+  /// - Parameters:
+  ///   - messageType: Message type to get all paths from.
+  ///   - fieldNumbers: Field numbers of paths to be included.
+  /// - Returns: Field mask that include paths of corresponding field numbers.
```
thomasvl, line 201: Probably need to document it throws if the field number isn't on the message

Sources/SwiftProtobuf/Message+FieldMask.swift
```diff
@@ -0,0 +1,136 @@
...
+  ///   - fieldMask: FieldMask specifies which fields should be merged.
+  public mutating func merge(
+    with source: Self,
+    fieldMask: Google_Protobuf_FieldMask,
+    mergeOption: Google_Protobuf_FieldMask.MergeOptions = .init()
+  ) throws {
+    var visitor = PathVisitor<Self>()
+    try source.traverse(visitor: &visitor)
+    let values = visitor.values
+    for path in fieldMask.paths {
```
thomasvl, line 78: Maybe put a note/todo here, since this seems like this could be a performance issue in the future; if there is a way to traverse once setting the any requested fields that seems like it would be better.

Sources/SwiftProtobuf/Message+FieldMask.swift
```diff
@@ -0,0 +1,136 @@
...
+      )
+    }
+  }
+}
+
+extension Message where Self: Equatable, Self: _ProtoNameProviding {
+
+  // TODO: Re-implement using clear fields instead of copying message
+
+  @discardableResult
```
thomasvl, line 92: Move this down right before the `public mutating...`

Sources/SwiftProtobuf/NameMap.swift
```diff
@@ -278,4 +278,11 @@ public struct _NameMap: ExpressibleByDictionaryLiteral {
     let n = Name(transientUtf8Buffer: raw)
     return jsonToNumberMap[n]
   }
+  
+  /// Returns all proto names
+  internal var names: [Name] {
```
thomasvl, line 283: Is this still used?  Seems like you might be using a map and description in other places?

Sources/SwiftProtobuf/PathDecoder.swift
```diff
@@ -0,0 +1,426 @@
...
+  mutating func decodeRepeatedMessageField<M>(
+    value: inout [M]
+  ) throws where M : Message {
+    try setRepeatedValue(&value)
+  }
+
+  mutating func decodeSingularGroupField<G>(
+    value: inout G?
+  ) throws where G : Message {
+    try setValue(&value, defaultValue: nil)
```
thomasvl, line 370: Why is this some different than decodeSingularMessageField?  Shouldn't they basically be the same as groups are still messages?

Sources/SwiftProtobuf/PathDecoder.swift
```diff
@@ -0,0 +1,426 @@
...
+    try setMapValue(&value)
+  }
+
+  mutating func decodeExtensionField(
+    values: inout ExtensionFieldValueSet,
+    messageType: any Message.Type,
+    fieldNumber: Int
+  ) throws {
+    preconditionFailure(
+      "Path decoder should never decode an extension field"
```
thomasvl, line 406: Prefix this with "Internal Error:" just so if it ever did show up it might be a little more clear for folks.

Sources/SwiftProtobuf/PathVisitor.swift
```diff
@@ -0,0 +1,277 @@
...
+
+  mutating func visitSingularEnumField<E: Enum>(value: E, fieldNumber: Int) throws {
+    visit(value, fieldNumber: fieldNumber)
+  }
+
+  mutating func visitSingularMessageField<M: Message>(value: M, fieldNumber: Int) throws {
+    guard var path = T.name(for: fieldNumber) else {
+      return
+    }
+    values[path] = value
```
thomasvl, line 111: In visit() you store with prevPath prepended, but here you don't, why?  Sounds like maybe a test doesn't cover the submessage case correctly to detect this?

thomasvl (commented):
Read back through everything to try and catch back up on the code.

Sources/SwiftProtobuf/Message+FieldMask.swift
pouyayarandi: It already gets all values of a message with only one traverse. But we can improve it with only getting required paths instead of all paths.

Sources/SwiftProtobuf/NameMap.swift
pouyayarandi: Yes it's still being used in `init(allFieldsOf message: M)` for initializing a FieldMask containing all paths of a message type `M`

pouyayarandi:
Comments are fixed

thomasvl (commented):
Looks pretty good, kicked off the CI on the current state of things.

@tbkka can you take a look again to provide final decision since it is a new addition to the library.

thomasvl:
@pouyayarandi looks like atleast one unittest test is failing.

pouyayarandi:
I changed my code a little bit, but I couldn't reproduce the failing test on my local machine and all tests passed. So it may fail again. Can you please re-run pipeline? @thomasvl

Sources/SwiftProtobuf/PathDecoder.swift
```diff
@@ -35,7 +35,9 @@ extension Message {
     guard let type = Self.self as? any _ProtoNameProviding.Type else {
       return nil
     }
-    return type._protobuf_nameMap.number(forJSONName: field)
+    return Array(field.utf8).withUnsafeBytes { bytes in
```
thomasvl, line 38: Isn't this trying to return an Array now instead of an Int?

pouyayarandi: The return type of `withUnsafeBytes` is the same as its closure return type and in the closure we return an Int.

[withUnsafeBytes documentation](https://developer.apple.com/documentation/swift/array/withunsafebytes(_:))

thomasvl: Sorry, should have gotten my morning coffee before looking at the change. 😄

thomasvl:
> It failed again 😢 Do you have any idea why it's happening? I run `make test` on my machine and everything seems okay. In addition, in some checks the failing test passes Is there any way to run the exact check on my machine to see the error?

So it's only failing on linux?

You could try using Docker to use do a local linux build (can't say I have any experience with that).

The failure seems to be around group merging?  You could try making a temporary test case out of it to try and zero in on things.  i.e. - make a test that just merges two messages with the fields in question, but doesn't use the FieldMask support, does that work as intended?  Ensure the calculated mask has what you expect, etc.

You can also tweak `.github/workflows/build.yml` to have it run on your branch name (and disable all but one of the checks needed), and enable CI on your own repo, then, then you can simply upload and let github test for you.  If you do this, maybe clone this branch into another one while you test things out, so each upload doesn't also update this PR, and go back to this branch when you've been able to figure it out.

thomasvl:
> Bug fixed @thomasvl
> 
> Problem was about `singularGroup` field of the message. Since this field name is `.unique` which means its proto name and json name differs, it caused the bug. I made some changes which just allow using proto names for merging purposes. Please re-run the tests they must fully passed now same as my forked repo.

Make sure you are synced up to `HEAD` on `main`, there as a change around groups as part of the Protobuf Editions work a few weeks ago, so more forms are accepted in TextFormat. So that could impact things.

pouyayarandi:
> Make sure you are synced up to HEAD on main

I fixed the bug on another branch which I rebased with synced fork of your `main` and then, cherry-picked the fix commit here. So I guess it should be synced with the latest changes

pouyayarandi:
Hi @thomasvl, just for follow-up; do you have any updates?

thomasvl:
> Hi @thomasvl, just for follow-up; do you have any updates?

I think we're still waiting on the Apple folks (@tbkka most likely) to look since this is a new addition to the library.

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -184,3 +179,195 @@ extension Google_Protobuf_FieldMask: _CustomJSONCodable {
...
+
+  /// Creates a FieldMasks with paths of the original FieldMask
+  /// that does not included in mask.
+  ///
+  /// - Parameter mask: FieldMask with paths should be substracted.
+  /// - Returns: FieldMask with all paths does not included in mask.
+  public func subtract(
+    _ mask: Google_Protobuf_FieldMask
+  ) -> Google_Protobuf_FieldMask {
+    let set = mask.pathsSet
```
tbkka, line 305: Minor style issue:  I'd prefer to just inline `Set<String>(mask.paths)` here rather than using the helper.

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -184,3 +179,195 @@ extension Google_Protobuf_FieldMask: _CustomJSONCodable {
...
+  }
+
+  /// Creates an intersection of two FieldMasks.
+  ///
+  /// - Parameter mask: FieldMask to intersect with.
+  /// - Returns: FieldMask with intersection of two path sets.
+  public func intersect(
+    _ mask: Google_Protobuf_FieldMask
+  ) -> Google_Protobuf_FieldMask {
+    let set = mask.pathsSet
```
tbkka, line 285: I'd prefer `Set<String>(mask.paths)` here rather than using the helper

Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
```diff
@@ -184,3 +179,195 @@ extension Google_Protobuf_FieldMask: _CustomJSONCodable {
...
+    for fieldMaskPath in paths {
+      if path.hasPrefix("\(fieldMaskPath).") || fieldMaskPath == path {
+        return true
+      }
+    }
+    return false
+  }
+
+  // Set containing paths of FieldMask
+  private var pathsSet: Set<String> {
```
tbkka, line 334: This helper does not add much.

tbkka (commented):
This looks good!  The code is well-organized and easy to read.  There are a few places where performance could be improved, but we can iterate on that later if that proves to be a problem for people.

I have just a few small suggestions:
* Inline a couple of trivial helpers
* A couple of small naming adjustments to clarify the public API.

It would also be good to update Documentation/API.md with information about this, but I think that should just be a separate PR.

Sources/SwiftProtobuf/Message+FieldMask.swift
```diff
@@ -0,0 +1,137 @@
...
+  // TODO: Re-implement using clear fields instead of copying message
+
+  /// Removes from 'message' any field that is not represented in the given
+  /// FieldMask. If the FieldMask is empty, does nothing.
+  ///
+  /// - Parameter fieldMask: FieldMask specifies which fields should be kept.
+  /// - Returns: Boolean determines if the message is modified
+  @discardableResult
+  public mutating func trim(
+    fieldMask: Google_Protobuf_FieldMask
```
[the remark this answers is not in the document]
tbkka: To be clear, I'm suggesting:
```
public mutating func trim(
  keeping fieldMask: Google_Protobuf_FieldMask
  ...
```

pouyayarandi:
Thank you @tbkka, all comments were fixed

tbkka:
Merged!    I'm looking forward to the next PR that updates Documentation/API.md with information about how people can use this.  ;-)

thomasvl:
@pouyayarandi thanks again for all the work here!

thomasvl:
@pouyayarandi & @tbkka – Not sure why I didn't think of it before, but this morning I got wondering if with the runtime now has should any of the [old initializers](https://github.com/apple/swift-protobuf/blob/main/Sources/SwiftProtobuf/Google_Protobuf_FieldMask%2BExtensions.swift#L117-L151) get revisited? Specifically –
1. Should the proto name based ones get deprecated to provide failable ones that do validation?
1. The JSON one has a call out around validation, can that now be done?

tbkka:
There are two kinds of validation we might do and a couple of different places to do it:
* We could validate that the paths are syntactically well-formed.  That is, reject `paths = ["😀"]`
* We could validate that the paths are valid for a particular message type

We can do any of the above validation in a few different places:
* We could assert in initializers
* We could have validating initializers that throw when they detect a problem
* We could validate when the FieldMask is serialized or deserialized
* We could validate when the FieldMask is used to mask a particular message

I think that asserting/validating initializers are probably the least important here, though they might simplify some use cases.  It's more important to validate when (de)serializing or masking.

thomasvl:
> I think that asserting/validating initializers are probably the least important here, though they might simplify some use cases. It's more important to validate when (de)serializing or masking.

I was about to say I don't think we can do this one, but it just dawned on me, I guess the initializers can't do it either.  The problem is we don't know the _Message_ the proto names/JSON names are supposed to match, we'd need all new initializers that also take a `Message.Type`, the current free string forms can't really do anything I think sort of the `paths = ["😀"]` type issues. 😦

tbkka:
People need to be able to build FieldMask protos without validating against a message type.  (Because there are always going to be cases where the message type that's going to be masked isn't actually on the local system.)

We _must_ validate against a message type when we use the FieldMask against an actual message.  I think in most cases, that's actually the right place to do that validation.

thomasvl:
Yea so I guess this work doesn't change anything.  But what we could do is validated all the strings to ensure they are valid field names according to the proto language spec.

PR #1509, opened 2023-11-22
Explicitly annotate existential usage of `Message` with the existential `any` keyword.

mrabiciu:
This PR separates out the addition of the existential `any` from https://github.com/apple/swift-protobuf/pull/1504 into its own distinct PR for easier consumption. 

Context: https://github.com/apple/swift-protobuf/pull/1504 adds Self constraints to Message with the addition of `static var _fields: Field<Self>`. With this change the swift compiler will require uses of the existential `Message` type to be explicitly annotated with the `any` keyword. This keyword is currently optional for existential types without Self constraints however it will be required in swift 6.

### DIFF
--- a/Sources/Conformance/main.swift
+++ b/Sources/Conformance/main.swift
@@ -84,7 +84,7 @@ func buildResponse(serializedBytes: [UInt8]) -> Conformance_ConformanceResponse 
         return response
     }
 
-    let msgType: SwiftProtobuf.Message.Type
+    let msgType: any SwiftProtobuf.Message.Type
     let extensions: SwiftProtobuf.ExtensionMap
     switch request.messageType {
     case "":
@@ -102,7 +102,7 @@ func buildResponse(serializedBytes: [UInt8]) -> Conformance_ConformanceResponse 
         return response
     }
 
-    let testMessage: SwiftProtobuf.Message
+    let testMessage: any SwiftProtobuf.Message
     switch request.payload {
     case .protobufPayload(let data)?:
         do {
--- a/Sources/SwiftProtobuf/BinaryDecoder.swift
+++ b/Sources/SwiftProtobuf/BinaryDecoder.swift
@@ -1090,7 +1090,7 @@ internal struct BinaryDecoder: Decoder {
 
     internal mutating func decodeExtensionField(
       values: inout ExtensionFieldValueSet,
-      messageType: Message.Type,
+      messageType: any Message.Type,
       fieldNumber: Int
     ) throws {
         if let ext = extensions?[messageType, fieldNumber] {
@@ -1104,7 +1104,7 @@ internal struct BinaryDecoder: Decoder {
     /// Helper to reuse between Extension decoding and MessageSet Extension decoding.
     private mutating func decodeExtensionField(
       values: inout ExtensionFieldValueSet,
-      messageType: Message.Type,
+      messageType: any Message.Type,
       fieldNumber: Int,
       messageExtension ext: AnyMessageExtension
     ) throws {
@@ -1131,7 +1131,7 @@ internal struct BinaryDecoder: Decoder {
 
     internal mutating func decodeExtensionFieldsAsMessageSet(
       values: inout ExtensionFieldValueSet,
-      messageType: Message.Type
+      messageType: any Message.Type
     ) throws {
         // Spin looking for the Item group, everything else will end up in unknown fields.
         while let fieldNumber = try self.nextFieldNumber() {
@@ -1175,7 +1175,7 @@ internal struct BinaryDecoder: Decoder {
 
     private mutating func decodeMessageSetItem(
       values: inout ExtensionFieldValueSet,
-      messageType: Message.Type
+      messageType: any Message.Type
     ) throws -> DecodeMessageSetItemResult {
         // This is loosely based on the C++:
         //   ExtensionSet::ParseMessageSetItem()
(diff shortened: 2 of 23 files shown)


thomasvl:
Can you expand the description for why this is needed (for history, etc.).

And looking at the other PR, `public struct Field<M: Message> { ... }` doesn't both with `any` there, which seems to say not all uses of `Message` need to change to `any Message`, so what were the concrete requirements and are all these changes actually needed?

mrabiciu:
Updated title and description

FranzBusch:
> This keyword is currently optional for existential types without Self constraints however it will be required in swift 6.

This statement is no longer true. The language workgroup has recently put out [a post](https://forums.swift.org/t/progress-toward-the-swift-6-language-mode/68315) where they stated that `ExistentialAny` is no longer going to become a default feature in Swift 6. So for now we should put this on hold.

Can you explain why your other PR needed this to be changed everywhere?

mrabiciu:
> This statement is no longer true. The language workgroup has recently put out [a post](https://forums.swift.org/t/progress-toward-the-swift-6-language-mode/68315) where they stated that ExistentialAny is no longer going to become a default feature in Swift 6
Ah TIL good to know!

> Can you explain why your other PR needed this to be changed everywhere?
I'm adding a `static var _fields: [Field<Self>] { get }` property to the `Message` protocol, which means this protocol now has `Self` constraints. With this change swift requires existential uses of `Message` to be explicitly annotated with the existential `any`.

Sources/Conformance/main.swift
```diff
@@ -84,7 +84,7 @@ func buildResponse(serializedBytes: [UInt8]) -> Conformance_ConformanceResponse
         return response
     }
 
-    let msgType: SwiftProtobuf.Message.Type
+    let msgType: any SwiftProtobuf.Message.Type
```
tbkka, line 87: I believe this is wrong.

Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -1090,7 +1090,7 @@ internal struct BinaryDecoder: Decoder {
 
     internal mutating func decodeExtensionField(
       values: inout ExtensionFieldValueSet,
-      messageType: Message.Type,
+      messageType: any Message.Type,
```
tbkka, line 1093: I believe this is wrong, as are the other `any Message.Type` below.

Sources/SwiftProtobuf/Decoder.swift
```diff
@@ -204,11 +204,11 @@ public protocol Decoder {
   // Decode extension fields
 
   /// Decode an extension field
-  mutating func decodeExtensionField(values: inout ExtensionFieldValueSet, messageType: Message.Type, fieldNumber: Int) throws
+  mutating func decodeExtensionField(values: inout ExtensionFieldValueSet, messageType: any Message.Type, fieldNumber: Int) throws
```
tbkka, line 207: More `any Message.Type`

Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
```diff
@@ -67,7 +67,7 @@ fileprivate final class UnsafeMutableTransferBox<Wrapped> {
 extension UnsafeMutableTransferBox: @unchecked Sendable {}
 
 // All access to this should be done on `knownTypesQueue`.
-fileprivate let knownTypes: UnsafeMutableTransferBox<[String:Message.Type]> = .init([
+fileprivate let knownTypes: UnsafeMutableTransferBox<[String: any Message.Type]> = .init([
```
tbkka, line 70: Again: `any Message.Type`

tbkka (commented):
I see a lot of `any Message.Type` and similar in here.  I believe that is wrong -- `Message.Type` is not being used as an existential here, so the `any` is not appropriate.

mrabiciu:
> I see a lot of any Message.Type and similar in here. I believe that is wrong -- Message.Type is not being used as an existential here, so the any is not appropriate.

@tbkka All of the places that I've annotated with `any` were labeled by the compiler as required when combining with my changes from https://github.com/apple/swift-protobuf/pull/1509

Screenshot 2023-11-27 at 3 34 50 PM

Sources/Conformance/main.swift
mrabiciu: I think this _is_ an existential use of `Message.Type` since the type assignment is dynamic below. `msgType` could be assigned to `ProtobufTestMessages_Proto3_TestAllTypesProto3` or `ProtobufTestMessages_Proto2_TestAllTypesProto2` so the concrete type implementing the `Message` protocol is not known.

tbkka:
Yep.  My mistake.

FranzBusch:
In general, I am happy with adopting `ExistentialAny` in this package since the language steering group has just deferred it to the next language revision. However, for consistency reason I would like to enable the `upcomingFeatureFlag` in our `Package.swift` then and just stick it everywhere. This way the compiler is going to enforce it for us.
@thomasvl @tbkka WDYT?

thomasvl:
> In general, I am happy with adopting `ExistentialAny` in this package since the language steering group has just deferred it to the next language revision. However, for consistency reason I would like to enable the `upcomingFeatureFlag` in our `Package.swift` then and just stick it everywhere. This way the compiler is going to enforce it for us. @thomasvl @tbkka WDYT?

Having the validation sounds good to me.

tbkka:
> @tbkka WDYT?

Sounds like a good plan to me.  Let's do that.

mrabiciu:
I found a workaround for this in https://github.com/apple/swift-protobuf/pull/1504 so this is no longer necessary for me. LMK if I should close this. I'm happy to merge it too if you'd like.

thomasvl:
I think we're good landing this, just need to do @FranzBusch 's suggestion to enable the workings in the Package.swift file.

Issue #713, opened 2018-01-18
Current use of serializedDataSize() in serializedData() causes additional runtime linear in the depth of the proto hierarchy


MrMage:
(Disclaimer: I know that Protobuf is not meant for large data structures, but I figured it wouldn't hurt to ask about this anyway.)

I have something similar to the following protobuf hierarchy:

    message Wrapper {
        Wrapped wrapped = 1;
    }
    
    message Wrapped {
        repeated string entry = 1; // the actual structure is a bit more complex, but that shouldn't really make a difference
    }

In my case, calling `serializedData()` on a message of type `Wrapped` takes about 0.75 seconds (it contains a _lot_ of data), of which about 0.25 seconds go towards computing the value of `wrapped.serializedDataSize()`.

Calling `serializedData()` on a message of type `Wrapper` wrapping the same `Wrapped` instance above takes about 1 second, of which about 0.25 seconds go towards computing the value of `wrapper.serializedDataSize()` on the highest (`Wrapper`) level and another 0.25 seconds to `wrapped.serializedDataSize()` (the same as above). The serialized size of `wrapper`, on the other hand is just 5 bytes more than that of `wrapped`. Each additional level would introduce another ~0.25 seconds of `serializedDataSize()`, essentially computing the same size over and over again.

If I change `Wrapper` to

    message Wrapper {
        bytes wrapped = 1;
    }

and manually store a serialized representation of `wrapped` in there, this takes about 0.75 seconds to encode `Wrapped` (as before), plus a negligible amount of overhead for copying that serialized representation around (much less than ~0.25 seconds).

This means that for cases in which `serializedDataSize()` is potentially costly, introducing extra hierarchy levels into my protobuf model can cause significant increases in computation time.

The C++ implementation avoids this problem by caching serialized sizes, as evidenced by the statement "Most of these are just simple wrappers around ByteSize() and SerializeWithCachedSizes()." on https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.message_lite#serialization.

Given that manipulating messages that are in the process of being serialized sounds like a _bad_ idea anyway, I wonder if it would be possible introduce similar caching into SwiftProtobuf? This could either be done by storing an extra `cachedSize` field in each message (see https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.message_lite#MessageLite.GetCachedSize.details). That would introduce an extra 8 bytes of overhead per message, though.

[...]

PR #1518, opened 2023-11-30
Binary encode from back-to-front

tbkka:
Our current front-to-back binary encoder has a subtle performance problem:  In order to allocate the correct number of bytes for the size of a sub-message, it needs to know the final size of the sub-message. This results in recursive calls to size each message, leading to overall performance that is quadratic in the depth of nesting.

To be clear:  this is not usually a serious problem since few people have messages with more than 2 or 3 layers of nesting.

This PR changes the encoder to encode from the end of the buffer back towards the front.  This allows us to write the size after we finish each sub-message, avoiding the recursive sizing calls.  With this version, we have only one sizing traversal in the initial top-level encoding request in order to properly allocate the output buffer.

Working from back to front does mean that individual fields get written in the opposite order.  This is _mostly_ not a problem: Protobuf explicitly states that decoders must accept fields in any order.  The one exception is for repeated fields, which we handle here by iterating the arrays backwards inside the encoder.

A few cases where order might matter:
* Unrecognized enum cases in repeated fields are treted as "unknown" fields which means that re-serializing puts them into a different place. Since this code puts the unknown fields at the beginning of the buffer rather than the end, this means that we've changed the resulting order after deserializing/reserializing.
* The conformance test has one test case that verifies merging behavior and seems very sensitive to the order of fields.  I suspect this is a bug in the test, but need to check further.


tbkka:
Finally sat down and hammered this out.  (It was easier than I thought.)   As noted above, there are a couple of issues that need to be further explored before this could be accepted as-is.  And of course, I'd have to work through and update the many tests that hard-code the exact bytes expected from the binary output.  (Or write a test helper that is smart enough to order-insensitively compare two binary serialized messages.)

Reworking this to sit on top of #1504 would allow us to iterate the fields in reverse order, which would let this encoder provide the same field ordering as the current encoding does.

tbkka:
And of course I also need to get the perf harness working again so I can verify the performance of this approach...

Sources/SwiftProtobuf/BinaryReverseEncoder.swift
```diff
@@ -0,0 +1,156 @@
...
+        // of contiguous storage, body is not called and nil is returned.
+        let isAvailable = utf8.withContiguousStorageIfAvailable { (body: UnsafeBufferPointer<UInt8>) -> Int in
+            let r = prepend(contentsOf: UnsafeRawBufferPointer(body))
+            putVarInt(value: body.count)
+	    return r
+        }
+        if isAvailable == nil {
+            precondition(false)
+            let count = utf8.count
+            putVarInt(value: count)
```
thomasvl, line 144: This seems backwards, but given the precondition, I'm guessing you aren't expecting this path to be needed?

thomasvl:
Nice!  I think waiting on the other cl so we could visit in reverse order would be best since it solves the ordering problem is the best bet vs. changing all the test/writing helpers/etc.

Performance wise…

One minor one, I think I can dig up a reference around doing the varints so they don't have to be recursive.

The sizing allocation/resizing is likely the biggest thing. To compute the buffer size initial means `serializedDataSize` and `BinaryEncodingSizeVisitor` for sub messages is also the bad performance case; we traverse the sub messages multiple times.  So the draft removes those multiple passes while encoding the bytes, but we're still playing the sizing costs. (and there is a non trivial number of usages way more than 3-4 levels deep 😄 )

Thinking about this, I see a few options:

1. Completely remove `BinaryEncodingSizeVisitor` and make binary encoding resize the buffer (likely in some chunk size) as it prepends bytes. Some of the "sub" encoders/visitors might make this complex.
2. Change `BinaryEncodingSizeVisitor` to over estimate things.  When it goes to compute a sub message size, just assume it will take the max number of bytes for a varint. Then once the full message has been written out, figure out how many wasted bytes are on the front and copy the data down to the front and resize the buffer.
3. Or, can we do the same trick in your reverse encoder within `BinaryEncodingSizeVisitor`?  Each time we need a sub message size, cache the current value for `serializedSize`, visit into the sub message with `self`, then compute the size using the cached value and the current value for `serializedSize`?  Thus we avoid the over visiting the sub messages.

Did any of those ideas make sense?

tbkka:
I really don't believe the initial sizing is a problem at all.   The problem with the old approach was that we did many sizing passes:
* The initial sizing walks the entire message, including all submessages, recursively.  Last time I profiled this code, this was actually quite fast.
* As we serialized, each time we looked at a submessage, we had to size that sub message _again_ by walking it and all of its children _again_.  This repeated size walk as we visit every sub message is where we get the quadratic behavior.

Google's C++ implementation avoids resizing each submessage by caching the size on each message object.  This is essentially your idea number 3 above.  It works, but it's awkward and requires some additional storage somewhere to store all the sizes.  (Google's C++ implementation adds a hidden field to every message for this purpose.)

I suspect resizing would be more expensive than doing a single sizing pass up front.  Not only is malloc a fairly expensive call, but keeping track of when a resize is needed adds a lot of bookkeeping logic.

(A more efficient way to serialize the varints would be nice.  This PR is still "draft" for a reason. 😉 )

Sources/SwiftProtobuf/BinaryReverseEncoder.swift
tbkka: Fixed.

tbkka: (I tested this locally by commenting out the contiguous handling and re-running the tests.   Everything still seems fine.)

tbkka:
Link this to #713

PR #1520, opened 2023-12-01
Make binary size calculation a single pass.

thomasvl:
Rather can using `serializedDataSize`, we can just save the size, traverse the sub message, and then query back the size to see how big the message was and then compute the number of bytes that varint encoding would take.


tbkka:
I'm fine with this, since I suppose it's a bit clearer than the previous form, though I'm curious whether you can actually see a difference in timing:   The old code called `subMessage.serializedDataSize()` which is just basically a wrapper around `subMessage.traverse()`.   So in essence you're manually inlining `serializedDataSize()` with this change.  We're still traversing the entire tree one time and adding up the size of every field in it.  Just in a slightly different order.

tbkka (approved):

thomasvl:
Your are completely correct, some how I got it in my head we were still descending after the call the `serializedDataSize()`.  At best, I guess we save a stack frame?  But since the api is internal, hopefully the compiler will inline things and make it moot.  I'll just close this.

Issue #605, opened 2017-06-20
perf_runner seems to leave local tree in an odd state


thomasvl:
Synced to d3b087912c8e1a77ce93e50364b35f7c5099a790, `Performance/perf_runner.sh -c 7b786f134df64c46baa9c7f63daced6af6211c01` finished with files in a locally modified state.

```
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.

Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	modified:   Protos/generated_swift_names_enum_cases.proto
	modified:   Protos/generated_swift_names_enums.proto
	modified:   Protos/generated_swift_names_fields.proto
	modified:   Protos/generated_swift_names_messages.proto
	modified:   Reference/generated_swift_names_enum_cases.pb.swift
	modified:   Reference/generated_swift_names_enums.pb.swift
	modified:   Reference/generated_swift_names_fields.pb.swift
	modified:   Reference/generated_swift_names_messages.pb.swift
	modified:   Sources/SwiftProtobuf/BinaryEncoder.swift
	modified:   Sources/SwiftProtobuf/BinaryEncodingVisitor.swift
	modified:   Sources/SwiftProtobuf/Message+BinaryAdditions.swift
	modified:   Tests/SwiftProtobufTests/generated_swift_names_enum_cases.pb.swift
	modified:   Tests/SwiftProtobufTests/generated_swift_names_enums.pb.swift
	modified:   Tests/SwiftProtobufTests/generated_swift_names_fields.pb.swift
	modified:   Tests/SwiftProtobufTests/generated_swift_names_messages.pb.swift

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   Protos/generated_swift_names_enum_cases.proto
	modified:   Protos/generated_swift_names_enums.proto
	modified:   Protos/generated_swift_names_fields.proto
	modified:   Protos/generated_swift_names_messages.proto
	modified:   Reference/generated_swift_names_enum_cases.pb.swift
	modified:   Reference/generated_swift_names_enums.pb.swift
	modified:   Reference/generated_swift_names_fields.pb.swift
	modified:   Reference/generated_swift_names_messages.pb.swift
	modified:   Sources/SwiftProtobuf/BinaryEncoder.swift
	modified:   Sources/SwiftProtobuf/BinaryEncodingVisitor.swift
[...]

PR #1521, opened 2023-12-02
Try to get the performance harness running again...

tbkka:
Ongoing changes to the protobuf build system broke the previous performance harness scripts.

I've rebuilt the C++ perf harness by using a CMake build script to build libprotobuf and the C++ harness together.  This uses more of the protobuf build machinery than I'd like -- it could probably be simplified and made more robust.

This has some requirements:
 * A protobuf source checkout that has all the submodules
 * abseil and googletest libraries installed on the local system and findable by CMake (On Mac, `brew install abseil googletest` seems to suffice)

### DIFF
--- a/Performance/Harness.h
+++ b/Performance/Harness.h
@@ -94,7 +94,7 @@ private:
    * from the main results.
    */
   template <typename Function>
-  typename std::result_of<Function()>::type measure_subtask(
+  typename std::invoke_result<Function>::type measure_subtask(
       const std::string& name, Function&& func);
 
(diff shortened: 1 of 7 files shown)


thomasvl:
Not sure if you want to look at #605 at the same time, or if there's enough changed it's already effectively resolved.

tbkka:
I changed the handling for comparison checkouts -- it should avoid #605 now, though I still need to do more testing.

thomasvl (approved):

PR #1524, opened 2023-12-05
Add `enableUpcomingFeature("ExistentialAny")` and update accordingly.

thomasvl:
Enable the feature on all targets, and then go through doing the fixits within Xcode 15 that are triggered as a result.

This basically replaces #1509

fyi - @mrabiciu

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -39,30 +39,41 @@ let package = Package(
         name: "SwiftProtobuf",
         exclude: ["CMakeLists.txt"],
         swiftSettings: [
+          .enableUpcomingFeature("ExistentialAny"),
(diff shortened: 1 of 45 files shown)


thomasvl:
My (foolish) original intent was just to add the flags and land that, didn't realize just how many things would have to be updated at the same time, hence this ended up being a replacement for the other PR (with the feature also getting enabled).

There are a few other targets we haven't enabled `StrictConcurrency=complete`, they end up needing some code changes, so I left that disable on the targets.  Might be something worth following up on at some point for completeness.

FranzBusch (approved):

PR #1526, opened 2023-12-13
Add option to include fields that are equal to the default value in the JSON encoding

mrabiciu:
This adds the ability to include fields that are equal to the default value in the json encoding. 

This is done by adding a `traversalOptions` var to `Visitor` as well as adding a `includeDefaultValues` to `JSONEncodingOptions` which controls the `JSONEncodingVisitor`'s `traversalOptions`

I've modified the generation of `traverse` to check `traversalOptions` before calling `visit` on each individual field.


mrabiciu:
This commit has all of the functional changes https://github.com/apple/swift-protobuf/pull/1526/commits/86f3e2bf1c54aed74a5dfe3be24ffa073ec22d38

thomasvl:
Do the other languages end up encoding the empty arrays when visiting default values?  i.e. - does it make sense to visit on those or just skip the visit calls still?  We probably want to make sure our JSON support ends up doing the same things the other upstream languages do.

In any case, since this is public api, we should spell out in the comments what the expected behavior is for all cases:
- repeated fields
- map fields
- extension fields (all types)
- proto2 syntax files (since non repeated field types have presence all have presence)
- proto3 syntax files where the field isn't optional (expect the visit)
- proto3 syntax files where a field is optional (likely should be the same as proto2 syntax files)

And then use a custom listener in the test to ensure all these behaviors for all the different message types.

thomasvl:
For the comments spelling out the behaviors, I guess we should use the terms from the upstream pages:
- https://protobuf.dev/programming-guides/proto2/#field-labels
- https://protobuf.dev/programming-guides/proto3/#field-labels

mrabiciu:
> Do the other languages end up encoding the empty arrays when visiting default values?

Yes I'm using the python package for comparison and if you set `including_default_value_fields=True` it includes empty lists and empty maps. 

https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html
> **including_default_value_fields** – If True, singular primitive fields, repeated fields, and map fields will always be serialized. If False, only serialize non-empty fields. Singular message fields and oneof fields are not affected by this option.

mrabiciu:
> And then use a custom listener in the test to ensure all these behaviors for all the different message types

I'm not sure what you mean by that. Is an encoding test that asserts the encoded json is as expected not enough?

thomasvl:
> > And then use a custom listener in the test to ensure all these behaviors for all the different message types
> 
> I'm not sure what you mean by that. Is an encoding test that asserts the encoded json is as expected not enough?

Haven't looked in detail, it might be.  I wasn't sure if json had any specific logic of it's own vs. ensure all the methods expected did/didn't get called.

thomasvl:
I finally got a chance to start looking at this, and I'm going to suggest we step back to hopefully won't be a bike shed –

As drafted, the PR calls the `JSONEncodingOptions` addition `includeDefaultValues`.  And `Visitor` gets `TraversalOptions` which has `visitDefaultValues`.

The looking at the C++ api, their JSON is [`always_print_primitive_fields`](https://github.com/protocolbuffers/protobuf/blob/9c7d2b9d83e0ad4e8ffeb8535d8c0a31a0a6a0b8/src/google/protobuf/json/json.h#L42-L46)

The Java api has: [`alwaysOutputDefaultValueFields` & `includingDefaultValueFields`](https://github.com/protocolbuffers/protobuf/blob/6ae76fd6ed6507ccab284ed16a430ae42fa64239/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java#L103-L112)

I don't think we want to eventually try to do the level of the api that Java has (the Set support would really complicate the visitor pattern).

So on the naming side, my 2¢ would be:

I think _default_ gets a little confusing around the behavior with repeated fields and `map<>` fields. So I think I sorta like the C++ naming because it is about _primitive_ fields, there still is a minor stretch in that a _repeated_ or _map_ field is _primitive_ (it's an _array_ or a _map_), but it a _single message_ field clearly isn't a _primitive_ field. So…

- `Visitor` gets `TraversalOptions` with `alwaysVisitPrimitiveFields`.
- `JSONEncodingOptions` gets `alwaysPrintPrimitiveFields`

The JSON value lines up with the two other names we already have that start `alwaysPrint…`, and the naming gnerally lines up with the C++ api (which is what I think we normally look to).

Thoughts?

@tbkka what makes sense to you?

tbkka:
I like Thomas' naming ideas here.

Tests/SwiftProtobufTests/Test_JSON.swift
```diff
@@ -970,6 +970,60 @@ final class Test_JSON: XCTestCase, PBTestHelpers {
         assertJSONDecodeFails("{")
         assertJSONDecodeFails("}")
     }
+  
+  func testEncodeWithDefaultValues() {
```
thomasvl, line 974: Looks like this maps to one of the C++ tests, maybe bring over a [few of the others](https://github.com/search?q=repo%3Aprotocolbuffers%2Fprotobuf+always_print_primitive_fields+path%3A*test.cc&type=code) also as they involve combining some of the options to help make sure there aren't indesired interactions?

Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -47,6 +47,10 @@ class MessageFieldGenerator: FieldGeneratorBase, FieldGenerator {
         return false
       }
     }
+  
+    var usesDefaultValueFlagForTraversal: Bool {
```
thomasvl, line 51: Generally, we don't want to have to check syntax any more (working towards Editions). Would just looking at say hasFieldPresence and if the field is repeated/map vs. singular work here?

Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -194,31 +198,59 @@ class MessageFieldGenerator: FieldGeneratorBase, FieldGenerator {
...
 
-        let varName = hasFieldPresence ? "v" : storedProperty
-
         var usesLocals = false
-        let conditional: String
+        let conditionals: [(condition: String, varName: String)]
         if isRepeated {  // Also covers maps
-            conditional = "!\(varName).isEmpty"
+            conditionals = [
+                (condition: "visitDefaultValues || !\(storedProperty).isEmpty", varName: storedProperty)
```
thomasvl, line 205: Realizing this as a total nit pick: should we flip this conditional?  The majority case is visitDefaultValues will be false, so should we check it second?

Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -194,31 +198,59 @@ class MessageFieldGenerator: FieldGeneratorBase, FieldGenerator {
...
+            ]
         } else if hasFieldPresence {
-            conditional = "let v = \(storedProperty)"
             usesLocals = true
+            if !isGroupOrMessage, fieldDescriptor.file.syntax == .proto2 {
+                conditionals = [
+                    (condition: "let v = \(storedProperty)", varName: "v"),
+                    (condition: "visitDefaultValues", varName: swiftName),
```
thomasvl, line 212: Does this need to be split to allow us the get the correct default value passed through since the stored value is nil at this point?  I wonder if we could avoid the two visit calls by crafting better generation here?  like can we make the `let v` somehow also get the default value (since the generator has it?  use `??` with a ternary?

Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -194,31 +198,59 @@ class MessageFieldGenerator: FieldGeneratorBase, FieldGenerator {
...
+            p.printIndented("try visitor.\(visitMethod)(\(traitsArg)value: \(conditional.varName), fieldNumber: \(number))")
+            if i == conditionals.count - 1 {
+                p.print("}\(suffix)")
+            }
 
-        p.print("\(prefix)if \(conditional) {")
-        p.printIndented("try visitor.\(visitMethod)(\(traitsArg)value: \(varName), fieldNumber: \(number))")
-        p.print("}\(suffix)")
+        }
+//        p.print("\(prefix)if \(conditional) {")
```
thomasvl, line 248: Leftover draft?

thomasvl (commented):
Also the test case added is in the regenerate commit, so I almost missed that addition.

Sources/protoc-gen-swift/MessageFieldGenerator.swift
mrabiciu: it would be done like this instead I think:
```swift
if visitDefaultValues || _someProperty != nil {
    try visitor.visitFloatSingularInt32(value: _someProperty ?? 0, fieldNumber: ...) 
}
```
This requires accessing `_someProperty` twice though so IDK if thats better or worse 🤷‍♂️

thomasvl: I guess I was thinking something like:

```swift
if let v = _someProperty ?? ( visitDefaultValues ? 0 : nil ) {
    try visitor.visitFloatSingularInt32(value: v, fieldNumber: ...) 
}
```

(typed in github comment field, no claims it is remotely close to valid code)

Sources/protoc-gen-swift/MessageFieldGenerator.swift
mrabiciu: There is a difference in behavior between proto2 and proto3 though. At least in the python lib which I was using for reference. In proto3 unset optional primitives are omitted from json, and in proto2 unset optional primitives are included. I'm not sure if theres a better way to do this.

thomasvl: The `Descriptor` support should already be tracking most of that: https://github.com/apple/swift-protobuf/blob/main/Sources/SwiftProtobufPluginLibrary/Descriptor.swift#L638-L642  What's the python code you're looking at for reference?

Sources/protoc-gen-swift/MessageFieldGenerator.swift
mrabiciu: Good callout, updated

Sources/protoc-gen-swift/MessageFieldGenerator.swift
mrabiciu: deleted

Sources/protoc-gen-swift/MessageFieldGenerator.swift
mrabiciu: Oh interesting! I'll try this

Sources/protoc-gen-swift/MessageFieldGenerator.swift
mrabiciu: I didn't look at python code, just observed how python behaves when serializing a message with every kind of field type.

mrabiciu: Here is an example of what I mean

Proto3 behavior

```protobuf
syntax = "proto3";

message Proto3Message {
    int32 singularInt32 = 1;
    repeated int32 repeatedInt32 = 2;
    optional int32 optionalInt32 = 3;
}
```

```python
print(MessageToJson(Proto3Message(), including_default_value_fields=True))
```
```
{
  "singularInt32": 0,
  "repeatedInt32": []
}
```

Proto2 behavior
```protobuf
message Proto2Message {
    required int32 singularInt32 = 1;
    repeated int32 repeatedInt32 = 2;
    optional int32 optionalInt32 = 3;
}
```
```python
print(MessageToJson(Proto2Message(), including_default_value_fields=True))
```
```
{
  "singularInt32": 0,
  "repeatedInt32": [],
  "optionalInt32": 0
}
```

`hasPresence` will be true for `optionalInt32` for both proto2 and proto3 but we want to print it in proto2 and omit it in proto3

Sources/SwiftProtobuf/Visitor.swift
```diff
@@ -460,10 +460,14 @@ public struct TraversalOptions {
   /// - map fields: Will always be visited, even if they are empty.
   /// - singular message and group fields: Will be visited only if they are set.
   /// - singular primitive fields: Will always be visited under proto2. Under proto3 unset optional primitive fields will not be visited.
-  public var visitDefaultValues: Bool
+  public var alwaysVisitPrimitiveFields: Bool
+    
+    public var visitDefaultValues: Bool {
```
thomasvl, line 465: Why include this?

Sources/protoc-gen-swift/MessageFieldGenerator.swift
thomasvl: I'm reaching out to the protobuf team for some advise on this to confirm the intended behavior.

Tests/SwiftProtobufTests/Test_JSON.swift
mrabiciu: Added some more tests intending to mirror what c++ is doing

Sources/SwiftProtobuf/Visitor.swift
mrabiciu: oops, left over from a refactor. Will delete.

Protos/SwiftProtobufTests/test_messages_proto3.proto
```diff
@@ -60,44 +60,65 @@ message TestAllTypesProto3 {
   }
 
   // Singular
-  int32 optional_int32 = 1;
```
thomasvl, line 63: This message and hence these fields are used for a bunch of tests, so while things pass, it likely means exact what was getting testing changes for a bunch of other things.  Probably makes sense to instead use a dedicated message (or file with message) for these specific tests so there's less chance of changing something else.

Tests/SwiftProtobufTests/Test_JSON.swift
```diff
@@ -970,7 +970,226 @@ final class Test_JSON: XCTestCase, PBTestHelpers {
...
+        + #""FieldName14":0,"#
+        + #""fieldName15":0,"#
+        + #""fieldName16":0,"#
+        + #""fieldName17":0,"#
+        + #""FieldName18":0"#
+        + "}"
+                
+        var encodingOptions = JSONEncodingOptions()
+        encodingOptions.alwaysPrintPrimitiveFields = true
+        assertJSONEncode(
```
thomasvl, line 1071: Since these don't always round trip (the configure == expected problem, it might be better/clearer to just skip the helpers changes and use do:


let msg = SwiftProtoTesting_Test3_TestAllTypesProto3()
let encoded = try msg.jsonString(options: encodingOptions)
XCTAssert(expected == encoded, "Did not encode correctly: got \(encoded) but expected \(expected)")

When when you do set some fields, use `with` when setting up `msg`.

Sources/SwiftProtobuf/Visitor.swift
```diff
@@ -445,8 +448,33 @@ public protocol Visitor {
   mutating func visitUnknown(bytes: Data) throws
 }
 
+/// Provides options for how visitor traversal should be carried out
+public struct TraversalOptions {
+  static let `default` = TraversalOptions()
```
thomasvl, line 453: @tbkka (and others) should we do this in all the different `*Options` classes?

I think we mostly just have the default values correct so an empty initializer works, but maybe this makes sense also? And should it be `public`?

Sources/SwiftProtobuf/Visitor.swift
```diff
@@ -445,8 +448,33 @@ public protocol Visitor {
...
+  /// Defaults to `false`.
+  ///
+  /// ### When set to true:
+  /// - repeated fields: Will always be visited, even if they are empty.
+  /// - map fields: Will always be visited, even if they are empty.
+  /// - singular message and group fields: Will be visited only if they are set.
+  /// - singular primitive fields: Will always be visited under proto2. Under proto3 unset optional primitive fields will not be visited.
+  public var alwaysVisitPrimitiveFields: Bool
+
+  public init(alwaysVisitPrimitiveFields: Bool = false) {
```
thomasvl, line 465: @tbkka (and others) - I think we've mostly left off doing an initializer like this because it could be somewhat fragile if options get added.  Should we just skip listing it and do `public var alwaysVisitPrimitiveFields: Bool = false` like we've done in all the other `*Options`?

Sources/protoc-gen-swift/MessageFieldGenerator.swift
thomasvl: Small progress update on this - So the initial response was that they team would have expected proto2 and proto3 syntax optionals to align, but as you noticed, they don't, and there are even some tests across the languages the essentially check that they don't match.  So I'm currently trying to see what they then means, are they going to bring things in line with that they actually intended or does one need to duplicate the current behavior.

So generally, I'd say I'm pretty happen with the CL, I'd like some of the other maintainers to wait in one some of the style things I asked about also.  Realistically, given the time I hear, I fear we might not get the answer about what behavior we actually should implement until after the holiday, but I'll relay along what I find out as soon as I can.

Sources/SwiftProtobuf/Visitor.swift
allevato: IMO, this is redundant with the no-argument constructor. A `default` static property makes more sense in class-based APIs like `FileManager` where it represents an instance that's lazily created and cached for future invocations. This is just a simple value construction, so it doesn't need it.

Sources/SwiftProtobuf/Visitor.swift
mrabiciu: As a user of the library I find it annoying to have to create a `var` in order to be able to set some of the flags. If you want to keep the empty init then would you be opposed to duplicating the `static func with(_ populator: (inout Self) -> ()) rethrows -> Self` pattern from Message here?

Sources/protoc-gen-swift/MessageFieldGenerator.swift
mrabiciu: That sounds good, thank you for the update! I was also surprised by this behavior too honestly. IMO the _correct_ behavior for proto2 should be to include fields if they are marked as `required` and exclude unset `optional` fields.

Sources/SwiftProtobuf/Visitor.swift
mrabiciu: Good point about it not being necessary for value types. I wasn't sure if that was the case. I'll remove.

Sources/SwiftProtobuf/Visitor.swift
mrabiciu: Added this utility ☝️. LMK if you'd like me to remove it.

Sources/SwiftProtobuf/Message.swift
```diff
@@ -31,7 +31,7 @@
 ///
 /// The actual functionality is implemented either in the generated code or in
 /// default implementations of the below methods and properties.
-public protocol Message: _CommonMessageConformances {
+public protocol Message: _CommonMessageConformances, PopulatorConstructable {
```
thomasvl, line 34: You might want to reference back to https://github.com/apple/swift-protobuf/pull/1498 and the things that references, I'm not sure if this would have size impacts we don't want or not.

Sources/SwiftProtobuf/PopulatorConstructable.swift
```diff
@@ -0,0 +1,46 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
+//
+
+
+import Foundation
+
+public protocol PopulatorConstructable {
```
thomasvl, line 13: I'll defer to others if it's worth doing this as a protocol, but the general idea does seem like it might make sense to support it on all of our `*Options` types (either via a protocol or by just manually adding it.

allevato: My own opinion:

I'm personally not a huge fan of ad hoc APIs like this (as the person who introduced the `with` helper for `Message`, I regret it today). There's been some recent discussion on the Swift forums about solving this in a more language-native way, and while this isn't something that's on the near horizon, I'd still prefer to not introduce another API that might have a better solution later on; having to declare a new `var` isn't _terribly_ onerous.

In general, when introducing a new protocol, a good question to ask is "what algorithms would I write generically over types that have that conformance requirement?" In this case, I think the answer is "none" aside from the `with` API. Providing this protocol tempts users to conform to it themselves or write generic code that uses it, which isn't something we want. In fact, I think it might be the case that a user seeing `.with` used in protobuf code would make it very easy for them to think the options types were _themselves_ protobufs by association, because of the existing inertia around that method.

tbkka: I agree with Tony about this protocol:  It doesn't really serve any purpose since there's little point in trying to build generic or existential code around it.  (The closure has to know exactly what fields the type supports, so the closure cannot itself be generic.)  Since every protocol conformance generates runtime metadata specifically to support existential or generic code, we should try to avoid protocols that aren't necessary. That's independent of whether `.with()` is a good API for these option structs.

Regarding `.with()`:   We added `.with()` to the message API in part because the obvious alternative -- using initializers with default arguments -- does not scale well to objects with dozens or hundreds of optional fields.  That's not an issue for these option structs.  The other motivating concern was around the appropriate ordering of such arguments -- again, that's not an issue for these option structs.  We can easily create static values of the option structs and/or initializers with default arguments and/or multiple initializers for common use cases.

(I'm sympathetic to Tony's argument that creating a `var` isn't that much friction, though I'm not entirely convinced I agree. :grin:)

mrabiciu: I still think theres value in being able to declare an immutable `*Options` instance in a single statement without having to resort to a closure. It doesn't really matter to me whether thats done through an initializer with default values or through a `with` constructor but I do think its valuable nonetheless. I think its easier to parse, and I think it provides safety guaranteeing that the value will not be accidentally mutated elsewhere in the function. 

The protocol was just an attempt to consolidate the logic but I'm happy to just duplicate the `with` implementation across different`*Options` or to create a non-empty initializer with default values.

mrabiciu: Removed the protocol for the sake of this PR

Sources/protoc-gen-swift/MessageFieldGenerator.swift
thomasvl: No update on this protobuf team side; with the holidays, that sorta makes sense. I'll relay things as soon as I hear them.

Re `required`, the name has always bugged me, because it only has meaning in the context of the _binary encoding format_.  For TextFormat (or JSON), it has no meaning, so we'll see where they end up, but those fields do have _presence_, it's just the extra validation for the binary format that really cares about `required`.

mrabiciu: I see okay, keep me posted on the desired functionality 🙏

mrabiciu: @thomasvl any updates?

thomasvl: I sent another ping today asking for any updates. 🤞

thomasvl: So it looks like the final direction is going to be the option should mean _Emit all without-presence fields_. So proto2 syntax `required` and `optional` don't get output, and proto3 syntax `optional` don't get output.  This would mean Python, C++, and Java aren't currently to spec, so they are scoping getting their implementations changed to match.  I've asked about when we might see a docs update to better capture the intent.

mrabiciu: Sounds good I can work on that. Thank you!

thomasvl: Looks like a docs rollout is happening:

https://protobuf.dev/programming-guides/proto2/#json-options

> **Always emit fields without presence:** Fields that don’t support presence and that have their default value are omitted by default in JSON output (for example, an implicit presence integer with a 0 value, implicit presence string fields that are empty strings, and empty repeated and map fields). An implementation may provide an option to override this behavior and output fields with their default values.
> 
> As of v25.x, the C++, Java, and Python implementations are nonconformant, as this flag affects proto2 `optional` fields but not proto3 `optional` fields. A fix is planned for a future release.

https://protobuf.dev/programming-guides/proto3/#json-options

> **Always emit fields without presence:** Fields that don’t support presence and that have their default value are omitted by default in JSON output (for example, an implicit presence integer with a 0 value, implicit presence string fields that are empty strings, and empty repeated and map fields). An implementation may provide an option to override this behavior and output fields with their default values.

mrabiciu: I'm confused by those docs for proto2
> for example, an implicit presence integer with a 0 value, implicit presence string fields that are empty strings, and empty repeated and map fields

Don't integer and string fields always have presence under proto2? Under what scenario would we actually emit these in proto2?

mrabiciu: I've updated the logic to just use `hasPresence` for proto2 the same way as in proto3. But I think those docs could use some clarification.

thomasvl: Please feel free to open an issue upstream with the feedback.  Some times it helps to hear directly from devs.

Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -36,6 +36,7 @@ class MessageFieldGenerator: FieldGeneratorBase, FieldGenerator {
 
     private var isMap: Bool {return fieldDescriptor.isMap}
     private var isPacked: Bool { return fieldDescriptor.isPacked }
+    private var isEnum: Bool { return fieldDescriptor.enumType != nil }
```
thomasvl, line 39: Is this still needed?

Sources/SwiftProtobuf/JSONEncodingOptions.swift
```diff
@@ -40,6 +40,13 @@ public struct JSONEncodingOptions: Sendable {
   /// by keys in lexographical order. This is an implementation detail
   /// and subject to change.
   public var useDeterministicOrdering: Bool = false
+    
+  /// Include fields that are equal to the default value in the JSON encoding.
+  ///
+  /// If `true`, singular primitive fields, repeated fields, and map fields will always be serialized (unset optional fields will not be serialized under proto3).
```
thomasvl, line 46: This probably needs a tweak to line up with the public docs (and behavior).

thomasvl:
@mrabiciu looks like you might need to sync also to resolve some conflicts.

mrabiciu:
Rebased on main and squashed all my commits

ChristopherDrum:
@thomasvl It's been 10 months now. Was there a blocker on this? It seemed at the time like it was ready to merge in. Has this been supplanted by another PR or methodology? Any guidance from the team on how we should deal with this situation in the meantime?

thomasvl:
New apis/additions to the library generally get signoff from someone at Apple since they own the repo.

Currently they want to ensure any changes/additions to the repo are non breaking as a major version bump could cause issues for downstream packages that also depend on this repo.

FranzBusch (approved):
This seems like a sensible change to me. This is only additive in adding new encoding options and it adds new test cases to check that this option is working correctly. The change is non breaking since we provide a default implementation of the new protocol requirement.

gjcairo:
@tbkka left a comment [here](https://github.com/apple/swift-protobuf/pull/1504#issuecomment-1828910103) in 2023 for another change that also only adds new APIs, but he mentioned it's breaking because it's breaking between the runtime and the generated code (kind of similarly to this PR, new API was added to the `Message` protocol, which means the generated code has to be re-generated to build).

I agree there are only API additions with default implementations on this PR, but if you have mismatching versions for the generated code and SwiftProtobuf (specifically, newer generated code running with an older swift-protobuf), things will break because the generated code now references a new field in the `Visitor`.

FWIW, I don't think these count as breaking changes (so I agree with @FranzBusch approving this): if someone went to the trouble of updating the generated code with a newer version of the plugin, why would they keep the old version of the runtime (or vice versa)? As long as users are updating both things simultaneously they should be safe, and I don't see why this wouldn't happen, to be honest.

@thomasvl In the README we suggest that you should try to keep the versions of `SwiftProtobuf` and `protoc-gen-swift` matching, but have we always been clear about what we officially support here?

thomasvl:
I guess it isn't a "break", but the potential catch I can see is if someone depends on protos compiled into another package, they would need those also updated for this option to work since the support is captured in the generated code.

The only other thing I can think of that has been a concern on additions is code size increase, and this does increase the generated code size (which I believe is what pushed some of the investigation into table based generation to try and factor things like this out).

Sources/SwiftProtobuf/Visitor.swift
```diff
@@ -445,8 +448,30 @@ public protocol Visitor {
...
+public struct TraversalOptions {
+
+  /// Determines if non-optional fields that are equal to their default values should be visited.
+  /// Defaults to `false`.
+  ///
+  /// ### When set to true:
+  /// - repeated fields: Will always be visited, even if they are empty.
+  /// - map fields: Will always be visited, even if they are empty.
+  /// - singular message and group fields: Will be visited only if they are set.
+  /// - singular primitive fields: Will always be visited under proto2. Under proto3 unset optional primitive fields will not be visited.
```
thomasvl, line 461: This doc string likely needs an update since everything is about Editions now upstream.

PR #1527, opened 2023-12-22
Make `anyTranscodeFailure` more descriptive

rebello95:
Right now it's quite difficult to identify _which type_ is missing from the `Google_Protobuf_Any` registry when the `anyTranscodeFailure` is thrown without having Xcode and breakpoints to debug the program. This change makes the error more descriptive and actionable by including the type URL that needs to be registered.

Errors now look like this:

> Swift/ErrorType.swift:200: Fatal error: Error raised at top level: SwiftProtobuf.BinaryEncodingError.anyTypeURLNotRegistered(typeURL: "type.googleapis.com/connectrpc.conformance.v1.IdempotentUnaryRequest")

### DIFF
--- a/Sources/SwiftProtobuf/BinaryEncodingError.swift
+++ b/Sources/SwiftProtobuf/BinaryEncodingError.swift
@@ -13,18 +13,18 @@
 // -----------------------------------------------------------------------------
 
 /// Describes errors that can occur when decoding a message from binary format.
-public enum BinaryEncodingError: Error {
+public enum BinaryEncodingError: Error, Equatable {
   /// `Any` fields that were decoded from JSON cannot be re-encoded to binary
   /// unless the object they hold is a well-known type or a type registered via
   /// `Google_Protobuf_Any.register()`.
+  case anyTypeURLNotRegistered(typeURL: String)
+  /// An unexpected failure when deserializing a `Google_Protobuf_Any`.
   case anyTranscodeFailure
-
   /// The definition of the message or one of its nested messages has required
   /// fields but the message being encoded did not include values for them. You
   /// must pass `partial: true` during encoding if you wish to explicitly ignore
   /// missing required fields.
   case missingRequiredFields
-
   /// Messages are limited to a maximum of 2GB in encoded size.
   case tooLarge
 }
--- a/Sources/SwiftProtobuf/JSONEncodingError.swift
+++ b/Sources/SwiftProtobuf/JSONEncodingError.swift
@@ -12,12 +12,12 @@
 ///
 // -----------------------------------------------------------------------------
 
-public enum JSONEncodingError: Error {
+public enum JSONEncodingError: Error, Equatable {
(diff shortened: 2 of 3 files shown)


Sources/SwiftProtobuf/BinaryEncodingError.swift
```diff
@@ -13,18 +13,18 @@
 // -----------------------------------------------------------------------------
 
 /// Describes errors that can occur when decoding a message from binary format.
-public enum BinaryEncodingError: Error {
+public enum BinaryEncodingError: Error, Equatable {
```
thomasvl, line 16: What caused this to be needed?

Sources/SwiftProtobuf/JSONEncodingError.swift
```diff
@@ -12,12 +12,12 @@
 ///
 // -----------------------------------------------------------------------------
 
-public enum JSONEncodingError: Error {
+public enum JSONEncodingError: Error, Equatable {
```
thomasvl, line 15: Same

thomasvl:
Forgot to mention, I'm good with this change, it's fine on `main` because we're working towards 2.x there, so changing the errors is safe.

@tbkka - you good with this change to the errors?

Sources/SwiftProtobuf/JSONEncodingError.swift
rebello95: Without it, I get the following error when doing `swift test`:

```
/Users/michael/Development/swift-protobuf/Tests/SwiftProtobufTests/Test_JSON_Conformance.swift:298:7: error: global function 'XCTAssertEqual(_:_:_:file:line:)' requires that 'JSONEncodingError' conform to 'Equatable'
      XCTAssertEqual($0 as? JSONEncodingError,
      ^
Swift.Optional:1:11: note: requirement from conditional conformance of 'Optional<JSONEncodingError>' to 'Equatable'
extension Optional : Equatable where Wrapped : Equatable {
          ^
```

Sources/SwiftProtobuf/BinaryEncodingError.swift
rebello95: https://github.com/apple/swift-protobuf/pull/1527#discussion_r1440116333 was the reason, though I only see an error when `JSONEncodingError` doesn't conform to `Equatable`. `BinaryEncodingError` could technically stay as-is, but I think this is a reasonable change to have parity

Sources/SwiftProtobuf/JSONEncodingError.swift
thomasvl: Thanks. I figured it was going to be implicit equatable support, but wanted to confirm.

tbkka:
Yeah, this is a nice improvement, but definitely needs to be held for 2.x.

thomasvl (approved):

thomasvl:
Thanks for the patch!

PR #1531, opened 2024-01-02
Don't allow implementationOnly with package visibility.

thomasvl:
Follow up to #1472, looks like we didn't realize this gate should also be updated.


thomasvl:
@gjcairo fyi.

FranzBusch (approved):

gjcairo (approved):
Good catch Thomas, thanks for the fix!

PR #1535, opened 2024-01-05
Module mappings and plugin options validation improvements

thomasvl:


Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
```diff
@@ -38,6 +38,13 @@ public struct ProtoFileToModuleMappings {
   /// access it to verify things.
   let mappings: [String:String]
 
+  /// A Boolean value that indicates that there were developer provided
+  /// mappings.
+  ///
+  /// Since `mappings` will have the bundled proto files also, this is used
+  /// to track there there are any provided mappings.
```
allevato, line 45: `to track whether there are...`

Sources/protoc-gen-swift/GeneratorOptions.swift
```diff
@@ -130,5 +130,24 @@ class GeneratorOptions {
...
+    if self.implementationOnlyImports && self.visibility != .internal {
+      throw GenerationError.message(message: """
+        Cannot use @_implementationOnly imports when the proto visibility is public or package.
+        Either change the visibility to internal, or disable @_implementationOnly imports.
+        """)
+    }
+
+    // The majority case is that if `self.protoToModuleMappings.hasMappings` is
+    // true, then `self.visibility` should be either `.public` or `.package`.
+    // However, it is possible for someone to put the "leaf" proto files in a
```
allevato, line 146: The word "leaf" here is confusing to me; I would think of the leaf files as the ones without any dependencies and would think of your "leaf" as the "root".

It might be clearer with just a concrete example ("A" depends on "B", B is public but A is internal).

allevato (approved):

Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
thomasvl: done

Sources/protoc-gen-swift/GeneratorOptions.swift
thomasvl: Revised the text including an example.

allevato (approved):

PR #1537, opened 2024-01-08
Update the proto files from upstream.

thomasvl:


allevato (approved):

thomasvl:
Updated commits to not update the files that have moved to editions, reached out to the protobuf team to check on the behavior change I'm seeing.

Issue #1540, opened 2024-01-17
WebAssembly build failure in AsyncMessageSequence


anuraaga:
I'm not sure if WebAssembly compatibility is a priority for the project but I do notice some `WASI` flags in the code so I am guessing there is some best effort support.

When compiling the latest HEAD to WebAssembly, there is a compiler failure in `AsyncMessageSequence`

```
7.006 /workspace/Sources/SwiftProtobuf/AsyncMessageSequence.swift:127:17: error: cannot find 'BinaryDelimited' in scope
7.006           throw BinaryDelimited.Error.malformedLength
7.006                 ^~~~~~~~~~~~~~~
7.006 /workspace/Sources/SwiftProtobuf/AsyncMessageSequence.swift:136:15: error: cannot find 'BinaryDelimited' in scope
7.006         throw BinaryDelimited.Error.truncated
7.006               ^~~~~~~~~~~~~~~
7.006 /workspace/Sources/SwiftProtobuf/AsyncMessageSequence.swift:158:19: error: cannot find 'BinaryDelimited' in scope
7.006             throw BinaryDelimited.Error.truncated
```

This makes sense because `BinaryDelimited` is flagged out on WASI.

https://github.com/apple/swift-protobuf/blob/0e7015750b9bd7b80f28cf997ae77ac0e6d37f0b/Sources/SwiftProtobuf/BinaryDelimited.swift#L15

Removing `AsyncMessageSequence.swift` seems to have no consequence and allows the build to pass. I wonder if that file should have a similar WASI guard on it.

You can see my build where I workaround the issue by just removing the file, and everything works swell.

https://github.com/wasilibs/go-protoc-gen-swift/blob/main/buildtools/wasm/Dockerfile#L10

thomasvl:
This was fixed via #1541

PR #1541, opened 2024-01-17
Expose the error to the `os(WASI)` build.

thomasvl:
This should let `AsyncMessageSequency.swift` build.

### DIFF
--- a/Sources/SwiftProtobuf/BinaryDelimited.swift
+++ b/Sources/SwiftProtobuf/BinaryDelimited.swift
@@ -14,6 +14,7 @@
 
 #if !os(WASI)
 import Foundation
+#endif
 
 /// Helper methods for reading/writing messages with a length prefix.
 public enum BinaryDelimited {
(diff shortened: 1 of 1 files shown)


thomasvl:
This might fix #1540

@anuraaga can you try this patch?  I don't know how to do a build to test it.

thomasvl:
@tbkka - should we try to set up any sorta CI for a WebAssembly build?  (or Windows I guess, I wonder if our CMake support is still correct/complete on `main` with all the shuffling we've been through)

anuraaga:
@thomasvl Thanks for the quick fix! I verified it with this patch and running docker build

https://github.com/wasilibs/go-protoc-gen-swift/compare/tvl-patch?expand=1

If wanting to add to CI here, a similar approach using the swiftwasm docker container is probably pretty easy to add here. I think a github actions job that sets the container to `ghcr.io/swiftwasm/swift:5.9-focal` and then just runs swift build would do it.

https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container

If there's interest, I would be happy to send a PR after this one too

allevato (approved):

PR #1544, opened 2024-02-07
Move to actions/checkout@v4 (on 1.x branch).

thomasvl:
Resolves the warnings seen on the actions output:

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,6 +69,8 @@ jobs:
       image: swift:${{ matrix.swift }}${{ matrix.ubuntu }}
     steps:
     - name: Checkout
+      # 5.1.5 and lower Swift versions end up on images where actions/checkout@v4
+      # fails because the image is too old.
       uses: actions/checkout@v3
       with:
         path: main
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -69,6 +69,8 @@ jobs:
       image: swift:${{ matrix.swift }}${{ matrix.ubuntu }}
     steps:
     - name: Checkout
+      # 5.1.5 and lower Swift versions end up on images where actions/checkout@v4
+      # fails because the image is too old.
```
thomasvl, line 73: I don't know if at some point github will be making these start to fail, so that might force the branch to drop some swift version coverage.

allevato: I guess we can wait until/if that happens. I wonder how many people would miss coverage of versions that old.

allevato (approved):

PR #1553, opened 2024-02-23
Update some comments to remove "syntax".

thomasvl:
With Editions, some behaviors aren't specific to syntax any more and instead based on the Features enabled instead, so update some of the comments to be correct going forward.


Tests/SwiftProtobufTests/Test_BinaryDecodingOptions.swift
```diff
@@ -267,8 +267,8 @@ final class Test_BinaryDecodingOptions: XCTestCase {
             XCTAssertTrue(msg2.optionalGroup.unknownFields.data.isEmpty)
         }
 
-        // An unknown enum value. proto2 syntax unknown enums uses a different code
-        // path to end up in unknown fields, so ensure that is honoring the option.
+        // An unknown enum value. Close enums uses a different code path and
```
allevato, line 270: Is the terminology close or closed? (Here and below)

allevato (approved):

PR #1558, opened 2024-02-26
`CodeGenerator` support for response edition information.

thomasvl:
- Add api to get supported editions range.
- If the generator signals editions support, add the range into the generation response.
- Make the protos Editions enum Comparable since the docs document it and it then will allow other code paths to do version comparisons.


allevato (approved):

thomasvl:
Oops, pulled back some of the extension apis and just marked them deprecated, so help eventual adoption.

Issue #1560, opened 2024-02-29
New concurrency warnings when building with Swift 5.10


allevato:
As a consequence of [SE-0412: Strict concurrency for global variables](https://github.com/apple/swift-evolution/blob/main/proposals/0412-strict-concurrency-for-global-variables.md), we're getting new concurrency warnings when building swift-protobuf. They can be narrowed down to two places:

Default instances of storage classes:

```
swift-protobuf/Sources/SwiftProtobuf/AnyMessageStorage.swift:144:14: error: static property 'defaultInstance'
is not concurrency-safe because it is not either conforming to 'Sendable' or isolated to a global actor; this
is an error in Swift 6
  static let defaultInstance = AnyMessageStorage()
             ^

swift-protobuf/Sources/SwiftProtobuf/descriptor.pb.swift:3111:16: error: static property 'defaultInstance' is
not concurrency-safe because it is not either conforming to 'Sendable' or isolated to a global actor; this is
an error in Swift 6
    static let defaultInstance = _StorageClass()
             ^
```

The `_NameMap` of each generated message:

```
swift-protobuf/Sources/SwiftProtobuf/any.pb.swift:195:21: error: static property '_protobuf_nameMap' is not
concurrency-safe because it is not either conforming to 'Sendable' or isolated to a global actor; this is an
error in Swift 6
  public static let _protobuf_nameMap: _NameMap = [
                    ^
```

SE-0412 added the requirements that:

> Under strict concurrency checking, require every global variable to either be isolated to a global actor or be both:
>
> 1. immutable
> 2. of `Sendable` type

Neither the storage classes nor `_NameMap` are `Sendable`. I need some more eyes who are better versed with Swift concurrency than I am 🙂 

## Storage classes

For the storage classes, we _could_ just drop the `defaultInstance` property entirely and just allocate a new empty instance. However, it _is_ a nice optimization to defer allocation of storage until the message is actually mutated. It's not super important if you're creating a new message and then immediately parsing or mutating it, but anywhere that we return an empty message as a default value of an unset message-typed field, it's very nice to not have to allocate storage if the user is only going to read from it.

Can we just mark the storage classes as `Sendable` on `@unchecked Sendable` since the messages themselves already conform to that, and the storage types can't be accessed from outside the message?

## _NameMap

[...]

PR #1563, opened 2024-02-29
For Swift 5.10, mark Storage defaultInstance `nonisolated(unsafe)`

thomasvl:
Part of the issues in #1560

### DIFF
--- a/Sources/protoc-gen-swift/MessageStorageClassGenerator.swift
+++ b/Sources/protoc-gen-swift/MessageStorageClassGenerator.swift
@@ -49,7 +49,15 @@ class MessageStorageClassGenerator {
       // readonly usage never causes the allocation.
       p.print("""
 
-          static let defaultInstance = _StorageClass()
+          #if swift(>=5.10)
+            // This property is used as the initial default value for new instances of the type.
+            // The type itself is protecting the reference to its storage via CoW semantics.
+            // This will force a copy to be made of this reference when the first mutation occurs;
+            // hence, it is safe to mark this as `nonisolated(unsafe)`.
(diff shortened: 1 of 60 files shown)


allevato (approved):
So unfortunate that we have to repeat the whole declaration. 🙁

tbkka (approved):

Sources/protoc-gen-swift/MessageStorageClassGenerator.swift
```diff
@@ -49,7 +49,11 @@ class MessageStorageClassGenerator {
       // readonly usage never causes the allocation.
       p.print("""
 
-          static let defaultInstance = _StorageClass()
+          #if swift(>=5.10)
+            static nonisolated(unsafe) let defaultInstance = _StorageClass()
```
FranzBusch, line 53: Can we generate a code comment here detailing why we use `nonisolated(unsafe)` and why that usage is in fact safe.

thomasvl: Sure, what would you like to it to say?

FranzBusch: Something like this
```
// This property is used as the initial default value for new instances of the type.
// The type itself is protecting the reference to its storage via CoW semantics. This will force a copy to be made
// of this reference when the first mutation occurs; hence, it is safe to mark this as `nonisolated(unsafe)`.
```

thomasvl: Done.

thomasvl:
Once we sort out rest of the warning, we'll probably want to pull these all to the 1.x branch.

Issue #1561, opened 2024-02-29
Revisit `_NameMap` in light of Swift 5 changing the default `String` encoding


thomasvl:
For a while now I've been wondering if `_NameMap` should get a revisit since [Swift 5 went UTF-8](https://www.swift.org/blog/utf8-string/).  We likely could drop all the allocations related to the `InternPool`, and we might actually end up at the point the compiler can just realized everything is concurrency safe again.

But given the Swift 5.10 new concurrency warnings around it (#1560), it seems like that might be even more reason to consider revisiting/simplifying it.

PR #1564, opened 2024-02-29
For Swift 5.10, mark Message/Enum _protobuf_nameMap `nonisolated(unsafe)`

thomasvl:
I think this is the rest of the issues in https://github.com/apple/swift-protobuf/issues/1560


tbkka (approved):

Sources/protoc-gen-swift/EnumGenerator.swift
```diff
@@ -140,18 +140,29 @@ class EnumGenerator {
   private func generateProtoNameProviding(printer p: inout CodePrinter) {
     let visibility = generatorOptions.visibilitySourceSnippet
 
-    p.print("\(visibility)static let _protobuf_nameMap: \(namer.swiftProtobufModulePrefix)_NameMap = [")
+    p.print("""
+        #if swift(>=5.10)
+          \(visibility)static nonisolated(unsafe) let _protobuf_nameMap: \(namer.swiftProtobufModulePrefix)_NameMap = _makeNameMap()
```
Lukasa, line 145: The fact that these might be public makes me a touch nervous. Why are these public again?

FranzBusch: I agree. Since they are public we cannot use `nonisolated(unsafe)` here since anybody could grab those; however, from looking at `_NameMap` this appears to be an immutable struct and the only reason that it isn't `Sendable` right now is due to the storage of a few `UnsafeRawBufferPointer`. We really should make the underlying type `Sendable` here by either coming up with an implementation that is `Sendable` by default or by double-checking that this is indeed immutable and then adding an `@unchecked Sendable` to the type.

thomasvl: I think revisiting the impl given how `String` has changed since the Swift 3 days (which is where I believe @tbkka `_NameMap` work goes back to) is likely a good idea, that's why I opened #1561.

If we push ahead with that in the coming weeks, this would likely just be sorter lived.

As for why they are `public`, it's because they are an api point that the Runtime needs to be able to invoke, and I don't believe we ever found a better way to handle that – i.e. - generated protocol conformance that the core runtime can call things on.

Lukasa: It's unfortunate that the need for these to satisfy a protocol witness forces them to be public. Ideally these conformances would be on a type that didn't need to be public, though I understand why they are.

thomasvl: > Until then I would recommend using `@unchecked Sendable` on the type instead of `nonisolated(unsafe) let` on the global

Is that likely to have to ripple through the support types also?  i.e. - `Name`, `InternPool`?

I'm doing this somewhat blind (need to do Xcode updates to locally test), so this PR might have to sit a little while until so I can look at switching it over.

thomasvl:
Closing this as I believe #1577 is the desired direction now.

PR #1565, opened 2024-03-01
Include "empty" privacy manifest file

### LINKED ISSUE #1529: xcprivacy file for this package
As per the new guidelines Third-party SDKs need to provide their own privacy manifest files. If this package doesn't have. Please let me know why it is not there and if the app review fails how can we justify this.

tbkka:
Include it as a resource in both SwiftProtobuf library and SwiftProtobufPluginLibrary when built via podspec or swiftpm

Resolves #1529

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -38,12 +38,14 @@ let package = Package(
     .target(
         name: "SwiftProtobuf",
         exclude: ["CMakeLists.txt"],
+	resources: [.copy("PrivacyInfo.xcprivacy")],
         swiftSettings: .packageSettings
     ),
     .target(
         name: "SwiftProtobufPluginLibrary",
         dependencies: ["SwiftProtobuf"],
         exclude: ["CMakeLists.txt"],
+	resources: [.copy("PrivacyInfo.xcprivacy")],
         swiftSettings: .packageSettings
     ),
     .target(
--- a/Package.swift
+++ b/Package.swift
@@ -38,14 +38,14 @@ let package = Package(
     .target(
         name: "SwiftProtobuf",
         exclude: ["CMakeLists.txt"],
-	resources: [.copy("PrivacyInfo.xcprivacy")],
+        resources: [.copy("PrivacyInfo.xcprivacy")],
         swiftSettings: .packageSettings
     ),
     .target(
         name: "SwiftProtobufPluginLibrary",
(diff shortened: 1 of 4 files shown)


nabs-m (approved):
Looks good. I'm just not sure if you meant to reference `Sources/SwiftProtobufPluginLibrary/PrivacyInfo.xcprivacy` somewhere, or just `Sources/SwiftProtobuf/PrivacyInfo.xcprivacy` like you did.

thomasvl (approved):
Does the other Package-*.swift support resources?  i.e. - do we need to add it to the other file?

SwiftProtobuf.podspec
```diff
@@ -16,6 +16,7 @@ Pod::Spec.new do |s|
   s.cocoapods_version = '>= 1.7.0'
 
   s.source_files = 'Sources/SwiftProtobuf/**/*.swift'
+  s.resource_bundle = {'SwiftProtobuf' => ['Sources/SwiftProtobuf/PrivacyInfo.xcprivacy']}
```
thomasvl, line 19: btw - Most things I've seen name this with `*_Privacy` (so `SwiftProtobuf_Privacy`), I'm not sure if it really matters or if you want to follow that convention.

Package.swift
```diff
@@ -38,12 +38,14 @@ let package = Package(
     .target(
         name: "SwiftProtobuf",
         exclude: ["CMakeLists.txt"],
+	resources: [.copy("PrivacyInfo.xcprivacy")],
```
thomasvl, line 41: Oh, looks there might be a tabs vs spacing things going on in this file.

SwiftProtobuf.podspec
tbkka: I really don't know.   The format I've used here seems to be commonly recommended in a few places.  I've yet to get any information from folks at Apple, though.

Package.swift
tbkka: Ooops.  Fixed.

tbkka:
> Does the other Package-*.swift support resources? i.e. - do we need to add it to the other file?

I don't think so.  Someone would have to be compiling an app for Apple's App Store with a Swift prior to 5.8 for it to make any difference.

tbkka:
I presume we also need this in the 1.x branch?

PR #1566, opened 2024-03-01
[1_x_release_branch] Include "empty" privacy manifest file

tbkka:
Include it as a resource in both SwiftProtobuf library and SwiftProtobufPluginLibrary when built via podspec or swiftpm

Cherry-pick from #1565


FranzBusch (approved):

tbkka:
This seems to DTRT for Swift 5.6 and later.  The older `Package.swift` can't support resources, but that shouldn't be an issue for folks submitting to Apple's App Store (which is what the privacy manifests are for).

thomasvl (approved):

Issue #1573, opened 2024-03-14
The fuzz regression tests seem to have stopped working.


thomasvl:
As seen on #1570 -

_fuzzing_regressions release_ passed on the first attempt of #1570: https://github.com/apple/swift-protobuf/actions/runs/8282282853, but on the second push it failed.

Repeating a test on `main` show the same failure for fuzz release: https://github.com/apple/swift-protobuf/actions/runs/8283116214

thomasvl:
From #1574, this might actually be something in the base image and not specific to the Swift bits.

thomasvl:
#1579 changes the `runs-on` image to address the issue.

Issue #1571, opened 2024-03-14
The "thread" sanitizer testing is failing


thomasvl:
Attempting to run with 5.10 on the github CI fails for both debug and release with:

```
FATAL: ThreadSanitizer: unexpected memory mapping 0x644ed6848000-0x644ed7cbe000
```

thomasvl:
From #1574, this is looking more like a base image issue and not specific to 5.10.

thomasvl:
Google Search for _ThreadSanitizer: unexpected memory mapping_ in the last week turned up: https://www.mail-archive.com/kernel-packages@lists.launchpad.net/msg522188.html

Lukasa:
That seems highly likely to be relevant.

thomasvl:
https://github.com/google/sanitizers/issues/1716 has a bunch of reference that might be useful to work around things.

thomasvl:
https://github.com/actions/runner-images/issues/9491 also might be useful to track.

mkruskal-google:
FWIW, I've traced Protobuf's similar issue back to github's 20240310.1.0 release of ubuntu-22. Downgrading to ubuntu-20 seems to workaround it. Historically, there have been a handful of linux kernel updates that broke sanitizers in this way, so it seems like a likely culprit since they did upgrade to 6.5.0-1016-azure

Lukasa:
That seems like a likely diagnosis.

PR #1574, opened 2024-03-14
Try pinning some testing to Swift 5.9.2.

thomasvl:
With the 5.10 release, we're seeing some test fail (#1573 and #1571), try pinning to 5.9.2 since CI from last week where `swift:latest` was that still worked.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -141,7 +141,9 @@ jobs:
         swiftpm_config: ["debug", "release"]
     container:
       # Test on the latest Swift release.
-      image: swift:latest
+      # image: swift:latest
+      # As of 5.10.0, the "thread" testing fails. Back to what last worked:
+      image: swift:5.9.2-jammy
     steps:
     - uses: actions/checkout@v4
     - name: Test
@@ -170,7 +172,9 @@ jobs:
         swiftpm_config: ["debug", "release"]
     container:
       # Test on the latest Swift release.
-      image: swift:latest
+      # image: swift:latest
+      # As of 5.10.0, the "thread" testing fails. Back to what last worked:
+      image: swift:5.9.2-jammy
     steps:
     - uses: actions/checkout@v4
     - name: Build


thomasvl:
@FranzBusch @Lukasa @tbkka -

Now I'm even more confused, on `main` pinning to Swift 5.9.2 also doesn't work, but now _fuzzing_regressions_ fails for _debug_. And even the _address_ version of _sanitizer_testing_ fails.

I'm guessing this isn't a good sign for fuzz testing in generally and the oss-fuzz tests might have trouble in the future. (But that might take someone manually updating their images to have a newer Swift, I don't fully recall that process at the moment.)

Lukasa:
Sort of amazingly, locally I see failures in `(address, debug)` when pinning to 5.9.2. That appears to be the only configuration that _doesn't_ fail in CI.

Lukasa:
I definitely don't reproduce the _segfault_ here though.

Lukasa:
So weird. The TSAN build fails with an error that _looks_ like google/sanitizers#503. But that was from an ancient kernel, with the implication that LLVM upstream fixed this long ago.

Lukasa:
I don't reproduce the TSAN issue in local docker on arm64. I reproduce something very similar to it, but that is probably a Rosetta issue. Let me see if I can find an Intel machine.

Lukasa:
Ok, I do not reproduce the TSAN issue on Intel either (run inside a VM, but _not_ a VM running the same kernel as the github actions runners.

So I wonder if the TSAN issue is related to a kernel update?

Lukasa:
Even if it's not a kernel update, the _most likely_ assumption is that something weird has changed in `ubuntu-latest`. I think it's most likely to be a kernel issue, but the linker is another tempting suspect, at least for tsan.

thomasvl:
> Even if it's not a kernel update, the _most likely_ assumption is that something weird has changed in `ubuntu-latest`. I think it's most likely to be a kernel issue, but the linker is another tempting suspect, at least for tsan.

I haven't tried a general search, but I just noted this: https://github.com/protocolbuffers/protobuf/pull/16185, so it may infact be a more general issue.

thomasvl:
> Hmm, any chance that upstream can weigh in? @thomasvl do you have anyone to ask for details about that issue?

I'm keeping an eye on it, but at this point I think they don't know either. They appear to bejust disabling the test so developers don't get slowed down by a failure they can't do anything about.

thomasvl:
I'm going to close this PR since it seems pinning the swift version isn't going to help here.  We should probably put any future investigations on #1571 and #1573

PR #1576, opened 2024-03-15
make fuzz and sanitizer tests not break CI.

thomasvl:
Mark the build steps as allowed to fail while we work out what's happening since it doesn't seem related to Swift.


thomasvl:
@Lukasa should this go to the 1.x branch also?

Lukasa:
May as well.

PR #1578, opened 2024-03-15
make fuzz and sanitizer tests not break CI.

thomasvl:
Mark the build steps as allowed to fail while we work out what's happening since it doesn't seem related to Swift.

This is pulling #1576 to the 1.x branch.


Lukasa (approved):

tkafka:
Disabling the sanitizer tests has a strong `xz backdoor` vibes, are we sure this is okay?

PR #1583, opened 2024-03-19
Pull the concurrency warnings fixes to the 1.x branch

thomasvl:


thomasvl:
I figured if we're likely to do a release soon we should pick these up as they impact generated code and thus could impact folks adopting the warnings in there own packages as the generated code will be compiled in those contexts.

Lukasa (approved):

PR #1586, opened 2024-03-19
Only include PrivacyInfo.xcprivacy on some platforms.

thomasvl:
AppleStore only requires the informatiion on iOS, watchOS, tvOS, visionOS; so exclude the resources for Linux and macOS.

As of Swift 5.10 (#1560), the generated Linux code when there are resources gets flagged with issues when using `StrictConcurrency=complete`; so this dodges the issue.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -11,6 +11,15 @@
 
 import PackageDescription
 
+// NOTE: Once all supported Swift versions support visionOS, this gate could be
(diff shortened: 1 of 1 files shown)


thomasvl:
Dropping this since cross compile from macOS to for linux doesn't really work with this, and since there are other issues for for a linux build anyway; it's not like this would enable -warnings-as-errors.

We can always revisit something like this in the future.

PR #1589, opened 2024-03-19
Fix some typos

mrs1669:
I found a few typos.
I would be grateful if you could confirm. I hope this helps!

### DIFF
--- a/CompileTests/MultiModule/Sources/ModuleA/a.pb.swift
+++ b/CompileTests/MultiModule/Sources/ModuleA/a.pb.swift
@@ -75,7 +75,7 @@ public struct A: SwiftProtobuf.ExtensibleMessage, Sendable {
 
 // MARK: - Extension Properties
 
-// Swift Extensions on the exteneded Messages to add easy access to the declared
+// Swift Extensions on the extended Messages to add easy access to the declared
 // extension fields. The names are based on the extension field name from the proto
 // declaration. To avoid naming collisions, the names are prefixed with the name of
 // the scope where the extend directive occurs.
--- a/CompileTests/MultiModule/Sources/ModuleA/a.pb.swift
+++ b/CompileTests/MultiModule/Sources/ModuleA/a.pb.swift
@@ -75,7 +75,7 @@ public struct A: SwiftProtobuf.ExtensibleMessage, Sendable {
 
 // MARK: - Extension Properties
 
-// Swift Extensions on the extended Messages to add easy access to the declared
+// Swift Extensions on the exteneded Messages to add easy access to the declared
 // extension fields. The names are based on the extension field name from the proto
 // declaration. To avoid naming collisions, the names are prefixed with the name of
 // the scope where the extend directive occurs.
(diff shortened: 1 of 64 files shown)


CompileTests/MultiModule/Sources/ModuleA/a.pb.swift
```diff
@@ -75,7 +75,7 @@ public struct A: SwiftProtobuf.ExtensibleMessage, Sendable {
 
 // MARK: - Extension Properties
 
-// Swift Extensions on the exteneded Messages to add easy access to the declared
+// Swift Extensions on the extended Messages to add easy access to the declared
```
thomasvl, line 78: This should be fixed in the a.proto file otherwise it will get lose on the next generation.

FuzzTesting/Sources/FuzzCommon/fuzz_testing.pb.swift
```diff
@@ -1818,7 +1818,7 @@ public struct SwiftProtoTesting_Fuzz_RepeatedGroup_ext: Sendable {
 
 // MARK: - Extension Properties
 
-// Swift Extensions on the exteneded Messages to add easy access to the declared
+// Swift Extensions on the extended Messages to add easy access to the declared
```
thomasvl, line 1821: Same, fix the .proto fuzz_testing.proto file.

thomasvl (commented):
Thanks!

The .pb.swift files need the fixes in the .proto files in the Protos/ sub directories so they won't get lost the next time we regenerated.

You'll also need to fix/update the copies in the Reference/ directory as the CI will then catch them not matching once the .proto files are updated.

mrs1669:
Thanks for review!
Please tell me the following point that I do not understand.

> The .pb.swift files need the fixes in the .proto files in the Protos/ sub directories so they won't get lost the next time we regenerated.

I am aware that there is no modification to the .proto file since the PR is a modification of the comment out in the .pb.swift file, is this correct?
It is my understanding that [this diff](https://github.com/apple/swift-protobuf/pull/1589/files#diff-579304a80fa4b82b64c568287087a932f09d2fb7d8c682111405bc3d681cb70bR159) fixes all "exteneded" in the repository when it is regenerated.

mrs1669:
> You'll also need to fix/update the copies in the Reference/ directory as the CI will then catch them not matching once the .proto files are updated.

https://github.com/apple/swift-protobuf/pull/1589/commits/edbc683a1499fe98e74f6c140b5e10ef8f7a78c1
This commit replaces all "extended" with "extended" in repositories, including Reference/ directory. Please let me know if I'm wrong about how to respond!

tbkka:
You should never directly edit any `.pb.swift` file or anything under `Reference`.  All such files are created from other files.  The original files should be edited instead and we will separately re-create those files to update their contents.

mrs1669:
Thank you, I understand.
I'll revert all `.pb.swift` file and `Reference/` directory diff.

thomasvl (approved):
Thank you!

thomasvl (changes requested):
Since one of the changes is in the generated text for files, you'll need to do a `make regenerate reference` to get things updated. That requires you have `protoc` on your path.

mrs1669:
@thomasvl 
I built protoc environment and also committed the diff generated by `make regenerate reference` command. 
please run CI and check it 🙇

thomasvl (approved):
Thank you again, looks good to go now.

PR #1590, opened 2024-03-20
Drop a test.

thomasvl:
This test was leveraging a questionable behavior of Foundation. The input isn't valid UTF-8, and there was a silent substitution going on, but trunk versions of the compiler (and hence Foundation) properly fail to make a String out of the input, so the force unwrap fails. So just drop the test as no longer valid.


allevato:
For posterity, [this change](https://github.com/apple/swift-corelibs-foundation/pull/4813) in swift-corelibs-foundation started causing this to fail. Instead of the invalid UTF-8 code units being substituted (with U+FFFD) as they were previously (and still are on Darwin, at least in macOS 14.4), the initializer returned nil instead.

I'm not sure if that's an intentional change in swift-corelibs-foundation, but the [documentation for `String(bytes:encoding:)`](https://developer.apple.com/documentation/swift/string/init(bytes:encoding:)) says nothing about its substitution behavior, so it's best we don't rely on it.

allevato (approved):

PR #1593, opened 2024-03-25
Update xcodeproj on 1.x branch

glbrntt:
Building with the xcodeproj fails because a number of files have been added/removed and these changes haven't been reflected in the xcodeproj. Additionally, the base config sets the macOS deployment target to 10.9, but this isn't valid as `Data` was made available in 10.10.


thomasvl:
> Additionally, the base config sets the macOS deployment target to 10.9, but this isn't valid as `Data` was made available in 10.10.

Huh.  Is Xcode complaining about this now?  Wonder what it hasn't tripped things up before.

thomasvl:
> > Additionally, the base config sets the macOS deployment target to 10.9, but this isn't valid as `Data` was made available in 10.10.
> 
> Huh. Is Xcode complaining about this now? Wonder what it hasn't tripped things up before.

The [podspec](https://github.com/apple/swift-protobuf/blob/1_x_release_branch/SwiftProtobuf.podspec#L12) says 10.11, might make sense to sync back up with that.

glbrntt:
> > Additionally, the base config sets the macOS deployment target to 10.9, but this isn't valid as `Data` was made available in 10.10.
> 
> Huh. Is Xcode complaining about this now? Wonder what it hasn't tripped things up before.

It was on Xcode 15.3, I didn't try an older version. I was a little surprised it hadn't been seen before.

thomasvl (approved):

PR #1595, opened 2024-03-26
Honor features and expose potential support to `CodeGenerator`

thomasvl:
This does not enable editions for Swift Protobuf, this is just more internals to build things out.

### DIFF
--- a/Sources/SwiftProtobufPluginLibrary/CodeGenerator.swift
+++ b/Sources/SwiftProtobufPluginLibrary/CodeGenerator.swift
@@ -47,6 +47,11 @@ public protocol CodeGenerator {
   /// the protocol buffer compiler.
   var supportedFeatures: [Google_Protobuf_Compiler_CodeGeneratorResponse.Feature] { get }
 
+  /// The Protobuf Edition range that this generator can handle. Attempting
+  /// to generate for an Edition outside this range will cause protoc to
+  /// error.
+  var supportedEditionRange: ClosedRange<Google_Protobuf_Edition> { get }
+
   /// If provided, the argument parsing will support `--version` and report
   /// this value.
   var version: String? { get }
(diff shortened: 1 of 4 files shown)


Sources/SwiftProtobufPluginLibrary/CodeGenerator.swift
```diff
@@ -139,6 +144,11 @@ extension CodeGenerator {
 // Provide default implementation for things so `CodeGenerator`s only have to
 // provide them if they wish too.
 extension CodeGenerator {
+  public var supportedEditionRange: ClosedRange<Google_Protobuf_Edition> {
+    // Default impl of unknow so generator don't have to provide this until
```
allevato, line 148:
```suggestion
    // Default impl of unknown so generator don't have to provide this until
```

Sources/SwiftProtobufPluginLibrary/Descriptor.swift
```diff
@@ -73,6 +84,10 @@ public final class DescriptorSet {
     featureSetDefaults: Google_Protobuf_FeatureSetDefaults,
     featureExtensions: [any AnyMessageExtension] = []
   ) {
+    precondition(Self.bundledEditionsSupport.contains(featureSetDefaults.minimumEdition),
+                 "Attempt to use a FeatureSetDefault miniumEdition that isn't supported by the library.")
```
allevato, line 88:
```suggestion
                 "Attempt to use a FeatureSetDefault minimumEdition that isn't supported by the library.")
```

allevato (approved):

Sources/SwiftProtobufPluginLibrary/CodeGenerator.swift
thomasvl: Done.

Sources/SwiftProtobufPluginLibrary/Descriptor.swift
thomasvl: Done

PR #1600, opened 2024-03-27
Update proto files on the 1.x branch

thomasvl:


thomasvl:
Upstream conformance tests got some new tests, so we need new proto files otherwise things fail because of the unknown fields.

allevato (approved):

PR #1606, opened 2024-04-01
Move to `loadUnaligned` and hopefully streamline some helpers

thomasvl:
Follow up from #1604, but realized we could also stream line some of the calling code.  Each of those follow ups is done as a commit so we could drop them if desired.

Looking at things, I think we could even merge the "FourByte"/"EightByte" into a single generic function since `MemoryLayout` can provide the size.


Sources/SwiftProtobuf/BinaryDecoder.swift
```diff
@@ -212,7 +212,7 @@ internal struct BinaryDecoder: Decoder {
         guard fieldWireFormat == WireFormat.fixed32 else {
             return
         }
-        try decodeFourByteNumber(value: &value)
```
thomasvl, line 215: Guess this was a bug if we ever were built for big endian.

allevato (approved):
This is so much cleaner.

thomasvl:
Added one more commit to merge into a single function.

allevato (approved):

thomasvl:
@tbkka gentle ping since you were the original author of a bunch of this.

PR #1609, opened 2024-04-03
Add missing `@preconcurrency` annotations to `Sendable`-conforming protocols

gjcairo:

Adding `Sendable` conformance is a breaking change unless the protocols are annotated with `@preconcurrency`. This PR adds the missing annotations.

### DIFF
--- a/Sources/SwiftProtobuf/Enum.swift
+++ b/Sources/SwiftProtobuf/Enum.swift
@@ -16,6 +16,7 @@
 // -----------------------------------------------------------------------------
 
 /// Generated enum types conform to this protocol.
+@preconcurrency
 public protocol Enum: RawRepresentable, Hashable, CaseIterable, Sendable {
   /// Creates a new instance of the enum initialized to its default value.
   init()
--- a/Sources/SwiftProtobuf/ExtensionFields.swift
+++ b/Sources/SwiftProtobuf/ExtensionFields.swift
@@ -22,6 +22,7 @@
 // equality with some other extension field; but it's type-sealed
 // so you can't actually access the contained value itself.
 //
+@preconcurrency
 public protocol AnyExtensionField: Sendable {
   func hash(into hasher: inout Hasher)
   var protobufExtension: any AnyMessageExtension { get }
@@ -46,7 +47,8 @@ extension AnyExtensionField {
 ///
 /// The regular ExtensionField type exposes the value directly.
 ///
-public protocol ExtensionField: AnyExtensionField, Hashable, Sendable {
+@preconcurrency
+public protocol ExtensionField: AnyExtensionField, Hashable {
   associatedtype ValueType
   var value: ValueType { get set }
   init(protobufExtension: any AnyMessageExtension, value: ValueType)
--- a/Sources/SwiftProtobuf/ExtensibleMessage.swift
+++ b/Sources/SwiftProtobuf/ExtensibleMessage.swift
@@ -13,6 +13,7 @@
 // -----------------------------------------------------------------------------
 
 // Messages that support extensions implement this protocol
+@preconcurrency
 public protocol ExtensibleMessage: Message {
     var _protobuf_extensionFieldValues: ExtensionFieldValueSet { get set }
 }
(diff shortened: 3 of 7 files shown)


Sources/SwiftProtobuf/Enum.swift
```diff
@@ -16,6 +16,7 @@
 // -----------------------------------------------------------------------------
 
 /// Generated enum types conform to this protocol.
+@preconcurrency
```
thomasvl, line 19: So are these really breaking?

Most of these are internal details and a Developer would never create something that manually conforms to the protocol. So do we really need to annotate them as `@preconcurrency`?

thomasvl: @FranzBusch @tbkka @Lukasa

gjcairo: These are `public` protocols, so even if it's extremely unlikely someone would go and manually implement types that conform to them, they _could_ do it. This means that public API has changed (because they now conform to `Sendable`) and thus should be annotated with `@preconcurrency`.
I don't think there's any downside to adding these annotations. Have you got any specific concerns?

Sources/SwiftProtobuf/ExtensionFields.swift
```diff
@@ -46,7 +47,8 @@ extension AnyExtensionField {
 ///
 /// The regular ExtensionField type exposes the value directly.
 ///
-public protocol ExtensionField: AnyExtensionField, Hashable, Sendable {
+@preconcurrency
+public protocol ExtensionField: AnyExtensionField, Hashable {
```
thomasvl, line 51: Can you also provide a little insight into why this one dropped `Sendable` (others didn't) and why some things that we hadn't tagged as `Sendable` needed the `@preconcurrency`?

gjcairo: > why this one dropped `Sendable`

`ExtensionField` conforms to `AnyExtensionField`, which already conforms to `Sendable`, making `ExtensionField` also conform to it transitively. I removed the explicit conformance just to tidy things up - it can be added again if you prefer.

> why some things that we hadn't tagged as `Sendable` needed the `@preconcurrency`

When a public protocol `A` conforms to another `Sendable` protocol `B`, protocol `A` inherits the `Sendable` conformance, meaning the `@preconcurrency` annotation has to be bubbled up.

thomasvl: Ack, thanks.

Sources/SwiftProtobuf/Enum.swift
FranzBusch: What @gjcairo said here is correct. These have been public protocols and adding the `Sendable` inheritance requires an `@preconcurrency` annotation to downgrade a potential error into a warning when Swift 6 comes around.

FranzBusch (approved):
LGTM!

FranzBusch:
The failing API breakage pipeline is producing false errors and those are actually not breaking changes.

```
  💔 API breakage: protocol Enum is now with @preconcurrency
  💔 API breakage: protocol ExtensibleMessage is now with @preconcurrency
  💔 API breakage: protocol AnyExtensionField is now with @preconcurrency
  💔 API breakage: protocol ExtensionField is now with @preconcurrency
  💔 API breakage: protocol ExtensionMap is now with @preconcurrency
  💔 API breakage: protocol FieldType is now with @preconcurrency
  💔 API breakage: protocol MapKeyType is now with @preconcurrency
  💔 API breakage: protocol MapValueType is now with @preconcurrency
  💔 API breakage: protocol Message is now with @preconcurrency
  💔 API breakage: protocol _MessageImplementationBase is now with @preconcurrency
  💔 API breakage: protocol AnyMessageExtension is now with @preconcurrency
```

thomasvl:
> The failing API breakage pipeline is producing false errors and those are actually not breaking changes.

Is there an upstream issue for this?

FranzBusch:
> Is there an upstream issue for this?

Not as far as I know

Sources/SwiftProtobuf/Enum.swift
thomasvl: > they _could_ do it

I don't think we've really historically worried about folks attempting to use these.

> to downgrade a potential error into a warning when Swift 6 comes around

What's the warnings that we aren't seeing yet?

FranzBusch: Is somebody would conform a type to a protocol that inherits from `Sendable` they will get a warning that their type is required to be `Sendable`. In Swift 6 mode that warning is going to be an error. Adding `@preconcurrency` makes sure that this stays a warning in Swift 6 mode.

PR #1610, opened 2024-04-03
Add a second api check against the last release.

thomasvl:
This way here is a report for what the tool thinks are api breaks between `main` (changes in the given PR), and and updated status of how things compare to the last release that was made.


thomasvl:
@FranzBusch @tbkka @Lukasa - Seems like this might be useful if we're revisiting how main compares to the latest release, thoughts?

@gjcairo fyi

PR #1612, opened 2024-04-05
Move new errors to extendable struct-based SwiftProtobufError

gjcairo:

Adding or removing cases to a `public enum` is an API-breaking change, and we've done both on `main`.

To get around this, it's common practice to have a struct-based `Error` type instead. A useful pattern is to have this error contain a `Code` (which defines the domain of the error), and a message explaining the error in more detail. This message can be a dynamic string, meaning it can contain additional information about the error, for example, the name of an unknown JSON field when decoding.

This PR adds a new `SwiftProtobufError`, which follows the above pattern.

Common errors are defined as `public static let`s, which means new cases can be added without breaking API. Errors can also be defined without the need for exposing a new constant - I have simply moved all of the existing errors (grouped by domain) to static constants for consistency, but this doesn't have to stay this way in the future.

Existing (i.e. already released) `enum`s have been left as-is to avoid any potential source-breaking changes where users are catching specific errors. New errors have been moved to `SwiftProtobufError`s - and so should any new errors moving forward. Docs have also been updated.


Sources/SwiftProtobuf/SwiftProtobufError.swift
```diff
@@ -0,0 +1,620 @@
...
+            Self(.anyUnpackError)
+        }
+        
+        public static var protoFileToModuleMappingError: Self {
+            Self(.protoFileToModuleMappingError)
+        }
+    }
+
+    /// A location within source code.
+    public struct SourceLocation: Sendable, Hashable {
```
thomasvl, line 153: Do we want/need this?

For the different decoding errors, it seems like having support for a byteoffset and maybe line/column could someday be useful?

Sources/SwiftProtobuf/SwiftProtobufError.swift
```diff
@@ -0,0 +1,620 @@
...
+// - MARK: Common errors
+
+extension SwiftProtobufError {
+    /// Errors arising from encoding protobufs into binary data.
+    public enum BinaryEncoding {
+        /// The definition of the message or one of its nested messages has required
+        /// fields but the message being encoded did not include values for them. You
+        /// must pass `partial: true` during encoding if you wish to explicitly ignore
+        /// missing required fields.
+        public static let missingRequiredFields = SwiftProtobufError(
```
thomasvl, line 201: If we wanted to add more info in the future (say what fields were missing), how would we do that?  Would adding a `static func` as the same name but with args be safe?

Sources/SwiftProtobuf/SwiftProtobufError.swift
```diff
@@ -0,0 +1,620 @@
...
+
+    private var storage: Storage
+    private mutating func ensureStorageIsUnique() {
+        if !isKnownUniquelyReferenced(&self.storage) {
+            self.storage = self.storage.copy()
+        }
+    }
+    
+    private final class Storage: Equatable {
+        var code: Code
```
thomasvl, line 30: Since `code` is basically what used to be the different error types, we only have a domain level for errors now, and for someone to tell the difference between two errors, they would have to match the `message`?

That feels very fragile as correctly a typo in a message is a breaking change.

With the existing enum system, when parsing delimited data, we added the new error types so you can tell the difference between when the framing part of the stream is bad vs. when the message within a frame had an issue. This way a caller can decide if the error forces the invalidation of the whole stream or if they can continue reading messages just discarding the bad one. But in this new model, they are all decoding errors, so they would have to inspect the message strings themselves to tell what specific error it is.

Should there be a model where they can act on specific errors, or do we have to group them all now to different `code` value where all would be acted on in the same way?

Looking at the tests, maybe they can compare the errors to the accessors instead, should we not expose `message` so they can't really check it and only dump the error in `description`, but still use it for equality to avoid the fragile ness?

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
```diff
@@ -136,18 +136,18 @@ final class Test_AsyncMessageSequence: XCTestCase {
...
         XCTFail("Shouldn't have returned a value for an invalid stream.")
       }
     } catch {
-      if error as! BinaryDecodingError == .tooLarge {
-        tooLargeThrown = true
-      }
+      self.assertSwiftProtobufError(
+        error,
+        code: .binaryDecodingError,
+        message: "Message too large: Bytes and Strings have a max size of 2GB."
```
thomasvl, line 148: Rather then duplicate the string, should it use `.BinaryDecoding.tooLarge`?

Tests/SwiftProtobufTests/Test_AllTypes.swift
```diff
@@ -1756,7 +1768,11 @@ final class Test_AllTypes: XCTestCase, PBTestHelpers {
           // Don't need all the bytes, want some to let the length issue trigger.
           0x01, 0x02, 0x03,
         ])) {
-          XCTAssertEqual($0 as! BinaryDecodingError, BinaryDecodingError.tooLarge)
+            self.assertSwiftProtobufError(
+                $0,
+                code: .binaryDecodingError,
+                message: "Message too large: Bytes and Strings have a max size of 2GB."
```
thomasvl, line 1774: Rather then duplicate the string, should it use `.BinaryDecoding.tooLarge`?

Tests/SwiftProtobufTests/Test_AllTypes.swift
```diff
@@ -1000,7 +1008,11 @@ final class Test_AllTypes: XCTestCase, PBTestHelpers {
...
```
thomasvl, line 1014: Rather then duplicate the string, should it use `.BinaryDecoding.tooLarge`?

Tests/SwiftProtobufTests/Test_AllTypes.swift
```diff
@@ -948,7 +952,11 @@ final class Test_AllTypes: XCTestCase, PBTestHelpers {
...
```
thomasvl, line 958: Rather then duplicate the string, should it use `.BinaryDecoding.tooLarge`?

Tests/SwiftProtobufTests/Test_AllTypes.swift
```diff
@@ -824,7 +824,11 @@ final class Test_AllTypes: XCTestCase, PBTestHelpers {
...
```
thomasvl, line 830: Rather then duplicate the string, should it use `.BinaryDecoding.tooLarge`?

Tests/SwiftProtobufTests/TestHelpers.swift
```diff
@@ -399,6 +399,16 @@ extension XCTestCase {
         XCTAssertTrue(actual.hasSuffix(expectedSuffix), fmt ?? "debugDescription did not match", file: file, line: line)
 #endif
     }
+    
+    func assertSwiftProtobufError(_ error: any Error, code: SwiftProtobufError.Code, message: String) {
```
thomasvl, line 403: Might not need this is everything moves to not making the error directly.

Sources/SwiftProtobuf/AsyncMessageSequence.swift
```diff
@@ -21,16 +21,15 @@ extension AsyncSequence where Element == UInt8 {
   ///
   /// - Parameters:
   ///   - messageType: The type of message to read.
-  ///   - extensions: An `ExtensionMap` used to look up and decode any extensions in
+  ///   - extensions: An ``ExtensionMap`` used to look up and decode any extensions in
   ///    messages encoded by this sequence, or in messages nested within these messages.
-  ///   - partial: If `false` (the default),  after decoding a message, `Message.isInitialized`
+  ///   - partial: If `false` (the default),  after decoding a message, ``Message/isInitialized-6abgi`
```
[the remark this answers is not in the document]
gjcairo: Yeah it's a reference. When there are func overloads and the names are the same, DocC automatically disambiguates with a suffix.

Sources/SwiftProtobuf/SwiftProtobufError.swift
```diff
@@ -0,0 +1,620 @@
...
+    /// A high level indication of the kind of error being thrown.
+    public struct Code: Hashable, Sendable, CustomStringConvertible {
+        private enum Wrapped: Hashable, Sendable, CustomStringConvertible {
+            case binaryEncodingError
+            case binaryDecodingError
+            case jsonEncodingError
+            case jsonDecodingError
+            case textFormatDecodingError
+            case anyUnpackError
+            case protoFileToModuleMappingError
```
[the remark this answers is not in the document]
gjcairo: You're right. I've changed this around a bit and got rid of `anyUnpackError` and `protoFileToModuleMappingError` an replaced them with `invalidArgument` and `internalError`. I think they're better as general error domains. What do you think?

Sources/SwiftProtobuf/AnyUnpackError.swift
```diff
@@ -21,6 +21,7 @@
 /// message.  At this time, any error can occur that might have occurred from a
 /// regular decoding operation.  There are also other errors that can occur due
 /// to problems with the `Any` value's structure.
+@available(*, deprecated, message: "This error type has been deprecated and won't be thrown anymore; it has been replaced by `SwiftProtobufError`.")
 public enum AnyUnpackError: Error {
```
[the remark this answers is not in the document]
FranzBusch: We have to differentiate here. This is not an API breaking change but might be a semantic breaking change. So just from a SemVer perspective this is fine. However, you are correct that if somebody expects a certain method to throw just a specific set of errors then we might break them. I don't think we ever made that promise nor do we make that promise in any of our other packages. Since Swift uses `any Error` (before typed throws came along) we never made any API level promises to constrain us.

I personally think this change is good and fine and nobody should ever really on specific errors being thrown from an API unless these APIs very explicitly state that.

Sources/SwiftProtobuf/SwiftProtobufError.swift
FranzBusch: The idea here is that `code` is unique per error type and we shouldn't really overload a single `code` with different `message`s. If that is correctly done then users are able to pattern match on the `code` which will give them back the exhaustive matching from before.

Sources/SwiftProtobuf/SwiftProtobufError.swift
FranzBusch: We introduced this in almost all of our error structs now for debugging purposes so that you know where an error comes from. It has been really useful over time.

FranzBusch: However is this actually used somewhere @gjcairo ?

Sources/SwiftProtobuf/SwiftProtobufError.swift
```diff
@@ -0,0 +1,620 @@
...
+            message: """
+                The definition of the message or one of its nested messages has required fields, \
+                but the message being encoded did not include values for them. \
+                Decode with `partial: true` if you wish to explicitly ignore missing required fields.
+            """
+        )
+        
+        /// Messages are limited to a maximum of 2GB in encoded size.
+        public static let tooLarge = SwiftProtobufError(
+            code: .binaryEncodingError,
```
FranzBusch, line 212: We really shouldn't use a single `code` for more than one error otherwise users can't match on the code. This means introducing a ton more cases in the underlying `code` enum but giving developers back better error matching.

Also all these let's should probably be `func`s if we want to pass through the `file/line/column` since these need to be created at the place where the error is created and not here where it is statically defined

gjcairo: Ah, I was using [`NIOFileSystem/FileSystemError`](https://github.com/apple/swift-nio/blob/fc63f0cf4e55a4597407a9fc95b16a2bc44b4982/Sources/NIOFileSystem/FileSystemError.swift#L29) as inspiration, where we only have a few `Code`s that group the error types in different categories, and then throw errors with different messages but the appropriate code.

Not sure I follow why users wouldn't be able to match on the code. This is possible:
```swift
do {
  try someOperationThatThrows()
} catch let error as SwiftProtobufError where error.code == .binaryEncodingError {
  // Do something
}
```
And so is this:
```swift
do {
  try someOperationThatThrows()
} catch let error as SwiftProtobufError {
  switch error.code {
  case .binaryEncodingError:
    // Do something
  case ...
  }
}
```

Regarding these having to be `func`s though, you're right, I was unsure about keeping the `SourceLocation` and did a refactor and I'm not really using it anyways but didn't properly clean up, my bad. I see there's another thread above discussing whether we want to keep it or not. If we do, I'll convert all of these into `func`s and add the location to the message. Otherwise I'll just get rid of it and keep these as `let`s.

Sources/SwiftProtobuf/SwiftProtobufError.swift
gjcairo: Yes my bad, I was refactoring this and ended up being unused for the common errors I created. If we want to keep it I'll change all of the `static let`s into `func`s so we can pass in the location, otherwise I can get rid of this.

Tests/SwiftProtobufTests/Test_AsyncMessageSequence.swift
gjcairo: Good catch, forgot to refactor this when I added the common error `static let`s.

Sources/SwiftProtobuf/SwiftProtobufError.swift
gjcairo: > For the different decoding errors, it seems like having support for a byteoffset and maybe line/column could someday be useful?

@thomasvl Are you suggesting that the decoding-related `Code`s have the offset or line/column of the data that caused the error so that users can match on those, or is it fine if they're just added to the message string?

Sources/SwiftProtobuf/AnyUnpackError.swift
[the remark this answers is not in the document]
thomasvl: Works for me.  I guess maybe we should better call it out in the comment message and then whenever it is released, we'll have to call it out for that release.

Sources/SwiftProtobuf/SwiftProtobufError.swift
thomasvl: I was thinking to the `Code`.  Generally I lean toward structured info vs. strings just because they tend to interoperate with other tooling better.  I'm also fine skipping it until we're actually going to provide the data.

Sources/SwiftProtobuf/SwiftProtobufError.swift
gjcairo: I agree we shouldn't use the `message` for equality, I really only conformed this error to `Equatable` for tests, but I've removed this now. I am asserting that both the code and message match in tests though, because I think it doesn't hurt to make sure the message is what we want it to be, but users will be forced to match on the code instead. I've left `description` unchanged, meaning the `message` is used there. 

I do think exposing the `message` provides no real value and on the contrary it opens the door to users matching on it (even though they shouldn't) so I would agree to making it `internal` instead.

With regards to the current enum system allowing callers to differentiate between the whole framing in a stream being bad vs a single message: we can add a new `Code` case, maybe something like `binaryStreamDecodingError`. You could then differentiate between each scenario by catching either `binaryDecodingError` or `binaryStreamDecodingError` error codes. How does that sound?

thomasvl: More codes to allow actually acting on the errors makes sense to me.

Sources/SwiftProtobuf/SwiftProtobufError.swift
gjcairo: Gotcha - that sounds good to me. Maybe we can take care of that bit in a follow up PR? I'm aware this one's already quite large.

WRT the original question about the `SourceLocation`, @thomasvl do you think it's worth having this as part of the message description or would you rather not have this for some reason?

thomasvl: > Gotcha - that sounds good to me. Maybe we can take care of that bit in a follow up PR? I'm aware this one's already quite large.

Makes sense.

> WRT the original question about the `SourceLocation`, @thomasvl do you think it's worth having this as part of the message description or would you rather not have this for some reason?

When you start to get into the space of client software, capturing `#file` can be risky as it allows someone to dump details out of the binary about the layout of the application's development. So I'm generally cautious of things like that.

Anyway, I'll have to defer to you all on this since it's your repo and the new Error interface should be what you all want.

FranzBusch: I think you bring up a very valid problem here with leaking file paths into potential logs or at worse user output. I just had to remind myself which magic identifier existed in the language today and I realised that we introduced the `#fileID` identifier exactly for this. @gjcairo can you make sure we rather use `#fileID` here since that only includes `ModuleName/FileName` instead of the full path.

gjcairo: Okay I've added the location information to all common errors, and only expose it in the error description.

Sources/SwiftProtobuf/SwiftProtobufError.swift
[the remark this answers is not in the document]
gjcairo: Made this change

Sources/SwiftProtobuf/SwiftProtobufError.swift
```diff
@@ -0,0 +1,1098 @@
...
+    public struct Code: Hashable, Sendable, CustomStringConvertible {
+        private enum Wrapped: Hashable, Sendable, CustomStringConvertible {
+            case binaryEncodingError
+            case binaryDecodingError
+            case binaryStreamDecodingError
+            case jsonEncodingError
+            case jsonDecodingError
+            case textFormatDecodingError
+            case invalidArgument
+            case internalError
```
thomasvl, line 99: So what's the recommendation going forward for when to add more codes?  If folks now can use the codes to figure out error types, should we keep the previous granularity?

gjcairo: What we've done in other open source projects is to have this code represent a domain, instead of a particular error - the rationale is that callers should be able to handle all errors from the same domain equally, without having so much unnecessary granularity that it ends up being a burden for users. 

For example, now all errors thrown when decoding JSON relating to the format of the received payload will have a `jsonDecodingError` code; errors relating to problems with some internal representations or other internal unrecoverable state will throw `internalError`; errors relating to failed binary encoding because of issues with the given message will throw `binaryEncodingError`, etc...

I followed this pattern here as well, which means you'd only add new cases here if a new domain category is warranted - other you'd reuse one of the existing codes for new errors if applicable. 

Does this make sense or have you got any concerns/questions?

thomasvl: Make sense.

Tests/SwiftProtobufTests/TestHelpers.swift
```diff
@@ -399,6 +399,10 @@ extension XCTestCase {
         XCTAssertTrue(actual.hasSuffix(expectedSuffix), fmt ?? "debugDescription did not match", file: file, line: line)
 #endif
     }
+    
+    func isSwiftProtobufErrorEqual(_ actual: SwiftProtobufError, _ expected: SwiftProtobufError) -> Bool {
```
thomasvl, line 403: What was the verdict, is this still needed if we don't want to lock down on the message strings?

gjcairo: I made the message `internal` and got rid of the `Equatable` conformance so that we don't make the message part of the public API for errors / so external callers cannot match on the message. 
However I decided to keep this assertion in unit tests just to make sure that we are surfacing the right error message in the right scenarios, because having accurate messages is still important IMO. I can get rid of this, but thought it was worth it.

gjcairo:
FYI @thomasvl just rebased on main and fixed a conflict after your changes on the JSONDecoder.

FranzBusch (approved):
This looks good to me know! I like the name spacing that you introduced @gjcairo

Sources/SwiftProtobuf/SwiftProtobufError.swift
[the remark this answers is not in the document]
thomasvl: Sorry, missed this reply -

anyUnpackError is likely still ok, as unpacking any is still within Source/SwiftProtobuf, it's the modulemapping that is only within the scope of another package, which is what seemed odd to me.  For the modulemapping, why not have a PluginLibError?  i.e. - have one error scope per package as you go up the stack?

Sources/SwiftProtobuf/AnyUnpackError.swift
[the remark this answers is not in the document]
thomasvl: fyi - I finally got a chance to go back to read semver - https://semver.org/#spec-item-8 is I think what I was remember. Changing the errors to completely throw different ones is where I could see someone saying we're in the "any backward incompatible changes" space, as existing code that error handled the errors we documented would no longer work.  But I guess until we fully have typed throws in the language, this stays a grey area.

Anyway, resolving this since it sounds like you all are good with this not requiring the bump.

Sources/SwiftProtobuf/SwiftProtobufError.swift
```diff
@@ -0,0 +1,1098 @@
...
+                message: "Messages are limited to a maximum of 2GB in encoded size.",
+                location: SourceLocation(function: function, file: file, line: line)
+            )
+        }
+        
+        /// `Any` fields that were decoded from JSON cannot be re-encoded to binary
+        /// unless the object they hold is a well-known type or a type registered via
+        /// `Google_Protobuf_Any.register()`.
+        public static func anyTypeURLNotRegistered(
+            typeURL: String,
```
thomasvl, line 247: I was going back through some of our internal code, and realized this new model for errors has a limitation when it comes to the general idea of details about an error.  So in the enum form, the `typeURL` was an associated value, so a developer can write code to extract what the type was and then act on it (log it, attempt to do something else about it, etc.)

In this new form, there is no way a developer can write code to actually get at this detail to do anything with it. If they describe the error all they get is the `Code`. So if someone wanted some degree of logging for how an application does in the field, how would they get these sorta information?  Snd I don't think we want the `message` expose for the previously mentioned reasons, as then folks will parse it making the string contents part of the api contract.

Are we headed down a path where we need something like `NSError`'s `userInfo` to do this completely generically to avoid future api issues? 😞

thomasvl: Quick follow up - I'm thinking cases like where say you have a server accepting connection, being able to get specific details about all the different decoding failures that are happening may help you better understand what if someone is trying to hack into your server vs. say just doing a DOS attack, etc.  So I'm starting to wonder if these "domains" are just too limited of info to provide something detailed enough to build monitoring on top of.

Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
thomasvl, line 1: Sorry for not thinking this fully through, lets revert this only file's change (and any support in `SwiftProtobufError`). At this layer, the error descriptions end up end user visible for problems when a developer invokes the plugin.  So lets just focus on the changes to the SwiftProtobuf module for the time being, and the PluginLibrary can get looked at (if desired) in a different PR.

Sources/SwiftProtobuf/SwiftProtobufError.swift
gjcairo: If users wanted to log the `typeURL` or any other information associated with the error, they can just get the error description and the `message` will be a part of it (which now makes me think it may make having made `message` internal moot, but let's stay focused for a minute :)).
So let's assume that the "something else" users want to do with these associated values is try to recover from the error in some way, and that they need more information to do so. It could be argued that maybe throwing errors to drive the program's logic isn't necessarily the best pattern. Using this `anyTypeURLNotRegistered` as an example: if it was a recoverable issue, you might want to add API to allow users to set a closure/have a delegate to handle "unregistered URLs" for instance.

However, I do understand the concern and I think it's still valid (seeing how changing the APIs in a more significant way would either be breaking or definitely out of scope). 

So thinking of solutions that would work for this... Having struct-based errors **does** allow for associated values, but it also requires having a more specific wrapped private enum instead of this more generic "domain" approach. We could add more specific error codes as a compromise for errors that definitely required associated values. 
For instance, we could have something like:

```swift
struct SwiftProtobufError: Error {
  private enum _Error {
    case malformedAnyField
    case anyTypeURLNotRegistered(typeURL: String)
    // There would obviously be more cases here, just writing these two for simplicity
  }

  private let _wrapped: _Error
  private init(_ wrapped: _Error) {
    self._wrapped = wrapped
  }

  public static let malformedAnyField = SwiftProtobufError(.malformedAnyField)

  public static func anyTypeURLNotRegistered(typeURL: String) -> SwiftProtobufError {
    SwiftProtobufError(.anyTypeURLNotRegistered(typeURL: typeURL))
  }

  public var unregisteredTypeURL: String? {
    switch self._wrapped {
      case .malformedAnyField:
        return nil
      case .anyTypeURLNotRegistered(let typeURL):
        return typeURL
    }
  }
}
```

Callers could then match the error and extract the associated value like this:
```swift
do {
  try somethingThatThrows()
} catch let error as SwiftProtobufError {
  if let unregisteredTypeURL = error.unregisteredTypeURL {
    // This error was thrown because there was an issue with an unregistered type URL:
    // do something with it.
  }
}
```

I know this isn't as ideal, but I think it's a good compromise. We've reached this current iteration of errors as structs after quite a lot of experimentation on our other libraries, and we think it's the best we can currently do.

What do you think? Would you want to have a getter for all associated values that were part of some error in the old model?

gjcairo: So it seems we'd only need two cases with associated values (and thus two getters): `anyTypeURLNotRegistered` and `unknownField`. Given the context, I think it's fine to compromise and go down this route so I'll just implement it, it's quick anyways. Others can say what they think once it's done and we can change it once again if need be.

thomasvl (approved):
I think I'm good with things, but I'd like Tim to look also as original author of a bunch of the library.

gjcairo:
Thanks @thomasvl, that's fair.

gjcairo:
Had to rebase after the latest changes.

@tbkka mind looking at this when you have a minute?

tbkka:
Finally getting to this, apologies for the delay.

There are a lot of details about the precise structure of the new errors ... it looks like Thomas and Gus have discussed those pretty thoroughly and I don't think there's much I could contribute.   I just have two fairly high-level issues:

1. This is a source-breaking change so should not be merged to 1.x.  (I presume that's obvious to everyone, just wanted to say it out loud.)

2. The changes to the throw sites within the library look pretty straightforward.  But I'm less clear on what we expect clients to do with this new structure.  Right now, I see this in the test suite:
```
                        XCTFail("Should not have succeed, pass: \(i), limit: \(limit)")
                    }
                } catch let error as SwiftProtobufError where self.isSwiftProtobufErrorEqual(error, .BinaryDecoding.messageDepthLimit()) {
                    ... deal with error ...
```

What should a client's `catch` statement look like in this new world?

FranzBusch:
> This is a source-breaking change so should not be merged to 1.x. (I presume that's obvious to everyone, just wanted to say it out loud.)

This shouldn't be a source breaking change. Why do you think it is?

tbkka:
> This shouldn't be a source breaking change. Why do you think it is?

So the systematic changes to the test suite weren't actually necessary?

Tests/SwiftProtobufTests/Test_Required.swift
```diff
@@ -254,8 +254,9 @@ final class Test_Required: XCTestCase, PBTestHelpers {
         do {
             let _: [UInt8] = try message.serializedBytes()
             XCTFail("Swift encode should have failed: \(message)", file: file, line: line)
-        } catch BinaryEncodingError.missingRequiredFields {
+        } catch let error as SwiftProtobufError where self.isSwiftProtobufErrorEqual(error, .BinaryEncoding.missingRequiredFields()) {
             // Correct error!
+            print("sdfdsf")
```
tbkka, line 259: Debugging leftover?

gjcairo:
@tbkka we discussed this some more offline, and we decided to only move new errors (i.e. those that were only in `main` and not part of any tags) to the new, extensible `SwiftProtobufError` type, but kept the old enums. This is less tidy, but non-breaking. Hope we're all okay with this approach.

Sources/SwiftProtobuf/AnyMessageStorage.swift
```diff
@@ -218,7 +218,7 @@ internal class AnyMessageStorage {
     case .contentJSON(let contentJSON, let options):
       // contentJSON requires we have the type available for decoding
       guard let messageType = Google_Protobuf_Any.messageType(forTypeURL: _typeURL) else {
-          throw BinaryEncodingError.anyTypeURLNotRegistered(typeURL: _typeURL)
+        throw SwiftProtobufError.BinaryEncoding.anyTypeURLNotRegistered(typeURL: _typeURL)
```
Lukasa, line 221: Ok, I think I'd like us to chat about this a bit. @thomasvl and @tbkka your opinions would be valuable.

In Protobuf 1.26, at this call site we used to throw `BinaryEncodingError.anyTranscodeFailure`. I'm inclined to say that where we don't need to change the thrown error type, we probably shouldn't. It's not an API break, so semver doesn't prevent us from making this change, but it does risk causing some trouble with users who actually were trying to catch this error.

However, this would mean these new errors are essentially removed from the tree. How do people feel about that?

thomasvl: You might need to go back to the PR, I believe this got added to fix a bug where someone wanted the better visibility into the failure.  So reverting it would likely also mean reopening the bug as we still wouldn't be providing the requested information.

thomasvl: Ah, ok, now I follow (I guess my ☕ needed more time 😄).

Historically (in the 1.x releases), we have added error codes, but I don't know that we ever changed a code, and have the existing enums, folks could catch _all_ the errors of that type easily.

But yea, I guess this is different, not only could someone trying to catch this fail due to the change, I guess since it is a new type throw, it would completely escape their `catch` based on the enum. But if we stick to that train of thought pedantically, it means even throwing the new error types is problematic.

I guess I can(am) arguing both outcomes. I'd like to be able to correct errors, add better errors without having to go to a new major version, but I'm not sure where exactly to set the _bar_.

When/How did Nio decide to change the errors over and what was the reasoning behind that? I don't think you all did a major bump when it landed, correct?

Lukasa: Generally speaking we don't believe we're beholden to never change the errors we throw. But we need to decide to do it mindfully, rather than doing it without careful consideration. It's _possible_ for us to keep these errors as-is, we just need to decide if we think people are actually planning to catch them.

gjcairo: So to summarise, options are:

1. We keep this change, which we've asserted is not-technically-but-de-facto-breaking, meaning users relying on catching the errors originally thrown in 1.26.0 will have to update their code to catch this error instead.
2. We roll this back to the errors thrown in 1.26.0, which means we'll need issues to track https://github.com/apple/swift-protobuf/pull/1527 and any other (if any) changes we've made to make errors more descriptive, and postpone them until(/if) we release a v2.

What are the next steps here? 

Even though I don't know if everyone finds it acceptable to indefinitely postpone having better errors, based on the conversations we've had in the past weeks, it seems to me that we're quite keen on avoiding any sort of breakage - which makes me think we want option 2.
Do we want to wait for @tbkka to weigh in?

Lukasa: My lean is probably to do option 2, but I'd like to see if we can get a better advocate for option 1. So far @FranzBusch is the best advocate for it, I wonder if he's up for strengthening his position?

FranzBusch: I thought it was a good moment to cut over to a new error format all together since we know the enum based errors are not evolvable and we already had a break in `main`. I personally don't share the same strong feeling about changing the errors that we throw since we never promised that we do this.
However, I am also not feeling strong enough to fight you all on this 😀 My main goal right now is to get `main` back into a releasable state without requiring a major. Let's get to that point as fast as possible and then we can re-litigate the error discussion.

thomasvl: Should we open a new issue to reference all the things being reverted to provide a clean clearing house of all the errors we'll want to ensure we change whenever we finally do this change?  i.e. - I don't want to accidentally lose a minor detail that we did work out and fail to recreate it whenever we do make the changes.

Lukasa: Seems sensible to me.

gjcairo: Okay, just made the change we discussed. I believe we're fully non-breaking now.

I'll open an issue listing the things we've reverted from this and other PRs.

gjcairo: https://github.com/apple/swift-protobuf/issues/1657

Lukasa (approved):

gjcairo:
@thomasvl are you okay with getting this merged since it's been approved on our end?

thomasvl:
> @thomasvl are you okay with getting this merged since it's been approved on our end?

Yea, I'm good.  I wasn't sure if @tbkka should look again since he'd made some comment in the past and things were adjusted based on those comments.

tbkka:
I'm good.   👍

PR #1613, opened 2024-04-05
Add a --breakage-allowlist-path list for the diff against release.

thomasvl:
This is needed incase there are more changes to descriptor, but also include the list from
https://github.com/apple/swift-protobuf/pull/1609#issuecomment-2034982265 which aren't actually breaks.


thomasvl:
@gjcairo fyi - incase we have any other false entries going forward, you should be able to add them at the same time.

FranzBusch (approved):

gjcairo (approved):

thomasvl:
Landing this, I'll work on the conformance tests in a few minutes here.

PR #1616, opened 2024-04-05
Follow upstream and add `isGroupLike` construct

thomasvl:
The 2024 edition will have to have a feature to really deal with this, but this should help get things some what in line.

PR #1619, opened 2024-04-06
Update the conformance tests for upstream changes

thomasvl:
There will have to be follow up work to fully pass things, but this will get the needed plumbing done to get a better picture for what needs work.

I have some ideas for the actual fixes, the catch is going to be trying to make something work that isn't breaking because of some new expectations arounds TextFormat group support.


thomasvl:
This is currently stacked on #1616, so once that's merged, I'll rebase this and mark it ready for review.

thomasvl:
@gjcairo fyi - since these are likely also getting in your way of posting PRs.

allevato (approved):

PR #1620, opened 2024-04-08
Ignore non breaking API changes

gjcairo:
There are a few more changes reported as API-breaking by the SPM's diagnosis tool that actually aren't, so this PR updates `known_api_breaks.txt`.


thomasvl (approved):
I thought some of these I was seeing with 5.9, but not 5.10, I guess we just keep listing everything?

gjcairo:
Oh I am seeing these with 5.10 actually.

PR #1632, opened 2024-04-18
Fix oneof primitive types decoding

### LINKED ISSUE #1608: Primitive one-of type always passes nil as its argument for decoder
When I implement a decoder for decoding a one-of type, if its associated value of one-of is primitive (string, double, etc) it always pass nil for decoder method that I've implemented as `Decoder` protocol requirement.

Please be sure to include:

* OS: macOS 14.0

* Xcode version: 15.3

* Swift version: 5.10

* Branch: main

* `.proto` snippet:
```proto
oneof o {
  int32 oneof_int32 = 61;
  int64 oneof_int64 = 62;
}
```

pouyayarandi:
This PR Fixes #1608


Tests/SwiftProtobufTests/Test_OneodFields_Decoding.swift
```diff
@@ -0,0 +1,72 @@
...
+///
+/// Exercises the dedcode for primitive fields within a oneof.
+///
+// -----------------------------------------------------------------------------
+
+import Foundation
+import XCTest
+import SwiftProtobuf
+
+final class Test_OneofFields_Decoding: XCTestCase {
```
thomasvl: There's already Test_Merge and Test_ParsingMerge, so I'd suggest extending one of those (maybe they should merge).

We also want to make sure everything is covered
- non repeating fields of pod types override
- non repeating fields of message types recursively merge
- repeating fields append
- oneof fields for pod overrides to the new value 
- oneof fields for a message merges
- there's likely cases with extension fields also (repeating/non repeating, pod/message)

I also don't think you need the custom decoder here, just create a two messages, and then use the merge api on the binary data from the second message.  There's some examples in the other files.

Sources/protoc-gen-swift/OneofGenerator.swift
```diff
@@ -345,6 +345,7 @@ class OneofGenerator {
               hadValueTest = "hadOneofValue"
           } else {
               p.print("var v: \(field.swiftType)?")
+              p.print("if case \(field.dottedSwiftName)(let _v) = \(storedProperty) {v = _v}")
```
thomasvl, line 348: Can you split this change out, since this is the actual generator change, it makes it much easier to see.

Sources/Conformance/conformance.pb.swift
```diff
@@ -491,6 +491,7 @@ extension Conformance_ConformanceRequest: SwiftProtobuf.Message, SwiftProtobuf._
       switch fieldNumber {
       case 1: try {
         var v: Data?
+        if case .protobufPayload(let _v) = self.payload {v = _v}
         try decoder.decodeSingularBytesField(value: &v)
         if let v = v {
           if self.payload != nil {try decoder.handleConflictingOneOf()}
```
thomasvl, line 497: This part I'm also no longer sure of.

If the oneof was already set to the same failed, is it always a potentially conflicting oneof? It likely is up to the decoder, but there is no longer enough info to say if it was already set to the same value or was set to a different oneof value.

pouyayarandi:
I checked `Test_Merge` and realized that it works with oneof fields. So I think the problem is my own code (which I was using decoder for capturing values in #1505).

I think we can close this PR and its related issue as my problem solved with changing my code implementation and without any necessary addition in generated codes.

PR #1642, opened 2024-04-29
Remove 'u' suffix support from TextFormat.

thomasvl:
https://protobuf.dev/reference/protobuf/textformat-spec/#numeric and https://protobuf.dev/reference/protobuf/textformat-spec/#literals don't seem to show any signs that a 'u' suffix was ever valid. I also don't see any references in the C++ code. It also seems like we would parse `-12u` as valid which seems a little odd since it clearly isn't an unsigned value.

So dropping the 'u' support as it seems out of spec.

### DIFF
--- a/Sources/SwiftProtobuf/TextFormatScanner.swift
+++ b/Sources/SwiftProtobuf/TextFormatScanner.swift
@@ -609,10 +609,6 @@ internal struct TextFormatScanner {
                         val = UInt64(digit - asciiLowerA + 10)
                     case asciiUpperA...asciiUpperF:
                         val = UInt64(digit - asciiUpperA + 10)
-                    case asciiLowerU: // trailing 'u'
-                        p += 1
-                        skipWhitespace()
-                        return n
                     default:
                         skipWhitespace()
                         return n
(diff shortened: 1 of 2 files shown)


thomasvl:
@tbkka - I haven't been able to figure out where this support came from, but it doesn't seem to be to spec.  This isn't an api change, but it is a behavior change, so if you'd rather keep it unless we do want to do breaking changes, we can just add a comment that it isn't to spec and it should be removed in the future. Up to you.

tbkka:
An early version of this file (commit 0aa510909 from Dec 2016) has a comment:
```
                        case "f", "u":
                            // proto1 allowed floats to be suffixed with 'f'
                            // and unsigned integers to be suffixed with 'u'
                            // Just ignore it:
                            return .number(s)
``` 
If I understand correctly, proto1 was never used outside of Google and I presume is no longer used inside of Google, so I'm fine just removing this.

thomasvl:
Ok, since the C++ doesn't support it either, then, yea, guess safe to drop.

Did you want to go ahead and stamp this with approval?

PR #1643, opened 2024-04-29
TextFormat decoding options to skip unknown fields/extensions.

thomasvl:
Modeled after the upstream C++, provide two new decoding options to skip unknown fields while decoding TextFormat.

### DIFF
--- a/Sources/SwiftProtobuf/DoubleParser.swift
+++ b/Sources/SwiftProtobuf/DoubleParser.swift
@@ -18,9 +18,12 @@ import Foundation
 internal class DoubleParser {
     // Temporary buffer so we can null-terminate the UTF-8 string
     // before calling the C standard library to parse it.
+    //
     // In theory, JSON writers should be able to represent any IEEE Double
     // in at most 25 bytes, but many writers will emit more digits than
-    // necessary, so we size this generously.
+    // necessary, so we size this generously; but we could still fail to
+    // parse if someone crafts something really long (especially for
+    // TextFormat due to overflows (see below)).
     private var work =
       UnsafeMutableBufferPointer<Int8>.allocate(capacity: 128)
 
@@ -49,6 +52,15 @@ internal class DoubleParser {
 
         // Fail if strtod() did not consume everything we expected
         // or if strtod() thought the number was out of range.
+        //
+        // NOTE: TextFormat specifically calls out handling for overflows
+        // for float/double fields:
+        // https://protobuf.dev/reference/protobuf/textformat-spec/#value
+        //
+        // > Overflows are treated as infinity or -infinity.
+        //
+        // But the JSON protobuf spec doesn't mention anything:
+        // https://protobuf.dev/programming-guides/proto3/#json
         if e != work.baseAddress! + bytes.count || !d.isFinite {
             return nil
         }
--- a/Sources/SwiftProtobuf/TextFormatDecoder.swift
+++ b/Sources/SwiftProtobuf/TextFormatDecoder.swift
@@ -27,6 +27,10 @@ internal struct TextFormatDecoder: Decoder {
     private var fieldNameMap: _NameMap?
     private var messageType: any Message.Type
 
+    internal var options: TextFormatDecodingOptions {
+      return scanner.options
+    }
+
     internal var complete: Bool {
         mutating get {
             return scanner.complete
@@ -63,27 +67,17 @@ internal struct TextFormatDecoder: Decoder {
     }
 
     mutating func nextFieldNumber() throws -> Int? {
-        // Per https://protobuf.dev/reference/protobuf/textformat-spec/#fields, every field can be
-        // followed by a field separator, so if we've seen a field, remove the separator before
-        // checking for the terminator.
         if fieldCount > 0 {
             scanner.skipOptionalSeparator()
         }
-        if let terminator = terminator,
-           scanner.skipOptionalObjectEnd(terminator) {
-            return nil
-        }
-        if let fieldNumber = try scanner.nextFieldNumber(names: fieldNameMap!, messageType: messageType) {
+        if let fieldNumber = try scanner.nextFieldNumber(names: fieldNameMap!,
+                                                         messageType: messageType,
+                                                         terminator: terminator) {
             fieldCount += 1
             return fieldNumber
-        } else if terminator == nil {
-            return nil
         } else {
-            // If this decoder is looking for at a terminator, then if the scanner failed to
-            // find a field number, something went wrong (end of stream).
-            throw TextFormatDecodingError.truncated
+            return nil
         }
-
     }
 
     mutating func decodeSingularFloatField(value: inout Float) throws {
@@ -559,6 +553,7 @@ internal struct TextFormatDecoder: Decoder {
         var keyField: KeyType.BaseType?
         var valueField: ValueType.BaseType?
         let terminator = try scanner.skipObjectStart()
+        let ignoreExtensionFields = options.ignoreUnknownExtensionFields
         while true {
             if scanner.skipOptionalObjectEnd(terminator) {
                 if let keyField = keyField, let valueField = valueField {
@@ -568,14 +563,20 @@ internal struct TextFormatDecoder: Decoder {
                     throw TextFormatDecodingError.malformedText
                 }
             }
-            if let key = try scanner.nextKey() {
+            if let key = try scanner.nextKey(allowExtensions: ignoreExtensionFields) {
                 switch key {
                 case "key", "1":
                     try KeyType.decodeSingular(value: &keyField, from: &self)
                 case "value", "2":
                     try ValueType.decodeSingular(value: &valueField, from: &self)
                 default:
-                    throw TextFormatDecodingError.unknownField
+                    if ignoreExtensionFields && key.hasPrefix("[") {
+                        try scanner.skipUnknownFieldValue()
+                    } else if options.ignoreUnknownFields && !key.hasPrefix("[") {
+                        try scanner.skipUnknownFieldValue()
+                    } else {
+                        throw TextFormatDecodingError.unknownField
+                    }
                 }
                 scanner.skipOptionalSeparator()
             } else {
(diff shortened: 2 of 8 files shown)


Sources/SwiftProtobuf/TextFormatDecoder.swift
```diff
@@ -568,14 +563,20 @@ internal struct TextFormatDecoder: Decoder {
...
-            if let key = try scanner.nextKey() {
+            if let key = try scanner.nextKey(allowExtensions: ignoreExtensionFields) {
                 switch key {
                 case "key", "1":
                     try KeyType.decodeSingular(value: &keyField, from: &self)
                 case "value", "2":
                     try ValueType.decodeSingular(value: &valueField, from: &self)
                 default:
-                    throw TextFormatDecodingError.unknownField
+                    if ignoreExtensionFields && key.hasPrefix("[") {
```
tbkka, line 573: Can we ever legally see an unknown field or extension field inside a _map_?   The map encoding is pretty tightly specified.

Sources/SwiftProtobuf/TextFormatDecoder.swift
```diff
@@ -666,14 +675,20 @@ internal struct TextFormatDecoder: Decoder {
...
-            if let key = try scanner.nextKey() {
+            if let key = try scanner.nextKey(allowExtensions: ignoreExtensionFields) {
                 switch key {
                 case "key", "1":
                     try KeyType.decodeSingular(value: &keyField, from: &self)
                 case "value", "2":
                     try decodeSingularMessageField(value: &valueField)
                 default:
-                    throw TextFormatDecodingError.unknownField
+                    if ignoreExtensionFields && key.hasPrefix("[") {
```
tbkka, line 685: Same question.

Sources/SwiftProtobuf/TextFormatDecoder.swift
```diff
@@ -617,14 +619,20 @@ internal struct TextFormatDecoder: Decoder {
...
-            if let key = try scanner.nextKey() {
+            if let key = try scanner.nextKey(allowExtensions: ignoreExtensionFields) {
                 switch key {
                 case "key", "1":
                     try KeyType.decodeSingular(value: &keyField, from: &self)
                 case "value", "2":
                     try decodeSingularEnumField(value: &valueField)
                 default:
-                    throw TextFormatDecodingError.unknownField
+                    if ignoreExtensionFields && key.hasPrefix("[") {
```
tbkka, line 629: Same question.

Sources/SwiftProtobuf/DoubleParser.swift
```diff
@@ -18,9 +18,14 @@ import Foundation
...
     // Temporary buffer so we can null-terminate the UTF-8 string
     // before calling the C standard library to parse it.
+    //
     // In theory, JSON writers should be able to represent any IEEE Double
     // in at most 25 bytes, but many writers will emit more digits than
     // necessary, so we size this generously.
+    //
+    // This is also used for TextFormat, there the spec says decimal numbers
+    // can actually go over the limit can get mapped to [-]inifinity; so in
+    // theory this needs to be larger, but we haven't worried about it yet.
```
tbkka, line 28: This comment conflates two different issues:
* Overflow and underflow handling.  Overflow (e.g., "1e9999") and underflow (e.g., "1e-9999") handling are pretty well specified by IEEE754 and related standards, so there's not really much need to talk about it here because there's really not much variation in how people handle it.
* The real concern here is how many characters we should parse, which is limited by the size of this buffer.  The comment above about JSON parsing applies equally to TextFormat.

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -1109,54 +1106,239 @@ internal struct TextFormatScanner {
...
+                while p != end {
+                    let c = p[0]
+                    if c >= asciiZero && c <= asciiNine {
+                        fieldNum = fieldNum &* 10 &+ (Int(c) - Int(asciiZero))
+                    } else {
+                        break
+                    }
+                    p += 1
+                }
+                skipWhitespace()
```
tbkka, line 1159: What is the limit on protobuf field numbers?   We should record `pstart = p` at the top of the above loop and then `p - pstart` is the total number of digits.  If there are too many digits, we should fail here with `unknownField`.

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -1109,54 +1106,239 @@ internal struct TextFormatScanner {
...
+            }
+            if c == asciiPeriod {
+                return true  // "(-)0." : clearly a float
+            }
         }
+
+        // At this point it is all digits until we hit something else:
+        // - If it contains an '.' or an 'e'/'E' (1e10) it's a float
+        // - A trailing 'f' or 'F' means that's a float.
+        while true {
```
tbkka, line 1247: I don't really understand this loop.   It looks like it will never return `false`, so why do we spend time looking at the characters here at all?

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -1109,54 +1106,239 @@ internal struct TextFormatScanner {
...
+        if c == asciiPeriod {
+            return true  // "(-)." : clearly a float
+        }
+
+        if c == asciiZero {
+            scan += 1
+            if scan == end { return false }  // "(-)0[end]" : call it not a float
+            c = scan[0]
+            if c == asciiLowerX ||  // "(-)0x" : hex - not a float
+                (c >= asciiZero && c <= asciiSeven) {  // "(-)0[0-7]" : octal - not a float
```
tbkka, line 1236: Should "089" be considered a valid decimal floating-point value or an invalid octal integer?

tbkka (commented):
Adding the options seems like a good idea.  There's some refactoring in here that I don't quite understand, though.

Sources/SwiftProtobuf/TextFormatDecoder.swift
thomasvl: Yea, I don't think you should, but since anyone can hand code one, the question becomes what should we do?  As far as I can tell, with the C++ code, if you turn on their options, they too will skip these fields.

(And it seems sorta odd if one enabling ignoring unknown fields that you'd ever actually get an unknownField error)

Sources/SwiftProtobuf/TextFormatScanner.swift
thomasvl: fyi - Despite what the diff looks like, this isn't new code, it just got indented so it can loop over unknown field/values.

Fields are 1-536,870,911, so we could add logic to cap that.  I'm happy to do that in a follow up since it's not directly related to unknown fields.

thomasvl: https://protobuf.dev/programming-guides/proto2/#assigning for the sepc

Sources/SwiftProtobuf/TextFormatScanner.swift
thomasvl: https://protobuf.dev/reference/protobuf/textformat-spec/#numeric

So it is invalid.  Rather that have this api also throw for detecting invalid, I was trying to just have this pick off if the number is a float or not, and then the actual parsing apis are what do the final validations.

Long term, I think we might want to look at bring over the C++ tokenizer concept, as it is the common base for their JSON and TextFormat parsing, so it could provide a good model to collapsing some common code between the two, and actually might make picking off some of this malformed things easier.

Sources/SwiftProtobuf/TextFormatScanner.swift
thomasvl: Ha, you are correct, so this used to have a trailing `u` case what would return `false`, but I cleaned that out before sending because of #1642, I'll go back and clean this up.

thomasvl: Updated.

Sources/SwiftProtobuf/DoubleParser.swift
thomasvl: I was actually trying (poorly) to capture both points are a problem.

TextFormat specifically calls out handling for overflows for float/double fields:
https://protobuf.dev/reference/protobuf/textformat-spec/#value

> Overflows are treated as infinity or -infinity.

But the JSON protobuf spec doesn't mention anything: https://protobuf.dev/programming-guides/proto3/#json

So the current finite check isn't completely correct, when called for TextFormat.

The other, is since things can be really long and thus overflow, the buffer size could also be a problem.

thomasvl: Split/updated the comment to hopefully better capture this.

Sources/SwiftProtobuf/TextFormatDecoder.swift
thomasvl: Here's the spec part for map - https://protobuf.dev/reference/protobuf/textformat-spec/#map, it's basically just a repeated message field, so I went with the same sorta support there for unknown fields.

tbkka: That makes sense.  I suppose we'll need test cases for maps with extensions in various places?

thomasvl: Yup, the changes to Tests/SwiftProtobufTests/Test_TextFormat_Map_proto3.swift adds that, testing fields and extensions independently to be sure things fail/pass as expected.

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: I agree.  This does sound like a good follow-up task.

Sources/SwiftProtobuf/DoubleParser.swift
tbkka: Much better.   Keep these two examples in mind:
* Numeric overflow can happen even with short tokens:  "1e9999" is too large a value for Double and thus will parse to Infinity (in the default rounding mode) even though it's only 6 characters long.
* Overly long tokens aren't necessarily numeric overflows: "123.999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999" is a very long token but does not overflow when correctly parsed as a floating-point number.

The code here has an arbitrary limit of 128 characters for a floating-point value.  I personally think it entirely acceptable to consider any numeric token with more characters to be malformed, as there's never a good reason to have a number token this long.  (This is the sort of corner case that spec writers never think about.  :shrug:)

thomasvl: Ok to resolve this, or did you want me to pull in any addition into into the code comment?

tbkka (approved):

PR #1647, opened 2024-05-02
Add a fail out if raw field numbers are too long.

thomasvl:
Follow up from the comment on adding unknown field skipping.

Since there is a max field length, we can fail out if the field number is too many digits.

Added tests to also cover this.

### DIFF
--- a/Sources/SwiftProtobuf/TextFormatScanner.swift
+++ b/Sources/SwiftProtobuf/TextFormatScanner.swift
@@ -68,6 +68,11 @@ private let asciiLowerY = UInt8(ascii: "y")
 private let asciiLowerZ = UInt8(ascii: "z")
 private let asciiUpperZ = UInt8(ascii: "Z")
 
+// https://protobuf.dev/programming-guides/proto2/#assigning
+// Fields can be between 1 and 536,870,911. So we can stop parsing
+// a raw number if we go over this (it also avoid rollover).
+private let maxFieldNumLength: Int = 9
+
 private func fromHexDigit(_ c: UInt8) -> UInt8? {
   if c >= asciiZero && c <= asciiNine {
     return c - asciiZero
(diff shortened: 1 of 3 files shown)


Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -1157,6 +1180,9 @@ internal struct TextFormatScanner {
                         break
                     }
                     p += 1
+                    if p - start > maxFieldNumLength {
```
thomasvl, line 1183: In this case, we could also directly check fieldNum against the value instead of doing something length based, not sure if you'd rather that.

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -1090,9 +1095,26 @@ internal struct TextFormatScanner {
...
+        case asciiOne...asciiNine:  // 1...9 (field numbers are 123, not 0123)
+            let start = p
+            p += 1
+            while p != end {
+                let c = p[0]
+                if c < asciiZero || c > asciiNine {
+                    break
+                }
+                p += 1
+                if p - start > maxFieldNumLength {
```
thomasvl, line 1109: This one could also do the calculation if desired and check the specific values instead of being length based.  But most map<> caes will use the key names, the performance impact should be small.

tbkka: I don't need to do that.  I just wanted to be sure we didn't leave an opening for someone to force an integer overflow in this code.

Sources/SwiftProtobuf/TextFormatScanner.swift
tbkka: As above.

tbkka (approved):

PR #1651, opened 2024-05-09
Support "-nan" in TextFormat.

thomasvl:
As odd as it sounds, upstream supports this and there is a unittest that ensure it parses:

  https://github.com/protocolbuffers/protobuf/blob/3c03e9351c57081d0dffae120ed37497017f105c/src/google/protobuf/compiler/parser_unittest.cc#L464

It seems to have come from:

  https://github.com/protocolbuffers/protobuf/pull/15017

So make sure Swift is also able to parse it.

Also reflow some of the unknown field parsing as inf/nan don't need to be special cased with how the flow now works.

### DIFF
--- a/Sources/SwiftProtobuf/TextFormatScanner.swift
+++ b/Sources/SwiftProtobuf/TextFormatScanner.swift
@@ -897,8 +897,18 @@ internal struct TextFormatScanner {
 
     // If the next token is the identifier "nan", return true.
     private mutating func skipOptionalNaN() -> Bool {
-        return skipOptionalKeyword(bytes:
-                                  [asciiLowerN, asciiLowerA, asciiLowerN])
+        let start = p
+        // "-nan" doesn't mean anything, but upstream handles it, so skip
+        // over any leading minus when checking for "nan".
+        if p != end && p[0] == asciiMinus {
+            p += 1
+        }
+        if skipOptionalKeyword(bytes: [asciiLowerN, asciiLowerA, asciiLowerN]) {
+            return true
+        } else {
+            p = start  // It wasn't "nan", rewind incase we skipped a minus sign.
+            return false
+        }
     }
 
     // If the next token is a recognized spelling of "infinity",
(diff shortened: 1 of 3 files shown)


allevato (approved):

Sources/SwiftProtobuf/TextFormatScanner.swift
```diff
@@ -1243,40 +1253,40 @@ internal struct TextFormatScanner {
...
+        // Floats or decimals can have leading '-'
         if c == asciiMinus {
             scan += 1
             if scan == end { return false }
             c = scan[0]
         }
 
         if c == asciiPeriod {
-            return true  // "(-)." : clearly a float
+            return false  // "(-)." : clearly a float
```
tbkka, line 1270: So TextFormat doesn't accept `-.123` or `.123` as a floating-point number?

tbkka: Oh, I see.  This function is now checking for a decimal integer.  That's why the true/false swap.

Sources/protoc-gen-swift/Descriptor+Extensions.swift
```diff
@@ -352,13 +352,15 @@ extension FieldDescriptor {
         case "inf": return "Double.infinity"
         case "-inf": return "-Double.infinity"
         case "nan": return "Double.nan"
+        case "-nan": return "Double.nan"
```
tbkka, line 355: Should this return "-Double.nan"?

Sources/protoc-gen-swift/Descriptor+Extensions.swift
```diff
@@ -352,13 +352,15 @@ extension FieldDescriptor {
...
         default: return defaultValue
         }
       case .float:
         switch defaultValue {
         case "inf": return "Float.infinity"
         case "-inf": return "-Float.infinity"
         case "nan": return "Float.nan"
+        case "-nan": return "Float.nan"
```
tbkka, line 363: Likewise, "-Float.nan"?

Sources/protoc-gen-swift/Descriptor+Extensions.swift
thomasvl: There really isn't such a thing as negative nan, it's something the parsing is supposed to allow.

thomasvl: Question opened up with the protobuf-team about the tracking of the sign.

PR #1652, opened 2024-05-09
Implicit presence on floats should look at the bitpattern.

thomasvl:
https://protobuf.dev/programming-guides/proto3/#default ends with:

> If a float or double value is set to +0 it will not be serialized,
> but -0 is considered distinct and will be serialized.


allevato (approved):

thomasvl:
Upstream just landed a conformance test for this: https://github.com/protocolbuffers/protobuf/pull/16782

PR #1654, opened 2024-05-13
Add back removed public API

gjcairo:

Removing public API (either by deleting or changing the parameters of a function) is a breaking change, and this has been done in `main`.

This PR adds back the removed API to undo breaking changes. Some of these methods have been deprecated or annotated with `@_disfavoredOverload` to favour the new API instead.

Note: currently a draft because this depends on https://github.com/apple/swift-protobuf/pull/1612


Sources/SwiftProtobuf/AsyncMessageSequence.swift
```diff
@@ -179,15 +179,15 @@ public struct AsyncMessageSequence<
       }
       if messageSize == 0 {
         return try M(
-          serializedBytes: [],
+            contiguousBytes: [],
```
thomasvl, line 182: Indent also seems strange.

Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
```diff
@@ -49,6 +49,26 @@ extension Google_Protobuf_Any {
...
+  /// containing a serialized message in Protocol Buffer text format.
+  ///
+  /// - Parameters:
+  ///   - textFormatString: The text format string to decode.
+  ///   - extensions: An `ExtensionMap` used to look up and decode any
+  ///     extensions in this message or messages nested within this message's
+  ///     fields.
+  /// - Throws: an instance of `TextFormatDecodingError` on failure.
+  @_disfavoredOverload
+  public init(
```
thomasvl, line 63: Should we explicitly mark all of these legacy shims as inline?  Would it help keep the library size down if they aren't used?

Sources/SwiftProtobuf/Message+BinaryAdditions_Data.swift
```diff
@@ -31,14 +31,75 @@ extension Message {
   ///   - options: The ``BinaryDecodingOptions`` to use.
   /// - Throws: ``BinaryDecodingError`` if decoding fails.
   @inlinable
+  @available(*, deprecated, renamed: "init(contiguousBytes:extensions:partial:options:)")
```
thomasvl, line 34: I guess these might cause a lot of warnings for clients?  Should we do that one the first move to head, or wait one release to add these just to keep the noise of this migration down a little?

gjcairo: If we don't warn, users won't know we're planning on deprecating this and will not move. I think it's better to just get this out of the way now instead of having a couple of releases that may both cause some friction - I asked around the team and we all seem to agree on this.

thomasvl: Ugh, I didn't think about this before.  We've broken the api symmetry.

The original api design for all three formats was to have naming consistency between parsing and generation.  For the Binary format, we settled on _serialized_.  So with this change, generation will be `serializedBytes`, but parsing will be `continuousBytes`. 😞

Is there no way (with overloads, etc.) that we can keep the original naming symmetry to make such a common api easier to use?

thomasvl: @tbkka fyi

gjcairo: Ah, that's frustrating (but a fair point). I've worked around this by adding some new overloads - would you mind taking a look at the last commit and seeing if you agree with the workaround? If so I can regenerate the protos.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -75,11 +75,13 @@ extension Message {
     return visitor.serializedSize
   }
 
+
```
thomasvl, line 78: Might have picked up some extra blanks here.

Sources/SwiftProtobuf/Message+TextFormatAdditions.swift
```diff
@@ -53,6 +53,24 @@ extension Message {
...
+  ///   - textFormatString: The text format string to decode.
+  ///   - extensions: An ``ExtensionMap`` used to look up and decode any
+  ///     extensions in this message or messages nested within this message's
+  ///     fields.
+  /// - Throws: ``SwiftProtobufError`` on failure.
+  public init(
+    textFormatString: String,
+    extensions: (any ExtensionMap)? = nil
+  ) throws {
+    try self.init(textFormatString: textFormatString,
```
thomasvl, line 70: Should we have to a todo to eventually drop this at api breaking time?

Tests/SwiftProtobufPluginLibraryTests/Test_Descriptor.swift
```diff
@@ -26,7 +26,7 @@ extension Descriptor {
 final class Test_Descriptor: XCTestCase {
 
   func testParsing() throws {
-    let fileSet = try Google_Protobuf_FileDescriptorSet(serializedBytes: fileDescriptorSetBytes)
+    let fileSet = try Google_Protobuf_FileDescriptorSet(contiguousBytes: fileDescriptorSetBytes)
```
thomasvl, line 29: I _think_ the tests can stay on `serializedBytes` now? (so this whole PR would be a lot smaller?)

gjcairo: Yes, I held off from changing these and regenerating protos in case we didn't want to go for this approach, but I'll go ahead and make those changes now too

Tests/SwiftProtobufTests/Test_BinaryDecodingOptions.swift
```diff
@@ -199,14 +199,14 @@ final class Test_BinaryDecodingOptions: XCTestCase {
         //   }
         do {
             let msg1 = try SwiftProtoTesting_TestAllExtensions(
-              serializedBytes: inputSubMessage,
+                serializedBytes: inputSubMessage,
```
thomasvl, line 202: looks like some extra space in a few on the final diffs

thomasvl (approved):

Tests/SwiftProtobufTests/Test_BinaryDecodingOptions.swift
gjcairo: Some of the files have become messed up at some point: there are files exclusively with 2 spaces, some with exclusively 4, and some with mixed 2 and 4. This adds a lot of churn when using refactoring tools in Xcode (such as renaming funcs). At some point we should make sure the spacing across the whole repo is consistent, and/or consider introducing something such as swift-format.

PR #1656, opened 2024-05-18
Use repeat instead of break label for utf8ToString

kyulee-com:
This update doesn't change functionality, but improves the (LLVM) loop optimizer. Using labeled statement with an infinite loop created a complex nested loop, which led inefficient codegen. More importantly it could cause mis-compilation depending on a LTO flag. This patch uses a repeat command with a boolean check to make the loop simpler and more efficient.

### DIFF
--- a/Sources/SwiftProtobuf/StringUtils.swift
+++ b/Sources/SwiftProtobuf/StringUtils.swift
@@ -88,17 +88,18 @@ internal func utf8ToString(bytes: UnsafeRawPointer, count: Int) -> String? {
   // Verify that the UTF-8 is valid.
   var p = sourceEncoding.ForwardParser()
   var i = codeUnits.makeIterator()
-  Loop:
-  while true {
+  var repeatLoop: Bool
+  repeat {
+    repeatLoop = false
     switch p.parseScalar(from: &i) {
     case .valid(_):
       break
     case .error:
       return nil
     case .emptyInput:
-      break Loop
+      repeatLoop = true
     }
-  }
+  } while repeatLoop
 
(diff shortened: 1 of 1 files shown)


kyulee-com:
@thomasvl @gjcairo I'm not sure if this is a right process to request a PR here. But found this code was problematic in our code base in a certain combination of opts, and the suggested change produced much cleaner code.  Let me know how to test this code.

thomasvl:
If there compiler does better with the change, I think there are a few other loops in the library that maybe should be tweaked.

Lukasa (changes requested):
Gus's callout here is great, the logic here has completely been inverted.

I'm also curious about the bad codegen and the LLVM miscompile. Do you have an llvm issue for the latter, and can you show us what you meant for the former? The code here seems likely to compile absolutely trivially, the labelled break should be almost irrelevant from LLVM's perspective.

kyulee-com:
Thanks all for looking at this issue. It seems my first attempt was entirely off, but after further investigation, I've come up with another workaround. I've reported a bug in LLVM, which you can find at https://github.com/llvm/llvm-project/issues/93110. I don't intend to upstream this workaround in the source as it seems too disruptive, but perhaps use this workaround in our downstream. Anyhow, I'm sharing this with a comment for reference.

Lukasa:
Thanks so much for your diligent work root-causing this, and for providing a workaround. As we only see a severe issue in thin-lto mode I'm ok with us calling the current code good, and waiting for Swift to fix the apparent miscompile issue. I've flagged this for the Swift team.

PR #1663, opened 2024-06-10
Fix handling of map fields inheriting delimited encoding

mkruskal-google:
A conformance test will be added to cover this case in the protobuf repo as well.  Map fields should never use delimited encoding today.  This may change in a future edition, but the semantics are unclear due to the generated message type and proto2 never allowed group maps.


Sources/SwiftProtobufPluginLibrary/Descriptor.swift
```diff
@@ -791,7 +791,9 @@ public final class FieldDescriptor {
   static let kLastReservedNumber: Int = 19999
 
   /// Declared type of this field.
-  public let type: Google_Protobuf_FieldDescriptorProto.TypeEnum
+  public var type: Google_Protobuf_FieldDescriptorProto.TypeEnum { return _type }
+  private var _type: Google_Protobuf_FieldDescriptorProto.TypeEnum
```
tbkka, line 795: The more idiomatic expression here is:
```
public private(set) var type: Google_Protobuf_FieldDescriptorProto.TypeEnum
```

thomasvl: I think I've mainly been using the private `_` version to help avoid accidentally setting things in any other place within the implementation. But either works.

tbkka:
Tests are failing due to the missing Reference for the new .proto.   And there's an API breakage message that looks unrelated to this PR.  Hmmm...

mkruskal-google:
Fixed the reference, not sure about the API breakage

tbkka:
@FranzBusch Any thoughts about the API breakage reported here?

thomasvl:
> @FranzBusch Any thoughts about the API breakage reported here?

I believe that's existing, my CL bumping the CI version also got it: https://github.com/apple/swift-protobuf/actions/runs/9421756383/job/25956524755

thomasvl:
@gjcairo also for the api breaks report.

gjcairo:
Those two types have been conformed to `Sendable` and annotated with `@preconcurrency` (see https://github.com/apple/swift-protobuf/blob/main/Sources/SwiftProtobuf/ExtensionMap.swift#L25 and https://github.com/apple/swift-protobuf/blob/main/Sources/SwiftProtobuf/MessageExtension.swift#L19). 
I’ve also added them to the known breaks allow list: https://github.com/apple/swift-protobuf/blob/main/known_api_breaks.txt#L14 and https://github.com/apple/swift-protobuf/blob/main/known_api_breaks.txt#L14

To me this seems like a bug in the API breakage reporting tool.

Lukasa:
But why is _this_ PR triggering the API break?

Lukasa:
Oh, because we compare against last release, not against `main`. Not sure that's the right thing to do here.

tbkka:
The API break report certainly isn't specific to this PR, so I'll merge this PR.

thomasvl:
> Oh, because we compare against last release, not against `main`. Not sure that's the right thing to do here.

I'll rearrange those checks once we make a new release from `main` (and retire the 1.x branch).

PR #1666, opened 2024-06-25
Undo breaking changes in SwiftProtobufPluginLibrary

gjcairo:
When we were undoing breaking changes in `main` to tag the next minor release, we forgot to check for breaking changes on `SwiftProtobufPluginLibrary` and only fixed those on the `SwiftProtobuf` target.

This PR undoes all breaking changes in `SwiftProtobufPluginLibrary`. It mostly required re-adding removed properties (or undo renames) while still keeping the new ones, and deprecating the old ones.


Sources/SwiftProtobufPluginLibrary/Descriptor.swift
```diff
@@ -436,14 +503,71 @@ public final class Descriptor {
...
-    return self.oneofs.filter { !$0.isSynthetic }
+    return self.oneofs.filter { !$0._isSynthetic }
   }()
   /// The extension field defintions under this message.
   public let extensions: [FieldDescriptor]
 
   /// The extension ranges declared for this message. They are returned in
   /// the order they are defined in the .proto file.
-  public let extensionRanges: [ExtensionRange]
+  public let protoExtensionRanges: [ExtensionRange]
```
thomasvl, line 513: This doesn't look right.  This is prefix "proto", but it isn't the one return proto message objects.  Is there something we can do for better naming going forward?

Can you overload properties on return type and @_disfavoredOverload?

Sources/SwiftProtobufPluginLibrary/Descriptor.swift
```diff
@@ -452,6 +576,10 @@ public final class Descriptor {
   /// order they are defined in the .proto file.
   public let reservedNames: [String]
 
+  /// True/False if this Message is just for a `map<>` entry.
+  @available(*, deprecated, message: "Please open a GitHub issue if you think functionality is missing.")
```
thomasvl, line 580: Maybe some of these should get `renamed: options.mapEntry` annotations so they can just migrate to the new api?

Sources/SwiftProtobufPluginLibrary/Descriptor.swift
```diff
@@ -465,12 +593,16 @@ public final class Descriptor {
   // Storage for `file`, will be set by bind()
   private unowned var _file: FileDescriptor?
 
+  @available(*, deprecated, message: "Please open a GitHub issue if you think functionality is missing.")
+  public var useMessageSetWireFormat: Bool { return proto.options.messageSetWireFormat }
```
thomasvl, line 597: add renamed:?

Sources/protoc-gen-swift/FieldGenerator.swift
```diff
@@ -89,7 +89,7 @@ class FieldGeneratorBase {
         result = ".standard(proto: \"\(protoName)\")"
       } else {
         /// The library's generation didn't match, so specify this explicitly.
-        result = ".unique(proto: \"\(protoName)\", json: \"\(jsonName)\")"
+        result = ".unique(proto: \"\(protoName)\", json: \"\(String(describing: jsonName))\")"
```
thomasvl, line 92: What was the issue here?

Sources/SwiftProtobufPluginLibrary/Descriptor.swift
```diff
@@ -171,6 +217,14 @@ public final class FileDescriptor {
   /// The service defintions at the file scope level.
   public let services: [ServiceDescriptor]
 
+  /// The `Google_Protobuf_FileOptions` set on this file.
+  @available(*, deprecated, renamed: "options")
+  public var fileOptions: Google_Protobuf_FileOptions {
+    get {
```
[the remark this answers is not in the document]
gjcairo: Good catch, it's not

Sources/SwiftProtobufPluginLibrary/Descriptor.swift
gjcairo: > Can you overload properties on return type and @_disfavoredOverload?

Sadly this is not possible.

I don't really like the name either but I'm not sure what a better alternative is, so happy to receive suggestions :D 
A maybe obvious one is `descriptorExtensionRanges` - not nice to repeat `descriptor` (because uses will likely be `descriptor.descriptorExtensionRanges`) but not the end of the world and may be clearer.

Sources/SwiftProtobufPluginLibrary/Descriptor.swift
```diff
@@ -372,7 +439,7 @@ public final class Descriptor {
     public var fullName: String { return containingType.fullName }
 
     /// The .proto file in which this ExtensionRange was defined.
-    public var file: FileDescriptor { return containingType.file }
+    public var file: FileDescriptor! { return containingType.file }
```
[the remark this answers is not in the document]
gjcairo: This isn't possible :(

Sources/SwiftProtobufPluginLibrary/Descriptor.swift
thomasvl: Ugh.  maybe `extensionRangeDescriptors`?  Both options are non idea.

gjcairo: What about `messageExtensionRanges` (following the wording used in the documentation for the prop)?

thomasvl: That works and seems better then the other two.

Sources/SwiftProtobufPluginLibrary/Descriptor.swift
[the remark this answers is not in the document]
gjcairo:
```swift
if let file = descriptor.file {
  // do something with `file`
}
```
This will break if we change the type from `FileDescriptor!` to `FileDescriptor`: even though the former is implicitly unwrapped, it's still an `Optional` so it can be used like one everywhere (so switching over it would break too for example).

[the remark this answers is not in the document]
FranzBusch: This isn't a warning it just straight up generates an error

```swift
let nonOptional: Int = 1
let implicitOptiona: Int! = 1
if let i = nonOptional {} // error: Initializer for conditional binding must have Optional type, not 'Int'
if let i = implicitOptiona {}
```

I agree that this is annoying but it is still a breaking change and we never know if anybody out there does the `if let` dance.

PR #1667, opened 2024-06-27
Update API breakage GH actions to run for SwiftProtobufPluginLibrary

gjcairo:
We had missed to check that no API breaks had happened on the `SwiftProtobufPluginLibrary` before tagging the next minor from `main`.

Because `SwiftProtobufPluginLibrary` is a public product, I've modified the GH actions we've got to check API breaks to remove the explicit products flag, which means we'll check API breaks for all targets.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -127,7 +127,8 @@ jobs:
       # https://github.com/actions/checkout/issues/766
       run: git config --global --add safe.directory ${GITHUB_WORKSPACE}
     - name: Check for API breaking changes
-      run: swift package diagnose-api-breaking-changes origin/main --products SwiftProtobuf
+      run: |
+        swift package diagnose-api-breaking-changes origin/main --breakage-allowlist-path known_api_breaks.txt
 
   api-breakage-since-last-release:
     name: Api Breakage Compared to Last Release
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -127,7 +127,8 @@ jobs:
       # https://github.com/actions/checkout/issues/766
       run: git config --global --add safe.directory ${GITHUB_WORKSPACE}
     - name: Check for API breaking changes
-      run: swift package diagnose-api-breaking-changes origin/main --products SwiftProtobuf
+      run: |
+        swift package diagnose-api-breaking-changes origin/main --breakage-allowlist-path known_api_breaks.txt
```
thomasvl, line 131: I guess this means it is also checking the protoc plugin and the swiftpm plugin?  Guess that doesn't hurt or should we just list the two products we actually want checked?

FranzBusch: Yeah I asked Gus to change this. Since calling out specific products already proved to be not enough. In all of our other repos we run this tool for all target and haven't had issues with that yet. I would prefer just running it against all targets and occasionally force-merging over an expected break in a non-public target than forgetting to add a target to this check script.

thomasvl (approved):

PR #1670, opened 2024-07-08
Update the api check.

thomasvl:
- With `main` now being used for releases, just need to check against the last release.
- Remove all the known breaks issues as we've done a release and thus reset the list.


thomasvl:
@gjcairo @FranzBusch @tbkka fyi.

allevato (approved):

FranzBusch (commented):
I would recommend to do the inverse and keep the breaking change against `main` and remove the one against the latest release. We do this in our other repos. This is really useful because sometimes the tool reports ABI but not API breaks which we force-merge over. If we compare against the release we have to force merge until we cut another release.

gjcairo (approved):

thomasvl:
> I would recommend to do the inverse and keep the breaking change against `main` and remove the one against the latest release. We do this in our other repos. This is really useful because sometimes the tool reports ABI but not API breaks which we force-merge over. If we compare against the release we have to force merge until we cut another release.

Sorry, already hit submit.  Will follow up.

So you don't bother with a known api breaks file then?

FranzBusch:
> So you don't bother with a known api breaks file then?

Correct. We make the decision once in a PR where we have to merge over it and then consider it baseline again.

thomasvl:
> > So you don't bother with a known api breaks file then?
> 
> Correct. We make the decision once in a PR where we have to merge over it and then consider it baseline again.

Someone on the Apple side will have to do any configuration need to allow the force merge, I'm not sure if something needs to be tweaked in the project configuration on GitHub to allow that.

PR #1671, opened 2024-07-08
Revert "Use older base image for sanitizer and fuzz testing."

thomasvl:
This reverts commit b3c2279e1acd28f7a8c384daaea7430e255f3347.


thomasvl:
This original CL was done when the fuzz testing broke because of a change in the bast images that fuzz didn't work with. That issues has since been resolved, to moving back to more current images just to help avoid the need for future updates/maintenance from being pinned to older things.

allevato (approved):

PR #1672, opened 2024-07-08
Move api check to just against main.

thomasvl:
Then we don't need to track breaks, we explicit allow them when merging.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -109,13 +109,11 @@ jobs:
       working-directory: main
       run: make compile-tests PROTOC=../protobuf/cmake_build/protoc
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -110,12 +110,10 @@ jobs:
       run: make compile-tests PROTOC=../protobuf/cmake_build/protoc
 
   api-breakage-since-last-release:
```
FranzBusch, line 112:
```suggestion
  api-breakage:
```

thomasvl: Done

FranzBusch (approved):

PR #1679, opened 2024-07-15
Let CodeGenerators provide extension for CustomOptions.

### LINKED ISSUE #1677: New `CodeGenerator` protocol should allow customising the `Google_Protobuf_Compiler_CodeGeneratorRequest`
When building a code generator which conforms to the `CodeGenerator` protocol, we are asked to implement a `generate` function.
The `CodeGenerator` protocol defines an extension which gives us the `main` function, which calls our `generate` function. 
This `main` function creates a `Google_Protobuf_Compiler_CodeGeneratorRequest` under the hood, which is ultimately passed to `DescriptorSet` (see [here](https://github.com/apple/swift-protobuf/blob/37dc6513588ee3d310c020278d87c61569c99221/Sources/SwiftProtobufPluginLibrary/CodeGenerator.swift#L115-L117)).

There is currently no way to customise the `Google_Protobuf_Compiler_CodeGeneratorRequest`, for example to add compiler extensions:

```swift
let request = try Google_Protobuf_Compiler_CodeGeneratorRequest(
    serializedBytes: FileHandle.standardInput.readDataToEndOfFile(),
    extensions: [some extensions here]
)
```

This forces us to re-implement main, duplicating all the logic, just so that we can add extensions.

Note that, prior to 1.27, this problem did not exist because `CodeGenerator` did not exist. We had to manually build the entire code generation logic by hand. 
However, that path is now deprecated with warnings encouraging us to move to using `CodeGenerator`. Therefore, it would be ideal for `CodeGenerator` to be flexible enough to support all use-cases.

## Potential solutions

- The `CodeGenerator` protocol could allow the implementor to specify a list of extensions, similar to how we can currently [specify a list of supportedFeatures](https://github.com/apple/swift-protobuf/blob/37dc6513588ee3d310c020278d87c61569c99221/Sources/SwiftProtobufPluginLibrary/CodeGenerator.swift#L48).
- The `main` function logic could take an optional argument of a closure or other means of customising the request.

thomasvl:
This will get those extensions recognized when the request from protoc is parsed, and then a generator can use the existing SwiftProtobuf extension related apis to access them.

Fixes #1677

### DIFF
--- a/Sources/SwiftProtobufPluginLibrary/CodeGenerator.swift
+++ b/Sources/SwiftProtobufPluginLibrary/CodeGenerator.swift
@@ -52,6 +52,11 @@ public protocol CodeGenerator {
   /// error.
   var supportedEditionRange: ClosedRange<Google_Protobuf_Edition> { get }
 
+  /// A list of extensions that define Custom Options
+  /// (https://protobuf.dev/programming-guides/proto2/#customoptions) for this generator so
+  /// they will be exposed on the `Descriptor` options.
+  var customOptionExtensions: [any AnyMessageExtension] { get }
+
   /// If provided, the argument parsing will support `--version` and report
   /// this value.
   var version: String? { get }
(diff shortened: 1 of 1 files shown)


thomasvl:
@gjcairo how's this look for your use case?

Sources/SwiftProtobufPluginLibrary/CodeGenerator.swift
```diff
@@ -110,10 +115,28 @@ extension CodeGenerator {
...
+        // Don't include Google_Protobuf_FeatureSet, that will be handing via custom features.
+        precondition(e.messageType == Google_Protobuf_EnumOptions.self ||
+                     e.messageType == Google_Protobuf_EnumValueOptions.self ||
+                     e.messageType == Google_Protobuf_ExtensionRangeOptions.self ||
+                     e.messageType == Google_Protobuf_FieldOptions.self ||
+                     e.messageType == Google_Protobuf_FileOptions.self ||
+                     e.messageType == Google_Protobuf_MethodOptions.self ||
+                     e.messageType == Google_Protobuf_OneofOptions.self ||
+                     e.messageType == Google_Protobuf_ServiceOptions.self,
+                     "CodeGenerator `customOptionExtensions` must only extend the descriptor.proto 'Options' messages \(e.messageType).")
```
gjcairo, line 130: Shouldn't we also allow `Google_Protobuf_MessageOptions`?

Sources/SwiftProtobufPluginLibrary/CodeGenerator.swift
```diff
@@ -52,6 +52,11 @@ public protocol CodeGenerator {
   /// error.
   var supportedEditionRange: ClosedRange<Google_Protobuf_Edition> { get }
 
+  /// A list of extenions that define Custom Options
```
gjcairo, line 55: Nit:
```suggestion
  /// A list of extensions that define Custom Options
```

gjcairo:
Thanks @thomasvl ! Other than the comment about the missing `Google_Protobuf_MessageOptions` this looks good :)

Sources/SwiftProtobufPluginLibrary/CodeGenerator.swift
thomasvl: Oops, yup.

allevato (approved):

PR #1680, opened 2024-07-23
Fix FuzzTests unit tests

gjcairo:
One of the assertions in the TextFormat fuzz-originated unit tests checked that we failed to decode a string built from a sequence of bytes. However, this sequence of bytes had invalid UTF8 characters. `Foundation` has improved in recent betas and now fails to build a String from this sequence (which it didn't before). This is causing the test to fail.

If the String couldn't have been constructed during fuzz testing, the test would have passed - see:
https://github.com/apple/swift-protobuf/blob/601e03a9ccff83868833f2e4268ca2650a7503c4/FuzzTesting/Sources/FuzzTextFormat/main.swift#L20

We should tolerate the String not being constructed here without failing the test, as it's the correct (and current) behaviour. This means that after this PR, assertions will be skipped when a String cannot be constructed.
However, we shouldn't simply remove the failing assertion, because in the case `Foundation` introduces this bug again, we should make sure we're failing the TextFormat decoding.


thomasvl:
Fyi - #1590 is the PR that removed the original test case.  We'd pulled it in to be a unittest to help keep a better eye on the handling.

thomasvl (approved):

PR #1682, opened 2024-07-24
Update links for repositories moved to the swiftlang org on GitHub

lamtrinhdev:
### Summary:

Update links for repositories moved to the swiftlang org on GitHub

### Modifications:

Update the link
+ https://github.com/apple/swift/ => https://github.com/swiftlang/swift/
+ https://github.com/apple/swift-docc-plugin  =>  https://github.com/swiftlang/swift-docc-plugin

### Result:

Correct the reference
+ https://github.com/apple/swift/ => https://github.com/swiftlang/swift/
+ https://github.com/apple/swift-docc-plugin  =>  https://github.com/swiftlang/swift-docc-plugin

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -32,7 +32,7 @@ let package = Package(
     ),
   ],
   dependencies: [
-    .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
+    .package(url: "https://github.com/swiftlang/swift-docc-plugin", from: "1.0.0"),
   ],
   targets: [
     .target(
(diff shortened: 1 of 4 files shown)


Package.swift
```diff
@@ -32,7 +32,7 @@ let package = Package(
     ),
   ],
   dependencies: [
-    .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
+    .package(url: "https://github.com/swiftlang/swift-docc-plugin", from: "1.0.0"),
```
thomasvl, line 35: Anyone know if this causes problems if some project depending on us ends up with two things depending on docc via different urls?

thomasvl: I'm hoping the Apple folks can chime in on if this would be a problem or not.

FranzBusch: Good question. I haven't tried this out yet.

tbkka: We should update it.

If different URLs are a problem, that's true regardless of whether we update, since other people will.

tbkka (approved):

PR #1683, opened 2024-07-24
Support access level on import statements

Skoti:
This PR brings in the possibility to define access level on import statements generated through `ProtoPathModuleMappings` plugin option.
Access level on import statements is a new feature of Swift 6, but it can already be enabled in Swift 5.9 builds.

See https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md

### DIFF
--- a/Documentation/PLUGIN.md
+++ b/Documentation/PLUGIN.md
@@ -184,8 +184,9 @@ this will only work if the `Visibility` is set to `internal`.
 
 ##### Generation Option: `UseAccessLevelOnImports` - imports preceded by a visibility modifier (`public`, `package`, `internal`)
 
-By default, the code generator does not precede any imports with a visibility modifier.
-You can change this with the `UseAccessLevelOnImports` option:
+The default behavior depends on the Swift version the plugin is compiled with. 
+For Swift versions below 6.0 the default is `false` and the code generator does not precede any imports with a visibility modifier. 
+You can change this by explicitly setting the `UseAccessLevelOnImports` option:
 
 ```
 $ protoc --swift_opt=UseAccessLevelOnImports=[value] --swift_out=. foo/bar/*.proto mumble/*.proto
@@ -193,9 +194,12 @@ $ protoc --swift_opt=UseAccessLevelOnImports=[value] --swift_out=. foo/bar/*.pro
 
 The possible values for `UseAccessLevelOnImports` are:
 
-* `false` (default): Generates plain import directives without a visibility modifier.
+* `false`: Generates plain import directives without a visibility modifier.
 * `true`: Imports of internal dependencies and any modules defined in the module
 mappings will be preceded by a visibility modifier corresponding to the visibility of the generated types - see `Visibility` option. 
+
+**Important:** It is strongly encouraged to use `internal` imports instead of `@_implementationOnly` imports. 
+Hence `UseAccessLevelOnImports` and `ImplementationOnlyImports` options exclude each other. 
 
 
 ### Building your project
--- a/Documentation/PLUGIN.md
+++ b/Documentation/PLUGIN.md
@@ -182,6 +182,22 @@ exposed via public API, so even if `ImplementationOnlyImports` is set to `true`,
 this will only work if the `Visibility` is set to `internal`.
 
 
+##### Generation Option: `UseAccessLevelOnImports` - imports preceded by a visibility modifier (`public`, `package`, `internal`)
+
+By default, the code generator does not precede any imports with a visibility modifier.
+You can change this with the `UseAccessLevelOnImports` option:
+
+```
+$ protoc --swift_opt=UseAccessLevelOnImports=[value] --swift_out=. foo/bar/*.proto mumble/*.proto
+```
+
+The possible values for `UseAccessLevelOnImports` are:
+
+* `false` (default): Generates plain import directives without a visibility modifier.
+* `true`: Imports of internal dependencies and any modules defined in the module
+mappings will be preceded by a visibility modifier corresponding to the visibility of the generated types - see `Visibility` option. 
+
+
 ### Building your project
 
 After copying the `.pb.swift` files into your project, you will need
--- a/PluginExamples/Package.swift
+++ b/PluginExamples/Package.swift
@@ -7,14 +7,22 @@ let package = Package(
     dependencies: [
         .package(path: "../")
     ],
-    targets: [
+    targets: targets()
+)
+
+private func targets() -> [Target] {
+    var testDependencies: [Target.Dependency] = [
+        .target(name: "Simple"),
+        .target(name: "Nested"),
+        .target(name: "Import"),
+    ]
+#if compiler(>=5.9)
+    testDependencies.append(.target(name: "AccessLevelOnImport"))
+#endif
+    var targets: [Target] = [
         .testTarget(
             name: "ExampleTests",
-            dependencies: [
-                .target(name: "Simple"),
-                .target(name: "Nested"),
-                .target(name: "Import"),
-            ]
+            dependencies: testDependencies
         ),
         .target(
             name: "Simple",
@@ -44,4 +52,21 @@ let package = Package(
             ]
         ),
     ]
-)
+#if compiler(>=5.9)
+    targets.append(
+        .target(
+            name: "AccessLevelOnImport",
+            dependencies: [
+                .product(name: "SwiftProtobuf", package: "swift-protobuf"),
+            ],
+            swiftSettings: [
+                .enableExperimentalFeature("AccessLevelOnImport"),
+            ],
+            plugins: [
+                .plugin(name: "SwiftProtobufPlugin", package: "swift-protobuf")
+            ]
+        )
+    )
+#endif
+    return targets
+}
--- a/Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
+++ b/Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
@@ -20,7 +20,7 @@ private let defaultSwiftProtobufModuleName = "SwiftProtobuf"
 public struct ProtoFileToModuleMappings {
 
   /// Errors raised from parsing mappings
-  public enum LoadError: Error {
+  public enum LoadError: Error, Equatable {
     /// Raised if the path wasn't found.
     case failToOpen(path: String)
     /// Raised if an mapping entry in the protobuf doesn't have a module name.
--- a/Sources/protoc-gen-swift/GeneratorOptions.swift
+++ b/Sources/protoc-gen-swift/GeneratorOptions.swift
@@ -79,7 +79,11 @@ class GeneratorOptions {
     var visibility: Visibility = .internal
     var swiftProtobufModuleName: String? = nil
     var implementationOnlyImports: Bool = false
+#if swift(>=6.0)
+    var useAccessLevelOnImports = true
+#else
     var useAccessLevelOnImports = false
+#endif
     var experimentalStripNonfunctionalCodegen: Bool = false
 
     for pair in parameter.parsedPairs {
--- a/Sources/protoc-gen-swift/GeneratorOptions.swift
+++ b/Sources/protoc-gen-swift/GeneratorOptions.swift
@@ -47,19 +47,19 @@ class GeneratorOptions {
 
     var isAccessLevel: Bool {
         switch self {
-        case .accessLevel: true
-        default: false
+        case .accessLevel: return true
+        default: return false
         }
     }
 
     var snippet: String {
       switch self {
       case let .accessLevel(visibility):
-        "\(visibility) import"
+        return "\(visibility) import"
       case .plain:
-        "import"
+        return "import"
       case .implementationOnly:
-        "@_implementationOnly import"
+        return "@_implementationOnly import"
       }
     }
   }
@@ -169,10 +169,10 @@ class GeneratorOptions {
 
     self.experimentalStripNonfunctionalCodegen = experimentalStripNonfunctionalCodegen
 
-    self.importDirective = switch (implementationOnlyImports, useAccessLevelOnImports) {
-    case (false, false): .plain
-    case (false, true): .accessLevel(visibility)
-    case (true, false): .implementationOnly
+    switch (implementationOnlyImports, useAccessLevelOnImports) {
+    case (false, false): self.importDirective = .plain
+    case (false, true): self.importDirective = .accessLevel(visibility)
+    case (true, false): self.importDirective = .implementationOnly
     case (true, true): throw GenerationError.message(message: """
       When using access levels on imports the @_implementationOnly option is unnecessary.
       Disable @_implementationOnly imports.
--- a/Sources/protoc-gen-swift/GeneratorOptions.swift
+++ b/Sources/protoc-gen-swift/GeneratorOptions.swift
@@ -30,21 +30,36 @@ class GeneratorOptions {
     }
   }
 
-  enum Visibility {
+  enum Visibility: String {
     case `internal`
     case `public`
     case `package`
 
     init?(flag: String) {
-      switch flag.lowercased() {
-      case "internal":
-        self = .internal
-      case "public":
-        self = .public
-      case "package":
-        self = .package
-      default:
-        return nil
+      self.init(rawValue: flag.lowercased())
+    }
+  }
+
+  enum ImportDirective: Equatable {
+    case accessLevel(Visibility)
+    case plain
+    case implementationOnly
+
+    var isAccessLevel: Bool {
+        switch self {
+        case .accessLevel: true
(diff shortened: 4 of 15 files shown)


thomasvl:
This feels like a lot of overlap with #1393 that added `@_implementationOnly` support, at a minimum it would need to be rationalize with that so one can't mix the options and get invalid Swift code.

I'm also not sure this can be done on a per module bases.  If a given `.proto` file imports another one, then it's going to use the types for fields/adding extension fields/etc. So it seems like there would have to be some alignment between the imports and how the file being generated will declare things.  i.e. - if the file is generating with `public` visibility, then can it really do `internal` or `private` on the `import` statements for modules for types likely needed?

@FranzBusch @gjcairo your thoughts in this space since I think you were the ones to look the most at the impl only support.

Skoti:
> if the file is generating with public visibility, then can it really do internal or private on the import statements for modules for types likely needed?

It can't as such Swift code wouldn't compile, however, to me this is simply a misconfiguration. But yeah I could add some validation steps if that's required.

> This feels like a lot of overlap with https://github.com/apple/swift-protobuf/pull/1393 that added @_implementationOnly support, at a minimum it would need to be rationalize with that so one can't mix the options and get invalid Swift code.

That's right. Basically, the access level you can use depends on the level of types you generate.
If the files are generated with `public` access you can only use `@_exported public import` or `public import`.
If the files are generated with `package` access (it's not possible right now in SwiftProtobuf), you can use the same as above + `package import`.
If the files are generated with `internal` access you can use the same as above + `internal import`.

We could go further and make `ProtoPathModuleMappings` and `ImplementationOnlyImports` options compatible.
What `ImplementationOnlyImports` does right now is that it globally adds `@_implementationOnly` which means it changes the default statement from `import` to `@_implementationOnly import` for all imports.

We could introduce `DefaultAccessLevel` option that would do the same but allow you to specify what will be the default access level (not only one hardcoded) for all imports, and the `ProtoPathModuleMappings` would allow you to override this behavior for the specified modules.

`ImplementationOnlyImports` would be deprecated and translated to `DefaultAccessLevel` with `@_implementationOnly` for backward compatibility.

Any configurations that would produce invalid Swift code could either be validated or left to fail in compile-time. The choice is yours.

thomasvl:
I don't think I immediately see how this is how this is a per module decision then, seems like it would make more sense to module model it just a single option since it has to line up with the generated access level.  Maybe even just a boolean to say if access level imports should be generated and they should then be inferred from the generated access level?

Skoti:
> I don't think I immediately see how this is how this is a per module decision then, seems like it would make more sense to module model it just a single option since it has to line up with the generated access level. Maybe even just a boolean to say if access level imports should be generated and they should then be inferred from the generated access level?

This would restrict what you can do with access level on imports.
It must be per module, because you may wish to expose each imported module differently.
For example, this should be possible in the generated `pb.swift` file if the access is `internal` for the generated proto:
```swift
@_exported public import ModuleWithReExportedProtos
public import ModuleWithProtosWithPublicVisibility
import InternalProtos

struct ProtoGeneratedAsinternal {
  var a: MessageFromReExportedModule
  var b: MessageFromPublicModule
  var c: MessageFromInternalModule
}
```

FranzBusch:
I agree with @thomasvl here that I also don't fully see how this can be done on a per module level. In your last example why would you publicly import a module if the generated protos are internal? Can't we simply this and say that the generated access level should match the import access level? (Taking `@_implementationOnly` out of the question for a second)

Skoti:
> Can't we simply this and say that the generated access level should match the import access level? (Taking @_implementationOnly out of the question for a second)

We can't because the import access level means much more. You may still want to expose your transitive dependencies differently.
It's the same answer, doing this would restrict how you could use the access levels.

Imagine a scenario like this:

`PublicProtos` <- `Main` <- end-client

- A module named `PublicProtos` contains a set of protos that are shared publicly.
- A module named `Main` depends on `PublicProtos` - it has public APIs that can take/return protos from `PublicProtos`, and it contains internal protos (the generated ones in question) some of which rely on protos from `PublicProtos`.
- An end-client that uses `Main` and transitively `PublicProtos`

Now, because you're using `PublicProtos` in public APIs of `Main` you must import them at least with `public import`.
This means the `internal` protos you generate for the `Main` module also need to use `public import`, otherwise you'll get an error saying:
`Ambiguous implicit access level for import of 'ProtoName'; it is imported as 'public' elsewhere`

The same is if you additionally wish to re-export `PublicProtos` types from the `Main` module, so the end-client doesn't have to rely on importing `PublicProtos` as well, they can simply import `Main`.
In that case you would need to import `PublicProtos` inside `Main` module with `@_exported public import`, and the same kind of import must be used in the generated internal protos even though they're internal - otherwise you'll get the "ambiguous access level" error.

In the future the `Main` module may depend on another one named `MyProtos`, but this one will be used internally only, so you don't want to use `public import`.

So this is module-based, you decide how each dependency is exposed to the end-client.

thomasvl:
Can you post a small SwiftPM package example showing the case you say is getting a warning? You should be able to do it with some trivial sources that don't involve protos, just basic Swift.  It would be safer to make sure everyone is considering the _same_ cases when discussion things.

Is the warning the same between Swift 5.x and Swift 6?  It's possible the compiler has tweaked the behavior between versions.

I'm also sorta curing about the mention of _implicit_ in that error message. It makes me wonder if an _explicit_ `internal` for the generated proto sources would solve the problem.

FranzBusch:
> Now, because you're using PublicProtos in public APIs of Main you must import them at least with public import.
This means the internal protos you generate for the Main module also need to use public import, otherwise you'll get an error saying:
Ambiguous implicit access level for import of 'ProtoName'; it is imported as 'public' elsewhere

I think that's the key here and I wasn't aware that restriction existed. Now that makes total sense. The access level of imports is checked module wide and not on a per file basis.

Skoti:
> I'm also sorta curing about the mention of implicit in that error message. It makes me wonder if an explicit internal for the generated proto sources would solve the problem.

Oh, I have missed that, yes **explicit internal** solves this! Thanks, for pointing  that out!
I wonder if this is the same in Swift 6. I'm using Swift `5.10` but I did enable `InternalImportsByDefault` 🤷‍♂️.

> The access level of imports is checked module wide and not on a per file basis.

True, but I've just found out they're reducing it to the most visible declaration, and they only validate that declarations are not relying on imports with a lower access.


And I've found this in the proposal as a confirmation, and it works in the code too!

> The same dependency can be imported with different access levels by different files of a same module. At the module level, we only take into account the most permissive access level. For example, if a dependency is imported as package and internal from two different files, we consider the dependency to be of package visibility at the module level.

> Current type-checking enforces that declaration respect their respective access levels. It reports as errors when a more visible declaration refers to a less visible declaration.

Ok, I think we can go with a simpler approach here as was suggested. Would you like me to modify this PR?

FranzBusch:
> Ok, I think we can go with a simpler approach here as was suggested. Would you like me to modify this PR?

Yes please. I think we can just match the generated access level

Skoti:
> Maybe even just a boolean to say if access level imports should be generated and they should then be inferred from the generated access level?

> Yes please. I think we can just match the generated access level

Done ✅ 
Please take a look 🙂

Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
```diff
@@ -20,7 +20,7 @@ private let defaultSwiftProtobufModuleName = "SwiftProtobuf"
 public struct ProtoFileToModuleMappings {
 
   /// Errors raised from parsing mappings
-  public enum LoadError: Error {
+  public enum LoadError: Error, Equatable {
```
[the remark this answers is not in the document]
Skoti: It was added manually here anyway, I've just deleted this so it's synthesized by the compiler.
https://github.com/apple/swift-protobuf/blob/3a381048e04a37a441e61e5801568f8dad29e07f/Tests/SwiftProtobufPluginLibraryTests/Test_ProtoFileToModuleMappings.swift#L17-L27

Skoti:
> Does it still work with 5.8 or does it need newer Swift versions to allow us to do that?

It's available since Swift 5.9, but you need to enable it manually in the hosting package/project.
Example for swift package:
```swift
            swiftSettings: [
                .enableExperimentalFeature("AccessLevelOnImport"),
                .enableUpcomingFeature("InternalImportsByDefault") // optionally this too
            ]
```

Documentation/PLUGIN.md
```diff
@@ -182,6 +182,22 @@ exposed via public API, so even if `ImplementationOnlyImports` is set to `true`,
...
+You can change this with the `UseAccessLevelOnImports` option:
+
+```
+$ protoc --swift_opt=UseAccessLevelOnImports=[value] --swift_out=. foo/bar/*.proto mumble/*.proto
+```
+
+The possible values for `UseAccessLevelOnImports` are:
+
+* `false` (default): Generates plain import directives without a visibility modifier.
+* `true`: Imports of internal dependencies and any modules defined in the module
```
FranzBusch, line 197: I am wondering if we should make the default for Swift 6 be true

FranzBusch: I would just change the default based on with what version the protoc plugin is compiled with. Since going forward we should have access level on imports by default

Skoti: > We've never had an option option to tell us what version of Swift we're generating for.

That's why I've just set it to `false` for now, it could be changed to `true` once Swift 6 is released.
But we can go with what Franz proposed:

> I would just change the default based on with what version the protoc plugin is compiled with. 

In that case if one still wants to generate without access imports they can explicitly set it through the option.
It's just that the default will automatically change which I think is convenient.

thomasvl: I think he's suggesting something like:

```swift
var accessLevelOnImports: Bool? = // .. fetch the option

#if swift(<=5.10)
accessLevelOnImport = accessLevelOnImports ?? false
#else
accessLevelOnImport = accessLevelOnImports ?? true
#endif
```

Note: that was typed directly into github so no claim I'm getting things right, but that should be the general idea.

Skoti: That's also how I understood it

Documentation/PLUGIN.md
```diff
@@ -182,6 +182,22 @@ exposed via public API, so even if `ImplementationOnlyImports` is set to `true`,
 this will only work if the `Visibility` is set to `internal`.
```
gjcairo, line 182: Can we please add a note to both READMEs explaining how `ImplementationOnlyImports` interacts with `UseAccessLevelOnImports` (i.e. both can't be set/`internal` visibility for imports should be preferred over `@_implementationOnly`)?

gjcairo (commented):
@Skoti thanks for this, looks great. Just left a small comment about adding some clarification around the options to the READMEs.

@thomasvl can you trigger CI on the PR? I don't seem to have the required privileges.

Documentation/PLUGIN.md
gjcairo: I don't have a strong opinion on this. However FWIW, Swift 6 language mode doesn't automatically default imports to internal. As the [SE proposal states](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md#proposed-solution):
>To balance source compatibility and best practices, an import without explicit access level has an implicit access level of public in Swift 5 and Swift 6. It will be internal in a future language mode.

Sources/protoc-gen-swift/GeneratorOptions.swift
```diff
@@ -140,13 +163,22 @@ class GeneratorOptions {
       visibilitySourceSnippet = "package "
     }
 
-    self.implementationOnlyImports = implementationOnlyImports
     self.experimentalStripNonfunctionalCodegen = experimentalStripNonfunctionalCodegen
 
+    self.importDirective = switch (implementationOnlyImports, useAccessLevelOnImports) {
```
thomasvl, line 168: CI caught that this is too new for some of the toolchains we support.  You'll have to write it slightly differently.

Skoti:
I've updated and applied all your remarks, please take a look.

> I think you might be able to make some examples in https://github.com/apple/swift-protobuf/tree/main/PluginExamples by explicitly enabling the features in the Package.swift` files to actually have checked usages of this.

I've also noticed that the `Simple` and `Nested` scenarios are the same, you may want to remove or alter one of them.

Skoti:
I've missed fixing some switch expressions to compile for Swift 5.8, it should be all now.

Skoti:
it now fails on
```
Run make test-spm-plugin PROTOC=../protobuf/cmake_build/protoc
env PROTOC_PATH=/__w/swift-protobuf/swift-protobuf/protobuf/cmake_build/protoc-29.0.0 swift test --package-path PluginExamples
error: 'pluginexamples': product 'SwiftProtobuf' required by package 'plugin examples' target 'Simple' not found in package 'swift-protobuf'.
make: *** [Makefile:201: test-spm-plugin] Error 1
Error: Process completed with exit code 2.
```

not sure why, the target is there 🤔

gjcairo:
@thomasvl can we try re-triggering the CI? This works locally for me.

thomasvl:
> @thomasvl can we try re-triggering the CI? This works locally for me.

Triggered a rerun of the failed one and it failed again.

PluginExamples/Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.6
+// swift-tools-version: 5.8
```
gjcairo, line 1: @Skoti can you please remove the `name` argument from the `swift-protobuf` package declaration when you list it as a dependency below? Like so:

```swift
let package = Package(
    name: "PluginExamples",
    dependencies: [
        .package(path: "../")
    ],
```

Looks like SPM in 5.8 is having trouble finding products when specifying a package name.
If you remove the `name` argument, it will default the package name to match the directory's name, which in this case is also `swift-protobuf` so it should all just work.
This should fix the broken CI.

gjcairo: @Skoti just a friendly ping here :) you think you can take a shot at this soon? You're very close to landing the PR and it would be nice to get this change done so it can be merged.

Skoti: @gjcairo done

Skoti: It didn't help :(
now all 3 fail even though it works for me locally for Swift 5.10 🤷‍♂️

gjcairo: Ah, didn't realise we are setting the path to be `/swift-protobuf/main` in the checkout GH action: https://github.com/apple/swift-protobuf/blob/main/.github/workflows/build.yml#L43-L46

Can we get rid of the `path:` there? It's unnecessary. The build should work without this.

gjcairo: Yes sorry I didn't mean to do that in this PR, I should have probably tagged you @thomasvl - it was a general question.
I think we should be okay with removing the path as it's not required and I can't see it conflicting with anything else.

I can open a PR with this change and @Skoti can rebase once that's landed and get this merged.

gjcairo: @thomasvl  https://github.com/apple/swift-protobuf/pull/1689

Skoti: The access-level examples need to be put into a separate job that won't run for Swift 5.8.

The current setup is that `ExampleTests` module depends on each test-case module, so it will try to compile `AccessLevelOnImport` module, and explicit access levels won't compile in Swift 5.8.

This seems like more work that should be part of a separate PR. Is it ok if we strip out plugin examples from this PR?

thomasvl: > This seems like more work that should be part of a separate PR. Is it ok if we strip out plugin examples from this PR?

Fine by me, this fear was what I tried to alluded to before when I wasn't sure if the examples could be used because of it having to work with older versions.

(a `#if swift` within the Package.swift for the examples _might_ work, but that's potentially getting into the _harder to maintain_ space)

gjcairo: Yeah I was just writing that you could add the new `AccessLevelOnImport` module as a target conditionally based on the compiler version (`#if compiler(>=5.9)`), **or** wrap all the source files in that module with that same compiler flag so they would be effectively empty in 5.8.
I don't think it's harder to maintain necessarily given we'll stop supporting 5.8 in a short few months when 6.0 is out anyways.
You can split this up in a separate PR if you prefer but I think doing that isn't much additional work if you just want to get it all done here.

Skoti: > Yeah I was just writing that you could add the new AccessLevelOnImport module as a target conditionally based on the compiler version (#if compiler(>=5.9)), or wrap all the source files in that module with that same compiler flag so they would be effectively empty in 5.8.

Now it should all work. Thanks!

tbkka (approved):

gjcairo (approved):

thomasvl:
Thank you!

PR #1686, opened 2024-07-29
Range check timestamps on all parsing paths.

thomasvl:
Include the fuzz test that caught this. The bug here is this was being accepted as valid input, when it couldn't be serialized back out into JSON, so this move the parsing test to make sure if fails no matter how it is formatted at parsing time.


tbkka (approved):
Nice catch!

thomasvl:
yea, I'm really glad I made the most of the fuzz testers serialize back *out* forcing no errors, cause that's what caught this.

PR #1690, opened 2024-08-12
Add a plugin api for extracting back out the descriptor proto.

### LINKED ISSUE #1678: SwiftProtobufPluginLibrary.FileDescriptor.proto is deprecated
The `proto` property on `SwiftProtobufPluginLibrary.FileDescriptor` is deprecated. In grpc-swift we currently use this property to generate data for the grpc [reflection service](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md) (which effectively serves serialized file descriptor protos).

Would you be open to either un-deprecating or providing new API to support this use case? I'd be more than happy with a `fileDescriptor.serializedFileDescriptorProto()` or something similar.

thomasvl:
Modeled after the C++ Copy*To() apis, this should provide the building blocks for what is needed now and might be needed in the future.

Fixes #1678


thomasvl:
@glbrntt fyi

allevato (approved):

PR #1693, opened 2024-08-19
Documentation for fieldmask utilities

pouyayarandi:
This PR adds documentation for field mask utilities added in #1505

### DIFF
--- a/Documentation/API.md
+++ b/Documentation/API.md
@@ -617,14 +617,40 @@ public func +(lhs: Google_Protobuf_Timestamp, rhs: Google_Protobuf_Duration) -> 
 
 ### Google_Protobuf_FieldMask
 
-The `Google_Protobuf_FieldMask` contains a list of field paths, where each path is a string that identifies a particular field or nested field in the message. This is particularly useful for selectively updating or trimming messages.
+`Google_Protobuf_FieldMask` is used to specify which fields in 
+a protocol buffer message should be included in operations such as updates or merges. 
+It allows precise control over which parts of the message 
+are affected by defining a list of field paths.
+
+For example, consider a protocol buffer message with nested fields:
 
 ```protobuf
-message FieldMask {
-  // The set of field mask paths.
-  repeated string paths = 1;
+message ParentMessage {
+  string name = 1;
+  ChildMessage child = 2;
+
+  message ChildMessage {
+    string childName = 1;
+    int32 age = 2;
+  }
 }
 ```
+
+If you want to update only the `name` field of `ParentMessage` 
+and the `childName` field within `ChildMessage`, 
+you would define a `FieldMask` as follows:
+
+```swift
+let fieldMask = Google_Protobuf_FieldMask.with { 
+  $0.paths = ["name", "child.childName"] 
+}
+```
+
+In this example, the `paths` list includes `"name"` to target 
+the `name` field in `ParentMessage` and `"child.childName"` 
+to target the `childName` field inside the nested `ChildMessage`. 
+This setup allows you to perform operations that affect 
+only these specified fields while leaving others unchanged.
 
 ## Extensions
 
@@ -710,7 +736,13 @@ concern because of Swift's common use for mobile applications.
 
 ### Merging Two Messages
 
-The `merge(from:fieldMask:)` function in Swift Protobuf enables developers to selectively merge fields from one message into another, guided by a `Google_Protobuf_FieldMask`. This method is particularly useful when you need to update only specific fields in a message without affecting others. The `merge` function is available as a method on `Message` types and requires two parameters: the source message (`from`) containing the data to merge and the `fieldMask` that specifies which fields should be updated.
+The `merge(from:fieldMask:)` function in Swift Protobuf enables developers 
+to selectively merge fields from one message into another, guided by a `Google_Protobuf_FieldMask`. 
+This method is particularly useful when you need to update only specific 
+fields in a message without affecting others. 
+The `merge` function is available as a method on `Message` types and requires two parameters: 
+the source message (`from`) containing the data to merge 
+and the `fieldMask` that specifies which fields should be updated.
 
 For example, consider a message with the following structure:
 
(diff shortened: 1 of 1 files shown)


Documentation/API.md
```diff
@@ -696,6 +732,102 @@ Descriptor objects. It is something that could get revisited in the future,
...
+This API is a powerful tool for managing partial updates in protocol buffer messages, 
+providing developers with precise control over the merging process.
+
+### Trimming a Message
+
+The `trim(keeping:)` function allows developers to retain only specific 
+fields in a protocol buffer message while clearing the rest. 
+This function is particularly useful when you want to ensure that only certain 
+fields are preserved in a message, effectively "trimming" the message 
+to contain just the necessary data.
```
tbkka, line 794: The second sentence here is redundant and can be dropped.

Documentation/API.md
```diff
@@ -696,6 +732,102 @@ Descriptor objects. It is something that could get revisited in the future,
...
+
+```swift
+message.trim(keeping: fieldMask)
+```
+
+After this operation, the `bar` field in `message` will still have the value `"bar"`, 
+while the `foo` field will be cleared, resetting to its default value (an empty string, 
+in this case). The `trim(keeping:)` function is performed in-place, meaning it directly 
+modifies the original message. 
+This function is ideal for scenarios where it's necessary to remove all but a few
```
tbkka, line 823: I think the rest of this paragraph can be dropped -- it all really just re-states what you've already said above.

Documentation/API.md
```diff
@@ -696,6 +732,102 @@ Descriptor objects. It is something that could get revisited in the future,
...
+  string foo = 1;
+  string bar = 2;
+}
+```
+
+Assume we have two instances of `ExampleMessage`:
+
+```swift
+let message1: ExampleMessage = .with { $0.foo = "foo1" }
+let message2: ExampleMessage = .with { $0.bar = "bar2" }
```
tbkka, line 760: I think this example would be better if `message2` also set `foo` to `"foo2"`.

Documentation/API.md
```diff
@@ -696,6 +732,102 @@ Descriptor objects. It is something that could get revisited in the future,
...
+```
+
+To merge `message2` into `message1` but only update the `bar` field, 
+you can define a `Google_Protobuf_FieldMask` like this:
```
tbkka, line 764: s/define/use/

Documentation/API.md
```diff
@@ -696,6 +732,102 @@ Descriptor objects. It is something that could get revisited in the future,
...
+
+```swift
+let fieldMask = Google_Protobuf_FieldMask.with { $0.paths = ["bar"] }
+```
+
+Then, you apply the merge:
```
tbkka, line 770: You can drop this and merge the sample lines above and below into one code block, so this part looks like:
```
... you can use a `Google_Protobuf_FieldMask` like this:

```swift
let fieldMask = Google_Protobuf_FieldMask.with { $0.paths = ["bar"] }
try message1.merge(from: message2, fieldMask: fieldMask)
``​`
```

Documentation/API.md
```diff
@@ -696,6 +732,102 @@ Descriptor objects. It is something that could get revisited in the future,
...
+
+```swift
+try message1.merge(from: message2, fieldMask: fieldMask)
+```
+
+After this operation, `message1.bar` will have the value `"bar2"` from `message2`, 
+while `message1.foo` remains `"foo1"`. The `merge` function operates in-place, 
+meaning it directly modifies `message1`. 
+This targeted approach is beneficial when handling partial updates,
```
tbkka, line 779: I would drop the rest of this paragraph, as it mostly just repeats what you've said above.

Documentation/API.md
```diff
@@ -696,6 +732,102 @@ Descriptor objects. It is something that could get revisited in the future,
...
+The `merge` function is available as a method on `Message` types and requires two parameters: 
+the source message (`from`) containing the data to merge 
+and the `fieldMask` that specifies which fields should be updated.
+
+For example, consider a message with the following structure:
+
+```protobuf
+message ExampleMessage {
+  string foo = 1;
+  string bar = 2;
```
tbkka, line 752: It would be good to illustrate how FieldMask works when you have nested messages.   I think maybe expand the examples:
```protobuf
message Example {
  string foo = 1;
  string bar = 2;
  Example nested = 3;
}
```

You can then add a short sentence below explaining that `"nested"` in the FieldMask will affect all fields in the nested message, and `"nested.foo"` will only affect the `foo` field of the `nested` message.

tbkka (approved):
This is pretty well written overall.  I've marked a few places where I think it could be shortened.  I've also noted one place where I think expanding an example would help make things clearer.

thomasvl:
fyi - once the docs are in, I'll look at doing a release to get this and the access levels work out.

pouyayarandi:
All comments were fixed @tbkka

Documentation/API.md
```diff
@@ -696,6 +732,106 @@ Descriptor objects. It is something that could get revisited in the future,
 but will need careful consideration; the bloat/size issues is of the most
 concern because of Swift's common use for mobile applications.
 
+## FieldMask Utilities
+
+### Merging Two Messages
+
+The `merge(from:fieldMask:)` function in Swift Protobuf enables developers 
+to selectively merge fields from one message into another, guided by a `Google_Protobuf_FieldMask`.
```
tbkka, line 740: Instead of "enables developers to selectively merge", you should just say "selectively merges".

Documentation/API.md
```diff
@@ -696,6 +732,106 @@ Descriptor objects. It is something that could get revisited in the future,
...
+try message1.merge(from: message2, fieldMask: fieldMask)
+```
+
+After this operation, `message1.bar` will have the value `"bar2"` from `message2`, 
+and `message1.nested.qux` will have the value `"qux2"` from `message2`, 
+while `message1.foo` and `message1.nested.baz` remain `"foo1"` and `"baz1"`. 
+Be aware that including `"nested"` in the FieldMask paths will cause all fields 
+within `message1.nested` to be updated from `message2` (including `baz` and `qux`), 
+whereas adding `"nested.qux"` only affects the `qux` field in the `nested` message. 
+The `merge` function operates in-place, meaning it directly modifies `message1`.
```
tbkka, line 800: Nice explanation here!

Documentation/API.md
```diff
@@ -696,6 +732,106 @@ Descriptor objects. It is something that could get revisited in the future,
...
+
+### Trimming a Message
+
+The `trim(keeping:)` function allows developers to retain only specific
```
tbkka, line 804: "The `trim(keeping:)` function retains only specific ..."

tbkka:
Thanks for all the hard work here.  I'm thrilled to get this out!

pouyayarandi:
It was my pleasure! I'm just as excited to see it out there!

PR #1694, opened 2024-08-23
Regenerate the generated_swift_names_*.protos

thomasvl:
This picks up everything from the new field mask helpers.


thomasvl:
The failure seems to be https://github.com/swiftlang/swift-docc-plugin/issues/92

PR #1696, opened 2024-08-23
Pin docc to a version that works for older swift versions.

thomasvl:


thomasvl:
Another option is that we just drop this dependency since I think Xcode will still be able to build docs without out.

tbkka (approved):

PR #1697, opened 2024-08-23
Version bump for a release.

thomasvl:


thomasvl:
aside: the one thing I'm unsure about is we'll capture the docc pin in this version, not sure if that will be an issue for downstream users that want 1.4.  Maybe wait until Monday to cut a release on the hope that docc will get fixed?

tbkka:
We should do this today regardless of docc, I think.  (We can bump a new version whenever docc does get fixed.)

tbkka (approved):

PR #1700, opened 2024-08-26
Move some other tests off descriptor.proto.

thomasvl:
These don't need to use descriptor, they just needed some object hierarchy and the ability to create extensions; so create custom types for this instead so only tests that have to be on descriptor.proto are.


tbkka (approved):
Nice!

allevato (approved):

PR #1701, opened 2024-08-28
Fix Foundation import's access level in generated code

gjcairo:
https://github.com/apple/swift-protobuf/pull/1683 added support for access level modifiers on imports in the generated code.

`Foundation` (which is needed because `Data` is used extensively in generated protos) is imported by manually writing to the file instead of using the "compute imports" logic we use to deal with all other imports. This is because other imports are all imports coming from other generated proto files, whereas Foundation is "pure Swift" dependency.

We forgot to add the access level modifier to this `Foundation` import.
This causes issues when enabling the `InternalImportsByDefault` flag on modules that use generated protos (even when `swift-protobuf`'s `UseAccessLevelOnImports` option is enabled), because `Foundation` will be imported as `internal`, while we expose public properties of type `Data` in the generated code.

This PR adds the right `import Foundation` statement based on the import directive.

### DIFF
--- a/Sources/protoc-gen-swift/FileGenerator.swift
+++ b/Sources/protoc-gen-swift/FileGenerator.swift
@@ -96,7 +96,7 @@ class FileGenerator {
           }
         }
 
-        p.print("\(comments)import Foundation")
+        p.print("\(comments)\(generatorOptions.importDirective.snippet) Foundation")
 
         if fileDescriptor.isBundledProto {
             p.print("// 'import \(namer.swiftProtobufModuleName)' suppressed, this proto file is meant to be bundled in the runtime.")


Sources/protoc-gen-swift/FileGenerator.swift
```diff
@@ -96,7 +96,7 @@ class FileGenerator {
           }
         }
 
-        p.print("\(comments)import Foundation")
+        p.print("\(comments)\(generatorOptions.importDirective.snippet) Foundation")
```
thomasvl, line 99: At first glance, does it make sense for this to end up as `public import` or `@_implementationOnly import`?  i.e. - I can see why it should get a qualifier, but should it be the same as all the proto modules, or should it be a fixed/specific value?

gjcairo: We do expose properties of type `Data` when the generated visibility is public, so in those scenarios we do need this to be `public import`. This is why I went for using the import directive (so that we don't import it as public if we are generating with a different visibility).

However, a downside of this approach is that marking an import as `public` will generate warnings if it's not actually used in public API in the file - and this _can_ happen in generated code, because we're **always** importing Foundation. 

Ideally, we'd have some logic to check whether we actually need `Foundation` or not in a given file, and decide whether to include the import at all or not. But this sounds like a bit of a pain to do (or like it would be pretty slow if e.g. we'd check each generated file at the end to see if we've used `Data` or something similar)

thomasvl: ack, tx.

I'd suggest opening an issue to have the generator conditional add that import (I guess we'd just need it for `bytes` fields on a message or extension)

gjcairo:
@thomasvl according to [the proposal](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md#interactions-with-other-modifiers-on-imports), `@_exported` imports **must** be followed by `public` if using access level modifiers:
>With this proposal, `@_exported` is accepted only on `public import` declarations, both with the modifier or the default `public` visibility in current language modes.

So I'd say this is the correct behaviour.

thomasvl (approved):

Sources/protoc-gen-swift/FileGenerator.swift
gjcairo: https://github.com/apple/swift-protobuf/issues/1702

PR #1703, opened 2024-08-28
Less generation when no messages, enums, or extensions.

thomasvl:
There's no need to generate any `import` directives or the version check when the file is completely empty, and with different access levels on imports that can get flagged with a newer compilers.

This is a common case when a .proto file contains a service.

Regenerate to get the one file update.


thomasvl:
@gjcairo fyi

allevato (approved):

PR #1704, opened 2024-08-28
Conditional foundation import

thomasvl:


thomasvl:
@gjcairo fyi

allevato (approved):

Sources/protoc-gen-swift/Descriptor+Extensions.swift
```diff
@@ -15,6 +15,22 @@ extension FileDescriptor {
     return SwiftProtobufInfo.isBundledProto(file: self)
   }
 
+  // Returns true if the file will beed to import Foundation.
```
gjcairo, line 18:
```suggestion
  // Returns true if the file will need to import Foundation.
```

Sources/protoc-gen-swift/Descriptor+Extensions.swift
```diff
@@ -166,6 +182,22 @@ extension Descriptor {
   /// Returns true if the message should use the message set wireformat.
   var useMessageSetWireFormat: Bool { return options.messageSetWireFormat }
 
+  var needsFoundationImport: Bool {
```
gjcairo, line 185: Nit: can we add the same docs from FileDescriptor above to this?

gjcairo (commented):
LGTM but two doc nits.

Thanks Thomas!

thomasvl:
Oops, window didn't reload to show the comment until after I hit merge.  Follow up PR coming.

PR #1707, opened 2024-08-29
Revise empty file support for `import public`

thomasvl:
If the file defines no types, but it does have `import public` directives, then it still needs provide those imports so code only depending on this module gets the types.


CompileTests/MultiModule/Sources/ReExportAOnly/reexport_a_only.pb.swift
```diff
@@ -0,0 +1,18 @@
...
+// swift-format-ignore-file
+// swiftlint:disable all
+//
+// Generated by the Swift generator plugin for the protocol buffer compiler.
+// Source: Sources/ReExportAOnly/reexport_a_only.proto
+//
+// For information on using the generated types, please see the documentation:
+//   https://github.com/apple/swift-protobuf/
+
+import SwiftProtobuf
```
thomasvl, line 11: @gjcairo @FranzBusch I'm pretty sure as current coded here this is wrong, but wanted to double check what should happen when we're generating access level imports.

For each of the cases, should anything be generated here?
- `public` - Should this be here as a `public import SwiftProtobuf`?
- `internal` - I was thinking maybe nothing, but would `internal import SwiftProtobuf` actually be needed so folks don't break if this proto file used to have a message and peer sources only had were using something that was a reexport and general SwiftProtobuf apis?
- `package` - Like the previous, what if another module in the package was importing this, module but the only message that wasn't a public import goes away, do we need to do something to ensure things are still exposed to those other modules importing this?

thomasvl: Talked this over with @allevato and the conclusion seems to be we never need the to `import SwiftProtobuf` if the file is only doing the re-export job.

allevato (approved):

CompileTests/MultiModule/Sources/ReExportAOnly/reexport_a_only.pb.swift
gjcairo: Yeah, I think with explicit access level modifiers on imports, the rule should be to only include what's necessary. This is because, if the visibility is set to `public` and we do `public import Foundation/SwiftProtobuf`, then the compiler will emit warnings if those imports are not used in the file. This is why I brought up only including `Foundation` when actually needed.
I agree with your conclusion and this LGTM.

gjcairo (approved):

PR #1709, opened 2024-09-03
Add CompileTests for InternalImportsByDefault

gjcairo:
This PR adds a new CompileTest for checking that generated code builds successfully when `InternalImportsByDefault` is enabled.


CompileTests/InternalImportsByDefault/Package.swift
```diff
@@ -22,7 +22,7 @@ let package = Package(
                 .product(name: "SwiftProtobuf", package: "swift-protobuf"),
             ],
             swiftSettings: [
-                .enableUpcomingFeature("InternalImportsByDefault"),
+                .enableExperimentalFeature("InternalImportsByDefault"),
```
thomasvl, line 25: I think you need to use `AccessLevelOnImport` and not use `InternalImportsByDefault`, the original issue with Foundation was because we didn't have an access level, so we don't want any level by default, but to ensure we're always providing level and to enable the support on these 5.x compilers.

gjcairo: Thanks, I forgot to enable `AccessLevelOnImport` too. If `InternalImportsByDefault` is not enabled, we still get some default: it defaults to `public` instead. I came across this bug when having both features enabled.

There were also some issues with 5.8 not understanding these flags so I had to add some compiler guards but now finally CI is happy.

PR #1718, opened 2024-09-20
Add Swift 6 toolchains to CI.

thomasvl:
This does not update the Package.swift to also support a full Swift 6 mode, that can be done in the future. This just ensure things build in the current setup with the new toolchain as expected.

This does not drop 5.8 yet, we'd said policy wise we were going to only support three at a time, not sure if nio has done it yet, maybe do it with them also?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,6 +25,8 @@ jobs:
         # seems better. This should also ensure protobuf caching changes with
         # each new image incase system in the Swift image are changed/updated.
         swift:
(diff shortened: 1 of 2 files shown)


tbkka (approved):

thomasvl:
Oops, sorry, missed that you'd asked @Lukasa to also review.  😦

Lukasa:
We are beginning to drop 5.8, but we're not in a hurry. We usually do a "one last release" on 5.8 and then drop it.

PR #1721, opened 2024-09-23
Some swift-format lint fixes

thomasvl:
Seems like swift-format failed to fix some `return` cases, but also fix some things that were lint only without formatting support.


thomasvl:
Opened https://github.com/swiftlang/swift-format/issues/812 for the _OmitExplicitReturns_ things that weren't autofixed.

allevato (approved):

PR #1724, opened 2024-09-24
Allow block comments for the table in the code.

thomasvl:
Switching to trailing comments doesn't read as well, and swift-format doesn't actually align all the line trailing comments, so the one line that is shorted would get it's comment indented compared to the rest so it doesn't read as well.


allevato (approved):

thomasvl:
Opened https://github.com/swiftlang/swift-format/issues/817 to maybe some day flip to line endings.

Issue #1730, opened 2024-11-09
Why is this a force unwrap ? My app just crashed 


adirburke:
https://github.com/apple/swift-protobuf/blob/4782e2f11d52e87cff85518126695898695712d8/Sources/SwiftProtobuf/Message%2BJSONAdditions.swift#L35

I just had a 
Swift runtime failure: Unexpectedly found nil while unwrapping an Optional value 

The function is throwing couldn't you throw a nil error instead of force-unwrap

adirburke:
Complete Stack Trace

zaraapp-1  | swift-runtime: failed to suspend thread 60890 while processing a crash; backtraces will be missing information
zaraapp-1  | swift-runtime: failed to suspend thread 60890 while processing a crash; backtraces will be missing information
zaraapp-1  | 
zaraapp-1  | *** Signal 4: Backtracing from 0x5c76c4306a32... done ***
zaraapp-1  | 
zaraapp-1  | *** Swift runtime failure: Unexpectedly found nil while unwrapping an Optional value ***
zaraapp-1  | 
zaraapp-1  | Thread 32 "NIO-ELT-0-#7" crashed:
zaraapp-1  | 
zaraapp-1  |  0 [inlined] [system] 0x00005c76c4306a32 Swift runtime failure: Unexpectedly found nil while unwrapping an Optional value in issueServer at //<compiler-generated>
zaraapp-1  |  1                    0x00005c76c4306a32 Message.jsonString(options:) + 386 in issueServer at /ZaraFull/.build/checkouts/swift-protobuf/Sources/SwiftProtobuf/Message+JSONAdditions.swift:35:52
zaraapp-1  |  2 [ra]               0x00005c76c4c13398 VaporHTTPAdapter.sendJson<A>(_:) + 535 in issueServer at /ZaraFull/Sources/zaraDatabaseWrapper/Sources/ZaraService/Controllers/VaporProtcols/VaporHTTPAdapter+HTTPResponse.swift:72:54
zaraapp-1  |  3 [ra] [thunk]       0x00005c76c4c14410 protocol witness for HTTPResponse.sendJson<A>(_:) in conformance VaporHTTPAdapter + 15 in issueServer at //<compiler-generated>
[...]

PR #1731, opened 2024-11-11
Swap String(bytes:encoding:)! for String(decoding:as:)

thomasvl:
Avoids the force unwraps that were needed before.

From the comment in https://github.com/apple/swift-protobuf/issues/1730#issuecomment-2466138248.

### DIFF
--- a/Sources/SwiftProtobuf/JSONEncoder.swift
+++ b/Sources/SwiftProtobuf/JSONEncoder.swift
@@ -73,7 +73,7 @@ internal struct JSONEncoder {
 
     internal var stringResult: String {
         get {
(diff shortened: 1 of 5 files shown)


thomasvl:
Will update once #1732 is merged, as that appears to be the problem.

thomasvl:
@Lukasa I moved over all the calls I could find per you comment on the other issue.  Let me know if you want a subset of these instead.

thomasvl:
@Lukasa did you want to do this suggestion from that issue, or should I drop this?

Lukasa (approved):
Apologies Thomas, juggling too many balls right now. Thanks for the ping, this looks great.

PR #1734, opened 2024-11-20
Strip file-level comment changes during editions codegen tests

mkruskal-google:
This was missed by some earlier changes, but follows the same pattern.  During some upgrades from proto2/proto3, comments may get reformatted and result in unexpected codegen diffs from non-functional code


thomasvl (approved):
Thanks!

PR #1740, opened 2025-01-06
Use Swift 6 mode on the plugin examples to ensure generated code compiles in this mode

thomasvl:


thomasvl:
Tried having only the 5.8 file that was like the old file, but 5.9 fails (https://github.com/apple/swift-protobuf/actions/runs/12640958090), I guess SwiftPM won't use a versioned file that is ≤ the current version still. 😦

FranzBusch (approved):

thomasvl:
> > Tried having only the 5.8 file that was like the old file, but 5.9 fails (https://github.com/apple/swift-protobuf/actions/runs/12640958090), I guess SwiftPM won't use a versioned file that is ≤ the current version still. 😦
> 
> Yes you always have to specify one per version.

I guess 5.10 does support falling back to older versions, I didn't have to provide a file specific to that version and things worked.

PR #1742, opened 2025-01-07
During JSON parsing, validate `@type` to be minimally valid.

thomasvl:
Upstream added a recent conformance test to ensure things fail for an empty `@type` or one that doesn't have atleast a single slash. This updates the library to do that validations during parsing from JSON *only*.

The change for this enforcement is pretty small (just the code in Sources/SwiftProtobuf/AnyMessageStorage.swift). The rest is to add a new error in the new SwiftProtobufError approach and then to update all of the apis to document that there are an additional type of error that could be throw.


tbkka (approved):
Looks great!

PR #1744, opened 2025-01-08
Build out the validation of google.protobuf.Any JSON support.

thomasvl:
The upstream major languages directly use their type registries to eagerly validate `type_url`s. This means they will fail to decode or encode JSON for a lot of cases.

The conformance tests only cover some of this so I hacked in some C++ unittests to see what errors are thrown and then made the matching cases here and expanded our error handling to duplicate the results.

The bulk of the changes are really just update comments to cover what errors are now throw (not actual code changes).

### DIFF
--- a/Tests/SwiftProtobufTests/Test_Any.swift
+++ b/Tests/SwiftProtobufTests/Test_Any.swift
@@ -873,38 +873,287 @@ final class Test_Any: XCTestCase {
         XCTAssertEqual(rejson, start)
     }
 
-    func test_Any_invalid() throws {
-        // These come from the upstream conformace tests.
+    func test_Any_typeURLValidations() throws {
+        // Upstream most langauges end up validating the type_url during encoding as well
+        // as during decoding. Basically to do things, they end up looking up the type
+        // in their global registries to use reflection to do things. SwiftProtobuf doesn't
+        // have a registry we can reply on being complete. So instead during encoding we
+        // do the most basic of checks. These tests help ensure those checks are working as
+        // expected.
+        //
+        // The inspiration for the tests and the errors comes from editing upstream's
+        // json_test.cc and observing a few things:
+        //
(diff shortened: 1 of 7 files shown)


Tests/SwiftProtobufTests/Test_Any.swift
```diff
@@ -873,38 +873,287 @@ final class Test_Any: XCTestCase {
         XCTAssertEqual(rejson, start)
     }
 
-    func test_Any_invalid() throws {
-        // These come from the upstream conformace tests.
+    func test_Any_typeURLValidations() throws {
+        // Upstream most langauges end up validating the type_url during encoding as well
+        // as during decoding. Basically to do things, they end up looking up the type
+        // in their global registries to use reflection to do things. SwiftProtobuf doesn't
+        // have a registry we can reply on being complete. So instead during encoding we
```
gjcairo, line 880:
```suggestion
        // have a registry we can rely on being complete. So instead during encoding we
```

Tests/SwiftProtobufTests/Test_Any.swift
```diff
@@ -873,38 +873,287 @@ final class Test_Any: XCTestCase {
...
+                        .JSONDecoding.invalidAnyTypeURL(type_url: "")
+                    )
+                )
+            }
         }
     }
 
     func test_Any_nestedList() throws {
-        var start = "{\"optionalAny\":{\"x\":"
+        var start = "{\"optionalAny\":{\"@type\":\"type.googleapis.com/Someting\",\"x\":"
```
gjcairo, line 1156:
```suggestion
        var start = "{\"optionalAny\":{\"@type\":\"type.googleapis.com/Something\",\"x\":"
```

gjcairo (approved):
LGTM, thanks Thomas

allevato (approved):

PR #1745, opened 2025-01-14
Add conversions between `Swift.Duration` and `Google_Protobuf_Duration`.

allevato:
I saw one of our teams recently add this extension to their code base and it makes sense to have it here.

Instead of just calling libc's `round` internally like some of the other precision-lowering conversions, I decided to make the `Duration`-to-protobuf conversion more flexible by taking a `FloatingPointRoundingRule` and having it default to standard "school room" rounding. It might be a good idea to update the other precision-lowering conversions to match (deprecating and/or defaulting the old versions as appropriate).

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Duration+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Duration+Extensions.swift
@@ -169,6 +169,46 @@ extension Google_Protobuf_Duration {
     }
 }
 
+@available(macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0, *)
+extension Google_Protobuf_Duration {
+    /// Creates a new `Google_Protobuf_Duration` by rounding a `Duration` to
+    /// the nearest nanosecond according to the given rounding rule.
+    ///
+    /// - Parameters:
+    ///   - duration: The `Duration`.
+    ///   - rule: The rounding rule to use.
+    public init(
+        rounding duration: Duration,
+        rule: FloatingPointRoundingRule = .toNearestOrAwayFromZero
+    ) {
+        let secs = duration.components.seconds
+        let attos = duration.components.attoseconds
+        let fracNanos =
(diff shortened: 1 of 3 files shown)


thomasvl (approved):

thomasvl:
> Love it. I have just written this code myself today!

@FranzBusch – What do you think to @allevato question in the initial PR, should we also do a follow up to provide `rounding` arguments to any others?

Sources/SwiftProtobuf/Google_Protobuf_Duration+Extensions.swift
```diff
@@ -169,6 +169,46 @@ extension Google_Protobuf_Duration {
...
+extension Google_Protobuf_Duration {
+    /// Creates a new `Google_Protobuf_Duration` by rounding a `Duration` to
+    /// the nearest nanosecond according to the given rounding rule.
+    ///
+    /// - Parameters:
+    ///   - duration: The `Duration`.
+    ///   - rule: The rounding rule to use.
+    public init(
+        rounding duration: Duration,
+        rule: FloatingPointRoundingRule = .toNearestOrAwayFromZero
```
tbkka, line 182: `.toNearestOrEven` should be the default.  (This is the default rounding rule for IEEE 754, for instance, and should be preferred in almost all cases.)

allevato: I chose this to align with the no-argument form of `BinaryFloatingPoint.round()` and `.rounded()` in the standard library: https://github.com/swiftlang/swift/blob/8dd7669816faf5228e2e4723988bef6f8b982662/stdlib/public/core/FloatingPoint.swift#L1659-L1667

Since this is the default for the standard library's behavior, I'd be worried about making this behave subtly differently. Do you still feel it should be changed?

tbkka: Leave it as-is then, I guess.

PR #1747, opened 2025-01-16
Add a initializer to Google_Protobuf_Duration with rounding control.

thomasvl:
Provide a new initializer with explicit rounding controls and map the other initializer though the single code paths.


thomasvl:
I have an option question about deprecating the existing apis in favor of this this new one.  Thoughts?

allevato:
IMO, I would deprecate `init(timeInterval: TimeInterval)` and point users to the new one with `renamed:`, since the argument label in the new one more clearly spells out the behavior of the conversion.

thomasvl:
> IMO, I would deprecate `init(timeInterval: TimeInterval)` and point users to the new one with `renamed:`, since the argument label in the new one more clearly spells out the behavior of the conversion.

Pushed a second commit with that, I can also drop it and force push again if we don't want to do it.

allevato (approved):

tbkka (approved):

PR #1748, opened 2025-01-23
Add a initializer to Google_Protobuf_Timestamp with rounding control

thomasvl:
Provide a new initializer with explicit rounding controls and map the other initializer though the single code paths.
   
Deprecate the old apis to direct folks to the new one to acknowledge the rounding that is happening on their values.

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Timestamp+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Timestamp+Extensions.swift
@@ -306,10 +306,10 @@ extension Google_Protobuf_Timestamp {
         //
         // Even though this goes through a rounding api, the api doesn't bother
         // to expose any controls on it. `Date` doesn't seem to document the
-        // internal percision and looking at current (Jan 2025) Foundation
-        // sources, the percisions for just getting the _current_ time seems to
-        // result in different percisions between platforms. So it doesn't seem
-        // like exposing the control for something that vaugue is worth while.
+        // internal precision and looking at current (Jan 2025) Foundation
+        // sources, the precision of the operations used just to get the
+        // _current_ time are platform-dependent. So it doesn't seem
+        // like exposing the control for something that vague is worthwhile.
         self.init(roundingTimeIntervalSinceReferenceDate: date.timeIntervalSinceReferenceDate)
     }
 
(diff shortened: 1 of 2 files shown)


Sources/SwiftProtobuf/Google_Protobuf_Timestamp+Extensions.swift
```diff
@@ -234,10 +234,26 @@ extension Google_Protobuf_Timestamp {
...
+
+    /// Creates a new `Google_Protobuf_Timestamp` initialized relative to 00:00:00
+    /// UTC on 1 January 1970 by a given number of seconds, rounded to the nearest
+    /// nanosecond according to the given rounding rule.
+    ///
+    /// - Parameters:
+    ///   - imeIntervalSince1970: The `TimeInterval`, interpreted as
+    ///     seconds relative to 00:00:00 UTC on 1 January 1970.
+    ///   - rule: The rounding rule to use.
+    public init(
```
thomasvl, line 250: @allevato and I were discussion options here, and this seems to follow things like `Date` in that it includes _time interval_ in the name.  Other things we considered:

```swift
init(
    roundingSince1970 timeInterval: TimeInterval,
    rule: FloatingPointRoundingRule
)
```

Which provides a slightly shorter name than what is suggested here.

```swift
init(
    timeIntervalSince1970: TimeInterval,
    roundingRule: FloatingPointRoundingRule
)
```

Would mean we don't have to deprecate the current method for the _default_ rounding, but the naming of the _rule_ argument is slightly different than foundation apis.

Sources/SwiftProtobuf/Google_Protobuf_Timestamp+Extensions.swift
```diff
@@ -269,9 +303,14 @@ extension Google_Protobuf_Timestamp {
...
-        self.init(
-            timeIntervalSinceReferenceDate: date.timeIntervalSinceReferenceDate
-        )
+        //
+        // Even though this goes through a rounding api, the api doesn't bother
+        // to expose any controls on it. `Date` doesn't seem to document the
+        // internal percision and looking at current (Jan 2025) Foundation
+        // sources, the percisions for just getting the _current_ time seems to
+        // result in different percisions between platforms. So it doesn't seem
+        // like exposing the control for something that vaugue is worth while.
```
allevato, line 312:
```suggestion
        // internal precision and looking at current (Jan 2025) Foundation
        // sources, the precision of the operations used just to get the
        // _current_ time are platform-dependent. So it doesn't seem
        // like exposing the control for something that vague is worthwhile.
```

allevato (approved):

Sources/SwiftProtobuf/Google_Protobuf_Timestamp+Extensions.swift
Lukasa: I do quite like the new name, honestly.

thomasvl: Ok, I guess we go with these.  I'll give the PR another day so folks approve or provide any other thoughts before merging it.

PR #1749, opened 2025-01-24
Require Semantic Version labels on PR

FranzBusch:
# Motivation

In most of our other Swift Server related repos we have introduced a new labelling system that allows us to use the GitHub auto-generated release notes. With those labels the release notes are properly grouped into their SemVer impact.

# Modification

This PR adds a new pipeline that makes sure each PR has the appropriate label set before it is merged.

# Result

No more PRs without SemVer classification.

### DIFF
--- a/.github/workflows/pull_request_label.yml
+++ b/.github/workflows/pull_request_label.yml
@@ -1,0 +1,18 @@
+name: PR label
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


thomasvl:
I guess it works since it forced this PR to need a label also.  😄

thomasvl:
Is there a way when preping a release to see the tagging on the PRs the commits came from to help figure out what sorta version bump is needed?

PR #1751, opened 2025-02-01
Cleaned up generated code for closed Enums

nicked:
For closed enums there no special `UNRECOGNIZED` case, so a plain Swift `Int` enum with explicit values can be used directly.

This results in the generated `init?(rawValue:)` and `var rawValue` functions not being needed, since the compiler will create them automatically.

### DIFF
--- a/CompileTests/MultiModule/Sources/ModuleA/a.pb.swift
+++ b/CompileTests/MultiModule/Sources/ModuleA/a.pb.swift
@@ -20,31 +20,13 @@ fileprivate struct _GeneratedWithProtocGenSwiftVersion: SwiftProtobuf.ProtobufAP
   typealias Version = _2
 }
 
-public enum E: SwiftProtobuf.Enum, Swift.CaseIterable {
-  public typealias RawValue = Int
-  case unset // = 0
-  case a // = 1
-  case b // = 2
+public enum E: Int, SwiftProtobuf.Enum, Swift.CaseIterable {
+  case unset = 0
+  case a = 1
+  case b = 2
 
   public init() {
     self = .unset
-  }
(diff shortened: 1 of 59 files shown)


thomasvl:
Huh, never really thought out this; since the _open_ enums needed it we didn't consider if the _closed_ could avoid the extra generation.

@tbkka any thoughts?

tbkka (approved):
Looks like a nice simplification.

thomasvl:
Looks like the PR needs the Reference directory updated also for the plugin tests.

nicked:
Hi, any idea why protobuf itself would [fail to build](https://github.com/apple/swift-protobuf/actions/runs/13088016173/job/36674764940?pr=1751)?

thomasvl:
> Hi, any idea why protobuf itself would [fail to build](https://github.com/apple/swift-protobuf/actions/runs/13088016173/job/36674764940?pr=1751)?

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90415
https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=14597b680a24b6f7375e4470dea935da9c369feb

Appears to be a compiler bug. I've let the protobuf team upstream know, hopefully they can tweak their sources to avoid this since it will be an issue for their customers on older (but supported) versions.

nicked:
Think I've got the workflow figured out now, thanks for your help!

thomasvl:
Looks like an upstream fix is coming: https://github.com/protocolbuffers/protobuf/pull/20234

thomasvl:
And with a respin everything is happy.

Thanks for noticing this, merging.

PR #1752, opened 2025-02-02
add visionos to podspec

grachyov:


SwiftProtobuf.podspec
```diff
@@ -12,6 +12,7 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = '10.13'
   s.tvos.deployment_target = '11.0'
   s.watchos.deployment_target = '4.0'
+  s.visionos.deployment_target = "1.0"
 
   s.cocoapods_version = '>= 1.7.0'
```
thomasvl, line 17: Does this need to also get updated so `visionos` will be known by cocoapods?  i.e. - what version did they add visionOS support?

grachyov: ok yes, it was `1.13.0` https://github.com/CocoaPods/CocoaPods/issues/11961

thomasvl: Hopefully that's not too new for folks.  Thanks!

tbkka (approved):

thomasvl:
> @FranzBusch I'm not sure how to read the semantic version label check failure here?

I think we just have to add the label.

thomasvl:
I'll try to look at the upstream protobuf build issues later, but I'm have a non-normal week this week, so not sure when I'll get to look.

thomasvl:
> I'll try to look at the upstream protobuf build issues later, but I'm have a non-normal week this week, so not sure when I'll get to look.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90415
https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=14597b680a24b6f7375e4470dea935da9c369feb

Appears to be a compiler bug.  I've let the protobuf team upstream know, hopefully they can tweak their sources to avoid this since it will be an issue for their customers on older (but supported) versions.

thomasvl:
Merging since this podspec won't have anything to do with the failures.

Thanks for the patch!

PR #1764, opened 2025-03-25
Deprecate FieldDescriptor.isOptional.

thomasvl:
This is the start of tracking the upstream changes to move away from Label and the "optional" bit in there. See
https://github.com/protocolbuffers/protobuf/pull/20687 for the start of it.

Update the comments on isRequired and isRepeated also.


thomasvl:
Using semver/minor since it is deprecating an api.

allevato (approved):

PR #1766, opened 2025-03-25
Deprecate `FieldDescriptor` `label`

thomasvl:
Add a second storage that can be used to avoid deprecation warnings from the code within the file that does have to touch/inspect the value.

Finishes up tracking changes in protocolbuffers/protobuf#20687.


thomasvl:
Using semver minor again since it deprecates an api.

allevato (approved):

PR #1775, opened 2025-05-12
Drop Swift 5.9

rnro:
### Motivation:

Swift 5.9 is no longer supported, we should bump the tools version

### Modifications:

* Bump the Swift tools version to Swift 5.10
* Remove unsupported versions from CI in build.yml

### Result:

Code reflects the expected support window.


thomasvl:
ps - if you want, split out the changes to generation and one of us can look at doing that if you wanted to avoid having to build protoc also.

thomasvl:
@FranzBusch which semver tag should removing old Swift versions get?

FranzBusch:
> @FranzBusch which semver tag should removing old Swift versions get?

Should be a SemVer minor

FuzzTesting/Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.10
```
thomasvl, line 1: I'm guessing because of this change you need to swap some things to `executableTarget`.  See the output of the failing CI actions for the warning that I'm basing this on.

@gjcairo for any other ideas for the failing fuzz tests.

thomasvl: Another random guess - What if you edit `FuzzTesting/do_build.sh` to remove the pass through of `-parse-as-library`?

rnro: Thanks for the suggestion. I’m going to look in the morning which Swift versions reproduce this.

rnro: I'm still trying things out but it looks like there are issues with fuzzing and SwiftPM
https://github.com/swiftlang/swift/issues/75800
https://github.com/swiftlang/swift-package-manager/issues/4522
https://github.com/swiftlang/swift-package-manager/issues/4769

rnro: Lowering the tools version back to 5.3 seems to have resolved the immediate issue.

rnro: @thomasvl Are you happy with this approach?

thomasvl:
LGTM, but I'd like one of the core Apple maintainers to sign off since this is officially raising the minimum.

gjcairo (approved):

thomasvl (approved):

Issue #1784, opened 2025-05-29
Move off `@testable` once Fuzz builds with current Xcodes


thomasvl:
I have a draft CL to remove the use of `@testable import` and instead ouse `package` visibility (since we've moved the baseline up to Swift 5.9), however because of https://github.com/swiftlang/swift-package-manager/issues/8731, the `Package.swift` for building the fuzz binary still requires using an older language version, so we likely can't land it yet. Once there is a Swift release with fuzzing fixed, we should be able to make the migration.

PR #1785, opened 2025-05-29
Move two of the tests off `@testable import`

thomasvl:
Now that Swift 5.9 is the baseline, make some things `package` visible and move off the `@testable` import.

Opened #1784 to track doing this with for the main runtime as they can't be done yet as the Fuzz tests still have to build with an older Swift version.


thomasvl:
@FranzBusch should this be semver/patch instead?

PR #1789, opened 2025-05-29
Create a bytecode interpreter to compactly represent text/JSON name maps.

allevato:
This is something I've been cooking for a bit, as a way to reduce codegen.

The problem with the `ExpressibleByDictionaryLiteral` approach we're using today is that its codegen is unpredictable. For what is really just static data, the Swift compiler always generates sequences of instructions in debug mode, and in release mode it may or may not be outlined into global data depending on various things outside of the author's control.

`StaticString`s, on the other hand, are super predictable since they always become static read-only data (ignoring the oddball edge case involving a single code point). So, why not do what some other languages do and create blobs of binary data that we can process at runtime instead?

The catch is that since `StaticString`s must be valid UTF-8, we have to invent our own little bytecode format a bit, to avoid encodings that might be invalid. The format is a stream of bytes as follows:

* Integers are represented as varints with 6-bit shifts and an always cleared MSB. This is similar to protobuf's own 7-bit varints except we lose a bit to ensure that continuation bytes are in the range 0x40–0x7F instead of 0x80–0xFF.
* Strings are represented as null-terminated UTF-8.
* String arrays are represented by a count integer followed by that many null-terminated UTF-8 strings.
* The stream always starts with an integer "version" number, which future-proofs us if we want to change the encoding. Right now, it's always zero.

Layered on top of the bytecode stream is an interpreter, which abstracts "instructions" that are `RawRepresentable` as `UInt64`, and the interpreter loop just reads the next instruction and invokes a callback that is supposed to read and process the operands.

The bytecode is an "ABI" that we can add to, but never remove from (for a particular major version of the runtime). For example, we could add new instructions in the future to represent things differently, as long as the runtime can still correctly interpret the bytecode from older generated code.

~~There are definitely more optimizations that we could do to the representation here to make it more compact. The most obvious thing would be to add specialized opcodes for the most common cases. Other things we could do would be to sort the field numbers in numerical order and represent them as adjacent differences instead of their absolute value—that would shrink things a bit for dense messages/enums that have numbers greater than 63.~~ I went ahead and did this.

~~But for now, I just wanted to put up this draft as a vibe check.~~ I took one of our particularly large transitive proto closures internally and measured the optimized and dead-stripped output with and without this change, using a tiny binary that just imports a huge proto module, sets a single field in a message, and then black-holes the message:

```
Before:  3995696 bytes
After:   3823880 bytes
Diff:     171816 bytes (–4.3%)
```

But I think this is just the beginning of what we can do—namely, we could refactor some of the generated message storage internals and make visitation completely generalized over a similar bytecode that we would generate about a message's layout. **That**, I imagine, would yield far greater codegen savings (and likely significant compile time savings as well).

---

Side note: Many of the APIs presented by the bytecode interpreter and the generated code just **scream** for `Span` and `UTF8Span`. I suppose one day we'll have a high enough deployment target that we could use those 😭


Sources/SwiftProtobuf/BytecodeInterpreter.swift
```diff
@@ -0,0 +1,61 @@
...
+/// Bytecode interpreters are generic over an `Instruction` type, which must be something (typically
+/// an `enum`) that has `UInt64` raw values. The only restriction on these values is that the raw
+/// value zero is reserved, so all instructions must have raw values of one or greater.
+///
+/// The current version of the bytecode stream (program format 0) is represented as a `StaticString`
+/// where non-textual information is binary-encoded in a way that is still guaranteed to be valid
+/// UTF-8. Specifically,
+///
+/// - Integers are encoded in a varint-like format similar to protobuf, except that only the low
+///   7 bits are used. The most-significant bit is always clear, and the second-most-significant
```
thomasvl, line 22: This is since the underlying storage is utf-8 and so the high bit would bring on other meanings?

When _SE-0483`InlineArray` Literal Syntax_ get approved, I'm guessing it would make sense to support/move to that as it means a full 8bit encoding so things could get smaller?  Any idea how easy this code could be leverage to support that eventually?

Sources/SwiftProtobuf/BytecodeInterpreter.swift
```diff
@@ -0,0 +1,61 @@
...
+    /// - Parameter handleInstruction: The function that will be invoked for each instruction that
+    ///   is read from the bytecode stream. The function takes two arguments: the `Instruction` that
+    ///   was read, and an `inout BytecodeReader` that the function should use to read operands and
+    ///   advance the stream.
+    package func execute(handleInstruction: (Instruction, inout BytecodeReader<Instruction>) -> Void) {
+        guard program.hasPointerRepresentation else {
+            // The only way this could happen is if the program were a single byte, meaning that it
+            // only has a 6-bits-or-fewer format specifier and nothing else. In other words, there
+            // are no instructions, and we can simply return as there is nothing to execute.
+            return
```
thomasvl, line 51: I still wonder if this should fatalError since it shouldn't happen and if something were generated with one character something went wrong?

Sources/SwiftProtobufPluginLibrary/BytecodeWriter.swift
```diff
@@ -0,0 +1,105 @@
...
+
+    /// The contents of the Swift string literal representing the written bytecode, without the
+    /// delimiting quotes.
+    private var code: String = ""
+
+    /// Creates a bytecode writer that uses the given program format.
+    ///
+    /// - Parameter programFormat: The program format of the bytecode stream to write. Currently,
+    ///   the only valid value is zero.
+    public init(programFormat: Int = 0) {
```
thomasvl, line 30: Should there be a constant over in the main runtime library for the zero value that is used by these places also?

Sources/SwiftProtobufPluginLibrary/BytecodeWriter.swift
```diff
@@ -0,0 +1,105 @@
...
+        if programFormat != 0 {
```
thomasvl, line 31: guard?

Sources/SwiftProtobuf/BytecodeReader.swift
```diff
@@ -0,0 +1,139 @@
...
+        var shift: UInt64 = 6
+        while true {
+            let byte = remainingProgram.first!
+            remainingProgram = remainingProgram.dropFirst()
+            value |= UInt64(byte & 0x3f) << shift
+            precondition(byte & 0x80 == 0, "Invalid integer leading byte \(byte)")
+            if byte & 0x40 == 0 {
+                return value
+            }
+            shift += 6
```
thomasvl, line 97: some guard that `shift` doesn't get too large?

Sources/SwiftProtobufPluginLibrary/BytecodeWriter.swift
thomasvl, line 1: Do we want to keep this in the protoc library until we're sure it make sense for some other plugin to also use this?

Sources/SwiftProtobuf/BytecodeReader.swift
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /// Reads and returns the next array of length-delimited strings from the bytecode stream.
+    ///
+    /// - Precondition: The reader must not be at the end of the bytecode stream.
+    ///
+    /// - Returns: An array of `UnsafeBufferPointer`s containing the strings that were read from the
+    ///   bytecode stream. See the documentation of `nextString()` for details on the lifetimes of
+    ///   these pointers.
+    package mutating func nextNullTerminatedStringArray() -> [UnsafeBufferPointer<UInt8>] {
```
thomasvl, line 127: Doesn't look like this got a test (or a helper to write one of these)

Sources/SwiftProtobuf/NameMap.swift
```diff
@@ -71,6 +71,57 @@ private class InternPool {
...
+    /// ## Operands
+    /// * The name of a reserved field.
+    case reservedName = 5
+
+    /// Represents a range of reserved field numbers in a proto message.
+    ///
+    /// ## Operands
+    /// * An integer representing the lower bound (inclusive) of the reserved field number range.
+    /// * An integer representing the upper bound (exclusive) of the reserved field number range.
+    case reservedFields = 6
```
thomasvl, line 122: "reservedNumber" nameMap isn't specific to fields, and enums also have reserved numbers, we just haven't needed to capture them.

Sources/protoc-gen-swift/FieldGenerator.swift
```diff
@@ -110,6 +113,50 @@ class FieldGeneratorBase {
         }
     }
 
+    func writeProtoNameInstruction(to writer: inout ProtoNameInstructionWriter) {
```
thomasvl, line 116: Shouldn't `var fieldMapNames` come out now?

Sources/protoc-gen-swift/FieldGenerator.swift
```diff
@@ -110,6 +113,50 @@ class FieldGeneratorBase {
...
+        // actually uses here.
+        let protoName: String
+        if fieldDescriptor.internal_isGroupLike {
+            protoName = fieldDescriptor.messageType!.name
+        } else {
+            protoName = fieldDescriptor.name
+        }
+        let jsonName = fieldDescriptor.jsonName
+
+        // TODO: Create a separate instruction for the group being able to match
```
thomasvl, line 130: Seems like we should do this extra opcode now vs. keeping it as a TODO

Sources/SwiftProtobufPluginLibrary/BytecodeWriter.swift
```diff
@@ -0,0 +1,105 @@
...
+
+    /// Writes an unsigned 64-bit integer to the bytecode stream.
+    public mutating func writeUInt64(_ value: UInt64) {
+        func append(_ value: UInt64) {
+            // Print the normal scalar if it's ASCII-printable so that we only use longer `\u{...}`
+            // sequences for those that are not.
+            if isprint(Int32(truncatingIfNeeded: value)) != 0 {
+                self.append(escapingIfNecessary: UnicodeScalar(UInt32(truncatingIfNeeded: value))!)
+            } else {
+                code.append(String(format: "\\u{%x}", value))
```
thomasvl, line 62: Is it worth picking off `\0` as the one special case?  (tab, newline, etc don't seem worth it, but if `\0` is common, it would slightly shorten strings and there's no code on the decoding side.

Sources/SwiftProtobuf/BytecodeInterpreter.swift
allevato: An empty message produces a single code-point name map, since it's just the version prefix and then nothing else: `"\u{0}"`

thomasvl: That should just be `_NameMap()` like it is now, no?

Sources/SwiftProtobuf/BytecodeInterpreter.swift
allevato: I haven't tested it with newer toolchains, but I suspect there would still be heavy type-checking costs to using `InlineArray` since it would have to be initialized with an array literal. Given this:

```
static let foo: InlineArray<someHugeNumber, UInt8> = [0, 1, 47, 8, ...]
```

At a _minimum_, even if it's able to store that as readonly data with minimal codegen under all circumstances, the type checker still has to verify that every element of that array literal is a `UInt8` and that it has exactly `someHugeNumber` elements. That's a lot of work that `StaticString` doesn't have, since that string literal is just an atomic blob.

What we really need is a way to encode raw bytes as a special kind of string-ish literal, which could supports `\xFF`-style encoding. Maybe it makes sense to propose something like that once semi-related features like `@section` land.

Sources/SwiftProtobuf/BytecodeInterpreter.swift
allevato: We could, but do we care enough about special casing that in the generator? The overhead of interpreting the empty name map bytecode would be low, especially for such an uncommon situation.

thomasvl: It is somewhat common for things that use extensions over fields as a design pattern (don't ask), so it does happen. The main "savings" from my POV is just less "startup code" that has to run, yes, it's is less, but zero is always less then "a little" 😄

Sources/SwiftProtobuf/BytecodeInterpreter.swift
allevato: As I suspected, creating an `InlineArray` of, say, 100 elements generates ~100 type variables and ~100 constraints in the constraint system. https://godbolt.org/z/9ahWK6oj1

None of those constraints are _hard to solve_—they're unambiguous since there aren't any overloads or casts/bridging involved—but the compiler still has to run through them all to prove that it's correct. I wouldn't feel comfortable putting that much load on the type checker when `StaticString` doesn't have any of that baggage.

But that's not all—in debug builds, the compiler *still* generates in non-optimized builds a scales-linearly-with-the-size-of-the-array "one-time initialization function" for the `InlineArray` for the property's `unsafeMutableAddressor`, even though it also generates it as static RO data. I don't really know what's going on there, but it's more unpredictability that we should avoid.

allevato: The extra codegen in debug builds seems silly and is probably an area with room for improvement, but I don't think there's a way around the type checker part since _something_ has to validate that the code is good.

Sources/SwiftProtobuf/BytecodeInterpreter.swift
allevato: I thought about this some more and I think we still need to support an empty (other than the version number) bytecode program. Even though we can simplify the `NameMap` case (and I've done so; now empty enums/messages generate a call the default initializer), I can see a future where we encode more descriptor-like information in these programs and I don't think we want to require every interpreter of those to handle the empty case separately.

Sources/SwiftProtobufPluginLibrary/BytecodeWriter.swift
allevato: Done.

Sources/SwiftProtobufPluginLibrary/BytecodeWriter.swift
allevato: Moved it to the `protoc-gen-swift` target.

Sources/SwiftProtobuf/BytecodeReader.swift
allevato: Added a couple tests.

Sources/SwiftProtobuf/NameMap.swift
allevato: Renamed to `reservedNumbers`.

Sources/protoc-gen-swift/FieldGenerator.swift
allevato: Done.

Sources/protoc-gen-swift/FieldGenerator.swift
allevato: Added a new opcode for groups.

Sources/SwiftProtobufPluginLibrary/BytecodeWriter.swift
allevato: Yeah, added `\0` as a special case now.

Sources/SwiftProtobuf/BytecodeReader.swift
allevato: Done.

Sources/SwiftProtobufPluginLibrary/BytecodeWriter.swift
allevato: I ended up defining this and removing the argument from `BytecodeWriter.init` entirely, because I can't think of a reason that we'd ever want the plugin to write a version of the bytecode stream that's _not_ the latest version. It's only the reader that needs to support multiple versions (unless we make an ABI break).

allevato:
I tweaked the `NameMap` bytecode a bit more. In order to keep field/case numbers as small as possible, I encode them as a delta from the previous number (they're encoded in numeric order); this frequently keeps the numbers within 6 bits. Additionally, I created special opcodes for the cases where the field/case is +1 from the previous one, since this will be the even more common case, and we don't encode a separate delta in that case.

@tbkka Do you have any thoughts on this? I'd love to land this infrastructure and start using it to experiment with packing descriptors that we can use for table-driven parsing/serializing.

CompileTests/MultiModule/Sources/ModuleA/a.pb.swift
```diff
@@ -103,18 +103,14 @@ public let Extensions_ext_str = SwiftProtobuf.MessageExtension<SwiftProtobuf.Opt
 // MARK: - Code below here is support for the SwiftProtobuf runtime.
 
 extension E: SwiftProtobuf._ProtoNameProviding {
-  public static let _protobuf_nameMap: SwiftProtobuf._NameMap = [
-    0: .same(proto: "E_UNSET"),
-    1: .same(proto: "E_A"),
-    2: .same(proto: "E_B"),
-  ]
+  private static let _protobuf_nameMap_bytecode: Swift.StaticString = "\0\u{2}\0E_UNSET\0\u{1}E_A\0\u{1}E_B\0"
```
tbkka, line 106: Seems this could be encoded as `\0\u{1}E_UNSET\0\u{1}E_A\0...`. That is, if the first field is zero, shouldn't we be able to use `sameNext` ?

tbkka (approved):
Looks promising.

tbkka:
CC: @Lukasa

CompileTests/MultiModule/Sources/ModuleA/a.pb.swift
allevato: The counter starts at 0, so using `sameNext` on the first value would pre-increment it and set `E_UNSET` to 1. So this works perfectly for field numbers that typically start at 1, but it's an awkward edge case for many enums.

We could resolve it by adding new `sameZero` and `aliasesZero` opcodes (since those are the only naming schemes used by enums). I didn't do it because it's additional runtime complexity for exactly one byte per enum, but I'm not opposed to making the change either.

tbkka: Ah.  Yes, that makes sense.  Good as-is then.

allevato:
Tagging this as `semver/minor` since it adds a new technically public `_NameMap.init(bytecode:)` API.

Sources/SwiftProtobuf/BytecodeInterpreter.swift
thomasvl: Does that mean we should still check the one byte to ensure it is still the right version number?

Sources/SwiftProtobuf/BytecodeReader.swift
```diff
@@ -0,0 +1,150 @@
...
+        let byte = UInt64(remainingProgram.first!)
+        remainingProgram = remainingProgram.dropFirst()
+        precondition(byte & 0x80 == 0, "Invalid integer leading byte \(byte)")
+
+        if byte & 0x40 == 0 {
+            return byte
+        }
+        var value: UInt64 = byte & 0x3f
+        var shift: UInt64 = 6
+        while true {
```
thomasvl, line 96: Total optimization question - should any/all of these use the arithmetic ops that don't check for overflow just to shrink/speed them up a hair?  The guard on the shift should take care of overflow, not sure if the compiler is smart enough to leave all the checks out as a result.

Sources/SwiftProtobuf/BytecodeReader.swift
```diff
@@ -0,0 +1,150 @@
...
+    /// This is provided as its own primitive operation because 32-bit values are extremely common
+    /// as field numbers (0 to 2^29-1) and enum cases (-2^31 to 2^31-1). In particular for enum
+    /// cases, using this function specifically for those cases avoids making mistakes involving
+    /// sign- vs. zero-extension between differently-sized integers.
+    ///
+    /// - Precondition: The reader must not be at the end of the bytecode stream.
+    ///
+    /// - Returns: The signed 32-bit integer that was read from the bytecode stream.
+    package mutating func nextInt32() -> Int32 {
+        // `Int32`s are stored by converting them bit-wise to a `UInt32` and then zero-extended to
```
thomasvl, line 68: Should this assert if the value was more than 32 bits?  I'm thinking of some safety net to ensure even negative numbers don't get expanded to64 bits.

Sources/protoc-gen-swift/BytecodeWriter.swift
```diff
@@ -0,0 +1,102 @@
...
+    /// Writes a signed 32-bit integer to the bytecode stream.
+    ///
+    /// This is provided as its own primitive operation because 32-bit values are extremely common
+    /// as field numbers (0 to 2^29-1) and enum cases (-2^31 to 2^31-1). In particular for enum
+    /// cases, using this function specifically for those cases avoids making mistakes involving
+    /// sign- vs. zero-extension between differently-sized integers.
+    package mutating func writeInt32(_ value: Int32) {
+        // `Int32`s are stored by converting them bit-wise to a `UInt32` and then zero-extended to
+        // `UInt64`, since this representation is smaller than sign-extending them to 64 bits.
+        writeUInt64(UInt64(UInt32(bitPattern: value)))
```
thomasvl, line 58: Same sorta question as on the parse, do we want any sanity check to ensure this never accidentally adds 32bits?

Sources/protoc-gen-swift/BytecodeWriter.swift
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /// Writes an unsigned 64-bit integer to the bytecode stream.
+    package mutating func writeUInt64(_ value: UInt64) {
+        func append(_ value: UInt64) {
+            // Print the normal scalar if it's ASCII-printable so that we only use longer `\u{...}`
+            // sequences for those that are not.
+            if value == 0 {
+                code.append("\\0")
+            } else if isprint(Int32(truncatingIfNeeded: value)) != 0 {
```
thomasvl, line 68: Doesn't this also need a check that the value is 32bit to begin with?  i.e. - if this got called with something over 32bits, but where the bottom 32bit happened to be pass `isprint`, the wrong thing could happen?

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -250,61 +250,25 @@ class MessageGenerator {
...
+        for f in fields {
+            f.writeProtoNameInstruction(to: &writer)
+        }
+        for name in descriptor.reservedNames {
+            writer.writeReservedName(name)
+        }
+        for range in descriptor.reservedRanges {
+            writer.writeReservedNumbers(range)
+        }
+        if fields.isEmpty && descriptor.reservedNames.isEmpty && descriptor.reservedRanges.isEmpty {
```
thomasvl: Rather than checking the fields and reserved bits, does it make sense to expose something on the writer to detect if it is empty except for the version marker? `writer.UseEmptyInitializer` or something like that?

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -250,61 +250,25 @@ class MessageGenerator {
...
-                p.withIndentation { p in
-                    for f in fields {
-                        for n in f.fieldMapNames {
-                            p.print("    \(f.number): \(n),")
-                        }
-                    }
-                }
-                p.print("])")
-            }
+            p.print(
```
thomasvl, line 266: `print` can take two args if you want to shorten this up or just use a multiline string to do both lines instead of two args.

Is there any code size tradeoffs from having two things here vs. just inlining the value directly to the initializer?

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -250,61 +250,25 @@ class MessageGenerator {
...
-                    for f in fields {
-                        for n in f.fieldMapNames {
-                            p.print("\(f.number): \(n),")
-                        }
-                    }
-                }
-                p.print("]")
-            }
+        var writer = ProtoNameInstructionWriter()
+        for f in fields {
```
thomasvl: This should probably be `fieldsSortedByNumber`, I'll bet we're getting some negative values with real world protos for the delta that pads things out to more bytes.

Sources/protoc-gen-swift/ProtoNameInstructionWriter.swift
```diff
@@ -0,0 +1,94 @@
...
+/// into bytecode streams.
+struct ProtoNameInstructionWriter {
+    private(set) var bytecode = BytecodeWriter<ProtoNameInstruction>()
+
+    /// The previous field or case number written to the stream, which is used to compute deltas.
+    private var previousNumber: Int32 = 0
+
+    mutating func writeSame(number: Int32, name: String) {
+        let delta = number - previousNumber
+        previousNumber = number
```
thomasvl: If you want a safety check for the sorta case I flagged, I believe with what we're doing now, we could put asserts in there to chat if the delta ever went negative.

Sources/SwiftProtobuf/BytecodeInterpreter.swift
allevato: Done.

Sources/SwiftProtobuf/BytecodeReader.swift
allevato: That's a good idea, done. (We should probably also follow up with the same in our varint encoding/decoding, if we have guards elsewhere.)

Sources/SwiftProtobuf/BytecodeReader.swift
allevato: Done.

Sources/protoc-gen-swift/BytecodeWriter.swift
allevato: I don't think that's needed here; since we receive an `Int32` and then bitwise convert it directly into a `UInt32`, getting something larger out of calling `UInt64(...)` on that would be a pretty big standard library bug.

Sources/protoc-gen-swift/BytecodeWriter.swift
allevato: The math below guarantees that the value received here will always be between 0 and 0x7f; I just use `UInt64` as the type here to avoid unnecessary casting/truncation at the call sites.

Sources/protoc-gen-swift/MessageGenerator.swift
allevato: Done.

Sources/protoc-gen-swift/MessageGenerator.swift
allevato: Done.

Sources/protoc-gen-swift/MessageGenerator.swift
allevato: Oops, I thought I had done that already... fixed.

Sources/protoc-gen-swift/ProtoNameInstructionWriter.swift
allevato: Done.

allevato:
As we talked about offline, I also added some tests as a safety check that the opcode values don't accidentally get changed in the future, and that those tests cover all the opcodes for a particular instruction set.

thomasvl (commented):
Looks good to me.

PR #1809, opened 2025-07-10
Use unchecked math operations.

thomasvl:
These are already checking for bad input, so no need for the compiler to include the overflow/underflow checks also.


thomasvl:
Since we did this in the NameMap change, figured it wouldn't hurt to do the same kind of change here.

allevato (approved):

PR #1816, opened 2025-07-21
Mark the legacy _NameMap initializer as deprecated.

thomasvl:
No one calls these directly, but the old generated code code. This will hopefully nudge folks to regenerate their sources to take advantage of the newer generation that comes from smaller codegen and has some better memory overhead during startup.

### DIFF
--- a/Sources/SwiftProtobuf/NameMap.swift
+++ b/Sources/SwiftProtobuf/NameMap.swift
@@ -304,6 +304,11 @@ public struct _NameMap: ExpressibleByDictionaryLiteral {
     #else  // !REMOVE_LEGACY_NAMEMAP_INITIALIZERS
 
     /// Build the bidirectional maps between numbers and proto/JSON names.
+    @available(
+        *,
+        deprecated,
+        message: "Please regenerate your .pb.swift files with the current version of the SwiftProtobuf protoc plugin."
+    )
     public init(
         reservedNames: [String],
         reservedRanges: [Range<Int32>],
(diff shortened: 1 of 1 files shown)


thomasvl:
Does this make sense to all?

FranzBusch (approved):
LGTM

freak4pc:
Sorry to bump this super old thread, we're actually dependent on an older library that uses the init(dictionaryLiteral:) variation (probably wrongfully). We'd like to fix it and contribute the fix back, but I'm not sure what's the "right" way, given NameMap doesn't seem to have any public initializers anymore? 

Would super appreciate guidance here :) 

https://github.com/centrifugal/centrifuge-swift/blob/91a5c356840ecf0dbc949b0d5599ca39f208b4d6/Sources/SwiftCentrifuge/client.pb.swift#L1028-L1032

Thanks!

allevato:
Those are generated messages, so presumably that project just needs to bump its version of the swift-protobuf dependency and regenerate them?

thomasvl:
Also, the old method are deprecated, but I believe they are [still there](https://github.com/apple/swift-protobuf/blob/main/Sources/SwiftProtobuf/NameMap.swift#L329), just deprecated.

freak4pc:
Right, my bad, wasn't too familiar with swift-protobuf before today ^^ 

Pushed a PR for them here: 

https://github.com/centrifugal/centrifuge-swift/pull/123

Thanks both!

PR #1820, opened 2025-07-31
Remove `import Swift`, doesn't seem to be needed.

thomasvl:


thomasvl:
@Lukasa @FranzBusch @tbkka Any ideas why we had this `import`?  It traces back to when the file first landed (#178), but doesn't say what it was needed.

tbkka:
I don't think it's needed, probably never was.  There was a convention early in Swift of including either `import Swift` or `import Foundation` in every Swift source file.

Issue #1821, opened 2025-08-05
Failing new upstream conformance test for Duration and Timestamp fail


thomasvl:
Upstream added some new conformance tests that currently fail:
https://github.com/protocolbuffers/protobuf/commit/a6bdd0a7b23d5d6e141f3592c13062e3785fc2fa#diff-4c6ba3d87bce4389f7883a1cffd8ec741f76a4ec05bba4a38a5018a8484cbad9

thomasvl:
Fixed that's to those two PRs.

PR #1823, opened 2025-08-06
Validate Duration before generating JSON to match upstream.

thomasvl:
- Add some tests.
- Document a few places where/why things get normalized.
- Update some tests specific to normalization that don't rely on constructing something that isn't otherwise "right" (json serialization would fail).

Progress on #1821.

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Duration+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Duration+Extensions.swift
@@ -18,6 +18,8 @@ import Foundation
 
 private let minDurationSeconds: Int64 = -maxDurationSeconds
 private let maxDurationSeconds: Int64 = 315_576_000_000
+private let minDurationNanos: Int32 = -maxDurationNanos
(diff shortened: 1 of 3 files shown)


thomasvl:
Calling this _semver minor_ even though there is no new public api because the behavior of the json support is changing (to match the spec).

Sources/SwiftProtobuf/Google_Protobuf_Duration+Extensions.swift
```diff
@@ -18,6 +18,8 @@ import Foundation
 
 private let minDurationSeconds: Int64 = -maxDurationSeconds
 private let maxDurationSeconds: Int64 = 315_576_000_000
+private let minDurationNanos: Int32 = -999_999_999
```
allevato: Nit: We could write this as `-maxDurationNanos` like the one above to reduce some repetition.

thomasvl: Done.

PR #1825, opened 2025-08-07
Support "edition2024".

thomasvl:
- Regenerate the feature set tables.
- Have the generator report the support.
- Tell the conformance runner to include 2024 checks.


thomasvl:
Using semver minor as the generator will be reporting a new edition as supported.

`edition2024` is enabled on upstream main, and they've started moving testing protos over to it. So the edition will likely be line in their next release.

allevato (approved):

PR #1835, opened 2025-08-12
Add automated protoc artifact bundle generation for releases

FranzBusch:
## Motivation

Currently the build plugin distributed from swift-protobuf and other repos that offer protoc plugins such as grpc-swift, are hardly usable since they require the user to install protoc and then point SwiftPM or the IDE to its location.

## Modifications

This PR adds a new step in our release flow that bundles the upstream protoc binaries into an artifactbundle and uploads them to the latest draft release. It also adds to the release steps to update the `Package.swift`

## Result

We now have an ecosystem wide distributing of protoc that can be used directly from the package manager without requiring users to install it manually. This makes build and command plugins that rely on protoc vastly more useful.

### DIFF
--- a/.github/scripts/upload_artifactbundle.sh
+++ b/.github/scripts/upload_artifactbundle.sh
@@ -1,0 +1,104 @@
+#!/bin/bash
+
+# This script generates an artifactbundle for protoc. This artifactbundle
+# is used by the Swift package manger. The script is run by a GitHub action
+# when a new pre-release is created for swift-protobuf.
+
+set -eux
+
+# Fetch the latest stable release from protocolbuffers/protobuf
+AUTH="Authorization: token $GITHUB_TOKEN"
+response=$(curl -sH "$AUTH" "https://api.github.com/repos/protocolbuffers/protobuf/releases/latest")
+TAG=$(echo "$response" | grep -m 1 '"tag_name":' | cut -d '"' -f 4)
+
+# Remove 'v' prefix if present
+TAG="${TAG#v}"
+
+if [[ ! "$TAG" =~ ^[0-9]+\.[0-9]+$ ]]; then
+    echo "Error: $TAG does not match the expected pattern"
+    exit 1
+fi
+
+# Fetch all protoc release assets from protocolbuffers/protobuf
+curl -LJ --output protoc-$TAG-osx-x86_64.zip -H 'Accept: application/octet-stream' https://github.com/protocolbuffers/protobuf/releases/download/v$TAG/protoc-$TAG-osx-x86_64.zip
+curl -LJ --output protoc-$TAG-osx-aarch_64.zip -H 'Accept: application/octet-stream' https://github.com/protocolbuffers/protobuf/releases/download/v$TAG/protoc-$TAG-osx-aarch_64.zip
+curl -LJ --output protoc-$TAG-linux-aarch_64.zip -H 'Accept: application/octet-stream' https://github.com/protocolbuffers/protobuf/releases/download/v$TAG/protoc-$TAG-linux-aarch_64.zip
+curl -LJ --output protoc-$TAG-linux-x86_64.zip -H 'Accept: application/octet-stream' https://github.com/protocolbuffers/protobuf/releases/download/v$TAG/protoc-$TAG-linux-x86_64.zip
+curl -LJ --output protoc-$TAG-win64.zip -H 'Accept: application/octet-stream' https://github.com/protocolbuffers/protobuf/releases/download/v$TAG/protoc-$TAG-win64.zip
+
+# Fetch and validate license from protocolbuffers/protobuf
+curl -LJ --output LICENSE -H 'Accept: application/vnd.github.v3.raw' https://api.github.com/repos/protocolbuffers/protobuf/contents/LICENSE
+LICENSE_HASH=$(sha256sum LICENSE | cut -d ' ' -f 1)
+EXPECTED_HASH="6e5e117324afd944dcf67f36cf329843bc1a92229a8cd9bb573d7a83130fea7d"
+
+if [ "$LICENSE_HASH" != "$EXPECTED_HASH" ]; then
+    echo "Error: License file has changed. Expected hash: $EXPECTED_HASH, Got: $LICENSE_HASH"
+    exit 1
+fi
+
+# Unzip all assets
+mkdir protoc-$TAG.artifactbundle
+unzip -d protoc-$TAG.artifactbundle/protoc-$TAG-osx-x86_64 protoc-$TAG-osx-x86_64.zip
+unzip -d protoc-$TAG.artifactbundle/protoc-$TAG-osx-aarch_64 protoc-$TAG-osx-aarch_64.zip
+unzip -d protoc-$TAG.artifactbundle/protoc-$TAG-linux-aarch_64 protoc-$TAG-linux-aarch_64.zip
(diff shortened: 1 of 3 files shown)


Documentation/RELEASING.md
```diff
@@ -18,7 +18,7 @@ When doing a release:
    and include a reference '#XYZ', this tells you what pull request it was part of.  This is useful
    for doing the release notes in the next section.
 
-1. Update the version on _main_
+2. Update the version on _main_
```
thomasvl: fyi: when [rendered](https://github.com/apple/swift-protobuf/blob/main/Documentation/RELEASING.md), markdown gets the numbers right, use "1" means reorders/additions never have to update numbers.

I'm fine changing them, I know some repos prefer they be numbered right.

Documentation/RELEASING.md
```diff
@@ -48,9 +48,45 @@ When doing a release:
    everything based on the PR descriptions and _semver_ tags in the repo. Just read
    though was was generate to see if any tweaks are needed.
 
-   When everything is good, click on _Publish release_.
+   **Important:** Save this as a **draft** release first (do not publish yet).
+
+4. Generate protoc artifact bundle
+
+   Once you have created the draft release, trigger the "Upload protoc artifactbundle"
```
thomasvl: Can any person with commit access to the repo do this?

What if we have two SwiftProtobuf releases without a new protobuf release happening, do we skip this step?

Documentation/RELEASING.md
```diff
@@ -48,9 +48,45 @@ When doing a release:
...
+   workflow from the [Actions tab](https://github.com/apple/swift-protobuf/actions/workflows/prerelease_protoc_artifactbundle.yml).
+   
+   This workflow will:
+   - Fetch the latest stable protoc release from protocolbuffers/protobuf
+   - Create a Swift Package Manager compatible artifact bundle
+   - Upload it to your draft release
+   
+   Wait for the workflow to complete successfully before proceeding.
+
+5. Update Package.swift with new artifact bundle
```
thomasvl: This is an update on `main`?  Does it then need to get picked into the already tagged release in some way?

.github/workflows/prerelease_protoc_artifactbundle.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  upload-artifactbundle:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@v4
+      - name: Upload artifactbundle
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
thomasvl, line 14: Is this something on the repo level that used, or will the person running the action need to do something for this to work?

.github/scripts/upload_artifactbundle.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
+# This script generates an artifactbundle for protoc. This artifactbundle
+# is used by the Swift package manger. The script is run by a GitHub action
+# when a new pre-release is created for swift-protobuf.
+
+set -ex
+
+# Fetch the latest stable release from protocolbuffers/protobuf
```
thomasvl, line 9: Scrolling through https://github.com/protocolbuffers/protobuf/releases, I currently see non pre-releases (in order):

- v31.1
- v29.5
- v25.8
- v31.0

So I guess they had some old releases to fix things, I'm guessing that means depending on when this runs, we could catch the not really _latest_.  Is there something that we should document doing if the _literal latest_ isn't the _logical latest_?

.github/scripts/upload_artifactbundle.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
+# This script generates an artifactbundle for protoc. This artifactbundle
+# is used by the Swift package manger. The script is run by a GitHub action
+# when a new pre-release is created for swift-protobuf.
+
+set -ex
```
thomasvl: `-eux` ?  i.e. - fail for unknown variable references?

Documentation/RELEASING.md
```diff
@@ -48,9 +48,45 @@ When doing a release:
...
+
+   Create a pull request that updates the `Package.swift` file to reference the new 
+   artifact bundle. You'll need to update two things:
+   
+   - **URL**: Change to point to your new release tag
+   - **Checksum**: Download the artifact bundle and calculate its SHA256 hash
+   
+   Example update:
+   ```swift
+   .binaryTarget(
```
thomasvl, line 79: I guess this will exist in the Package.swift once we do our first release with this in place, but for now it doesn't yet exist?

Once we have this, could we then use it for all our Test targets and not check in the generated files for development?  We'd still need them in the runtime and plugin library, but would reduce the need for the tests?  (and I guess Reference/ still needs them all)  And the plugin/compile tests that use the plugin could also make use of it?

Documentation/RELEASING.md
FranzBusch: I didn't know that. Thanks!

Documentation/RELEASING.md
FranzBusch: > Can any person with commit access to the repo do this?
Yes 

> What if we have two SwiftProtobuf releases without a new protobuf release happening, do we skip this step?
Good point. I added a section to check if there is a new release and only then run this.

Documentation/RELEASING.md
FranzBusch: This happens automatically with a draft release since that points to the latest `main` and hasn't tagged yet

.github/workflows/prerelease_protoc_artifactbundle.yml
FranzBusch: Yes this is Github provided by default so it will just work

.github/scripts/upload_artifactbundle.sh
FranzBusch: This shouldn't be a problem since from what I can tell the Google folks will always tag the latest release with the "latest" GitHub which we are using to fetch below. You can get to this yourself by going to https://github.com/protocolbuffers/protobuf/releases/latest which as of right now points to `v31.1`

Documentation/RELEASING.md
FranzBusch: Yes correct. I have some of those changes ready for a subsequent PR :)

.github/scripts/upload_artifactbundle.sh
thomasvl: I was thinking "latest" was auto managed by github.  Guess we'll see if something does cause an issue.

Documentation/RELEASING.md
```diff
@@ -48,7 +48,47 @@ When doing a release:
...
+   Wait for the workflow to complete successfully before proceeding.
+
+1. Update Package.swift with new artifact bundle
+
+   If there was a new protoc release and you uploaded a new artifact bundle in
+   the previous step. Create a pull request that updates the `Package.swift` file
+   to reference the new artifact bundle. You'll need to update two things:
+   
+   - **URL**: Change to point to your new release tag
+   - **Checksum**: Download the artifact bundle and calculate its SHA256 hash
```
thomasvl, line 75: Whats the right command line to do that?

FranzBusch: You can do it via the command line by doing `swift package compute-checksum <path>` but the GH UI will also show it to you

PR #1836, opened 2025-08-13
Add `protoc` product

FranzBusch:
## Motivation

This follows my previous PR in adding support to distribute the `protoc` from this package.

## Modifications

Adds a new binary target and product for `protoc`. Adapts our PluginExample tests to use it instead of the CI build one

## Result

Ecosystem wide solution for `protoc`


thomasvl:
So 5.10 wasn't able to do the binary artifact?

FranzBusch:
> So 5.10 wasn't able to do the binary artifact?

Yeah it sadly was not. We landed support for products backed by binary targets in 6.0 it appears

PR #1837, opened 2025-08-14
Create separate protoc releases

FranzBusch:
# Motivation

In a previous PR, I added a Github workflow and scripts to attach a protoc artifactbundle to the latest draft release. This created a chicken and egg problem since we needed to reference the protoc URL in our Package.swift but the URL wasn't public until we published a release.

# Modifications

This PR changes the workflow and script so that we create separate protoc-artifactbundle releases with a separate tagging scheme that won't conflict with our normal tags. These releases are only used to distribute protoc artifactbundles.

# Result

We can now generate the protoc artifactbundle releases that we can reference in our Package.swift without creating this chicken and egg problem.


.github/scripts/draft_release_protoc_artifactbundle.sh
```diff
@@ -85,20 +109,34 @@ EOF
...
-    echo "Error: No draft release found"
+# Create a new draft release for protoc-vXXX
+echo "Creating draft release protoc-v$TAG"
+create_response=$(curl -sH "$AUTH" -X POST "$GITHUB_API_URL/repos/$GITHUB_REPOSITORY/releases" \
+  -d "{
+    \"tag_name\": \"protoc-artifactbundle-v$TAG\",
+    \"name\": \"protoc v$TAG artifactbundle\",
+    \"body\": \"Protoc artifactbundle for version $TAG\",
+    \"draft\": true,
+    \"prerelease\": false
```
glbrntt: We should make sure this isn't set as the latest release. Looks like you need to set `make_latest` to `false` (https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release)

PR #1841, opened 2025-08-15
Deprecate another helper.

thomasvl:
Like the others, these shouldn't be needed if using CodeGenerator.


thomasvl:
Looking at github search results, seems like everything is off these old helpers, so just deprecating the last of them.

PR #1843, opened 2025-08-18
Prepare for a 1.31.0 release and update docs

FranzBusch:
# Motivations

In the previous PRs, I landed support for distributing `protoc` via this repo which finally allows us to broadly recommend the SPM plugin.

# Modifications

This PR bumps the versions and updates the docs for the SPM plugin and a tiny addition to the RELEASING doc to ensure the latest check is set.

# Result

Unblocking the ecosystem in broadly adopting the SPM plugin.

### DIFF
--- a/Sources/protoc-gen-swift/Docs.docc/spm-plugin.md
+++ b/Sources/protoc-gen-swift/Docs.docc/spm-plugin.md
@@ -105,20 +105,16 @@ is generating Swift types for the `Bar.proto` file with the `public` visibility.
 invocation is using the `pathToUnderscores` file naming option. This option can be used to solve
 problems where a single target contains two or more proto files with the same name.
 
-### Defining the path to the protoc binary
+### Overriding the path to the protoc binary
 
-The plugin needs to be able to invoke the `protoc` binary to generate the Swift types. There are several ways to achieve this.
+The plugin needs to be able to invoke the `protoc` binary to generate the Swift types.
+By default the plugin will use the latest `protoc` binary distributed as an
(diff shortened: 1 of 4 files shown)


Sources/protoc-gen-swift/Docs.docc/spm-plugin.md
```diff
@@ -105,20 +105,16 @@ is generating Swift types for the `Bar.proto` file with the `public` visibility.
...
-The plugin needs to be able to invoke the `protoc` binary to generate the Swift types. There are several ways to achieve this.
+The plugin needs to be able to invoke the `protoc` binary to generate the Swift types.
+By default the plugin will use the latest `protoc` binary distributed as an
+artifact bundle from the [swift-protobuf releases](https://github.com/apple/swift-protobuf/releases).
 
-First, by default, the package manager looks into the `$PATH` to find binaries named `protoc`.
-This works immediately if you use `swift build` to build your package and `protoc` is installed
-in the `$PATH` (`brew` is adding it to your `$PATH` automatically).
-However, this doesn't work if you want to compile from Xcode since Xcode is not passed the `$PATH`.
+There are several ways for you to override the path to the `protoc` binary.
```
thomasvl, line 114: Just checking - will a protoc on $PATH _always_ override the one pointed too by our Package.swift?

thomasvl: i.e. - should we document the order things are done in the code so it is less likely to be confusing what "wins"

FranzBusch: No $PATH will no longer be used at all since we supply protoc now and that's the fallback if none of the others are specified

PR #1846, opened 2025-08-25
Changes to embed source code

amralaa-MSFT:


allevato:
These declarations are `package` level by intent. What is your goal with this change?

amralaa-MSFT:
@allevato Sorry, the pull request wasn't intended for public use.

PR #1849, opened 2025-09-08
Move from 6.1.2 to 6.1.3 on CI.

thomasvl:


thomasvl:
@FranzBusch Looks like you might have required all the current checks for PRs as the 6.1.2 is listed even though it doesn't exist any more with this change. Is there a way to change the configuration on github so it won't fail the checks like this?

thomasvl:
We could delete the patch version from our numbers so we just list `major.minor` and the links will lets things float, but if something ever breaks on a patch version, we won't really see it without digging into log and by looking to see if a new patch rolled out since the last time out CL ran.

We'd also still have this sorta problem of the github config needing to be be updated when we add/remove versions (to require the new one or drop the old one).

thomasvl:
> @FranzBusch Looks like you might have required all the current checks for PRs as the 6.1.2 is listed even though it doesn't exist any more with this change. Is there a way to change the configuration on github so it won't fail the checks like this?

@Lukasa @tbkka incase either of you can tweak the config on github for this.

Lukasa:
Amended, you're good to go.

thomasvl:
> Amended, you're good to go.

Thanks, still need one of you to review. :)

For future reference, is there a way to not have to do this tweaking with each release?  Do we want to switch to `major.minor` and float like I mentioned?

Lukasa:
That's what we do on swift-nio and friends, we just float the minor version.

Lukasa (approved):

PR #1850, opened 2025-09-09
Let users opt of fetching protoc

glbrntt:
Motivation:

In some build environments fetching the protoc binary artifact doesn't work as expected (e.g. using '--build-system xcode'). This isn't an issue with SwiftProtobuf but this workardound unblocks users broken by the inclusion of protoc.

Modifications:

- Allow users to specify 'PROTOBUF_NO_PROTOC=true' or 'PROTOBUF_NO_PROTOC=1' to remove protoc from the package manifest.

Result:

Unblocks broken users

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -18,10 +18,6 @@ let package = Package(
             name: "protoc-gen-swift",
             targets: ["protoc-gen-swift"]
         ),
-        .executable(
-            name: "protoc",
-            targets: ["protoc"]
-        ),
         .library(
             name: "SwiftProtobuf",
             targets: ["SwiftProtobuf"]
@@ -54,12 +50,6 @@ let package = Package(
             dependencies: ["SwiftProtobuf"],
             swiftSettings: .packageSettings
         ),
-        .binaryTarget(
-            name: "protoc",
-            url:
-                "https://github.com/apple/swift-protobuf/releases/download/protoc-artifactbundle-v31.1/protoc-31.1.artifactbundle.zip",
-            checksum: "f18bf2cfbbebd83133a4c29733b01871e3afdc73c102d62949a841d4f9bab86f"
-        ),
         .executableTarget(
             name: "protoc-gen-swift",
             dependencies: ["SwiftProtobufPluginLibrary", "SwiftProtobuf"],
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
I'm happy with this, but I'd like @FranzBusch to review for sure, and ideally one of @allevato or @thomasvl.

Package.swift
```diff
@@ -96,6 +86,37 @@ let package = Package(
...
 )
 
+// Including protoc as a binary artifact can cause build issues in some environments. As a
+// temporary measure offer an opt-out by setting PROTOBUF_NO_PROTOC=true.
+let includeProtoc: Bool
+if let noProtoc = Context.environment["PROTOBUF_NO_PROTOC"] {
+    includeProtoc = !(noProtoc.lowercased() == "true" || noProtoc == "1")
+} else {
+    includeProtoc = true
+}
```
thomasvl: nit: this block might be more discoverable by folks if it was at the top of the file

thomasvl:
> In some build environments fetching the protoc binary artifact doesn't work as expected (e.g. using '--build-system xcode').

Does that mean the plugin support doesn't work for Xcode users?  That seems like it might have been a major use case.

> I'm happy with this, but I'd like @FranzBusch to review for sure, and ideally one of @allevato or @thomasvl.

I'll defer to you all, since we don't use SwiftPM within Google (just on some OSS things), this is less of an issue for us.

glbrntt:
> > In some build environments fetching the protoc binary artifact doesn't work as expected (e.g. using '--build-system xcode').
> 
> Does that mean the plugin support doesn't work for Xcode users? That seems like it might have been a major use case.

No, that still works.

It's only problematic if you're invoking the Xcode build system via SwiftPM (e.g. `swift build --build-system xcode`).

FranzBusch:
> Does that mean the plugin support doesn't work for Xcode users? That seems like it might have been a major use case.

As @glbrntt said this only is a problem when using Swift PM support for Xcodes build system which was never an officially supported build system but is still used by some folks. The Swift PM maintainers are currently working on a new swift-build based build system which is intended to replace both the current native and Xcode build system.

FranzBusch (approved):

thomasvl:
@FranzBusch do you all need a 1.31.1 release with this? https://github.com/apple/swift-protobuf/compare/1.31.0...main

glbrntt:
> @FranzBusch do you all need a 1.31.1 release with this? [1.31.0...main](https://github.com/apple/swift-protobuf/compare/1.31.0...main)

Yes please!

PR #1853, opened 2025-09-18
Include Swift 6.2 in the CI config.

thomasvl:
- Add it into the mix for builds.
- Move conformance config to 6.2.

NOTE: Not removing 5.10 at this time. Per policy it can be removed, but waiting until nio has so we stay in sync on supported OSes for grpc.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,6 +23,8 @@ jobs:
         # could use different Ubuntu releases. At the moment they are all the "noble",
         # which is also what would be desired, so we don't bother listing explicit ones.
         swift:
+        - version: "6.2"
+          hook: "SWIFT_BUILD_TEST_HOOK=\"-Xswiftc -warnings-as-errors\""
         - version: "6.1"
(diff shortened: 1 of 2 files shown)


thomasvl:
I'm guessing you will want to add the 6.2 build to the required checks in the GitHub config?

thomasvl:
@FranzBusch @Lukasa to the 5.10 point - https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/install_swift.sh still using 5.10, but looking at https://github.com/google/oss-fuzz/tree/master/infra/base-images/base-builder-swift, they maybe want to update the base images (they seem to now support two).  Any chance someone on the nio side can help updated things there to say 6.2 so we're good for a while again?

thomasvl:
@FranzBusch @Lukasa ping on getting this landed (independent of the question over when to drop 5.10 and fuzz testing)

I believe the config needs to get uploaded to block for this as well as needing approval.

FranzBusch:
Updated the rules for PR checks

PR #1857, opened 2025-10-05
Add basic RawSpan support for initializing and merging messages

pawelmajcher:
On platforms targeting Swift 6.2 and later this allows a `RawSpan` in lieu of `SwiftProtobufContiguousBytes`-conforming type when creating or merging a protobuf message. This will improve memory usage by reducing allocations, for example when parsing a large document that includes protobuf messages as part of its binary. Unlike the internal `_merge` function, this does not expose unsafe API while retaining its benefits.

- I decided to make a twin declaration of the initializer and `merge` function instead of conforming `RawSpan` to `SwiftProtobufContiguousTypes`, because
  - a) `RawSpan` is not mutable, and
  - b) `Span` types are non-owning and cannot conform to initializers.
- I used the conditional compilation block and `@available` directive per analogy to [this implementation in SwiftNIO](https://github.com/apple/swift-nio/blob/main/Sources/NIOCore/ByteBuffer-core.swift#L1066).

### DIFF
--- a/Sources/SwiftProtobuf/Message+BinaryAdditions.swift
+++ b/Sources/SwiftProtobuf/Message+BinaryAdditions.swift
@@ -100,6 +100,34 @@ extension Message {
         try merge(serializedBytes: bytes, extensions: extensions, partial: partial, options: options)
     }
 
+    #if compiler(>=6.2)
+    /// Creates a new message by decoding the bytes provided by a `RawSpan`
+    /// containing a serialized message in Protocol Buffer binary format.
+    ///
+    /// - Parameters:
+    ///   - serializedBytes: The `RawSpan` of binary-encoded message data to decode.
+    ///   - extensions: An ``ExtensionMap`` used to look up and decode any
+    ///     extensions in this message or messages nested within this message's
+    ///     fields.
+    ///   - partial: If `false` (the default), this method will check
+    ///     ``Message/isInitialized-6abgi`` after decoding to verify that all required
+    ///     fields are present. If any are missing, this method throws
+    ///     ``BinaryDecodingError/missingRequiredFields``.
+    ///   - options: The ``BinaryDecodingOptions`` to use.
+    /// - Throws: ``BinaryDecodingError`` if decoding fails.
+    @inlinable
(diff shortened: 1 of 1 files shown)


Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -100,6 +100,34 @@ extension Message {
...
+    ///     extensions in this message or messages nested within this message's
+    ///     fields.
+    ///   - partial: If `false` (the default), this method will check
+    ///     ``Message/isInitialized-6abgi`` after decoding to verify that all required
+    ///     fields are present. If any are missing, this method throws
+    ///     ``BinaryDecodingError/missingRequiredFields``.
+    ///   - options: The ``BinaryDecodingOptions`` to use.
+    /// - Throws: ``BinaryDecodingError`` if decoding fails.
+    @inlinable
+    @available(macOS 26, iOS 26, tvOS 26, visionOS 26, *)
```
Lukasa: You can use an older availability guard here. NIO's is a leftover from being added before the API shipped, but RawSpan was backported. Its correct availability per the docs is:

```swift
@available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0)
```

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -131,6 +159,39 @@ extension Message {
...
```
Lukasa: Same note on availability here.

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
pawelmajcher: Thanks for noticing! Updated the attributes to support the backport.

Lukasa (changes requested):
As an API this makes sense to me, but I'd like to see some tests here.

pawelmajcher:
I added test coverage for both initializing and merging messages with RawSpan, but I'm not sure if more extensive testing is needed here.

PR #1859, opened 2025-10-07
Be explicit about the type and bound we're checking in `BytecodeReader.nextInt32`.

### LINKED ISSUE #1858: "nextInt32() read a value larger than 32 bits" assertion failure on Apple Watch (S8 SoC)
I'm getting an assertion failure "nextInt32() read a value larger than 32 bits" when running on an Apple Watch Ultra (gen 1) watchOS 26.0.2 (23R362) at [BytecodeReader.swift:71](https://github.com/apple/swift-protobuf/blob/05f940e45a8d7759a18f0d643f70e5128c44a487/Sources/SwiftProtobuf/BytecodeReader.swift#L71).

No issue in the simulator.

I looked pretty hard for anything that could be maxing out Int32 and couldn't find anything.

I regenerated and ran again using v1.30.0 (before BytecodeReader) and haven't encountered any issues.

Is this a 32-bit Int sneaking in when running on older watch hardware? `1` as Int32 in `uint64Value < 1 &<< 32`?

allevato:
Thanks to a heterogeneous comparison operator on `BinaryInteger`, the expression `someUInt64 < 1 &<< 32` was compiling incorrectly on 32-bit architectures, because the right-hand side was being inferred as `Int` instead of `UInt64`. And on a 32-bit system, `1 &<< 32` is zero.

Fixes #1858.

### DIFF
--- a/Sources/SwiftProtobuf/BytecodeReader.swift
+++ b/Sources/SwiftProtobuf/BytecodeReader.swift
@@ -68,7 +68,7 @@ package struct BytecodeReader<Instruction: RawRepresentable> where Instruction.R
         // `Int32`s are stored by converting them bit-wise to a `UInt32` and then zero-extended to
         // `UInt64`, since this representation is smaller than sign-extending them to 64 bits.
         let uint64Value = nextUInt64()
-        assert(uint64Value < 1 &<< 32, "nextInt32() read a value larger than 32 bits")
+        assert(uint64Value < UInt64(0x1_0000_0000), "nextInt32() read a value larger than 32 bits")
         return Int32(bitPattern: UInt32(truncatingIfNeeded: uint64Value))
     }
 


thomasvl:
Aside - there's no public api addition here, but I do wonder if this should be `semver/minor` due to the bug in question.

allevato:
> Aside - there's no public api addition here, but I do wonder if this should be `semver/minor` due to the bug in question.

The bytecode was introduced in 1.31.0, so if we release this fix as the next patch release (1.31.2), then users who are using `upToNextMinor` in their SwiftPM dependencies would get the fix automatically. Likewise, if we released it as 1.32.0, then those users _wouldn't_ get the fix without manually updating, which would be undesirable.

thomasvl:
> > Aside - there's no public api addition here, but I do wonder if this should be `semver/minor` due to the bug in question.
> 
> The bytecode was introduced in 1.31.0, so if we release this fix as the next patch release (1.31.2), then users who are using `upToNextMinor` in their SwiftPM dependencies would get the fix automatically. Likewise, if we released it as 1.32.0, then those users _wouldn't_ get the fix without manually updating, which would be undesirable.

And I guess the bug only happens to folks using TextFormat or JSON on watchOS.  Binary on watchOS is fine, and TextFormat/JSON on everything else is 64bit and won't run into this problem.

FranzBusch:
I think this is fine as a patch release since this looks like a relatively straightforward bug-fix.

PR #1861, opened 2025-10-07
[table-driven-branch] Initial definition of table-driven message layout and in-memory storage.

allevato:
This is the first chunk of work driving toward table-driven messages. The best way to view this would be commit-by-commit, rather than the whole PR diff. I've opened a separate `table-driven-branch` for all this work so that I can iterate without interfering with `main` development.

As discussed offline, the general idea is that we do something similar to what we did for `NameMap` by introducing a `StaticString`-encoded description that compactly represents the layout of the message. By combining this with a single common storage class that manages a raw memory buffer based on that layout, we can dramatically reduce codegen size (we're no longer generating unique metadata for each private storage class, and eventually we won't be generating unique methods for decode, traverse, `==`, and so forth).

This PR introduces the following concepts:

* `_MessageLayout`, a helper that extracts data from the `StaticString` description for a particular message type
* `_MessageStorage`, the common storage class that allocates space and manages slices of it based on a `_MessageLayout`. It contains accessors/mutators for fields based on their byte offset, and eventually will house the other fundamental operations like traverse, decode, equality, and so forth.
* The generator is updated to produce the new implementation for accessors and the message layout string, and I've carved out a tiny subset of the unittest.proto files to generate them in isolation so that I can verify that the generated code compiles and I can write some basic unit tests as I build out more functionality.

There are still a lot of missing parts here, like maps, oneofs, extensions, etc.

Happy to answer questions about the gritty details throughout; some of them were in my original design doc but I've made a few changes since that write-up based on more prototyping.


Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -0,0 +1,335 @@
...
+///     match µpb's layout constraints, this value will never be larger than 2^16 - 1.
+/// *   Bytes 8-9: Information about the field's presence. Specifically,
+///     *   If this field is a member of a `oneof`, then the bitwise inverse of this value is the
+///         byte offset into in-memory storage where the field number of the populated `oneof`
+///         field is stored.
+///     *   Otherwise, the value is the index of the has-bit used to store the presence of the
+///         field.
+/// *   Bytes 10-11: For message/group fields, an opaque index as a base-128 integer used to
+///     request the metatype of the submessage from the containing message's submessage accessor.
+/// *   Bytes 12: The type of the field.
```
thomasvl: Nit: "Byte", it's just one.

Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -0,0 +1,335 @@
...
+    /// Creates a new message layout from the given values.
+    ///
+    /// This initializer is public because generated messages need to call it.
+    public init(_ layout: StaticString) {
+        precondition(
+            layout.hasPointerRepresentation,
+            "The layout string should have a pointer-based representation; this is a generator bug"
+        )
+        self.layout = UnsafeRawBufferPointer(start: layout.utf8Start, count: layout.utf8CodeUnitCount)
+        precondition(version == 0, "This runtime only supports version 0 message layouts")
```
thomasvl, line 140: Should there be any check that the overall size matches the requirements?  i.e. - is header plus a multiple of field size? the fieldCount matches that, etc.?

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -249,6 +253,36 @@ class MessageGenerator {
...
+            "",
+            "\(visibility)static func _protobuf_submessage(for token: SwiftProtobuf._MessageLayout.SubmessageToken) -> any SwiftProtobuf.Message.Type {"
+        )
+        p.withIndentation { p in
+            let submessageIndicesAndNames = messageLayoutCalculator.submessageIndicesAndNames
+            if submessageIndicesAndNames.isEmpty {
+                p.print("preconditionFailure(\"invalid submessage token; this is a generator bug\")")
+            } else {
+                p.print("switch token.index {")
+                for (index, type) in submessageIndicesAndNames {
```
thomasvl: I wonder if these could get to be too large to type check. :(

If there is only one type, do we want to use a switch or just a precondition to validate the index and direct return?

Sources/protoc-gen-swift/MessageLayoutCalculator.swift
```diff
@@ -0,0 +1,181 @@
...
+
+    /// The Swift string literals (without surrounding quotes) that encode the message layout in
+    /// the generated source.
+    var layoutLiterals: TargetSpecificValues<String> {
+        layoutWriters.map(\.layoutCode)
+    }
+
+    /// The mapping from index to fully-qualified name of all submessages used by the message whose
+    /// layout is being calculated, sorted by index.
+    var submessageIndicesAndNames: [(index: Int, name: String)] {
```
thomasvl: Do we need the indexes in here?  Shouldn't it always start with 1 and be dense?

Sources/protoc-gen-swift/MessageLayoutCalculator.swift
```diff
@@ -0,0 +1,181 @@
...
+                break
+            }
+            lastFieldNumber = field.number
+        }
+        let denseBelow = lastFieldNumber + 1
+
+        let requiredFieldsFirst = fieldsSortedByNumber.sorted { $0.isRequired && !$1.isRequired }
+        var requiredCount = 0
+        for field in requiredFieldsFirst {
+            field.presence = fieldCount
```
thomasvl: I'm not follow this at first glance, why are we setting all these values to the fieldCount?

Sources/protoc-gen-swift/MessageLayoutCalculator.swift
```diff
@@ -0,0 +1,181 @@
...
+        // TODO: Compute the presence information for each oneof by reserving a 32-bit integer for
+        // each oneof in the in-memory storage, after the has-bits.
+
+        // Compute the byte offset of each field in storage. We will start at the offset that is
+        // a byte past the last has-bit.
+        let fieldsSortedByStorage = fieldsSortedByNumber.sorted { $0.storageKind < $1.storageKind }
+        for field in fieldsSortedByStorage {
+            let fieldSizes = field.storageKind.strides
+
+            // Make sure we're properly aligned for this type.
```
thomasvl, line 109: I don't think I saw it noted anyplace, but the previous source and this is specific to ensure a more packed layout?  small or large or reverse?  maybe document that where the relevant parts are?

Sources/protoc-gen-swift/MessageLayoutCalculator.swift
```diff
@@ -0,0 +1,181 @@
...
+
+/// Manages the generation of a message layout string for a single platform.
+private struct MessageLayoutWriter {
+    /// Contains the Swift string literal (without quotes) that encodes the message layout in the
+    /// generated source.
+    var layoutCode: String = ""
+
+    /// Appends the given integer to the encoded layout literal in base 128 format, using the given
+    /// number of bytes to represent it.
+    mutating func writeBase128Int(_ value: UInt64, byteWidth: Int) {
```
thomasvl, line 149: Do you want tests for any of these parsers/writers?

Sources/protoc-gen-swift/TargetSpecificValues.swift
```diff
@@ -0,0 +1,145 @@
+// Sources/protoc-gen-swift/PointerWidthPair.swift - A pair of values based on pointer width
```
thomasvl: Name here doesn't match file name.

Sources/protoc-gen-swift/TargetSpecificValues.swift
```diff
@@ -0,0 +1,145 @@
...
+    /// The value for platforms that use 64-bit pointers.
+    case pointerWidth64
+
+    /// The value for platforms that use 32-bit pointers.
+    case pointerWidth32
+
+    /// The compiler condition that corresponds to this value choice.
+    var compilerCondition: String {
+        switch self {
+        case .pointerWidth64: return "_pointerBitWidth(_64)"
```
thomasvl, line 57: aside - should we note any place why we only care about pointer width?  i.e. - watchOS with it's odd size of ints/pointers?

andrep:
This is exciting Tony, thanks! One question — I couldn't find the design doc that you mentioned at the end of your PR description. Do you have a link to that? 🙏🏼

allevato:
> This is exciting Tony, thanks! One question — I couldn't find the design doc that you mentioned at the end of your PR description. Do you have a link to that? 🙏🏼

I had only shared it among the SwiftProtobuf maintainers. Here's a public-friendly version: https://gist.github.com/allevato/546f0408e297f96f924f00409139b572

There's already been some divergence from that document—for example, I'm no longer using a tail-allocated `ManagedBuffer` but instead just wrapping a `class` (in order to get unique reference checking for CoW) over a separately allocated `UnsafeMutableRawBufferPointer`. The former, I've heard, is less friendly toward new security features like typed allocators.

Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -0,0 +1,304 @@
...
+        return oldValue != 0
+    }
+}
+
+// MARK: - Field readers
+
+// The field reader functions have some explicit specializations (both concrete and more-constrained
+// generic) for cases where we want to encode a "default default value". This reduces the amount of
+// code generation that we need to do for the common case (fields without presence, where the
+// default value is the zero/empty value for the field's type).
```
thomasvl, line 109: Do you want to document the inline/emitintoclient directives and why they are used?

Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -0,0 +1,335 @@
...
+//
+// -----------------------------------------------------------------------------
+///
+/// The in-memory layout description for the fields of a table-driven message.
+///
+// -----------------------------------------------------------------------------
+
+/// Defines the in-memory layout of the storage for a message.
+///
+/// This type is public because it needs to be referenced and initialized from generated messages.
```
Lukasa, line 17: At this stage should we be putting these behind `@_spi` annotations? Just to add a little speed-bump to finding these.

Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -0,0 +1,335 @@
...
+/// *   Bytes 7-9: The number of required fields defined by the message, as a base-128 integer.
+/// *   Bytes 10-12: The largest field number `N` for which all fields in the range `1..<N` are
+///     inhabited, as a base-128 integer. We only need three bytes to represent this because the
+///     largest possible number is 65537; otherwise, it would imply that the message had more than
+///     65536 fields in violation of the bound above.
+///
+/// ### Field layouts
+///
+/// After the header above, starting at byte offset 13, is a sequence of encoded field layouts.
+/// Each field layout is 13 bytes long and they are written in field number order. Each entry
```
Lukasa, line 65: Have we measured the impact of this slightly tricky size? Conceptually using 16-byte units should lead to a few microarchitectural benefits:

- loads can be done using either vector units or arm64's funny `ldp` instruction
- such loads are always aligned
- field layouts rarely (possibly never) span pages
- computation of addresses for offsets doesn't require a multiplication but can be done entirely with shifts and adds

This is definitely in "microoptimization" territory and shouldn't be thought of as blocking this PR, but I wanted to suggest it might be worth following up on, particularly for very large types.

Lukasa: The natural downside is that the fields are bigger but even that may not hurt too much, as we can minimize the growth by "unpacking" some things that are currently packed, which may win us back some code size in the manipulation.

Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -0,0 +1,335 @@
...
+///         field is stored.
+///     *   Otherwise, the value is the index of the has-bit used to store the presence of the
+///         field.
+/// *   Bytes 10-11: For message/group fields, an opaque index as a base-128 integer used to
+///     request the metatype of the submessage from the containing message's submessage accessor.
+/// *   Bytes 12: The type of the field.
+public struct _MessageLayout: @unchecked Sendable {
+    // Using `UnsafeRawBufferPointer` requires that we declare the `Sendable` conformance as
+    // `@unchecked`. Clearly this is safe because the pointer obtained from a `StaticString` is an
+    // immortal compile-time constant and we only read from it.
```
Lukasa, line 91: Is there any possibility to make this pointer `private`? It makes this assertion easier to validate.

Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -0,0 +1,335 @@
...
+
+    /// The encoded layout of the fields of the message.
+    let layout: UnsafeRawBufferPointer
```
Lukasa: A question for us: is there a way for us to use a typealias and some compatible APIs to make this a `RawSpan` on platforms and compiler versions that have access to that type, while falling back to the pointer when we don't? It'd be nice to have the bounds-checking.

Sources/SwiftProtobuf/_MessageLayout.swift
Lukasa: If not, can we consider making it harder to construct one? Maybe remove the `public init` and instead give it a leading-underscore `static` factory function.

Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -0,0 +1,335 @@
...
+        self.layout = UnsafeRawBufferPointer(start: layout.utf8Start, count: layout.utf8CodeUnitCount)
+        precondition(version == 0, "This runtime only supports version 0 message layouts")
+    }
+}
+
+/// The size, in bytes, of the header the describes the overall message layout.
+private let messageLayoutHeaderSize = 13
+
+/// The size, in bytes, of an encoded field layout in the static string representation.
+private let fieldLayoutSize = 13
```
Lukasa: Nit: it tends to be nicer to make these computed `var` so that the compiler doesn't accidentally end up allocating storage for these.

Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -0,0 +1,335 @@
...
+
+extension _MessageLayout {
+    /// Calls the given function for each field defined in the message, in order, passing it the
+    /// layout of the field.
+    func forEachField(body: (FieldLayout) -> Void) {
```
Lukasa: Can this be made into an iterator rather than a foreach?

Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -0,0 +1,335 @@
...
+        for index in stride(from: messageLayoutHeaderSize, to: layout.count, by: fieldLayoutSize) {
+            body(FieldLayout(slice: layout[index..<(index + fieldLayoutSize)]))
+        }
+    }
+
+    /// Calls the given function for the field with the given number in the message, passing it the
+    /// layout of the field.
+    ///
+    /// - Precondition: The field must be defined.
+    func forField(withNumber number: UInt32, body: (FieldLayout) -> Void) {
```
Lukasa: Same note here: can this be a subscript rather than a closure-taking API?

Sources/protoc-gen-swift/FieldGenerator.swift
```diff
@@ -199,6 +187,28 @@ class FieldGeneratorBase {
...
+    var storageOffsetExpression: String {
+        // If all the values are the same, generate cleaner code by just passing the single value
+        // directly.
+        if let valueIfAllEqual = storageOffsets.valueIfAllEqual {
+            return "\(valueIfAllEqual)"
+        }
+
+        // Otherwise, generate a call to the helper function that chooses the right value based on
+        // target platform.
+        var result = "SwiftProtobuf._fieldOffset("
```
thomasvl: Aside - I wonder if `reduce` would be a good pattern for building up strings like this, see the initial value, and then just keep adding, etc. You could even see with `""`,  and use that input as the signal for the initial block vs. the `", "` to add.

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -133,8 +122,7 @@ class MessageGenerator {
         }
 
         // Messages that have a storage class will always need @unchecked.
```
thomasvl: Comment can go now.

Although it seems like we did "unchecked" when there was storage, and now we don't need that?

Sources/protoc-gen-swift/MessageLayoutCalculator.swift
```diff
@@ -60,22 +60,44 @@ struct MessageLayoutCalculator {
...
             }
         }
 
-        var byteOffsets = TargetSpecificValues<Int>(forAllTargets: fieldCount / 8)
-        if fieldCount % 8 != 0 {
-            byteOffsets.add(.init(forAllTargets: 1))
+        // Compute the byte offset following the has-bits.
+        var byteOffset = fieldCount / 8 + (fieldCount % 8 != 0 ? 1 : 0)
+
+        // If any oneofs are present in the message, allocate a `UInt32` for each one that will be
```
thomasvl, line 84: Given the max limit on the number of fields, we could use uint16, but then we couldn't use the field numbers as the values, you'd have to do your own mapping of values.

Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -0,0 +1,304 @@
...
+        hasBit: HasBit
+    ) -> Bool {
+        guard isPresent(hasBit: hasBit) else { return defaultValue }
+        return (buffer.baseAddress! + offset).bindMemory(to: Bool.self, capacity: 1).pointee
+    }
+
+    /// Returns the integer value at the given offset in the storage, or the default value if the
+    /// value is not present.
+    @_alwaysEmitIntoClient @inline(__always)
+    public func value<Value: BinaryInteger & FixedWidthInteger>(
```
Lukasa: How does this work with non-trivial types?

Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -0,0 +1,304 @@
...
+        hasBit: HasBit
+    ) -> Data {
+        guard isPresent(hasBit: hasBit) else { return defaultValue }
+        return (buffer.baseAddress! + offset).bindMemory(to: Data.self, capacity: 1).pointee
+    }
+
+    /// Returns the value at the given offset in the storage, or the default value if the value is
+    /// not present.
+    @_alwaysEmitIntoClient @inline(__always)
+    public func value<C: RangeReplaceableCollection>(
```
Lukasa: This seems unlikely to be right: does the raw storage contain a Swift `Array`?

Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest_import_public.pb.swift
```diff
@@ -0,0 +1,115 @@
...
+  var e: Int32 {
+    get { return _storage.value(at: 4, default: 0, hasBit: (0, 1)) }
+    set { _uniqueStorage().updateValue(at: 4, to: newValue, willBeSet: true, hasBit: (0, 1)) }
+  }
+  /// Returns true if `e` has been explicitly set.
+  var hasE: Bool { return _storage.isPresent(hasBit: (0, 1)) }
+  /// Clears the value of `e`. Subsequent reads from it will return its default value.
+  mutating func clearE() { _uniqueStorage().clearValue(at: 4, type: Int32.self, hasBit: (0, 1)) }
+
+  var unknownFields = SwiftProtobuf.UnknownStorage()
```
thomasvl, line 69: Aside - i do still wonder about moving these and extension storage if that would simplify parsing/serialization.

Lukasa (commented):
I took a bit of a spin through @allevato. I'm generally very excited for this change, thanks so much for tackling it! I've left some notes in the diff.

Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
```diff
@@ -0,0 +1,4920 @@
...
+  }
+
+  static func ==(lhs: SwiftProtoTesting_TestOneof.FooGroup, rhs: SwiftProtoTesting_TestOneof.FooGroup) -> Bool {
+    fatalError("table-driven == not yet implemented")
+  }
+}
+
+extension SwiftProtoTesting_TestRequiredOneof: SwiftProtobuf.Message, SwiftProtobuf._MessageImplementationBase, SwiftProtobuf._ProtoNameProviding {
+  static let protoMessageName: String = _protobuf_package + ".TestRequiredOneof"
+  stat
```
thomasvl, line 4901: Does a default impl for things like this make sense to we only have to generate it if it is needed?  Would that be any cost savings, or does that force runtime costs?  (I guess always generating it could force constant string data sizes also in other places)

Sources/protoc-gen-swift/FieldGenerator.swift
allevato: I realized after writing this that the loop could just be a simple `join`. I think `join` would be clearer than `reduce` in this case, and `join` specifically handles the nature of something "between each **pair** of elements" rather than something "for each element".

Sources/protoc-gen-swift/MessageGenerator.swift
allevato: We needed `@unchecked` before because the internal storage class wasn't `Sendable`.

Technically the new `_MessageStorage` isn't safely `Sendable` either, but I declared it as `@unchecked Sendable` because the only permitted use of it is by the generated messages, who ensure that it never has a data race because of the CoW logic.

I'm starting to question that choice though, to make sure it doesn't burn us in the future. I'll go back to not declaring any `Sendable`-ity on `_MessageStorage` and make the structs themselves `@unchecked Sendable`. That's safer in the event that we ever want to use `_MessageStorage` for something else dynamically.

Sources/protoc-gen-swift/MessageGenerator.swift
allevato: I can't recall, did we see type-checking problems from switches over integers, or just over enums? I _think_ only the latter should encounter difficulty from the exhaustiveness check.

This is a situation where it would be nice to just generate an array and index into it, but without constant-initialized data, that puts us back into the "lots of codegen to initialize a collection" situation that we're trying to avoid.

Sources/protoc-gen-swift/MessageLayoutCalculator.swift
allevato: Good point. Made it a simple array.

Sources/protoc-gen-swift/MessageLayoutCalculator.swift
allevato: Right. µpb reserves 32 bits per `oneof` to store field numbers here, so it seems simplest to do the same.

Sources/protoc-gen-swift/MessageLayoutCalculator.swift
allevato: It's documented in the comment on `FieldStorageKind` in `FieldGenerator.swift`. I'll add another comment here that points to that.

Sources/protoc-gen-swift/MessageLayoutCalculator.swift
allevato: The tests are somewhat gross to write because we're generating a Swift string literal in source code form (which includes backslash escapes). That makes it difficult to round-trip a string through the calculator -> writer -> layout decoder because we'd have to effectively write a small Swift string parser to turn something like `"\0"` into a single NUL code unit.

I'll think about whether there's a good way to do this, but I might just end up having to hardcode strings in each direction for the assertions.

Sources/protoc-gen-swift/TargetSpecificValues.swift
allevato: Would love if we could delete all these...

Sources/protoc-gen-swift/TargetSpecificValues.swift
allevato: I think the information here and in `FieldStorageKind` already cover what we need to say here.

I've designed this type to be more extensible in case we need to add conditions that aren't purely based on pointer/integer width, but for now, this should cover all the known platforms that we support. (It also matches what µpb supports.)

Sources/SwiftProtobuf/_MessageLayout.swift
allevato: Yeah, I'm happy to hide these behind `@_spi`. Our underscore-prefix convention came from the days before `@_spi` existed or was stable, but it should be safe to use now on all the toolchains we support.

(For this type in particular, the worst thing someone could do is initialize one, but then there are literally no other public operations they can invoke on it. `_MessageStorage` on the other hand benefits more from being hidden.)

Sources/SwiftProtobuf/_MessageLayout.swift
allevato: I agree, these are good ideas to explore. One of the tricky parts is that we still need to work at the level of individual bytes because of the 7-bit encoding, but there are likely still benefits from reading larger aligned values and doing the shifting/masking within that larger value instead of doing separate loads of individual bytes.

I'll plan to look into this further when I set up some simple benchmarks. Since it's relatively easy to update the generator and layout while I'm iterating on it, I can make changes like you're suggesting and measure the runtime of iterating over the fields of a message, say 1000 times. What's currently in this PR seems like a good baseline where the changes you're suggesting would likely be strictly improvements.

Sources/SwiftProtobuf/_MessageLayout.swift
allevato: Done.

Sources/SwiftProtobuf/_MessageLayout.swift
allevato: Done.

Sources/SwiftProtobuf/_MessageLayout.swift
allevato: That's likely possible but probably not straightforward right now, since the API surfaces are still fairly different.

For a `RawSpan`, we'd want to use the `extracting(...)` family of APIs for slicing, and then `unsafeLoad(...)` to read values therein. [SE-0488](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0488-extracting.md) was supposed to add the `extracting(...)` APIs to the various `Unsafe*BufferPointer` types in Swift 6.2, but oddly I only see them present on `Unsafe(Mutable)BufferPointer`, not any of the `Raw` variants. (As review manager of that proposal, I should follow up on this separately.)

Even if we had the full `extracting(...)` APIs, `RawSpan` differs from `UnsafeRawBufferPointer` by providing an API named `unsafeLoad(...)` instead of `load(...)`. So we'd have smooth over those API differences using additional extensions.

Sources/SwiftProtobuf/_MessageLayout.swift
allevato: Added a precondition check for the size of the layout.

Sources/SwiftProtobuf/_MessageLayout.swift
allevato: Done.

Sources/SwiftProtobuf/_MessageLayout.swift
allevato: Yes, it can now! My original implementation was storing the `StaticString` itself and calling `withUTF8Buffer`, so that forced a closure-based implementation. But now that I'm just storing the buffer pointer directly (which is both cheaper and safe since we know the layout string will always be more than 1 `Unicode.Scalar`), I can build an iterator around this instead.

Sources/SwiftProtobuf/_MessageLayout.swift
allevato: Likewise, done.

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: Elaborated on this here.

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: It wouldn't, but we're never going to generate any code that would operate on a hypothetical non-trivial integer type.

I could add `& BitwiseCopyable` to be more correct about this, but that would require us to bump our minimum Swift compiler version to 6.0. Maybe that's worth doing for a number of other reasons, but we should have that discussion first.

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: Yes, the raw storage would contain an `Array`.

Can you elaborate on what's incorrect about it? The reference count of the value will be handled in `updateValue`, where we call `initialize`/`deinitialize` as needed when modifying the values. This should simply retrieve the existing value out of the raw storage and then return a copy of it.

Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest_import_public.pb.swift
allevato: Yes, I plan to move these into `_MessageStorage`, but there's no urgency to do that until I start implementing the operations that touch them.

Sources/protoc-gen-swift/MessageLayoutCalculator.swift
allevato: I think this was a rebase/fixup problem on my end. This was an earlier version, but if you look at the state at the final commit, it's correct.

Sources/protoc-gen-swift/MessageGenerator.swift
thomasvl: ack on the array part and hopefully ints won't have the problem.

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: I just had an epiphany about this. While _we_ would never generate any code that used nontrivial integer types, nothing prevents someone from doing this:

```protobuf
package foo;

message Int96 {
  uint64 low = 1;
  int32 high = 2;
}
```

and then in Swift:

```swift
extension Foo_Int96: BinaryInteger, FixedWidthInteger {
  // implement the requirements
}
```

If the protos were generated into the same module as that extension, then the compiler might choose the `<BinaryInteger & FixedWidthInteger>` overloads for something that really needs to be accessed as a message.

So to avoid that, I've written the manual specializations for `(U?)Int(32|64)` and for `Float` and `Double`. I've also removed the `RangeReplaceableCollection` one in favor of explicit `Array` and `Dictionary` overloads.

It's possible that all of the `value(at: ...)` functions could be collapsed into a single unconstrained generic function, since they all do the same thing. Right now, I'm more confident that they'll optimize well if they're manually written out and they're zero-cost to the runtime since they're `@_alwaysEmitIntoClient`. But once I start measuring binary sizes, I can test that out and adjust as necessary.

(I may also move these to be subscripts instead of functions so that we can use `_modify` accessors, which would benefit in-place array/dictionary mutation—something that wasn't possible in my original `ManagedBuffer`-based prototype. So that might impact the shape of these APIs as well.)

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: I've updated this to be explicitly an overload on `[Element]` and added a separate one for dictionaries. I'm still curious about your concerns here, since AFAIK this should be safe since the initialized state of the space storing the collections is kept in sync.

There's probably a separate discussion to be had about whether we should create our own collections instead of leaning on `Array`/`Dictionary`. Probably more relevant for `Dictionary` since hashing collections make the use of Swift protos with `map` fields nondeterministic. But that would be a major API breaking change for clients.

Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
allevato: I need to flesh out the implementation some more to see whether/how these will get hoisted up into protocol requirements on `Message`. If so, then it would make sense to be a default implementation there.

Sources/SwiftProtobuf/_MessageStorage.swift
Lukasa: Oh the concern here was really only trying to understand how the memory management was lined up, and it wasn't clear that there were plausibly references to other storage from this buffer.

thomasvl:
> > This is exciting Tony, thanks! One question — I couldn't find the design doc that you mentioned at the end of your PR description. Do you have a link to that? 🙏🏼
> 
> I had only shared it among the SwiftProtobuf maintainers. Here's a public-friendly version: https://gist.github.com/allevato/546f0408e297f96f924f00409139b572
> 
> There's already been some divergence from that document—for example, I'm no longer using a tail-allocated `ManagedBuffer` but instead just wrapping a `class` (in order to get unique reference checking for CoW) over a separately allocated `UnsafeMutableRawBufferPointer`. The former, I've heard, is less friendly toward new security features like typed allocators.

Does it make sense to add maybe a "Design Notes" type directory (under "Documentation"?) and land this doc as a historical reference to provide background on the implementation?  (I can argue for/against it, just figured I'd ask)

allevato:
> Does it make sense to add maybe a "Design Notes" type directory (under "Documentation"?) and land this doc as a historical reference to provide background on the implementation? (I can argue for/against it, just figured I'd ask)

I think that's a good idea, but rather than using this doc as it is today, it would be better to use an updated version once the design/implementation settles. The background information inspired by µpb that's in there today is valuable, but I don't think it would be helpful to commit the parts that are still in flux and might differ from the eventual implementation that we land.

allevato:
Since this work is going on a separate branch, I'm going to go ahead and land these changes to keep moving forward. I still encourage comments on PRs even after they've been merged; if there's anything I've overlooked or clarifications I can make, that's still useful feedback that I can apply to later work.

Issue #1856, opened 2025-10-03
Swift complier fails in `traverse` function for large `oneof` fields


pm-dev:
Given a `oneof` field with ~600 options, the generated `traverse` function will fail to compile because swift is unable to check the switch is exhaustive in reasonable time

Ex/
```proto
message Action {
  oneof action {
    tracking.actions.feed.AddReaction add_reaction = 1;
    tracking.actions.feed.RemoveReaction remove_reaction = 2;
    ... // 600 options
```

For now, I've been able to work around this issue by implementing a similar solution to #904 by breaking up the switch statement.
Ex/
```swift
public func traverse<V: SwiftProtobuf.Visitor>(visitor: inout V) throws {
  switch self.action {
    case .addReaction?: try {...}()
    // ... ~500 cases
    default: break
  }
  switch self.action {
    // ... remaining cases
    case nil: break
    default: break
  }
  try unknownFields.traverse(visitor: &visitor)
}
```
Env:
swift: 6.2
swift-protobuf: 1.31.1

thomasvl:
Note: quick skim says the required fields/extensions validation hooks could need this also.

thomasvl:
> For now, I've been able to work around this issue by implementing a similar solution to https://github.com/apple/swift-protobuf/issues/904 by breaking up the switch statement.

Do you want to propose a PR for this to get credit for it?

thomasvl:
I believe this was fixed in the referenced PR.

PR #1866, opened 2025-10-08
Update OneofGenerator to support breaking up large switch statements

pm-dev:
Given a oneof field with ~600 options, the generated traverse function will fail to compile because swift is unable to check the switch is exhaustive in reasonable time. See https://github.com/apple/swift-protobuf/issues/1856

This PR follows similar logic to #904 which allows switch statements to be broken up into multiple switch statements (using a `default: break` case).

Instead of hardcoding 500, I've added a new GeneratorOption to support setting the "maxCasesInSwitch" which allows the same number to be used in `EnumGenerator` and `OneofGenerator` and allows clients to set this number to whatever suits them based on Swift version or other considerations. This number could also be set to something very high to effectively turn off the feature.

The second commit adds tests, I asked claude to help out with this part so please help me in double checking those tests. I was able to run tests successfully locally.

### DIFF
--- a/Tests/SwiftProtobufTests/Test_OneofSwitchSplit.swift
+++ b/Tests/SwiftProtobufTests/Test_OneofSwitchSplit.swift
@@ -20,38 +20,38 @@
 // -----------------------------------------------------------------------------
 
 import Foundation
-import XCTest
 import SwiftProtobuf
+import XCTest
 
 final class Test_OneofSwitchSplit: XCTestCase {
 
     func testChunkBoundaries() throws {
         var msg = SwiftProtoTesting_SwitchSplit_SwitchSplitMessage()
-        
+
         // Test chunk 1: fields 1-250 (before regular_field_251)
         msg.field001 = 1
         XCTAssertEqual(msg.field001, 1)
         msg.field250 = 250
         XCTAssertEqual(msg.field250, 250)
-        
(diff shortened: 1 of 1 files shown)


thomasvl:
Haven't looked at the PR yet, but the comment about the generation option for the size of switches gives be a little pause.  I follow the logic on the idea, but I also worry since every option is something we have to continue to support going forward, and to my knowledge, no one has come back with the 500 limit being too large/small yet.

@tbkka any thoughts on exposing this as a generation option vs. say just making it a shared constant between the two places?

pm-dev:
Would be easy enough to remove the option, but keep it as a constant on GeneratorOptions for sharing.
It did make testing easier because the proto file could be much smaller vs having 500+ oneof options

tbkka:
I too am reluctant to expose unnecessary options.

pm-dev:
I've updated my branch to remove the "MaxCasesInSwitch" option in favor of hardcoding a 500 case limit.

I've also updated tests, which now require oneof with 500+ options

pm-dev:
Updated with suggestions

pm-dev:
Anything remaining on my end to get this across the finish line?

thomasvl:
@tbkka any other thoughts on this one, or good to go?

PR #1867, opened 2025-10-10
[table-driven-branch] Move unknown fields and extension storage into `_MessageStorage`.

allevato:
These will likely change further as the table-driven approach develops,
but my goal with this change is to get everything into the common
storage type so that message types being a thin wrapper around a
single pointer. (These values would need to be in the storage anyway to
be reachable from the common message operations.)

### DIFF
--- a/Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
+++ b/Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
@@ -609,7 +609,13 @@ struct SwiftProtoTesting_TestAllTypes: @unchecked Sendable {
     set {oneofField = .oneofBytes(newValue)}
   }
 
-  var unknownFields = SwiftProtobuf.UnknownStorage()
+  var unknownFields: SwiftProtobuf.UnknownStorage {
+    get { _storage.unknownFields }
+    _modify {
+      _ = _uniqueStorage()
+      yield &_storage.unknownFields
+    }
+  }
 
   /// For oneof test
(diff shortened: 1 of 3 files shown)


allevato:
This is a tiny refactor (not much design impact here) so I'm going to go ahead and merge this. I don't expect that the final state of things will look like just these two fields hoisted up, but this helps simplify some future work.

Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -41,6 +41,13 @@ import Foundation
     /// The memory buffer that contain's the data for the message's fields.
     @usableFromInline internal let buffer: UnsafeMutableRawBufferPointer
 
+    /// The storage used for unknown fields.
+    public var unknownFields: UnknownStorage
+
+    /// The storage used for extension field values.
+    /// TODO: This will very likely change as the table-driven implementation evolves.
+    public var extensionFieldValues: ExtensionFieldValueSet
```
thomasvl, line 49: Not all messages need to support extensions. Not sure if you want to note that, and I guess we do the enforcement via if we provide the forwarding apis in the generated code?

allevato: Yeah, messages that don't need extensions just don't generate the public accessor for this right now.

I didn't want to belabor that too much here because `_MessageStorage` doesn't know from where it's being used. We could potentially do something with generics to conditionalize features that are only needed by some messages, or we could just say that all message storage reserves this small bit to avoid additional complexity. I'll need to see how the implementation shakes out.

PR #1868, opened 2025-10-13
[table-driven-branch] Implement deinitialization and copying.

allevato:
Note: `oneof`s and `map`s are still not implemented.

For most cases, deinitialization of message storage is fairly straightforward. For scalar trivial values, we needn't do anything. For nontrivial types like `String`, `Data`, and arrays of fields, we have to tell the compiler statically what the actual type is so that it can generate the right code for the operation.

Copying message storage (to implement CoW) is similar. We ignore scalar values at first, because they (along with has-bits) are stored together at the front of the buffer. While we iterate over the fields, we copy-initialize any non-trivial fields (again telling the compiler what the actual type is), and we record the offset of the first non-trivial field we saw. That lets us do a single "memcpy" at the end to copy all the trivial fields. This is a small optimization but it seems worthwhile to take advantage of this characteristic of the layout.

> NOTE: There are still be some cases that we could optimize. For example, a message with only trivial fields would still iterate over all of the field layouts, do nothing in the loop, and then do nothing else (for deinit) or just do a memcpy (for copying). So it might be worthwhile to use a bit somewhere in the layout to describe whether a message is only trivial fields. Leaving this for a future PR.

What is less straightforward: we need a way to deal with arbitrary message types (singular and repeated). The original plan was to use the generated submessage accessor function for this, but that doesn't work well for arrays of messages. After chatting with @thomasvl offline, I'm going to look at replacing the submessage accessor concept with a set of "submessage operations" that would be a sort of callback where `_MessageStorage` can delegate an operation on a value of an arbitrary complex type to the message, who would have more information generated about the types in question.

### DIFF
--- a/Sources/SwiftProtobuf/_MessageLayout.swift
+++ b/Sources/SwiftProtobuf/_MessageLayout.swift
@@ -226,8 +226,11 @@ struct FieldLayout {
         case oneOfMember(Int)
 
         fileprivate init(rawValue: Int) {
+            // The raw value needs to be treated as a 14-bit signed integer where the MSB (bit 13)
+            // acts as the sign bit. Therefore, we need to check the range of the value to
+            // determine if it's a oneof (0x2000...0x3fff) or not (0x0000...0x1fff), then
+            // sign-extend it to 16 bits so that we can correctly take its inverse.
             if rawValue >= 0x2000 {
-                // Sign-extend the original 14-bit value so that we can correctly take its inverse.
                 self = .oneOfMember(Int(~(UInt16(rawValue) | 0xc000)))
             } else {
                 self = .hasBit(byteOffset: rawValue >> 3, mask: 1 << UInt8(rawValue & 7))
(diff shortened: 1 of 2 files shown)


Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -215,6 +215,26 @@ extension _MessageLayout {
...
+    enum Presence: Sendable, Equatable {
+        /// The byte offset and mask of the has-bit for this field that is not a oneof member.
+        case hasBit(byteOffset: Int, mask: UInt8)
+
+        /// The byte offset of the 32-bit integer that holds the field number of the currently set
+        /// oneof member.
+        case oneOfMember(Int)
+
+        fileprivate init(rawValue: Int) {
+            if rawValue >= 0x2000 {
```
thomasvl, line 233: Worth a constant for this so the valid is less "magic" and self documenting?

allevato: This is the only place it's used and the comment below describes it. I've moved the comment up and added more detailed, but I try to avoid single-use-named-constants in general because the indirection often harms readability more than it just showing what the number is.

PR #1869, opened 2025-10-14
[table-driven-branch] Deinitialize and copy submessage fields.

allevato:
In order to perform the low-level typed operations on the raw memory, the protobuf runtime needs to be able to refer statically to the type of a submessage field. This is inherently challenging when we're trying to use a single storage class to represent all possible messages.

To achieve this, we generate trampoline functions that are passed into the `_MessageLayout` initializer and stored. Then, for example, when deinitializing a submessage field, the storage class calls the deinitializer trampoline with the appropriate submessage index and the trampoline function immediately calls back into the storage class but passes it the type hint so that the memory can be bound to the correct type. Likewise for copying a submessage field.

This replaces the original notion of a submessage accessor function that relied on returning the metatype as an existential value. That wouldn't be compatible with embedded Swift; these type hints should be.

I've also updated the generator to avoid generating these trampoline functions if a message doesn't have any submessages. In this case, we can just pass a placeholder function baked into the runtime that does a precondition failure, so we save a bit on codegen.

### DIFF
--- a/Sources/protoc-gen-swift/MessageGenerator.swift
+++ b/Sources/protoc-gen-swift/MessageGenerator.swift
@@ -265,8 +265,8 @@ class MessageGenerator {
         }
 
         let submessageNames = messageLayoutCalculator.submessageNames
+        p.print()
         p.print(
-            "",
             "private static let _protobuf_messageLayout = SwiftProtobuf._MessageLayout(layout: _protobuf_messageLayoutString",
             newlines: false
         )
--- a/Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
+++ b/Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
@@ -4119,6 +4119,7 @@ extension SwiftProtoTesting_TestAllTypes: SwiftProtobuf.Message, SwiftProtobuf._
   #else
     #error("Unsupported platform")
   #endif
+
   private static let _protobuf_messageLayout = SwiftProtobuf._MessageLayout(layout: _protobuf_messageLayoutString, deinitializeSubmessage: _protobuf_deinitializeSubmessage, copySubmessage: _protobuf_copySubmessage)
 
   private static func _protobuf_deinitializeSubmessage(for token: SwiftProtobuf._MessageLayout.SubmessageToken, field: SwiftProtobuf.FieldLayout, storage: SwiftProtobuf._MessageStorage) {
@@ -4184,6 +4185,7 @@ extension SwiftProtoTesting_TestAllTypes.NestedMessage: SwiftProtobuf.Message, S
   #else
     #error("Unsupported platform")
   #endif
+
   private static let _protobuf_messageLayout = SwiftProtobuf._MessageLayout(layout: _protobuf_messageLayoutString)
 
   public var isInitialized: Bool {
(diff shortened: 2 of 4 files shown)


Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
```diff
@@ -4111,15 +4111,45 @@ extension SwiftProtoTesting_TestAllTypes: SwiftProtobuf.Message, SwiftProtobuf._
...
+    @_alwaysEmitIntoClient @inline(__always)
```
thomasvl, line 4114: why the directives here since this is already in the generated code?  Does it force it into the init call?

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -257,33 +257,68 @@ class MessageGenerator {
     }
 
     private func generateMessageLayout(printer p: inout CodePrinter) {
+        let submessageNames = messageLayoutCalculator.submessageNames
+        let deinitializeSubmessageName: String
+        let copySubmessageName: String
+        if submessageNames.isEmpty {
```
thomasvl: I'm not sure if it would save anything, but you could put two initializers, one with hooks and one without, and then have the runtime carry the one without the hooks that just defaults the arguments. it might even let you make those default private within the runtime so theres a few less symbols to expose.

allevato: Done. That makes the generator a lot cleaner and it also means we don't have to expose those entry points outside of the runtime.

Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
allevato: So as part of my deep diving into codegen, if this was just a regular property, we'd still generate a separate accessor for it. That accessor serves no purpose, because the string is only referenced from one place. `@_alwaysEmitIntoClient` ensures that the symbol for the entry point is never generated.

Now that the `_MessageLayout` initializer can take other functions, I didn't want to duplicate all of that in each branch of the `#if`. So this change lets us break out the layout string into a separate property and make that be the only part that's conditionally compiled, and then the `_MessageLayout` initializer just references the symbol _at no cost_ because it will be emitted directly at the call site.

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -257,32 +257,32 @@ class MessageGenerator {
...
                 #"private static var _protobuf_messageLayoutString: StaticString { "\#(value)" }"#
             )
         }
+
+        let submessageNames = messageLayoutCalculator.submessageNames
         p.print(
             "",
-            "private static let _protobuf_messageLayout = SwiftProtobuf._MessageLayout(layout: _protobuf_messageLayoutString, deinitializeSubmessage: \(deinitializeSubmessageName), copySubmessage: \(copySubmessageName))"
+            "private static let _protobuf_messageLayout = SwiftProtobuf._MessageLayout(layout: _protobuf_messageLayoutString",
+            newlines: false
```
thomasvl, line 271: With `newlines: false`, doesn't that make the empty string also in this print a noop?

allevato: Yeah... fixed.

Rewriting these printer APIs is something else I'd like to add to my todo list.

thomasvl: Yea, the new c++ `Emit()` apis are sorta interesting to copy, but it also feels like sting interpolation could lead to something else.

allevato:
Merging this since I have `oneof`s coming up, stacked on top of this.

PR #1870, opened 2025-10-15
[table-driven-branch] Add support for `oneof`s.

allevato:
This commit adds support for in-memory storage (and getters/setters) of `oneof`s and their members.

We keep the same API as before: we generate an `enum` with associated values to represent the `oneof` itself, and we also provide getters and setters for the individual members directly on the containing message.

Though this is invisible to the end user, the implementation has "flipped". In the current main branch, the value we store is the enum value, and the member getters/setters extract the value from that. With the table-driven approach, we store the values in memory as a `UInt32` that tracks the currently set field along with whatever storage a field of each member type would normally require. So the member accessors are now "fast" and directly manipulate storage, while the enum accessor does a `switch` to determine which field needs to be manipulated.

Possible future optimization: Today, we allocate separate storage in memory for each oneof field, even though only one of
them can be set at any time. We could do what µpb does and have the storage overlap (choosing the max size among all the member types).

Since this is the first time the binary-search subscript is being used in `_MessageLayout`, I've also fixed a bug there where I had a comparison reversed. I blame this on µpb naming both of their operands `num` and `number`...

### DIFF
--- a/Sources/protoc-gen-swift/OneofGenerator.swift
+++ b/Sources/protoc-gen-swift/OneofGenerator.swift
@@ -26,6 +26,7 @@ class OneofGenerator {
         let swiftName: String
         let dottedSwiftName: String
         let swiftType: String
+        let hasExplicitDefaultValue: Bool
         let swiftDefaultValue: String
         let protoGenericType: String
         let comments: String
@@ -72,6 +73,7 @@ class OneofGenerator {
             swiftName = names.name
             dottedSwiftName = names.prefixed
             swiftType = descriptor.swiftType(namer: namer)
+            hasExplicitDefaultValue = descriptor.defaultValue != nil
             swiftDefaultValue = descriptor.swiftDefaultValue(namer: namer)
             protoGenericType = descriptor.protoGenericType
             comments = descriptor.protoSourceCommentsWithDeprecation(generatorOptions: generatorOptions)
@@ -251,13 +253,24 @@ class OneofGenerator {
         let visibility = generatorOptions.visibilitySourceSnippet
         let oneofPresence = "(\(field.oneofOffset), \(field.number))"
 
+        // Only generate a default value expression for the getter if the proto contained an
+        // explicitly written default value (or if it is a message field, since we don't have a
+        // suitable default value in that overload).
+        let defaultValueArgument: String
+        switch field.rawFieldType {
+        case .group, .message:
+            defaultValueArgument = "default: \(field.swiftDefaultValue), "
+        default:
+            defaultValueArgument = field.hasExplicitDefaultValue ? "default: \(field.swiftDefaultValue), " : ""
+        }
+
         // The individual member accessors manipulate the underlying storage directly.
         p.print(
             "",
             "\(field.comments)\(visibility)var \(field.swiftName): \(field.swiftType) {"
         )
         p.printIndented(
-            "get { return _storage.value(at: \(field.storageOffsetExpression), default: \(field.swiftDefaultValue), oneofPresence: \(oneofPresence)) }",
+            "get { return _storage.value(at: \(field.storageOffsetExpression), \(defaultValueArgument)oneofPresence: \(oneofPresence)) }",
             "set { _uniqueStorage().updateValue(at: \(field.storageOffsetExpression), to: newValue, oneofPresence: \(oneofPresence)) }"
         )
         p.print("}")
--- a/Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_TableDriven.swift
+++ b/Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_TableDriven.swift
@@ -59,6 +59,7 @@ final class Test_TableDriven: XCTestCase {
             .with { $0.d = 20 },
             .with { $0.d = 30 },
         ]
+        msg.oneofField = .oneofString("oneof string")
 
         var msgCopy = msg
         msgCopy.optionalBool = false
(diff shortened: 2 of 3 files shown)


Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -484,9 +489,238 @@ extension _MessageStorage {
...
+        let offsetPointer = (buffer.baseAddress! + presence.offset).bindMemory(to: UInt32.self, capacity: 1)
+        let oldFieldNumber = offsetPointer.pointee
+        offsetPointer.pointee = presence.fieldNumber
+        return oldFieldNumber
+    }
+
+    /// Returns the `Bool` value at the given offset in the storage if it is the currently
+    /// populated member of its containing oneof, or the default value otherwise.
+    @_alwaysEmitIntoClient @inline(__always)
+    public func value(at offset: Int, default defaultValue: Bool = false, oneofPresence: OneofPresence) -> Bool {
```
thomasvl, line 522: Does anything make use of these defaults for `defaultValue`?  Is it worth having second copies of these without the argument instead that inline the value?  i.e. - would it make slightly smaller code with the defaults are zero/false/empty?

Sources/protoc-gen-swift/OneofGenerator.swift
```diff
@@ -197,93 +201,65 @@ class OneofGenerator {
...
-    private func gerenateOneofEnumProperty(printer p: inout CodePrinter) {
+    private func gerenateOneofEnumProperty(printer p: inout CodePrinter, oneofOffset: UInt16) {
         let visibility = generatorOptions.visibilitySourceSnippet
-        p.print()
         p.print(
-            "\(comments)\(visibility)var \(swiftFieldName): \(swiftFullName)? = nil"
+            "",
+            "\(comments)\(visibility)var \(swiftFieldName): \(swiftFullName)? {"
         )
+        // TODO: Investigate whether we need to split any of these switches up for very large oneofs
```
thomasvl, line 216: Since we're switching on a UInt32, I thought this wasn't going to be a problem?  Or I guess it only would be for the setter?  Yea, that's likely to be a problem then. :(

Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
```diff
@@ -575,38 +575,47 @@ struct SwiftProtoTesting_TestAllTypes: @unchecked Sendable {
...
+      case 111: return .oneofUint32(oneofUint32)
+      case 112: return .oneofNestedMessage(oneofNestedMessage)
+      case 113: return .oneofString(oneofString)
+      case 114: return .oneofBytes(oneofBytes)
+      default: preconditionFailure("Internal logic error; populated oneof field \(populatedField) is not a member of this oneof")
+      }
     }
-    set {oneofField = .oneofUint32(newValue)}
+    set {
+      switch newValue {
```
thomasvl, line 591: Could you store the raw field number and then switch on that instead?  I guess that would be a little more code since you still need to extract from the new value after the fact.

Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_TableDriven.swift
```diff
@@ -137,4 +137,149 @@ final class Test_TableDriven: XCTestCase {
         XCTAssertEqual(msg.repeatedImportMessage[0].d, 99)
         XCTAssertEqual(msg.repeatedImportMessage[1].d, 20)
     }
+
+    func testOneofModifyMembers() {
```
thomasvl, line 149: Shouldn't you also test oneofs work through a copy and copyonwrite?

Sources/protoc-gen-swift/OneofGenerator.swift
allevato: Right, the setter was mainly what I had in mind here, since this will involve the exhaustiveness checker. The previous implementation didn't because each accessor only checked a single case using `if case`.

Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
allevato: I'm afraid I'm not following; can you be more specific about what you had in mind? Since the incoming value is an instance of the enum, we can't avoid switching over that.

allevato: Chatted offline; that would mean moving the `switch` into the field number accessor, which doesn't really save much.

Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_TableDriven.swift
allevato: Done.

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: For regular fields, this is controlled by `hasFieldPresence`, which we don't really have for oneof members. Instead I updated the generator to check whether the field has an explicit default value specified, and if not, to just omit the argument and use the default encoded into the `value` function.

thomasvl: ack - I wasn't sure if having two versions with `_MessageStorage` (with and without default) and hardcoding the values made for any smaller binary sizes since there's one less argument to pass.

allevato: I think these things are definitely worth measuring once we get the fundamentals in place, since they're simple generator changes. But right now I want to focus on getting the core functionality implemented.

PR #1871, opened 2025-10-16
[table-driven-branch] Fix some copy/deinit bugs.

allevato:
For copying, we were computing the size of the bitwise-copied range incorrectly. We were tracking the offset of the first non-trivial field we found, but this isn't correct because we iterate over them in *field number* order, not based on their *storage kind*. Instead, we need to track the *lowest* offset. It may be worth it to store this in the layout after all so that we don't have to recompute it each time.

For both copy and deinit, I had to fill in the missing oneof support. It's surprising that the tests passed without this, but I think that was by coincidence with the bug above, since it was causing us to bitwise-copy *more* data than we otherwise would have, and things just happened to line up correctly.

### DIFF
--- a/Sources/SwiftProtobuf/_MessageStorage.swift
+++ b/Sources/SwiftProtobuf/_MessageStorage.swift
@@ -229,10 +229,13 @@ extension _MessageStorage {
             }
         }
 
-        // Copy all of the trivial values (including has-bits) in bitwise fashion.
-        if firstNontrivialStorageOffset != 0 {
-            destination.buffer.copyMemory(from: .init(rebasing: buffer[..<firstNontrivialStorageOffset]))
-        }
+        // Copy all of the trivial field values, has-bits, and any oneof tracking in bitwise
+        // fashion.
+        destination.buffer.copyMemory(from: .init(rebasing: buffer[..<firstNontrivialStorageOffset]))
+
+        destination.unknownFields = unknownFields
+        // TODO: Handle extension fields.
+
         return destination
     }
 


Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -235,22 +239,14 @@ extension _MessageStorage {
     /// Copy-initializes the field associated with the given layout information in the destination
     /// storage using its value from this storage.
     public func copyField<T>(_ field: FieldLayout, to destination: _MessageStorage, type: T.Type) {
-        switch field.presence {
-        case .oneOfMember:
-            // TODO: Support oneof fields.
-            break
+        guard isPresent(field) else { return }
```
thomasvl, line 245: Worth an any asset to confirm we only call this on arrays or non pod fields?

Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -220,12 +229,7 @@ extension _MessageStorage {
...
         }
 
-        // Copy all of the trivial values (including has-bits) in bitwise fashion. Note that if we
-        // never set `firstNontrivialStorageOffset`, then the whole message must have been trivial
-        // fields. (This also works if the message is empty; that is, zero bytes.)
-        if firstNontrivialStorageOffset == 0 {
-            firstNontrivialStorageOffset = layout.size
-        }
+        // Copy all of the trivial values (including has-bits) in bitwise fashion.
         if firstNontrivialStorageOffset != 0 {
```
thomasvl: Thinking some more about this, is it even possible for this to be zero now?  Seems like that could only happen with an empty message.

- If there's only complex fields (messages, strings, repeated), won't there still be has bits?
- If all the fields are in a oneof, no matter what the type of fields, won't there still be the storage for the "active" oneof item?

So maybe skip the gate since we don't need to optimize that case (before I was thinking all complex would skip this, but I forgot about the hasbits).

Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -220,12 +229,7 @@ extension _MessageStorage {
             }
         }
 
-        // Copy all of the trivial values (including has-bits) in bitwise fashion. Note that if we
-        // never set `firstNontrivialStorageOffset`, then the whole message must have been trivial
-        // fields. (This also works if the message is empty; that is, zero bytes.)
-        if firstNontrivialStorageOffset == 0 {
-            firstNontrivialStorageOffset = layout.size
-        }
+        // Copy all of the trivial values (including has-bits) in bitwise fashion.
```
thomasvl: I keep reading this as "all the trivial [field] values", so maybe just call it out:

> // Copy all of the trivial field values, has-bits, and any oneof selection tracking in bitwise fashion.

thomasvl (approved):
Not sure if you want to build out test cases - setting a single field of each type per copy might have shaken out the bugs.

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: Done.

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: Yeah, this would only happen for empty messages, so I've dropped the comparison since it's the very statistically unlikely one.

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: This is called on all types of fields. Even for POD fields that don't have explicit presence, we use has-bits because they're cheap and avoid having different code paths.

thomasvl: Chatted offline, this was about adding a check as the first line of copyFields, not about isPresent. Tony's going to consider options for adding assertions/invariant checks in some way.

PR #1872, opened 2025-10-17
[table-driven-branch] Implement message equality.

allevato:
This is stacked on top of https://github.com/apple/swift-protobuf/pull/1871.

### DIFF
--- a/Sources/SwiftProtobuf/_MessageStorage.swift
+++ b/Sources/SwiftProtobuf/_MessageStorage.swift
@@ -735,6 +735,15 @@ extension _MessageStorage {
     /// default defined to be 100.
     @inline(never)
     public func isEqual(to other: _MessageStorage) -> Bool {
+        if self === other {
+            /// Identical message storage means they must be equal.
+            return true
+        }
+        // TODO: If we store the offset of the first non-trivial field in the layout, we can make
+        // this extremely fast by doing the memcmp up front and failing fast. Likewise, we can also
+        // avoid the loop entirely if the message contains only trivial fields. We could see similar
+        // performance wins for copy and deinit.
+
         // Loops through the fields, checking equality of any that are non-trivial types. We ignore
         // the trivial ones here, instead tracking the byte offset of the first non-trivial field
         // so that we can bitwise-compare those as a block afterward.
@@ -841,14 +850,10 @@ extension _MessageStorage {
     ) -> Bool {
         let isSelfPresent = isPresent(field)
         let isOtherPresent = other.isPresent(field)
-        if !isSelfPresent && !isOtherPresent {
-            // If the field is not present in both messages, they are equal.
-            return true
-        }
-        if isSelfPresent != isOtherPresent {
-            // If the field presence is different between the two messages, they are unequal (even
-            // if their effective values would be the same).
-            return false
+        guard isSelfPresent && isOtherPresent else {
+            // If the field isn't present in both messages, then it must be absent in both to be
+            // considered equal.
+            return isSelfPresent == isOtherPresent
         }
         // The field is present in both messages, so compare their values.
         let selfPointer = (buffer.baseAddress! + field.offset).bindMemory(to: T.self, capacity: 1)
--- a/Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_TableDriven.swift
+++ b/Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_TableDriven.swift
@@ -297,7 +297,7 @@ final class Test_TableDriven: XCTestCase {
             $0.repeatedInt64 = [1, 2, 3]
             $0.oneofString = "hello"
         }
-        // XCTAssertTrue(lhs == lhs)
+        XCTAssertTrue(lhs == lhs)
 
         let rhs = MessageTestType.with {
             $0.optionalInt32 = 50


Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -719,6 +719,141 @@ extension _MessageStorage {
...
+    ///
+    /// Message equality in SwiftProtobuf includes presence. That is, a message with an integer
+    /// field set to 100 is not considered equal to one where that field is not present but has a
+    /// default defined to be 100.
+    @inline(never)
+    public func isEqual(to other: _MessageStorage) -> Bool {
+        // Loops through the fields, checking equality of any that are non-trivial types. We ignore
+        // the trivial ones here, instead tracking the byte offset of the first non-trivial field
+        // so that we can bitwise-compare those as a block afterward.
+        var firstNontrivialStorageOffset = layout.size
```
thomasvl, line 750: If we know the sizeof the hasbits (and oneof storage), a memcmp of that up from would be a way faster fail fast as it would be way faster.  I guess this just adds to maybe we should store the offset of that data and the primative fields.

ps - I guess it would also let us skip the loop in copy/dealloc if the offset == storage size.

Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -719,6 +719,141 @@ extension _MessageStorage {
...
+    /// Returns whether the given field in the receiver is equal to the same field in the other
+    /// storage, given the expected type of that field.
+    public func isField<T: Equatable>(
+        _ field: FieldLayout,
+        equalToSameFieldIn other: _MessageStorage,
+        type: T.Type
+    ) -> Bool {
+        let isSelfPresent = isPresent(field)
+        let isOtherPresent = other.isPresent(field)
+        if !isSelfPresent && !isOtherPresent {
```
thomasvl: This might be too cute/magic (assuming I got it right), but it streamlines the both set cases.

```
// If they both aren't set, then they are only equal if they both aren't set.
guard isSelfPresent && isOtherPresent else {
  return isSelfPresent == isOtherPresent
}
```

Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_TableDriven.swift
```diff
@@ -290,4 +290,27 @@ final class Test_TableDriven: XCTestCase {
...
         XCTAssertEqual(msg.oneofNestedMessage.bb, 0)
     }
+
+    func testEquality() {
+        let lhs = MessageTestType.with {
+            $0.optionalInt32 = 50
+            $0.repeatedInt64 = [1, 2, 3]
+            $0.oneofString = "hello"
+        }
+        // XCTAssertTrue(lhs == lhs)
```
thomasvl: Did you want to uncomment this?

Actually, should this fast track any place?

Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
```diff
@@ -4637,7 +4684,7 @@ extension SwiftProtoTesting_TestRequiredMessage: SwiftProtobuf.Message, SwiftPro
   }
 
   static func ==(lhs: SwiftProtoTesting_TestRequiredMessage, rhs: SwiftProtoTesting_TestRequiredMessage) -> Bool {
-    fatalError("table-driven == not yet implemented")
+    return lhs._storage.isEqual(to: rhs._storage)
```
thomasvl, line 4687: Actually, can this move fully into the runtime or do we need to generate it into every message since it's now the same?

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: I think that's reasonable, along with a comment to make it seem less magical.

Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_TableDriven.swift
allevato: Oops, uncommented.

And yeah, I just added a check for `self === other` in `_MessageStorage.isEqual(to:)` to short circuit this case. Funny, after the recent looooooong discussion about the `is[Trivially]Identical(to:)` API on Swift Evolution, I should have had that in the forefront of my mind.

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: I've added a TODO to track updating the layout with this information because you're right, this would be a good improvement for many messages.

Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
allevato: We can't move it into the runtime as a protocol extension implementation without exposing `_storage` as a protocol requirement, which I'd like to avoid since it has a pretty large `public` API surface (even guarded by SPI).

We could have the compiler synthesize it if we made `_MessageStorage` conform to `Equatable`, but that wouldn't change anything in terms of compiled code size—it would just be the compiler generating it for each message instead of us.

thomasvl: ack.

PR #1874, opened 2025-10-21
Support `isInitialized`.

allevato:
This PR implements `isInitializedShallow` on the storage class, which is the fast-path that just checks the has-bits using a bitwise-compare. The public `isInitialized` is built on top of that—it does the shallow check first, and if that succeeds, then it walks the layout and checks the submessage fields recursively.

I'm growing a bit weary of how many of these "trampoline" functions I've had to introduce. I think the `isInitialized` one could be redesigned to be something that just returns a layout and a storage, but some of the others need to be able to propagate explicit type hints through the API layers. I'd like to find a way to generalize them that doesn't rely on existentials/too many generics so that we don't risk designing something that either doesn't work on embedded or gets code bloat from too many explicit specializations, but I haven't come up with anything yet. Maybe once I've implemented the full spread of what we need, an idea will come to mind.

### DIFF
--- a/Sources/protoc-gen-swift/MessageGenerator.swift
+++ b/Sources/protoc-gen-swift/MessageGenerator.swift
@@ -440,7 +440,7 @@ class MessageGenerator {
     private func generateIsInitialized(printer p: inout CodePrinter) {
         // If the message does not have any fields that need this check, don't generate it; we'll
         // use the protocol extension default.
-        guard messageLayoutCalculator.needsIsInitializedGeneration else {
+        guard descriptor.containsRequiredFields() else {
             return
         }
 
--- a/Sources/protoc-gen-swift/MessageLayoutCalculator.swift
+++ b/Sources/protoc-gen-swift/MessageLayoutCalculator.swift
@@ -41,9 +41,6 @@ struct MessageLayoutCalculator {
         submessageCollector.usedSubmessages.sorted { $0.value.index < $1.value.index }.map { $0.value }
     }
 
-    /// Indicates whether we need to generate an `isInitialized` property for this message.
-    private(set) var needsIsInitializedGeneration = false
-
     /// Creates a new message layout calculator for a message containing the given fields and for
     /// a platform with the given pointer bit-width.
     init(fieldsSortedByNumber: [any FieldGenerator]) {
@@ -73,9 +70,6 @@ struct MessageLayoutCalculator {
                 // The presence is just the has-bit index.
                 field.presence = .hasBit(hasBitIndex)
                 hasBitIndex += 1
-            }
-            if !needsIsInitializedGeneration && field.needsIsInitializedGeneration {
-                needsIsInitializedGeneration = true
             }
             if field.isRequired {
                 requiredCount += 1
--- a/Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
+++ b/Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
@@ -4394,6 +4394,10 @@ extension SwiftProtoTesting_TestAllExtensions: SwiftProtobuf.Message, SwiftProto
 
   private static let _protobuf_messageLayout = SwiftProtobuf._MessageLayout(layout: _protobuf_messageLayoutString)
 
+  public var isInitialized: Bool {
+    return _storage.isInitialized
+  }
+
   mutating func decodeMessage<D: SwiftProtobuf.Decoder>(decoder: inout D) throws {
     fatalError("table-driven decodeMessage not yet implemented")
   }
@@ -4825,6 +4829,10 @@ extension SwiftProtoTesting_TestFieldOrderings: SwiftProtobuf.Message, SwiftProt
     return true
   }
 
+  public var isInitialized: Bool {
+    return _storage.isInitialized
+  }
+
   mutating func decodeMessage<D: SwiftProtobuf.Decoder>(decoder: inout D) throws {
     fatalError("table-driven decodeMessage not yet implemented")
   }
@@ -5140,6 +5148,10 @@ extension SwiftProtoTesting_TestPackedExtensions: SwiftProtobuf.Message, SwiftPr
   #endif
 
   private static let _protobuf_messageLayout = SwiftProtobuf._MessageLayout(layout: _protobuf_messageLayoutString)
+
+  public var isInitialized: Bool {
+    return _storage.isInitialized
+  }
 
   mutating func decodeMessage<D: SwiftProtobuf.Decoder>(decoder: inout D) throws {
     fatalError("table-driven decodeMessage not yet implemented")


Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -862,6 +862,80 @@ extension _MessageStorage {
...
+    /// state.
+    @inline(never)
+    private var isMessageInitializedShallow: Bool {
+        // A message with no required fields is trivially considered initialized.
+        guard layout.requiredCount > 0 else { return true }
+
+        // The has-bits for the required fields have been ordered first in storage, so we can
+        // quickly determine whether a message is initialzed using a simple `memcmp` (with at most
+        // one additional masked byte comparison for overflow bits).
+        let requiredByteCount = layout.requiredCount / 8
```
thomasvl, line 880: Is it worth worrying about when this is zero?  i.e. - fastpath?

Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -862,6 +862,80 @@ extension _MessageStorage {
...
+                )
+                guard isSubmessageInitialized else { return false }
+
+            default:
+                // Nothing to do for other types of fields; they've already been considered by the
+                // shallow check.
+                break
+            }
+        }
+        return true
```
thomasvl, line 924: What about extension fields?

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: Done. No reason to allocate zero bytes and call `memcmp` on it.

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: Added a TODO for extension fields. (I could call through to the object I'm nesting right now but I plan to look over extensions more holistically later.)

allevato:
> So generally this will result in more work than needed.
>
> If a message has no required fields, and it has no fields of type message and no extension ranges, there's no need to generate.
>
> The old code didn't recurse this but relied on calling isinitialized, but given you have to generate the trampolines, it might be worth doing the recursive eval on message used as field types - i.e. - if they don't have required fields or extension or have subfields with it, don't put it in the trampoline and shrink things some.
> 
> Likewise if you do this recursive eval, you could just generate the toplevel return True and never call through to the layout objects in the first place.

Goods points. I've updated message layout calculation to track more information about submessages—right now, whether they need an `isInitialized` check. If some submessage doesn't, I return true instead of calling back into the storage. If _no_ submessages need the check, I just generate a single `return true` for the whole trampoline function.

Lastly, if a message itself doesn't have any required fields (recursively), I drop the `isInitialized` generation completely and fall back to the protocol default.

Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -36,6 +36,10 @@ class MessageFieldGenerator: FieldGeneratorBase, FieldGenerator {
 
     var oneofIndex: Int? { nil }
 
+    var needsIsInitializedGeneration: Bool {
+        isRequired || (isGroupOrMessage && fieldDescriptor.messageType!.containsRequiredFields())
```
thomasvl, line 40: Check if the message in the field contains extension ranges, if yes you'll have to recurse through it.

Actually, the better thing might be to stick a method on Descriptor that is "needsIsInitialized" and that checks for a required field, extension ranges, and then recurses calling the same thing on all the message/group fields. That would be the exhaustive case and handle Mesage A has field of Mesage B has field of Mesage C and C has a required filed.  That means A, B, and C all need to generate isInitialized.

Sources/protoc-gen-swift/OneofGenerator.swift
```diff
@@ -33,6 +33,11 @@ class OneofGenerator {
 
         var presence: FieldPresence = .oneofMember(0)
 
+        var needsIsInitializedGeneration: Bool {
+            // oneof fields can't be required, so we only have to consider the submessage case.
+            isGroupOrMessage && fieldDescriptor.messageType!.containsRequiredFields()
```
thomasvl, line 38: Same, check if the message has extension ranges.

Sources/protoc-gen-swift/MessageFieldGenerator.swift
thomasvl: And ignore me, this is what containsRequiredFields already done.  :)

Sources/protoc-gen-swift/MessageLayoutCalculator.swift
```diff
@@ -37,10 +37,13 @@ struct MessageLayoutCalculator {
...
     /// The first element in this array corresponds to the submessage with index 1, and the rest
     /// increase accordingly.
-    var submessageNames: [String] {
-        submessages.usedSubmessages.sorted { $0.value < $1.value }.map { $0.key }
+    var submessages: [SubmessageInfo] {
+        submessageCollector.usedSubmessages.sorted { $0.value.index < $1.value.index }.map { $0.value }
     }
 
+    /// Indicates whether we need to generate an `isInitialized` property for this message.
+    private(set) var needsIsInitializedGeneration = false
```
thomasvl: I guess the one thing missing is feeding in if there are extension ranges.  While the layout code is a TODO, we also need to factor it into if we need to generate the override.

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -419,6 +438,12 @@ class MessageGenerator {
     ///
     /// - Parameter printer: The code printer.
     private func generateIsInitialized(printer p: inout CodePrinter) {
+        // If the message does not have any fields that need this check, don't generate it; we'll
+        // use the protocol extension default.
+        guard messageLayoutCalculator.needsIsInitializedGeneration else {
```
thomasvl: Actually, could this just use containsRequiredFields to check if it need to generate it?

allevato: Hah, yes, that's simpler. And after making that change, the diff showed that it picked up generation for a few messages that I was missing—messages with no required fields but with extensions.

Sources/protoc-gen-swift/MessageLayoutCalculator.swift
allevato: I deleted this since it was only used by `MessageGenerator` and that was replaced with just `containsRequiredFields`.

Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest.pb.swift
```diff
@@ -5401,8 +5386,8 @@ extension SwiftProtoTesting_TestParsingMerge.RepeatedFieldsGenerator.Group2: Swi
     }
   }
 
-  public var isInitialized: Bool {
-    fatalError("table-driven isInitialized not yet implemented")
+  private static func _protobuf_isSubmessageInitialized(for token: SwiftProtobuf._MessageLayout.SubmessageToken, field: SwiftProtobuf.FieldLayout, storage: SwiftProtobuf._MessageStorage) -> Bool {
```
thomasvl, line 5389: I wonder if it is worth doing a initializer with out this argument and defaulting the impl for this to `return true`. Then things just just have extensions don't actually have to provide this shim and a single common one could be used for all messages with extension support but no required fields and no subfields that need recursion through.

Maybe just something to consider in the future.

allevato: Yeah, that's a possibility. I'm hoping I can find a way to streamline all of these trampolines into something common though, which might simplify everything even further.

PR #1876, opened 2025-10-24
Build `protoc` from source

FranzBusch:
## Motivation

In a recent addition, we included changes to provide `protoc` as a binary target from the official releases from the upstream `protobuf` repository. While this solved one of the largest issues in our ecosystem where folks currently needed to install `protoc` through a system package manager, it also introduced another set of problems that binary targets bring with them.

## Modifications

This PR adds `protobuf` and `abseil-cpp` as a submodule to this project and includes a new target that builds both together from source. It includes the minimal set of files from both to successfully build `protoc` and also passes the right compiler and linker flags to build those warning free.

## Result

We are no longer using a binary target for `protoc` but instead build it from source which alleviates the problems that we introduced with the binary target.

### DIFF
--- a/.github/workflows/update_protobuf.yml
+++ b/.github/workflows/update_protobuf.yml
@@ -1,0 +1,119 @@
+name: Update Protobuf and Abseil Submodules
+
+permissions:
+  contents: write
+  pull-requests: write
+
+on:
+  schedule:
+    # Every night at 2am UTC
+    - cron: '0 2 * * *'
+  workflow_dispatch:
+
+jobs:
+  check-and-update:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+        submodules: true
+        token: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Check for new protobuf release
+      id: check_release
+      run: |
+        # Get the latest protobuf release tag
+        LATEST_TAG=$(curl -s https://api.github.com/repos/protocolbuffers/protobuf/releases/latest | jq -r '.tag_name')
+        echo "Latest protobuf release: $LATEST_TAG"
+
+        # Get current protobuf submodule commit
+        cd Sources/protobuf/protobuf
+        CURRENT_TAG=$(git describe --tags --exact-match 2>/dev/null || echo "none")
+        echo "Current protobuf version: $CURRENT_TAG"
+        cd ../../..
+
+        # Check if update is needed
+        if [ "$LATEST_TAG" != "$CURRENT_TAG" ] && [ "$LATEST_TAG" != "null" ]; then
+          echo "needs_update=true" >> $GITHUB_OUTPUT
+          echo "latest_tag=$LATEST_TAG" >> $GITHUB_OUTPUT
+          echo "Update needed: $CURRENT_TAG -> $LATEST_TAG"
+        else
+          echo "needs_update=false" >> $GITHUB_OUTPUT
+          echo "No update needed"
+        fi
+
+    - name: Update protobuf submodule
+      if: steps.check_release.outputs.needs_update == 'true'
+      run: |
+        cd Sources/protobuf/protobuf
+        git fetch --tags
+        git checkout ${{ steps.check_release.outputs.latest_tag }}
+        cd ../../..
+
+    - name: Determine required abseil version
+      if: steps.check_release.outputs.needs_update == 'true'
+      id: check_abseil
+      run: |
+        # Extract abseil commit from protobuf_deps.bzl
+        ABSEIL_COMMIT=$(grep -A 10 '"com_google_absl"' Sources/protobuf/protobuf/protobuf_deps.bzl | grep 'commit' | head -1 | sed 's/.*"\([a-f0-9]*\)".*/\1/')
+
+        if [ -z "$ABSEIL_COMMIT" ]; then
+          echo "Error: Could not determine abseil commit"
+          exit 1
+        fi
+
+        echo "Required abseil commit: $ABSEIL_COMMIT"
+        echo "abseil_commit=$ABSEIL_COMMIT" >> $GITHUB_OUTPUT
+
+    - name: Update abseil submodule
+      if: steps.check_release.outputs.needs_update == 'true'
+      run: |
+        cd Sources/protobuf/abseil
+        git fetch
+        git checkout ${{ steps.check_abseil.outputs.abseil_commit }}
+        cd ../../..
+
+    - name: Stage submodule updates
+      if: steps.check_release.outputs.needs_update == 'true'
+      run: |
+        git add Sources/protobuf/protobuf
+        git add Sources/protobuf/abseil
+
+    - name: Get abseil version info
+      if: steps.check_release.outputs.needs_update == 'true'
+      id: abseil_info
+      run: |
+        cd Sources/protobuf/abseil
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -59,6 +54,464 @@ let package = Package(
...
+                "protobuf/src/google/protobuf/compiler/cpp/field_generators/string_view_field.cc",
+                "protobuf/src/google/protobuf/compiler/cpp/file.cc",
+                "protobuf/src/google/protobuf/compiler/cpp/generator.cc",
+                "protobuf/src/google/protobuf/compiler/cpp/helpers.cc",
+                "protobuf/src/google/protobuf/compiler/cpp/message.cc",
+                "protobuf/src/google/protobuf/compiler/cpp/message_layout_helper.cc",
+                "protobuf/src/google/protobuf/compiler/cpp/parse_function_generator.cc",
+                "protobuf/src/google/protobuf/compiler/cpp/service.cc",
+                "protobuf/src/google/protobuf/compiler/cpp/tracker.cc",
+                "protobuf/src/google/protobuf/compiler/csharp/csharp_doc_comment.cc",
```
allevato: Can the csharp and java files be moved down to relevant subsections?

In fact, do we need the other language generators at all? IIRC there's a build flag that should do a minimal protoc that doesn't include any of the other language plugins. If we're saying that this `protoc` build should only be considered as an implementation detail of our plugin and not a general purpose distribution, then that seems fine—I don't think we'd want to claim that we're maintaining a full compiler anyway.

FranzBusch: Do you know the build setting by any chance? Right now the symbols are required. I just added files that were needed due to symbol availability.

allevato: I misremembered, it's not a build flag. It's if you build with [main_no_generators.cc](http://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/compiler/main_no_generators.cc) instead of `main.cc`, the former doesn't pull in any of the other plugins.

FranzBusch: Thanks. I was able to remove all language specific plugins this way.

allevato:
Can you update the README with any new information for end users? For example, adding `--recurse-submodules` to their `git clone` command or `git submodule init/update` for folks who already have a checkout?

If another package depends on swift-protobuf (whether only using the runtime or using the runtime + plugin), I assume SwiftPM does the right thing when it clones the dependency to ensure that submodules get pulled in?

Likewise, should we update our [RELEASING.md](https://github.com/apple/swift-protobuf/blob/main/Documentation/RELEASING.md) with instructions about how to bump the submodules when needed?

Otherwise, this seems like a really positive direction, but I'd get @thomasvl's feedback on it too.

.gitmodules
```diff
@@ -0,0 +1,8 @@
+[submodule "Sources/protobuf/protobuf"]
+	path = Sources/protobuf/protobuf
+	url = https://github.com/protocolbuffers/protobuf
+	branch = main
```
thomasvl, line 4: While I like tracking form a breakage pov, but we likely need releases to be pinned or a release could break when it gets a new protobuf that needs changes to still build.

Maybe the right process is document how/when to update both of these but then also create a weekly cron CI step that updates both branches and tries building just to track when main changes in a way that would be a problem.  Just so we get early warning when something will need to be done.

.github/workflows/build.yml
```diff
@@ -46,6 +46,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         path: swift-protobuf
+        submodules: true
```
thomasvl, line 43: Aren't we checking out protobuf twice now? (and building it twice)

thomasvl (commented):
Should the Makefile get updated to use the protobuf that will be checked out?

.gitmodules
FranzBusch: The submodules are pointing to the v33.0 release and the abseil release used by that.

FranzBusch:
@thomasvl I addressed your comments. Let me know what you think about the changes to the CI and the `RELEASING.md`

Makefile
allevato: Could we add building our copy of `protoc` using `swift build` as another Make dependency of the plugin, and use that copy of `protoc` for tasks like regenerating the protos? Basically anywhere that `protoc` is used today by this file.

It would be nice to also still allow `protoc` to be overridden, just in case, but the default behavior could be to build this one.

Documentation/RELEASING.md
```diff
@@ -69,41 +69,68 @@ When doing a Swift Protobuf library release:
...
-     - Download protoc binaries for all supported platforms
-     - Create a Swift Package Manager compatible artifact bundle
-     - Create a new draft release tagged `protoc-vX.Y`
-     - Upload the artifactbundle to the draft release
-   - If versions match, exit early (no action needed)
+   Visit the [protobuf releases page](https://github.com/protocolbuffers/protobuf/releases)
+   to check if a new version is available.
 
-1. **Publish the release**
+2. **Update the protobuf submodule**
```
thomasvl: fyi - you can level these all as `1.` and when rendered they will be correct, that way edits never have to renumber.

.github/workflows/build.yml
```diff
@@ -34,59 +34,22 @@ jobs:
...
-          ;;
-        esac
+      run: apt-get update && apt-get install -y make g++ cmake
     - name: Build
-      working-directory: swift-protobuf
       run: make build ${{ matrix.swift.hook }}
     - name: Test runtime
-      working-directory: swift-protobuf
       run: make test-runtime ${{ matrix.swift.hook }}
-    - name: Cache protobuf
```
thomasvl, line 72: I guess if builds get too slow we can bring back the protobuf caching.

thomasvl:
If possible, it would be nice to do a follow up with a cron CI job that checks regularly to see if there is a new protobuf release and would nudge us to update. Generally I worry about the maintenance on this if it isn't kept up to date.

thomasvl:
@FranzBusch assuming you all are good having to keep this updated, this seems about ready to go.  I'd suggest that cron based CI to give yourselves some warnings about when that work would be needed.

I think you'll also have to update the required Checks on github as that might still block this being submitted.

Makefile
FranzBusch: I pushed a commit that does that.

FranzBusch:
> @FranzBusch assuming you all are good having to keep this updated, this seems about ready to go. I'd suggest that cron based CI to give yourselves some warnings about when that work would be needed.

I added a CI workflow that updates to the latest release and opens a PR. I haven't tested this yet so let's see how it works once we merge it.

.github/workflows/update_protobuf.yml
```diff
@@ -0,0 +1,119 @@
...
+      uses: actions/checkout@v4
+      with:
+        submodules: true
+        token: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Check for new protobuf release
+      id: check_release
+      run: |
+        # Get the latest protobuf release tag
+        LATEST_TAG=$(curl -s https://api.github.com/repos/protocolbuffers/protobuf/releases/latest | jq -r '.tag_name')
```
thomasvl, line 27: Looking at https://github.com/protocolbuffers/protobuf/tags – Is "latest" going to be safe?  Should we filter away anything with "-rc" in the name?  Should we try to skip any language specific ones?

FranzBusch: The latest is applied to `releases` not `tags` and from what I see is that the Protobuf team is marking all `-rc` as `pre-release` and not `latest`

PR #1877, opened 2025-10-27
[table-driven-branch] Create a general trampoline function.

allevato:
Some of the operations we're going to implement, like `serializedBytesSize` and `serializedBytes`, really just need
a way to invoke an arbitrary operation on the underlying storage of a submessage.

`isInitialized` can also be implemented on top of this. It means that we don't short-circuit as quickly as we could, but I don't think the minor performance boost of skipping a dispatch is worth the complexity of having another trampoline function.

### DIFF
--- a/Sources/SwiftProtobuf/Message.swift
+++ b/Sources/SwiftProtobuf/Message.swift
@@ -180,7 +180,9 @@ extension Message {
 /// or uses them as `Dictionary` keys.
 @preconcurrency
 public protocol _MessageImplementationBase: Message, Hashable {
-    // This is an implementation detail of the runtime.
+    // This is an implementation detail of the runtime; users should not call it. The return type
+    // is a class-bound existential because the true SPI type cannot be used in a protocol
+    // requirement.
     func _protobuf_messageStorage(accessToken: _MessageStorageToken) -> AnyObject
 
     // Legacy function; no longer used, but left to maintain source compatibility.


Sources/SwiftProtobuf/Message.swift
```diff
@@ -180,12 +180,20 @@ extension Message {
 /// or uses them as `Dictionary` keys.
 @preconcurrency
 public protocol _MessageImplementationBase: Message, Hashable {
+    // This is an implementation detail of the runtime.
+    func _protobuf_messageStorage(accessToken: _MessageStorageToken) -> AnyObject
```
thomasvl, line 186: why bother with the AnyObject here vs. just directly use the correct type since it is know? Or is that just because of the spi usage for the import?

Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest_swift_required_fields.pb.swift
```diff
@@ -957,10 +971,15 @@ extension SwiftProtoTesting_NoneRequired: SwiftProtobuf.Message, SwiftProtobuf._
     }
   }
 
-  private static func _protobuf_isSubmessageInitialized(for token: SwiftProtobuf._MessageLayout.SubmessageToken, field: SwiftProtobuf.FieldLayout, storage: SwiftProtobuf._MessageStorage) -> Bool {
-    return true
+  private static func _protobuf_performOnSubmessageStorage(for token: SwiftProtobuf._MessageLayout.SubmessageToken, field: SwiftProtobuf.FieldLayout, storage: SwiftProtobuf._MessageStorage, perform: (SwiftProtobuf._MessageStorage) throws -> Bool) throws -> Bool {
+    switch token.index {
```
thomasvl, line 975: aside - does it ever make sense to move these to an `if/else` when there is a single case?  I guess that comes when looking at final code gen.

Sources/SwiftProtobuf/Message.swift
thomasvl: i.e. - maybe document the reasoning?

allevato: Fleshed out the comment. Yeah, the reason is because an SPI type can't be used in a requirement of a protocol that's not SPI. Embedded supports class-bound existentials, so we can do this safely and just do the casting inside the runtime.

Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest_swift_required_fields.pb.swift
allevato: I would expect the generated object code to be the same in either case, so it's probably not worth complicating the generator.

Sources/SwiftProtobuf/Message.swift
allevato: I did try making the type itself non-SPI and only apply SPI to the methods themselves, but that causes more fallout that isn't worth it (limitations around how things get accessed/inlined).

PR #1879, opened 2025-10-27
Optimize the CI setup some now that `protoc` is built by SwiftPM

thomasvl:
- Don't override the `PROTOC` value when running tests, let the in tree one be used.
- For conformance tests, only build the runner in the protobuf checkout (no need to force a `protoc` build also)
- Move the protobuf conformance runner build and the conformance tests to be after all the other steps, might as well let everything else finish first to catch issues before waiting on protobuf to build.


thomasvl:
And we can't do this since the `protoc` target isn't in the `Package@swift-5.10.swift` file.  Guess this will have to wait for 5.10 to be dropped which is also dependent on oss-fuzz also being updated.

PR #1880, opened 2025-10-28
[table-driven-branch] Implement binary serialization.

allevato:
This PR implements `serializedBytesSize()` and `serializedBytes()` on `_MessageStorage`. Eventually, we'll be able to put the call to the latter completely in the runtime (in `_MessageImplementationBase`), but for now I'm generating the method directly because we still need the old WKTs and plugin protos to work until we migrate those to be table-driven.

I've also fixed a bug in the earlier `isInitialized` logic; we were checking *any* message/group field to see if it was initialized,
including optional message/group fields. Fixed it to allow unset optional messages/groups and only recurse into them when they're present.

I've brought over the subset of the `TestAllTypes` encoding tests that are able to pass today. This should cover scalar and repeated fields (packed and unpacked), submessages/groups, and oneofs. I also brought over the subset of required field encoding tests to verify the expected throwing and partial behavior.

Lastly, I've created separate source files to hold the `+BinarySize` and `+BinaryEncoding` extensions on `_MessageStorage`. I'll go back and do the same for some of the other large chunks of `_MessageStorage` so that that class is easier to navigate.

### DIFF
--- a/Sources/SwiftProtobuf/FieldTag.swift
+++ b/Sources/SwiftProtobuf/FieldTag.swift
@@ -53,7 +53,7 @@ package struct FieldTag: RawRepresentable {
     /// Note that this size is always independent of the wire format because the low three bits do
     /// not change the size of a varint.
     package static func encodedSize(ofTagWithFieldNumber number: Int) -> Int {
-        return Varint.encodedSize(of: UInt32(truncatingIfNeeded: number) << 3)
+        return Varint.encodedSize(of: UInt32(truncatingIfNeeded: number) &<< 3)
     }
 
     /// Creates a new tag from its raw numeric representation.
--- a/Sources/SwiftProtobuf/_MessageStorage+BinarySize.swift
+++ b/Sources/SwiftProtobuf/_MessageStorage+BinarySize.swift
@@ -70,9 +70,9 @@ extension _MessageStorage {
             case .float:
                 return fixedWidthRepeatedFieldSize(for: fieldNumber, at: offset, isPacked: isPacked, as: Float.self)
 
-            case .group, .message:
+            case .group:
                 precondition(!isPacked, "a packed message/group field should not be reachable")
-                return serializedByteSize(ofMessageOrGroupField: field, fieldNumber: fieldNumber)
+                return serializedByteSize(ofGroupField: field, fieldNumber: fieldNumber)
 
             case .int32:
                 let values = assumedPresentValue(at: offset, as: [Int32].self)
@@ -89,6 +89,10 @@ extension _MessageStorage {
                 return isPacked
                     ? tagSize + Varint.encodedSize(of: Int64(dataSize)) + dataSize
                     : (tagSize * values.count) + dataSize
+
+            case .message:
+                precondition(!isPacked, "a packed message/group field should not be reachable")
+                return serializedByteSize(ofMessageField: field, fieldNumber: fieldNumber)
 
             case .sfixed32:
                 return fixedWidthRepeatedFieldSize(for: fieldNumber, at: offset, isPacked: isPacked, as: Int32.self)
@@ -168,8 +172,8 @@ extension _MessageStorage {
             case .float:
                 return fixedWidthSingularFieldSize(for: fieldNumber, as: Float.self)
 
-            case .group, .message:
-                return serializedByteSize(ofMessageOrGroupField: field, fieldNumber: fieldNumber)
+            case .group:
+                return serializedByteSize(ofGroupField: field, fieldNumber: fieldNumber)
 
             case .int32:
                 return FieldTag.encodedSize(ofTagWithFieldNumber: fieldNumber)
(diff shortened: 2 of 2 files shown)


Sources/SwiftProtobuf/FieldTag.swift
```diff
@@ -47,6 +47,15 @@ package struct FieldTag: RawRepresentable {
         Varint.encodedSize(of: rawValue)
     }
 
+    /// Returns the size in bytes of the tag with the given field number when it is encoded in
+    /// binary format.
+    ///
+    /// Note that this size is always independent of the wire format because the low three bits do
+    /// not change the size of a varint.
+    package static func encodedSize(ofTagWithFieldNumber number: Int) -> Int {
+        return Varint.encodedSize(of: UInt32(truncatingIfNeeded: number) << 3)
```
thomasvl: `&<<`?

Sources/SwiftProtobuf/_MessageStorage+BinarySize.swift
```diff
@@ -0,0 +1,274 @@
...
+            case .bool:
+                return fixedWidthRepeatedFieldSize(for: fieldNumber, at: offset, isPacked: isPacked, as: Bool.self)
+
+            case .bytes:
+                precondition(!isPacked, "a packed bytes field should not be reachable")
+                let values = assumedPresentValue(at: offset, as: [Data].self)
+                let tagSize = FieldTag.encodedSize(ofTagWithFieldNumber: fieldNumber)
+                let dataSize = values.reduce(into: 0) { result, value in
+                    let count = value.count
+                    result += Varint.encodedSize(of: Int64(count)) + count
```
thomasvl, line 53: This seems wrong, you don't seem actually adding the number of bytes within the field, just the space to encode that length.

Sources/SwiftProtobuf/_MessageStorage+BinarySize.swift
```diff
@@ -0,0 +1,274 @@
...
+                    ? tagSize + Varint.encodedSize(of: Int64(dataSize)) + dataSize
+                    : (tagSize * values.count) + dataSize
+
+            case .string:
+                precondition(!isPacked, "a packed string field should not be reachable")
+                let values = assumedPresentValue(at: offset, as: [String].self)
+                let tagSize = FieldTag.encodedSize(ofTagWithFieldNumber: fieldNumber)
+                let dataSize = values.reduce(into: 0) { result, value in
+                    let count = value.utf8.count
+                    result += Varint.encodedSize(of: Int64(count)) + count
```
thomasvl, line 125: Same, what about the actually bytes for the string?

Sources/SwiftProtobuf/_MessageStorage+BinarySize.swift
```diff
@@ -0,0 +1,274 @@
...
+    /// layout, since that has already been done by the caller.
+    private func serializedByteSize(ofMessageOrGroupField field: FieldLayout, fieldNumber: Int) -> Int {
+        var messageSize = 0
+        _ = try! layout.performOnSubmessageStorage(
+            _MessageLayout.SubmessageToken(index: field.submessageIndex),
+            field,
+            self
+        ) {
+            // Include the size of the appropriate tags.
+            if field.rawFieldType == .message {
```
thomasvl: Is it worth moving this outside the performOnSubmessageStroage and avoid having to repeat this check/branch while looping over repeated values?

Sources/SwiftProtobuf/_MessageStorage+BinarySize.swift
```diff
@@ -0,0 +1,274 @@
...
+                messageSize +=
+                    FieldTag.encodedSize(ofTagWithFieldNumber: fieldNumber)
+                    + Varint.encodedSize(of: UInt64(messageSize))
```
thomasvl: This looks suspect to me.  messageSize is going to get the sum of all the repeated messages, why ware we re-encoding that sum here?  You need to encode the size that isn't computed until down on line 269, so this is too early.  This might be reason for doing the message/group branch before the loop to simplify the two closures.

Sources/SwiftProtobuf/_MessageStorage+BinarySize.swift
thomasvl: Sorry, missed the word wrapped " + count"

Sources/SwiftProtobuf/FieldTag.swift
allevato: Done. Would be good to go back through the other old code and update these, but I'll do that later in a dedicated PR.

Sources/SwiftProtobuf/_MessageStorage+BinarySize.swift
allevato: Yep, you're right, I messed this up when I was trying to rearrange things. I guess none of the tests caught this because we don't encode any nested messages that are larger than 127 bytes.

I've updated the code with fixed logic and better variable names to ensure it's clear what's being processed.

Sources/SwiftProtobuf/_MessageStorage+BinarySize.swift
allevato: I found it was cleaner to just have separate functions for message and group fields. Even though it ends up duplicating the `performOnSubmessageStorage` bit, it keeps everything else clear.

thomasvl: Yea, that's much easier to follow and hopefully the compiler/optimizer will be even happier with it.

Issue #1881, opened 2025-10-29
Target 'protoc' in product 'SwiftProtobufPlugin' contains unsafe build flags


madsodgaard:
Looks like the `1.33.0` release broke using the Protobuf plugin, because we now get the following error at compile time:

```
the target 'protoc' in product 'SwiftProtobufPlugin' contains unsafe build flags
```

works fine with `1.32.0`

rnro:
Thanks for raising this issue! It looks like @FranzBusch has this one.

PR #1882, opened 2025-10-29
Use the new warning control instead of `unsafeFlag`s

FranzBusch:
## Motivation

In my recent work to build `protoc` from source I mistakenly disabled warnings by using unsafe flags. Unsafe flags make packages essentially unusable.

## Modifications

Disable the warnings by using the new warning control settings introduced in 6.2 instead.

## Result

Fixes https://github.com/apple/swift-protobuf/issues/1881

### DIFF
--- a/Package@swift-6.0.swift
+++ b/Package@swift-6.0.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:6.2
+// swift-tools-version:5.10
 
 // Package.swift
 //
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.10
+// swift-tools-version:6.2
```
allevato, line 1: This will cause everything to compile in Swift 6 language mode as well, right? That's probably fine (if we've fixed all the relevant issues), but should we add in `swiftLanguageModes: [.v5]` for now so that we don't change the language mode as part of an otherwise unrelated change, and then we can go back and audit it for Swift 6 support later?

Lukasa: Sounds sensible to me.

FranzBusch: I was aware and wanted to see what the CI says. It looks like we are compiling cleanly in Swift 6 mode. I will change it back to Swift 5 mode for now.

FranzBusch: Just when I wrote this comment I saw that we are actually setting the language mode already in all of our manifests.

allevato: Well that's what I get for not expanding the diff far enough.

PR #1888, opened 2025-10-31
[table-driven-branch] Initial implementation of binary parsing. 

allevato:
To keep things manageable, I've only implemented parsing so far for singular scalar fields (sans enum, which we still need to support throughout).

I've tweaked the `Message` protocol definition slightly to let me inject generated `_merge` methods for the new table-driven protos while keeping the old protos working for the generator plugin. That will all get cleaned up once we have enough support that we can move those over to table-driven as well.

### DIFF
--- a/Sources/SwiftProtobuf/WireFormatReader.swift
+++ b/Sources/SwiftProtobuf/WireFormatReader.swift
@@ -232,7 +232,7 @@ struct WireFormatReader {
             available >= length,
             "Internal error: tried to advance \(length) bytes with only \(available) available"
         )
-        available -= length
+        available &-= length
         pointer! += length
     }
 
@@ -240,7 +240,7 @@ struct WireFormatReader {
     ///
     /// - Throws: `BinaryDecodingError` if the recursion limit has been exceeded.
     private mutating func incrementRecursionDepth() throws {
-        recursionBudget -= 1
+        recursionBudget &-= 1
         if recursionBudget < 0 {
             throw BinaryDecodingError.messageDepthLimit
         }
@@ -248,7 +248,7 @@ struct WireFormatReader {
 
     /// Tracks that the reader is exiting a message or group.
     private mutating func decrementRecursionDepth() {
-        recursionBudget += 1
+        recursionBudget &+= 1
         // This should never happen. If it does, something is probably corrupting memory, and
         // simply throwing doesn't make much sense.
         if recursionBudget > messageDepthLimit {


Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -56,7 +56,10 @@ import Foundation
 ///     (https://github.com/protocolbuffers/protobuf/commit/90824aaa69000452bff5ad8db3240215a3b9a595)
 ///     since larger messages would overflow various data structures.
 /// *   Bytes 7-9: The number of required fields defined by the message, as a base-128 integer.
-/// *   Bytes 10-12: The largest field number `N` for which all fields in the range `1..<N` are
+/// *   Bytes 10-12: The number of fields that have explicit presence, as a base-128 integer. Note
+///     that this will always be greater than or equal to the required count, because required
+///     fields also have explicit presence.
```
thomasvl, line 61: Does it make sense to add asserts for these sorta things some place in debug builds? Or just keep them in the generator itself?

Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -540,6 +547,158 @@ extension _MessageStorage {
     // - Enums
 }
 
+// MARK: - Field mutators used for parsing and reflection APIs
+
+// Unlike the above APIs, these only take a `FieldLayout` as an argument. These are used when
+// parsing messages and in reflection APIs, where we don't already know at generation time (as we
```
thomasvl: Not sure I follow this.  We always know all this at "generation" time, is this about not knowing it within say the generic table code that parses binaries?  i.e. - it will have to be looked up at _runtime_ because we don't _generate_ things specifically for serialization?

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
```diff
@@ -0,0 +1,177 @@
...
+            break
+
+        case .array:
+            // TODO: Support repeated fields.
+            break
+
+        case .scalar:
+            switch field.rawFieldType {
+            case .bool:
+                guard tag.wireFormat == .varint else { return false }
```
thomasvl, line 77: I wonder if using a mapping like `var wireFormat: WireFormat` you added to `_MessageLayout` would make sense, because then it could be one `guard` for all scalars.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
```diff
@@ -0,0 +1,177 @@
...
+                updateValue(of: field, to: try reader.nextVarint())
+
+            default:
+                preconditionFailure("Unreachable")
+            }
+
+        default:
+            preconditionFailure("Unreachable")
+        }
+        return true
```
thomasvl, line 167: todo make this false?  i.e. - it shouldn't get here?

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
```diff
@@ -0,0 +1,177 @@
...
+    ///
+    /// - Parameters:
+    ///   - buffer: The binary-encoded message data to decode.
+    ///   - partial: If `false` (the default), this method will check
+    ///     ``Message/isInitialized-6abgi`` after decoding to verify that all required
+    ///     fields are present. If any are missing, this method throws
+    ///     ``BinaryDecodingError/missingRequiredFields``.
+    ///   - options: The ``BinaryDecodingOptions`` to use.
+    /// - Throws: ``BinaryDecodingError`` if decoding fails.
+    public func merge(
```
thomasvl, line 28: This probably will need the extension map, no?

Sources/SwiftProtobuf/_MessageLayout.swift
allevato: Doing it in the generator makes sense. Less so here, I think, because if you got this far and the assertion were to fail then it means you've hand-modified the generated code in a way that's invalid? (If it's a situation like a layout change between versions, we can handle that with the version check.)

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: Right, this was poor phrasing on my part. Updated it.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
allevato: Yeah. The encoder doesn't take it yet either; I'll add it in once I'm ready to support extensions.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
allevato: Repeated fields would benefit less from that because they (but only some) would also need to check `lengthDelimited` to support the packed case, so they we couldn't share a single guard for those. But even in spite of that, I think checking against a single compile-time-known wire format here is likely to perform better than squeezing it into a common mapping that has to do a branch or at least an indirect load against a `switch` table.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
allevato: All of the branches above end without returning in the successful case so that they fall through down here to the `return true`. They only return early when the wire format mismatches or if they throw.

PR #1889, opened 2025-10-31
[table-driven-branch] Ensure that merge respects copy-on-write.

allevato:
Merging binary data into an existing proto was not ensuring unique storage, clobbering both the original proto and the copy.

This change fixes that by defensively uniquing the storage at the beginning of the merge, even if the message would not actually change (for example, if merging an empty sequence of bytes). We could optimize that case if we really wanted to, but it's probably not worth the complexity. Defensively cloning once is also likely to be better than checking if it needs to be cloned on each field decode -- this is actually very hard to do anyway because by that point we're already inside the storage class and we don't have a good way to update the storage reference of the containing message struct.

When we drop the generated `merge` function in favor of a default protocol implementation, we'll need to add something like `ensureUnique()` as another protocol requirement that it can call. This is a little awkward (as it will be public API) but should be fine; we can use the parameter-that-can-only-be-instantiated-by-the-runtime trick if we really want to prevent clients from calling it.

When we get to the point where we're supporting reflective mutation of messages, we'll likely need to do the same defensive copying there, for the same reason.

### DIFF
--- a/Sources/protoc-gen-swift/MessageGenerator.swift
+++ b/Sources/protoc-gen-swift/MessageGenerator.swift
@@ -256,9 +256,9 @@ class MessageGenerator {
                 p.print("return try _storage.serializedBytes(partial: partial, options: options)")
             }
             p.print("}")
-            p.print("\(visibility)func _merge(rawBuffer body: UnsafeRawBufferPointer, extensions: (any ExtensionMap)?, partial: Bool, options: BinaryDecodingOptions) throws {")
+            p.print("\(visibility)mutating func _merge(rawBuffer body: UnsafeRawBufferPointer, extensions: (any ExtensionMap)?, partial: Bool, options: BinaryDecodingOptions) throws {")
             p.withIndentation { p in
-                p.print("try _storage.merge(byReadingFrom: body, partial: partial, options: options)")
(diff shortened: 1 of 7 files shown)


Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -256,9 +256,9 @@ class MessageGenerator {
                 p.print("return try _storage.serializedBytes(partial: partial, options: options)")
             }
             p.print("}")
-            p.print("\(visibility)func _merge(rawBuffer body: UnsafeRawBufferPointer, extensions: (any ExtensionMap)?, partial: Bool, options: BinaryDecodingOptions) throws {")
+            p.print("\(visibility)mutating func _merge(rawBuffer body: UnsafeRawBufferPointer, extensions: (any ExtensionMap)?, partial: Bool, options: BinaryDecodingOptions) throws {")
             p.withIndentation { p in
-                p.print("try _storage.merge(byReadingFrom: body, partial: partial, options: options)")
+                p.print("try _uniqueStorage().merge(byReadingFrom: body, partial: partial, options: options)")
```
thomasvl, line 261: oops.  didn't think about it.  sorry.

PR #1891, opened 2025-11-03
Tweak the CI setup some now that `protoc` is built by SwiftPM

thomasvl:
- Don't override the `PROTOC` value when running tests, let the in tree one be used.
- For conformance tests, only build the runner in the protobuf checkout (no need to force a `protoc` build also)
- Move the protobuf conformance runner build and the conformance tests to be after all the other steps, might as well let everything else finish first to catch issues before waiting on another protobuf to build.
- Turn off the tests on 5.10 that need protoc – `Package@swift-5.10.swift` doesn't include the `protoc` target, so skip
the tests there that need it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,10 +49,16 @@ jobs:
     - name: Test runtime
       run: make test-runtime ${{ matrix.swift.hook }}
     - name: Test plugin
+      # The protoc build isn't part of Package@swift-5.10.swift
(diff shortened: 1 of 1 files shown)


thomasvl:
This does slightly lower our 5.10 coverage, but per the version policy we could drop that, we're keeping it at the moment because oss-fuzz is still on 5.10 and keeping that coverage seems useful.

PR #1894, opened 2025-11-03
Use implicit returns in generated code

glbrntt:
Motivation:

The generated code has explicit returns for computed properties in a
number of places.

Modifications:

- Removed explicit 'return' where possible

Result:

Generated code is ever-so-slightly smaller


glbrntt:
The actual changes are in https://github.com/apple/swift-protobuf/pull/1894/commits/1d7fbec8f73e0590d6c4857b28fd01c2ca6c822d, the changes from `make regenerate` are in https://github.com/apple/swift-protobuf/pull/1894/commits/b80f8121b61e1a9fcd4e54c2b502fcfba0eeb227

PR #1895, opened 2025-11-03
[table-driven-branch] Implement binary decoding for singular message/group fields.

allevato:
This introduces a couple closure-based helper methods on `WireFormatReader` that create sub-readers to parse submessages and groups. The sub-readers track their own termination state and then the parent reader is updated when the sub-reader is done.

In doing this, I removed the separate tracking of the maximum depth limit and remaining recursion budget. By only exposing sub-reader based APIs, we can just have a single budget that decreases each time a sub-reader is created.

I've introduced the notion of an abstract "submessage operation" that gets passed to `performOnSubmessageStorage` since we now need to handle mutation (and not only mutation, but first-time creation if a field isn't present). The plan is for this to also grow to support appending, for decoding repeated fields.

Also fixed another small binary search bug for field layouts. µpb's "denseBelow" actually means "denseAtMost", but I defined ours as strictly below, so I was hitting a boundary bug that just happened to coincide with the treatment of groups in our test protos.

### DIFF
--- a/Sources/SwiftProtobuf/WireFormatReader.swift
+++ b/Sources/SwiftProtobuf/WireFormatReader.swift
@@ -228,7 +228,7 @@ struct WireFormatReader {
         let subBuffer = UnsafeRawBufferPointer(start: pointer, count: length)
         var subReader = WireFormatReader(
             buffer: subBuffer,
-            recursionBudget: recursionBudget - 1,
+            recursionBudget: recursionBudget &- 1,
             trackingGroupFieldNumber: 0
         )
         try perform(&subReader)
@@ -255,7 +255,7 @@ struct WireFormatReader {
         let subBuffer = UnsafeRawBufferPointer(start: pointer, count: available)
         var subReader = WireFormatReader(
             buffer: subBuffer,
-            recursionBudget: recursionBudget - 1,
+            recursionBudget: recursionBudget &- 1,
             trackingGroupFieldNumber: fieldNumber
         )
         try perform(&subReader)
@@ -293,6 +293,7 @@ struct WireFormatReader {
             guard recursionBudget > 0 else {
                 throw BinaryDecodingError.messageDepthLimit
             }
+            recursionBudget &-= 1
             while hasAvailableData {
                 let innerTag = try nextTagWithoutUpdatingLastTagPointer()
                 if innerTag.wireFormat == .endGroup {
@@ -301,6 +302,7 @@ struct WireFormatReader {
                         // not balanced.
                         throw BinaryDecodingError.malformedProtobuf
                     }
+                    recursionBudget &+= 1
                     break
                 } else {
                     _ = try sliceBySkippingField(tag: innerTag)


Sources/SwiftProtobuf/WireFormatReader.swift
```diff
@@ -197,7 +290,9 @@ struct WireFormatReader {
             advance(by: size)
 
         case .startGroup:
-            try incrementRecursionDepth()
+            guard recursionBudget > 0 else {
```
thomasvl, line 293: I feel like we might have lost something.  If there is a group within a group, how are we tracking the depth?

I think there should be a test around unknowns and recursion depth that might help ensure this is working correctly.

Sources/SwiftProtobuf/WireFormatReader.swift
```diff
@@ -173,6 +214,58 @@ struct WireFormatReader {
...
+    /// - Throws: `BinaryDecodingError` if an error occurred while reading from the input.
+    mutating func withReaderForNextLengthDelimitedSlice(perform: (inout WireFormatReader) throws -> Void) throws {
+        guard recursionBudget > 0 else {
+            throw BinaryDecodingError.messageDepthLimit
+        }
+        let length = try nextVarintAsValidatedDelimitedLength()
+        let subBuffer = UnsafeRawBufferPointer(start: pointer, count: length)
+        var subReader = WireFormatReader(
+            buffer: subBuffer,
+            recursionBudget: recursionBudget - 1,
```
thomasvl: `&-` for these?

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
```diff
@@ -30,16 +30,42 @@ extension _MessageStorage {
...
+        while reader.hasAvailableData {
+            let tag = try reader.nextTag()
+            let consumed = try decodeNextField(
+                from: &reader,
+                tag: tag,
+                partial: partial,
+                discardUnknownFields: discardUnknownFields
+            )
+            if !consumed {
+                try decodeUnknownField(from: &reader, tag: tag)
```
thomasvl, line 61: Should this be honoring `discardUnknownFields` in some way?

Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -259,16 +259,36 @@ extension _MessageStorage {
...
+        switch operation {
+        case .read:
+            let submessage = (buffer.baseAddress! + field.offset).bindMemory(to: T.self, capacity: 1).pointee
+            return try perform(submessage.storageForRuntime)
+        case .mutate:
+            // If the submessage isn't already present, we need to initialize a new one first.
+            // Otherwise, ensure that the storage is unique before we mutate it for CoW.
+            let pointer = (buffer.baseAddress! + field.offset).bindMemory(to: T.self, capacity: 1)
+            if !isPresent(field) {
+                pointer.initialize(to: T.init())
```
thomasvl, line 280: Not sure I follow the order here, doesn't this mean we initialize it (and set bits) before doing the ensureUnique check below? Is that safe or could this modify something that someone else is also reading from?

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
allevato: This is the same code as before, just moved. I'm not doing anything with unknown fields (except skipping over them) until I'm ready to implement them throughout the whole system. That should come soon though; I should be able to do it after this one lands.

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: I _think_ this is safe as written, since callers of this function will ensure CoW.

For example, consider the case where we're merging into a message that has a nested message, with potentially multiple threads initially operating on the same shared storage. The initial call to `_merge` uniques that storage, so by the time we get to this line, we should know that the outer message is uniqued if necessary, so each thread would be initializing its own independent copy of the message storage. In that case, we don't need to unique the new nested message's storage, because we know it must be unique already (since we've just initialized it).

Likewise, if the message was already initialized (and perhaps there are multiple copies of it), that's when we go to the uniquing branch below. So that forces the storage of the nested messages to be unique, and then the closure is called with that now-unique storage.

One thing we should probably do later on is write some good multi-threading tests that exercise all of these code paths—even better if we can run them with TSan. (Not sure what the SwiftPM situation looks like for that off the top of my head.)

Sources/SwiftProtobuf/WireFormatReader.swift
allevato: Done.

Sources/SwiftProtobuf/WireFormatReader.swift
allevato: You're right, I oversimplified here. I've updated it to just manually update the budget here instead of having separate increment/decrement methods, since they would otherwise be unused.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
thomasvl: Yea, just explicitly passing it is what called it out to me. :)

Sources/SwiftProtobuf/_MessageStorage.swift
thomasvl: Ah, right "this" message vs. the "sub message".

PR #1896, opened 2025-11-03
Move the upstream proto check back to protobuf head.

thomasvl:
This will give early warning again instead of trigger after we do an update.


thomasvl:
Once we land this I can manually trigger it to confirm it is all correct.

PR #1897, opened 2025-11-03
[table-driven-branch] Add support for unknown fields.

allevato:
Also fix a small message layout bug where we were assuming we could read a `UInt32` at the end of the message layout and mask off the bits we didn't need. When the message is empty, this fails because there are only 3 bytes available.

### DIFF
--- a/Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
+++ b/Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
@@ -243,8 +243,8 @@ extension _MessageStorage {
     /// - Throws: `BinaryDecodingError` if an error occurred while reading from the buffer.
     private func decodeUnknownField(from reader: inout WireFormatReader, tag: FieldTag, discard: Bool) throws {
         let slice = try reader.sliceBySkippingField(tag: tag)
-        if !discard, let pointer = slice.baseAddress, slice.count > 0 {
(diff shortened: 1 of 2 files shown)


Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
```diff
@@ -234,10 +234,17 @@ extension _MessageStorage {
...
+    /// - Parameters:
+    ///   - reader: The `WireFormatReader` from which to read the next field.
+    ///   - tag: The tag representing the current field that was just read from the reader.
+    ///   - discard: If true, the field's data should be skipped. Otherwise, it will be stored in
+    ///     the unknown fields storage.
+    /// - Throws: `BinaryDecodingError` if an error occurred while reading from the buffer.
+    private func decodeUnknownField(from reader: inout WireFormatReader, tag: FieldTag, discard: Bool) throws {
+        let slice = try reader.sliceBySkippingField(tag: tag)
+        if !discard, let pointer = slice.baseAddress, slice.count > 0 {
+            unknownFields.append(protobufData: Data(bytes: pointer, count: slice.count))
```
thomasvl: Since I think `append` is already an internal api, it might make sense to have a version that can directly take the slice and skip making a `Data` from it.

allevato: Done.

PR #1898, opened 2025-11-03
Update proto files

thomasvl:


thomasvl:
Since the upstream file check got shuffled into a few loops, it looks like only the last thing failing fails the step, so the CI workflow will need some work, but I'll circle back to that in the future.

PR #1901, opened 2025-11-04
Revise CI running of conformance tests.

thomasvl:
- Have the `Makefile` provide better messaging in the conformance runners isn't found.
- Have `build.yml` run the conformance tests as it is in the protobuf submodule.
- Revise the cron based conformance run to be a protobuf HEAD run, so it happens on every PR and still happens on a cron as an early signal for when something might need attention.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,8 +40,12 @@ jobs:
       with:
         submodules: true
     - name: Update and install dependencies
-      # dependencies from https://github.com/protocolbuffers/protobuf/blob/main/src/README.md
-      run: apt-get update && apt-get install -y make g++ cmake
+      # Just ensure we have the latest `g++` for building protoc (since it is
+      # built by SwiftPM), full dependencies available at
+      # https://github.com/protocolbuffers/protobuf/blob/main/src/README.md
+      #
+      # `make` is needed for using our own `Makefile`.
+      run: apt-get update && apt-get install -y make g++
     - name: Build
       run: make build ${{ matrix.swift.hook }}
     - name: Test runtime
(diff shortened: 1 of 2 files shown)


thomasvl:
@FranzBusch @Lukasa 

The questions on this in my mind are:
- For the submodule protobuf, it would be nice if that was a different `job` instead of a `step` as it would be parallel and thus faster for total results, but if we want that still on all the supported Xcode versions, we'd have to duplicate the matrix. We could do it just on the _latest_ Swift release if we want the result and aren't worried about version specific conformance coverage. Any thoughts on changing that up?
- Or, do we still want to drop the conformance tests with the submodule protobuf on CI or do we just do it on HEAD?

ps - we could also update the `Makefile` to do the build of the conformance runner so folks don't have to manually do that.

FranzBusch:
> The questions on this in my mind are:
> 
> * For the submodule protobuf, it would be nice if that was a different `job` instead of a `step` as it would be parallel and thus faster for total results, but if we want that still on all the supported Xcode versions, we'd have to duplicate the matrix. We could do it just on the _latest_ Swift release if we want the result and aren't worried about version specific conformance coverage. Any thoughts on changing that up?
> * Or, do we still want to drop the conformance tests with the submodule protobuf on CI or do we just do it on HEAD?
> 
> ps - we could also update the `Makefile` to do the build of the conformance runner so folks don't have to manually do that.

IMO it's fine if we run the conformance tests only with the latest Swift release.

thomasvl:
> > The questions on this in my mind are:
> > 
> > * For the submodule protobuf, it would be nice if that was a different `job` instead of a `step` as it would be parallel and thus faster for total results, but if we want that still on all the supported Xcode versions, we'd have to duplicate the matrix. We could do it just on the _latest_ Swift release if we want the result and aren't worried about version specific conformance coverage. Any thoughts on changing that up?
> > * Or, do we still want to drop the conformance tests with the submodule protobuf on CI or do we just do it on HEAD?
> > 
> > ps - we could also update the `Makefile` to do the build of the conformance runner so folks don't have to manually do that.
> 
> IMO it's fine if we run the conformance tests only with the latest Swift release.

Amended.

Up to you all on if you want to make this new check required also.

thomasvl:
@Lukasa @FranzBusch ping

thomasvl:
@FranzBusch @Lukasa I went ahead and landed this, can one of you make the new step required for PRs since we probably want to ensure the conformance tests are always passing?

PR #1903, opened 2025-11-04
[table-driven-branch] Binary decoding for repeated fields.

allevato:
This adds a new `append` operation to `performOnSubmessageStorage` to handle the new distinction (we're appending to the array itself, not mutating all the elements of the array). But other than that, the implementation is fairly straightforward.

### DIFF
--- a/Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
+++ b/Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
@@ -380,6 +380,12 @@ extension _MessageStorage {
         unpackedWireFormat: WireFormat,
         decodeElement: (inout WireFormatReader) throws -> T
     ) throws -> Bool {
+        assert(
+            field.rawFieldType != .bytes && field.rawFieldType != .group && field.rawFieldType != .message
+                && field.rawFieldType != .string,
+            "Internal error: length-delimited singular values should not reach here"
+        )
+
         switch tag.wireFormat {
         case unpackedWireFormat:
             appendValue(try decodeElement(&reader), to: field)


Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -306,11 +311,34 @@ extension _MessageStorage {
...
+            let pointer = (buffer.baseAddress! + field.offset).bindMemory(to: [T].self, capacity: 1)
+            if !isPresent(field) {
+                pointer.initialize(to: [])
+                switch field.presence {
+                case .hasBit(let hasByteOffset, let hasMask):
+                    _ = updatePresence(hasBit: (hasByteOffset, hasMask), willBeSet: true)
+                case .oneOfMember(let oneofOffset):
+                    _ = updatePopulatedOneofMember((oneofOffset, field.fieldNumber))
+                }
+            }
```
thomasvl, line 332: Seems like this block is in a few places, I wonder if there's a way to share it vs. repeat it, but maybe not.

Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -718,6 +746,25 @@ extension _MessageStorage {
             updateValue(at: offset, to: newValue, oneofPresence: (oneofOffset, field.fieldNumber))
         }
     }
+
+    /// Appends the given value to the values already present in the field, initializing the field
+    /// if necessary.
+    func appendValue<T>(_ value: T, to field: FieldLayout) {
+        // If the field isn't already present, we need to initialize a new array first.
+        let pointer = (buffer.baseAddress! + field.offset).bindMemory(to: [T].self, capacity: 1)
+        if !isPresent(field) {
```
thomasvl, line 755: To append, doesn't it already have to have an array in there?  i.e. - you create the array above in a bunch of cases.  i.e. - when would it need to actually have to do this initialize and set of the has bit vs. be able to assume it is already done?

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
```diff
@@ -233,6 +360,94 @@ extension _MessageStorage {
...
+    ///   what was expected.
+    private func appendMaybePackedValues<T>(
+        from reader: inout WireFormatReader,
+        to field: FieldLayout,
+        tag: FieldTag,
+        unpackedWireFormat: WireFormat,
+        decodeElement: (inout WireFormatReader) throws -> T
+    ) throws -> Bool {
+        switch tag.wireFormat {
+        case unpackedWireFormat:
```
thomasvl, line 390: debug assertion that wireformat isn't message/group/bytes/string?

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: Decoding any value in a repeated field is an "append" operation since those are the merge semantics. Let's say you have a message with a repeated message field, and you create a new instance of it. Then, you decode/merge into it. When you find that message field on the wire, you decode it and "append" it to what's already stored there (which might be nothing). For the first one we encounter that means the array hasn't actually been initialized yet. So we have to handle that case.

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: The `pointer.initialize` bit differs subtly in each case, but the rest of it might be able to be collapsed into a private helper function. I'm inclined to look at it as follow-up optimization, though.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
allevato: Done.

Sources/SwiftProtobuf/_MessageStorage.swift
thomasvl: Chatted offline, figured out where I was misconnecting things.

PR #1906, opened 2025-11-06
Optimize the varint size calculations to be branchless.

thomasvl:
The logic for this comes from the C++ code:
https://github.com/protocolbuffers/protobuf/blob/195da42b90bd39aa8c0b6551188fccf77a83f506/src/google/protobuf/io/coded_stream.h#L1754-L1826

I've always thought Swift didn't have a way to get to the optimal instructions, but I was wrong, it has all the building blocks needed.

Looking at new code in godbolt, in `-O`, without or without the arithmetic overflow protection, the code is the same (tiny), but by turning off the arithmetic overflow checks, we also keep the debug code a bit smaller.

### DIFF
--- a/Sources/SwiftProtobuf/Varint.swift
+++ b/Sources/SwiftProtobuf/Varint.swift
@@ -19,20 +19,12 @@ package enum Varint {
     ///
     /// - Parameter value: The number whose varint size should be calculated.
     /// - Returns: The size, in bytes, of the 32-bit varint.
+    @usableFromInline
     package static func encodedSize(of value: UInt32) -> Int {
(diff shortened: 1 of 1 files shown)


allevato (approved):
I won't claim to understand the sorcery without staring at it a bit longer, but a nice improvement to be sure.

tbkka:
Cute.  Varint coding stores 7 bits per byte, so you this divides the number of bits (`UInt64.bitWidth - clz`) by 7 (9/64 ≅ 1/7) and rounds up (`+ 64`) to get the number of bytes needed.

PR #1907, opened 2025-11-06
[table-driven-branch] Implement support for enums.

allevato:
## Enum generators

While I was in here, I refactored the enum generators to create a common base class but separate "open" and "closed" generators. This cleaned up some of the generator code a bit, and gives us a foothold in the future if we want to change the representation of one of these, like turning open enums into `RawRepresentable` structs.

## Storage

As singular fields, we store the underlying `Int32` value of the enum in storage instead of the Swift enum value because it's
simpler. For repeated enums, since we don't want to penalize the performance of accessors that return Swift arrays, we store the `Array` of Swift enum values.

## Binary encoding/decoding

The above means that we need trampoline functions for some specific cases:

* Deinitializing and copying a repeated enum field.
* Checking whether two Swift enum arrays are equal.
* When encoding a repeated enum, we need to get the raw values.
* When decoding a singular or repeated enum, we need to validate the `Int32`s on the wire to make sure they're valid cases.

So this change is larger than I had hoped, because we needed to introduce a new trampoline (`performOnRawEnumValues`), and then it made sense to rename some of the existing trampoline infrastructure that no longer just serves submessages.

### DIFF
--- a/Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
+++ b/Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
@@ -474,8 +474,7 @@ extension _MessageStorage {
         // into unknown fields at the end.
         var invalidValues: [Int32] = []
 
-        // Recursion budget is irrelevant here because we should never recurse into groups or
-        // messages here; they're not supported as packed fields.
+        // Recursion budget is irrelevant here because we're only reading enums.
         var elementsReader = WireFormatReader(buffer: elementsBuffer, recursionBudget: 0)
 
         try layout.performOnRawEnumValues(
@@ -485,7 +484,7 @@ extension _MessageStorage {
             .append
         ) { outRawValue in
             guard elementsReader.hasAvailableData else { return false }
-            outRawValue = Int32(try elementsReader.nextVarint())
+            outRawValue = Int32(bitPattern: UInt32(truncatingIfNeeded: try elementsReader.nextVarint()))
             return true
         } /*onInvalidValue*/ _: {
             invalidValues.append($0)
--- a/Sources/SwiftProtobuf/_MessageStorage+BinaryEncoding.swift
+++ b/Sources/SwiftProtobuf/_MessageStorage+BinaryEncoding.swift
@@ -512,6 +512,7 @@ extension _MessageStorage {
                 length += Varint.encodedSize(of: $0)
                 return true
             } /*onInvalidValue*/ _: { _ in
+                assertionFailure("invalid value handler should never be called for .read")
             }
 
             encoder.startField(fieldNumber: fieldNumber, wireFormat: .lengthDelimited)
@@ -527,6 +528,7 @@ extension _MessageStorage {
                 encoder.putVarInt(value: Int64($0))
                 return true
             } /*onInvalidValue*/ _: { _ in
+                assertionFailure("invalid value handler should never be called for .read")
             }
         } else {
             // Iterate over the raw values and encode each as its own tag and varint.
@@ -540,6 +542,7 @@ extension _MessageStorage {
                 encoder.putVarInt(value: Int64($0))
                 return true
             } /*onInvalidValue*/ _: { _ in
+                assertionFailure("invalid value handler should never be called for .read")
             }
         }
     }
--- a/Sources/SwiftProtobuf/_MessageStorage.swift
+++ b/Sources/SwiftProtobuf/_MessageStorage.swift
@@ -421,9 +421,10 @@ extension _MessageStorage {
         case .append:
             let pointer = (buffer.baseAddress! + field.offset).bindMemory(to: [T].self, capacity: 1)
             var rawValue: Int32 = 0
+            var isFieldPresent = isPresent(field)
             while try perform(&rawValue) {
                 if let newValue = T(rawValue: Int(rawValue)) {
-                    if !isPresent(field) {
+                    if !isFieldPresent {
                         pointer.initialize(to: [])
                         switch field.presence {
                         case .hasBit(let hasByteOffset, let hasMask):
@@ -431,6 +432,7 @@ extension _MessageStorage {
                         case .oneOfMember(let oneofOffset):
                             _ = updatePopulatedOneofMember((oneofOffset, field.fieldNumber))
                         }
+                        isFieldPresent = true
                     }
                     pointer.pointee.append(newValue)
                 } else {


Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
```diff
@@ -398,7 +404,116 @@ extension _MessageStorage {
...
+        guard elementsBuffer.baseAddress != nil, elementsBuffer.count > 0 else {
+            return
+        }
+
+        // If we see any invalid values during decode them, save them here so we can write them
+        // into unknown fields at the end.
+        var invalidValues: [Int32] = []
 
+        // Recursion budget is irrelevant here because we should never recurse into groups or
+        // messages here; they're not supported as packed fields.
```
thomasvl: Teh "not supported in packed fields" is sorta not related, there's no submessage in a list of enums. 😄

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
```diff
@@ -398,7 +404,116 @@ extension _MessageStorage {
...
+        var elementsReader = WireFormatReader(buffer: elementsBuffer, recursionBudget: 0)
+
+        try layout.performOnRawEnumValues(
+            _MessageLayout.TrampolineToken(index: field.submessageIndex),
+            field,
+            self,
+            .append
+        ) { outRawValue in
+            guard elementsReader.hasAvailableData else { return false }
+            outRawValue = Int32(try elementsReader.nextVarint())
```
thomasvl: Why in the none packed one is it:

```
            outRawValue = Int32(bitPattern: UInt32(truncatingIfNeeded: try reader.nextVarint()))
```

i.e. - why don't you need the truncating part here?

Sources/SwiftProtobuf/_MessageStorage+BinaryEncoding.swift
```diff
@@ -494,4 +492,55 @@ extension _MessageStorage {
...
+        } else {
+            // Iterate over the raw values and encode each as its own tag and varint.
+            _ = try layout.performOnRawEnumValues(
+                _MessageLayout.TrampolineToken(index: field.submessageIndex),
+                field,
+                self,
+                .read
+            ) {
+                encoder.startField(fieldNumber: fieldNumber, wireFormat: .varint)
+                encoder.putVarInt(value: Int64($0))
```
thomasvl, line 542: Given one truncates on reading enums, we could probably do a 32bit version that doesn't have to sign extend for these, but it means we'd need to do the sizing also correctly do realize they will be shorter.

Sources/SwiftProtobuf/_MessageStorage+BinaryEncoding.swift
```diff
@@ -494,4 +492,55 @@ extension _MessageStorage {
...
+        if isPacked {
+            // First, iterate over the values to compute the packed length.
+            var length = 0
+            _ = try layout.performOnRawEnumValues(
+                _MessageLayout.TrampolineToken(index: field.submessageIndex),
+                field,
+                self,
+                .read
+            ) {
+                length += Varint.encodedSize(of: $0)
```
thomasvl, line 512: Does this need a cast?  i.e. - are we going to use the 32bit sizeof in which case will think negative values are only 5 bytes instead of 10?

Sources/SwiftProtobuf/_MessageStorage+BinaryEncoding.swift
```diff
@@ -494,4 +492,55 @@ extension _MessageStorage {
...
+                return true
+            } /*onInvalidValue*/ _: { _ in
+            }
```
thomasvl, line 516: assert can't happen?

Sources/SwiftProtobuf/_MessageStorage+BinarySize.swift
```diff
@@ -161,8 +160,8 @@ extension _MessageStorage {
                 return fixedWidthSingularFieldSize(for: fieldNumber, as: Double.self)
 
             case .enum:
-                // TODO: Support enums.
-                return 0
+                return FieldTag.encodedSize(ofTagWithFieldNumber: fieldNumber)
+                    + Varint.encodedSize(of: assumedPresentValue(at: offset, as: Int32.self))
```
thomasvl, line 164: Does this have a 32/64 bit size problem?

Sources/SwiftProtobuf/_MessageStorage+BinarySize.swift
```diff
@@ -247,6 +246,34 @@ extension _MessageStorage {
...
+        var totalEnumsSize = 0
+        var count = 0
+        _ = try! layout.performOnRawEnumValues(
+            _MessageLayout.TrampolineToken(index: field.submessageIndex),
+            field,
+            self,
+            .read
+        ) {
+            count += 1
+            totalEnumsSize += Varint.encodedSize(of: $0)
```
thomasvl, line 264: Int64 case?

Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -341,6 +335,110 @@ extension _MessageStorage {
...
+
+        case .mutate:
+            preconditionFailure("Internal error: repeated performOnRawEnumValues should not be called to mutate")
+
+        case .append:
+            let pointer = (buffer.baseAddress! + field.offset).bindMemory(to: [T].self, capacity: 1)
+            var rawValue: Int32 = 0
+            while try perform(&rawValue) {
+                if let newValue = T(rawValue: Int(rawValue)) {
+                    if !isPresent(field) {
```
thomasvl: I do wonder if we should have a local bool that caches isPresent so we don't keep doing all the work.  Hopefully the compiler/optimizer can be smart about it.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
allevato: Copy and pasted comment 🙂   Updated.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
allevato: Just a mistake on my part, which I guess the subset of tests I'm using didn't catch.

Sources/SwiftProtobuf/_MessageStorage+BinaryEncoding.swift
allevato: I just double-checked and we always sign-extend negative `Int32`s: https://github.com/apple/swift-protobuf/blob/a028b77456218b581f77f95d083682b82847bd1c/Sources/SwiftProtobuf/Varint.swift#L36-L39

That's after your optimization change yesterday, but the state of the file before that had similar logic.

Sources/SwiftProtobuf/_MessageStorage+BinaryEncoding.swift
allevato: Done throughout.

Sources/SwiftProtobuf/_MessageStorage+BinaryEncoding.swift
allevato: Longer-term, I'd like to turn `BinaryEncoder` into a purely low-level API that would be the opposite of `WireFormatReader` (`WireFormatWriter`), but that needs to wait until I've removed all the binary visitor code.

Sources/SwiftProtobuf/_MessageStorage+BinarySize.swift
allevato: It gets sign-extended, as above.

Sources/SwiftProtobuf/_MessageStorage+BinarySize.swift
allevato: Ditto

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: Since this does require switching over the presence-nature of the field (to know whether it's a has-bit check or a oneof check), we probably should do that. Added.

thomasvl: fyi - I feel like this improvement might apply in other specific helpers, but I didn't go looking.

PR #1910, opened 2025-11-11
[table-driven-branch] Add support for `map` fields.

allevato:
See individual commit messages for details of what's in each change.

### DIFF
--- a/Protos/Tests/SwiftProtobufTests/unittest_swift_required_fields.proto
+++ b/Protos/Tests/SwiftProtobufTests/unittest_swift_required_fields.proto
@@ -94,3 +94,7 @@ message NoneRequiredButNestedRequired {
     optional int64 opt_1 = 1;
     optional NestedRequired opt_2 = 2;
 }
+
+message MapWithNestedRequiredValues {
+    map<int32, NestedRequired> map_1 = 1;
+}
--- a/Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_RequiredFields.swift
+++ b/Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_RequiredFields.swift
@@ -200,6 +200,20 @@ final class Test_RequiredFields: XCTestCase {
         XCTAssertFalse(msg.isInitialized)
     }
 
+    func testMapWithRequiredFieldsInValues() {
+        var msg = SwiftProtoTesting_MapWithNestedRequiredValues()
+        XCTAssertTrue(msg.isInitialized)
+
+        msg.map1[0] = SwiftProtoTesting_NestedRequired()
+        XCTAssertFalse(msg.isInitialized)
+
+        msg.map1[0]!.req1 = 50
+        XCTAssertTrue(msg.isInitialized)
+
+        msg.map1[0] = nil
+        XCTAssertTrue(msg.isInitialized)
+    }
+
     // Helper to assert encoding fails with a not initialized error.
     fileprivate func assertEncodeFailsNotInitialized(
         _ message: SwiftProtoTesting_TestAllRequiredTypes,
--- a/Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest_swift_required_fields.pb.swift
+++ b/Tests/ExperimentalTableDrivenSwiftProtobufTests/unittest_swift_required_fields.pb.swift
@@ -667,6 +667,37 @@ struct SwiftProtoTesting_NoneRequiredButNestedRequired: @unchecked Sendable {
   }
 }
 
+struct SwiftProtoTesting_MapWithNestedRequiredValues: @unchecked Sendable {
+  // SwiftProtobuf.Message conformance is added in an extension below. See the
+  // `Message` and `Message+*Additions` files in the SwiftProtobuf library for
+  // methods supported on all messages.
+
+  var map1: [Int32: SwiftProtoTesting_NestedRequired] {
+    get { return _storage.value(at: SwiftProtobuf._fieldOffset(8, 4), hasBit: (0, 1)) }
+    set { _uniqueStorage().updateValue(at: SwiftProtobuf._fieldOffset(8, 4), to: newValue, willBeSet: !newValue.isEmpty, hasBit: (0, 1)) }
+  }
+
+  var unknownFields: SwiftProtobuf.UnknownStorage {
+    get { _storage.unknownFields }
+    _modify {
+      _ = _uniqueStorage()
+      yield &_storage.unknownFields
+    }
+  }
+
+  init() {}
+
+  private var _storage = SwiftProtobuf._MessageStorage(layout: Self._protobuf_messageLayout)
+
+  private mutating func _uniqueStorage() -> SwiftProtobuf._MessageStorage {
+    if !isKnownUniquelyReferenced(&_storage) { _storage = _storage.copy() }
+    return _storage
+  }
+  mutating func _protobuf_ensureUniqueStorage(accessToken: SwiftProtobuf._MessageStorageToken) {
+    _ = _uniqueStorage()
+  }
+}
+
 // MARK: - Code below here is support for the SwiftProtobuf runtime.
 
 fileprivate let _protobuf_package = "swift_proto_testing"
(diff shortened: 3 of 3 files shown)


Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -94,8 +94,14 @@ class MessageFieldGenerator: FieldGeneratorBase, FieldGenerator {
         storedProperty = "self.\(hasFieldPresence ? underscoreSwiftName : swiftName)"
 
         switch descriptor.type {
-        case .group, .message:
+        case .group:
             trampolineFieldKind = .message(swiftType)
+        case .message:
+            if descriptor.messageType?.mapKeyAndValue != nil {
```
thomasvl: Force unwrap since it can't fail?

Sources/protoc-gen-swift/Descriptor+Extensions.swift
```diff
@@ -294,7 +294,7 @@ extension FieldDescriptor {
         if case (let keyField, let valueField)? = messageType?.mapKeyAndValue {
             let keyType = keyField.swiftType(namer: namer)
             let valueType = valueField.swiftType(namer: namer)
-            return "Dictionary<" + keyType + "," + valueType + ">"
+            return "[\(keyType): \(valueType)]"
```
thomasvl, line 297: Just curious, why did this matter?

Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -182,7 +189,8 @@ import Foundation
         copyField: @escaping TrampolineCopier,
         areFieldsEqual: @escaping TrampolineEquater,
         performOnSubmessageStorage: @escaping SubmessageStoragePerformer,
-        performOnRawEnumValues: @escaping RawEnumValuesPerformer
+        performOnRawEnumValues: @escaping RawEnumValuesPerformer,
+        mapEntryLayout: @escaping MapEntryLayout
```
thomasvl: Note to self - maybe look in the future about having a few versions of init that leaves off this one and then auto fills it with the stub.  i.e. - keep codegen size down by at the expense of a few more things in the runtime for common cases for things that aren't generally needed. (in theory the enum one might also be something to specialize without)

Sources/protoc-gen-swift/MapEntryGenerator.swift
```diff
@@ -0,0 +1,57 @@
...
+    private let generatorOptions: GeneratorOptions
+    private let namer: SwiftProtobufNamer
+    private let entryLayoutCalculator: MessageLayoutCalculator
+
+    init(
+        descriptor: Descriptor,
+        generatorOptions: GeneratorOptions,
+        namer: SwiftProtobufNamer,
+    ) {
+        self.descriptor = descriptor
```
thomasvl, line 40: assert mapKeyAndValue just to confirm no one every adds this some other place by mistake?

Tests/ExperimentalTableDrivenSwiftProtobufTests/map_unittest.pb.swift
```diff
@@ -366,6 +366,68 @@ extension SwiftProtoTesting_TestMap: SwiftProtobuf.Message, SwiftProtobuf._Messa
...
+    case 4:
+      return "\0\u{18}\0\0\u{2}\0\0\0\0\0\0\0\0\u{3}\0\0\u{1}\0\0\0\0\u{8}\0\0\0\0\0\0\u{4}\u{2}\0\0\0\0\u{10}\0\0\u{1}\0\0\0\u{4}"
+    case 5:
+      return "\0\u{c}\0\0\u{2}\0\0\0\0\0\0\0\0\u{3}\0\0\u{1}\0\0\0\0\u{4}\0\0\0\0\0\0\u{5}\u{2}\0\0\0\0\u{8}\0\0\u{1}\0\0\0\u{2}"
+    case 6:
+      return "\0\u{10}\0\0\u{2}\0\0\0\0\0\0\0\0\u{3}\0\0\u{1}\0\0\0\0\u{4}\0\0\0\0\0\0\u{5}\u{2}\0\0\0\0\u{8}\0\0\u{1}\0\0\0\u{1}"
+    case 7:
+      return "\0\u{3}\0\0\u{2}\0\0\0\0\0\0\0\0\u{3}\0\0\u{1}\0\0\0\0\u{1}\0\0\0\0\0\0\u{8}\u{2}\0\0\0\0\u{2}\0\0\u{1}\0\0\0\u{8}"
+    case 8:
+      #if _pointerBitWidth(_64)
```
thomasvl, line 390: I was going to ask if it made sense to do one string and a list of offsets; but I guess since we need to do platform specific things, we'd need two strings and two sets of offsets, so maybe not.

Sources/SwiftProtobuf/MapEntryWorkingSpace.swift
```diff
@@ -0,0 +1,54 @@
...
+//
+// Copyright (c) 2014 - 2025 Apple Inc. and the project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+///
+/// Manages temporary message storage objects used to encode and decode map
```
thomasvl, line 11: will this likely be for all formats or just binary decode/encode?

Sources/SwiftProtobuf/MapEntryWorkingSpace.swift
```diff
@@ -0,0 +1,54 @@
...
+///
+/// To minimize unnecessary allocations, the encode/decode loop for a message creates an instance
+/// of this type to maintain a cache of `_MessageStorage` objects that it uses as temporary
+/// workspace before/after transferring the key and value into/out of the dictionary. This allows
+/// map entry serialization to be implemented in essentially the same fashion as other types.
+struct MapEntryWorkingSpace {
+    /// The layout of the message that contains the map field being encoded/decoded.
+    private let ownerLayout: _MessageLayout
+
+    /// The cache of `_MessageStorage` objects used for the map entries in this message.
```
thomasvl, line 31: keyed by field number or by the token for the type?  i.e. - do we collapse for repeated types?

Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -155,7 +155,19 @@ import Foundation
...
+
+    /// The function type for the generated function that is called to perform an arbitrary
+    /// operation on the elements of a map.
+    public typealias MapEntryPerformer = (
+        _ token: TrampolineToken,
+        _ field: FieldLayout,
+        _ storage: _MessageStorage,
+        _ workingSpace: _MessageStorage,
+        _ operation: TrampolineFieldOperation,
+        _ deterministicOrdering: Bool,
```
thomasvl, line 168: aside - I wonder if we should pass options directly to be more future proof on things that are public? I guess that runs into problems with json/binary/textformat having different options.  maybe we need a CommonOptions for things that this sorta common code could need and to map things between them?

Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -190,7 +205,8 @@ import Foundation
         areFieldsEqual: @escaping TrampolineEquater,
         performOnSubmessageStorage: @escaping SubmessageStoragePerformer,
         performOnRawEnumValues: @escaping RawEnumValuesPerformer,
-        mapEntryLayout: @escaping MapEntryLayout
+        mapEntryLayout: @escaping MapEntryLayout,
+        performOnMapEntry: @escaping MapEntryPerformer
```
thomasvl, line 209: yea, i guess with two maps ones, there's more validation for considering doing an initializer without for common cases.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
```diff
@@ -49,13 +49,15 @@ extension _MessageStorage {
         partial: Bool,
         discardUnknownFields: Bool
     ) throws {
+        var mapEntryWorkingSpace = MapEntryWorkingSpace(ownerLayout: layout)
```
thomasvl, line 52: aside - wonder if this should be optional so we can avoid creating it for all the messages that don't have map<> fields.  Guess we'd need a bit in our info for the message to know if it was needed unless we can lazy create it on demand.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
```diff
@@ -102,8 +105,21 @@ extension _MessageStorage {
...
+                _MessageLayout.TrampolineToken(index: field.submessageIndex),
+                field,
+                self,
+                mapEntryWorkingSpace.storage(for: field.submessageIndex),
+                .append,
+                // Deterministic ordering doesn't apply to decoding.
+                false
+            ) { workingSpace in
+                try reader.withReaderForNextLengthDelimitedSlice { subReader in
+                    try workingSpace.merge(byReadingFrom: &subReader, partial: true, discardUnknownFields: true)
```
thomasvl: Since you force `partial: true` who does that validation still happen afterwards?

Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_Map.swift
```diff
@@ -119,4 +119,381 @@ final class Test_Map: XCTestCase {
...
+        // Missing map value on the wire.
+        assertDecodeSucceeds(inputBytes: [10, 2, 8, 1], recodedBytes: [10, 4, 8, 1, 16, 0]) {
+            $0.mapInt32Int32 == [1: 0]
+        }
+        // Missing map key on the wire.
+        assertDecodeSucceeds(inputBytes: [10, 2, 16, 2], recodedBytes: [10, 4, 8, 0, 16, 2]) {
+            $0.mapInt32Int32 == [0: 2]
+        }
+        // Missing map key and value on the wire.
+        assertDecodeSucceeds(inputBytes: [10, 0], recodedBytes: [10, 4, 8, 0, 16, 0]) {
```
thomasvl, line 142: Can you add a test for `[10, 4, 8, 1, 16, 2, 10, 0]` which should recode as `[10, 4, 8, 1, 16, 2, 10, 4, 8, 0, 16, 0]`.  i.e. - two entries, but ensure we "clear" the key/value in the cache between decoding elements.

And likely do something like this for all the types you test? (and yes, we should have had a test for this already)

Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_Map.swift
```diff
@@ -119,4 +119,381 @@ final class Test_Map: XCTestCase {
...
+        }
+        // Missing map value on the wire.
+        assertDecodeSucceeds(inputBytes: [130, 1, 2, 8, 1], recodedBytes: [130, 1, 4, 8, 1, 16, 0]) {
+            $0.mapInt32Enum == [1: SwiftProtoTesting_MapEnum.foo]
+        }
+        // Missing map key on the wire.
+        assertDecodeSucceeds(inputBytes: [130, 1, 2, 16, 2], recodedBytes: [130, 1, 4, 8, 0, 16, 2]) {
+            $0.mapInt32Enum == [0: SwiftProtoTesting_MapEnum.baz]
+        }
+        // Missing map key and value on the wire.
```
thomasvl, line 322: aside - i wonder if it is a proto2 defined enum, it first entry can be non zero, so I wonder if we should be properly defaulting this.

Sources/protoc-gen-swift/Descriptor+Extensions.swift
allevato: It didn't, but I figured while I was making other changes I'd clean this up.

The only thing I can think of is that in prehistoric versions of Swift there were contexts where you couldn't use the type sugar and had to spell the full nominal type and we needed to protect against that, but those problems don't exist anymore.

Sources/protoc-gen-swift/MapEntryGenerator.swift
allevato: Like I have on line 36?

Sources/SwiftProtobuf/_MessageLayout.swift
allevato: That will depend on whether it's something that needs to control the dictionary access/iteration itself (like ordering), or if it's just something that the passed in closure can capture from the caller's local state. Looking at the current text/JSON options, I don't see anything else there that we would need.

_However_, in the interest of future proofing ourselves, I think it might be cleaner for each of these trampolines to take a single struct that describes all of its parameters, and possibly also return a struct. Then, the runtime could add things easily without all the generated code having to change. I'll do that as a separate change later.

Sources/SwiftProtobuf/_MessageLayout.swift
allevato: As a follow-up, I can create a version of the initializer that takes default arguments for each of these and update the generator to only pass the ones it needs.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
allevato: FWIW, creating this isn't very expensive (storing the message layout is just storing a single `StaticString`, and then creating an empty dictionary is a load/store of a singleton value), and it only happens at the beginning of any message encode/decode, so I don't think optimizing this specific bit is urgent. But I'm also not opposed to it if we update the layout to provide that info.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
allevato: I had it in my head that `partial` didn't matter here because the pseudo-message used to represent map entries didn't have required fields but I guess we still need it to handle values that are messages with required fields. Updated to pass the original through.

Sources/SwiftProtobuf/MapEntryWorkingSpace.swift
allevato: I think it might be just for binary, since that would be the only format where we need to traffic the entries through something that's equivalent to the wire format. Since text/JSON will be reflection-based, we can probably just provide more direct APIs for maps that avoid this.

Sources/SwiftProtobuf/MapEntryWorkingSpace.swift
allevato: It's by trampoline index (submessage token) for the type. Since maps can't hold repeated types, then this will only represent singular submessage fields.

Tests/ExperimentalTableDrivenSwiftProtobufTests/map_unittest.pb.swift
allevato: There might be ways to shrink this down further since the information we need here is so much smaller (the message layout doesn't really buy us anything, we just need a subset of the field layout), but I'm not sure how we would do that without complicating `MessageLayout`/`FieldLayout`/`MessageStorage`—we'd need to add either compile-time indirection or runtime indirection.

I think this still ends up being better than what upb currently generates. upb does a full distinct mini-table for each map field, even if their key/value types are the same. At least with these strings, there's the hope that the linker might fold them. I'm not sure if that happens for upb's more complex const structures (if you have tracing enabled, it definitely doesn't, because the synthesized descriptor name becomes part of the table).

One idea we could explore in the future would be whether it's worth it to just generate the full Cartesian product of possible layouts into the runtime and expose those as constants that this generated code would reference. If I've done my math correctly, that's 14 * 17 = 238 layout strings at 42 bytes each totalling 9996 bytes (not counting any related Swift metadata).

Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_Map.swift
allevato: I construct the default value by just calling `E()` for the `Enum`-conforming type, so as long as that initializer constructs the right value then I _think_ we do the right thing here.

Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_Map.swift
allevato: Done throughout. It turns out that weren't clearing the shared storage, and then when I added that, it uncovered a memory overrun bug when clearing any `enum` field, so that was a fun detour.

Sources/protoc-gen-swift/MapEntryGenerator.swift
thomasvl: I was going commit by commit. 😄

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
thomasvl: On a single simple message, yea, it's cheap.  I was thinking complex things with deep graphs and repeated message fields. We'd end up creating a lot of this was that graph gets created, and my guess is the common case is they are never needed in that entire graph.

Anyway, something we can revisit later.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
thomasvl: Actually, what about validation on serialization?  Maybe passing `false` here is fine, and `_MessageStorage.isInitialized()` needs to make sure it handles maps?  Because `_messageStorage.merge()` in the BinaryAdditions will also check `isInitialized`.

Protos/Tests/SwiftProtobufTests/unittest_swift_enum_clobbering.proto
```diff
@@ -0,0 +1,33 @@
...
+//
+// Copyright (c) 2014 - 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+// -----------------------------------------------------------------------------
+///
+/// Test Swift reserved words used as enum or message names
```
thomasvl: copy/pasted?

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
thomasvl: Edit, sorry passing "true" and blindly allowing partials here, because the later call will then validate it.

Protos/Tests/SwiftProtobufTests/unittest_swift_enum_clobbering.proto
allevato: Sadly it was.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
allevato: That should already be working. The first thing `serializedBytes` if `partial == false` is call `isInitialized`, which does a deep traversal, so we don't have to pass `partial` through to submessage parsing. For dictionaries with message values, that should hit the `performOnSubmessageStorage` trampoline used by `isInitialized` and it'll be handled there, not as part of the working space encoding.

Sources/protoc-gen-swift/MessageFieldGenerator.swift
allevato: Done.

Sources/SwiftProtobuf/_MessageStorage+BinaryDecoding.swift
thomasvl: Tweak the required tests to have map coverage?

allevato: Done.

lauradaniele088-pixel:
Hmm

PR #1912, opened 2025-11-12
Removed: warnings on swift build

agamdua:
* Adopted using URL instead of Path, which is deprecated.


FranzBusch:
Yeah this is a bit tricky since some of those APIs are not existing in all the Swift versions we currently support.

Lukasa:
Yeah, this may require specific compiler checks around these APIs.

0xTim:
Is there any update on this PR?

agamdua:
Hey @0xTim, this is currently dependent on #1913 to drop 5.10 support - the work for that is underway!

thomasvl:

PR #1913, opened 2025-11-13
Drop support for Swift 5.10

glbrntt:
- Remove 5.10 from CI and Makefile workarounds
- Remove 5.10 specific package manifests
- Fix deprecation warnings from 6.0
- Bump language mode to 6 and remove 'StrictConcurrency=Complete'

### DIFF
--- a/FuzzTesting/Package.swift
+++ b/FuzzTesting/Package.swift
@@ -1,6 +1,7 @@
 // swift-tools-version:5.3
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 // **NOTE**: 5.3 due to https://github.com/swiftlang/swift/issues/75800
+//   This is fixed in 6.2 but oss-fuzz is still using 6.1.
 
 import PackageDescription
 


thomasvl:
We can't do this until the oss-fuzz support for Swift is updated otherwise, that will start to fail to build.

thomasvl:
ps - I opened an issue with them because it isn't clear how to update them: https://github.com/google/oss-fuzz/issues/14033

glbrntt:
Ah okay, no problem.

Moving to `swift-tools-version: 6.0` is also what introduces the deprecation warnings in the SwiftPM plugin, so this PR depends on https://github.com/apple/swift-protobuf/pull/1912 getting merged which in turn depends on dropping 5.10 support (or `#if compiler` checking a lot of code).

thomasvl:
I should have said, the oss-fuzz side doens't prevent landing this, it's more a question of do you all (as owners of the repo) and to lose the fuzz coverage over there?

.github/workflows/build.yml
```diff
@@ -29,9 +29,6 @@ jobs:
           hook: "SWIFT_BUILD_TEST_HOOK=\"-Xswiftc -warnings-as-errors\""
         - version: "6.0"
           hook: "SWIFT_BUILD_TEST_HOOK=\"-Xswiftc -warnings-as-errors\""
```
thomasvl: Now that everything sets the hook you can probably drop it and just inline passing this in the two places that used it below.

README.md
```diff
@@ -90,7 +90,7 @@ your project as explained below.
 
 To use Swift with Protocol buffers, you'll need:
 
-* A Swift 5.10 or later compiler (or, if building with Xcode, Xcode 15.3 or later
+* A Swift 6.0 or later compiler (or, if building with Xcode, Xcode 15.3 or later
```
thomasvl: I think it was Xcode 16.0 that first had Swift 6.

glbrntt:
> I should have said, the oss-fuzz side doens't prevent landing this, it's more a question of do you all (as owners of the repo) and to lose the fuzz coverage over there?

We should sort out oss-fuzz before landing this, it'd be a shame to lose fuzzing coverage. I'll look into it...

thomasvl:
> Moving to `swift-tools-version: 6.0` is also what introduces the deprecation warnings in the SwiftPM plugin, so this PR depends on #1912 getting merged which in turn depends on dropping 5.10 support (or `#if compiler` checking a lot of code).

I think https://github.com/apple/swift-protobuf/pull/1958 fixes the SwiftPM plugin issues.

And https://github.com/google/oss-fuzz/pull/14486 went yesterday, I'm guessing we give that that weekend to ensure things are good, and then we can look at landing this.

Can you update this for the current state of `main` and also update the `Package.swift` in `FuzzTesting/` to the new minimum?

glbrntt:
> Can you update this for the current state of `main` and also update the `Package.swift` in `FuzzTesting/` to the new minimum?

I tried updating `FuzzTesting/` but ran into some build issues which I couldn't quickly resolve. Are you okay with leaving it for now? I'm not sure when I'll able to find the time to look into it for a little while.

It's a shame we can't use 6.2 as that simplifies building quite a lot (with `--sanitize=fuzzer`).

thomasvl:
> > Can you update this for the current state of `main` and also update the `Package.swift` in `FuzzTesting/` to the new minimum?
> 
> I tried updating `FuzzTesting/` but ran into some build issues which I couldn't quickly resolve. Are you okay with leaving it for now? I'm not sure when I'll able to find the time to look into it for a little while.

The referenced issue is closed, do we need a new one, or do we just need to update our note to say it's fixed in 6.2?

> It's a shame we can't use 6.2 as that simplifies building quite a lot (with `--sanitize=fuzzer`).

fyi @FranzBusch has a oss-fuzz cl that can be updated to hopefully move us to 6.2.

thomasvl:
fyi - forced a sync to get the conformance fixes into this, you'll need to do that locally also.

PR #1914, opened 2025-11-17
Update proto files

thomasvl:


thomasvl:
The "API Break" in expected since they added a new enum, but since nothing but the plugins use those descriptors, we're fine.

allevato (approved):

PR #1920, opened 2025-11-26
Add option to set custom proto path

madsodgaard:
Allows the developer to supply a custom directory that will be used as the base of the protofiles.

```json
{
    "invocations": [
        {
            "protoFiles": [
                "file1.proto",
                "file2.proto",
            ],
            "visibility": "public",
            "protoPath": "proto_repo_submodule/protos"
        }
    ]
}

```

This is useful if you are storing protofiles inside a Git repo and adding that as a submodule to other repos. Giving you a directory structure like:

```
MySwiftApp/
  - Sources
    - MyAppTarget
      - proto_repo_submodule
        - protos
          - file1.proto
          - file2.proto
```

and any proto files will be importing other proto files relative to that "protos" directory:

such as in **file1.proto**
```protobuf
import "file2.proto"
```


FranzBusch:
Thanks for opening this PR. Just for my understanding this is a convenience and you could achieve the same with the current options by just passing the full path in the `protoFiles` option right?

madsodgaard:
@FranzBusch Unfortunately not. The issue is when the protofiles use `import`. Any `import` statements will look for files relative to the proto-path passed to `protoc`. So If I pass:

```
{
    "invocations": [
        {
            "protoFiles": [
                "proto_repo_submodule/protos/file1.proto",
                "proto_repo_submodule/protos/file2.proto",
            ],
            "visibility": "public"
        }
    ]
}
```

and `file1.proto`, does `import "file2.proto2` (_which is the correct thing to do_), that won't work. Because we just pass `MyLibrary/Sources` as the protopath.

FranzBusch (approved):
This LGTM! @glbrntt can you check if that makes sense for the gRPC plugin as well?

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -86,6 +86,8 @@ struct SwiftProtobufPlugin {
             var implementationOnlyImports: Bool?
             /// Whether import statements should be preceded with visibility.
             var useAccessLevelOnImports: Bool?
+            /// Overrides the path to look for protobuf files
+            var protoPath: String?
```
glbrntt, line 100: This should likely be an array rather than a `String` as you can specify `-I`/`--proto_path` multiple times to `protoc`.

Given the example in the PR description it's not hard to imagine multiple git submodules being added.

madsodgaard: good idea, I changed it to array 👍

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -202,7 +210,7 @@ struct SwiftProtobufPlugin {
         for var file in invocation.protoFiles {
             // Append the file to the protoc args so that it is used for generating
             protocArgs.append("\(file)")
-            inputFiles.append(directory.appending(file))
+            inputFiles.append(protoDirectory.appending(file))
```
glbrntt, line 222: Shouldn't this still be `directory`?

madsodgaard:
@glbrntt @FranzBusch In terms of using `protoPaths` as an array, this has some implications for specifying the input files to SwiftPM.

Consider:
```
MySwiftApp/
  - Sources
    - MyAppTarget
      - proto_repo_submodule1
        - protos
          - file1.proto
          - file2.proto
      - proto_repo_submodule2
        - protos
          - file3.proto
          - file4.proto
```
using
```json
{
    "invocations": [
        {
            "protoFiles": [
                "file1.proto",
                "file2.proto",
                "file3.proto",
                "file4.proto"
            ],
            "visibility": "public",
            "protoPaths": ["proto_repo_submodule1/protos", "proto_repo_submodule2/protos"]
        }
    ]
}
```

we need to pass the `.proto` files as inputs of the SwiftPM plugin. Should we just pass all possible combinations of file paths, so?
* `proto_repo_submodule1/protos/file1.proto`
* `proto_repo_submodule2/protos/file1.proto`
* etc...

Alternatively, we should just provide the `protoPath` as a `String` (so a single directory), and the files in `protoFiles` are relative to that. And if you have a different submodule/folder structure, that would be a separate `invocation`?

FranzBusch:
> Alternatively, we should just provide the `protoPath` as a `String` (so a single directory), and the files in `protoFiles` are relative to that. And if you have a different submodule/folder structure, that would be a separate `invocation`?

I am leaning towards that. @thomasvl would love to get your take on this as well

thomasvl:
> > Alternatively, we should just provide the `protoPath` as a `String` (so a single directory), and the files in `protoFiles` are relative to that. And if you have a different submodule/folder structure, that would be a separate `invocation`?
> 
> I am leaning towards that. @thomasvl would love to get your take on this as well

Within Google, the `proto_library` rule basically is a different invocation, so it seems like modeling a single value might make the most sense. The only place I can see this still falling short for folks is if `invocations[1]` needs to `import` a `.proto` file from `invocations[0]`, i.e. - the files compose. I don't think there's a way to model that, and it would take collapsing to a single invocation which then hits the multiple proto path problem.

So maybe a single string is good enough for now, and a future option would be to allow the invocations to reference a previous one just for proto path arguments. But the there's also the more general problem of two targets having protos that need to reference each other, and current SwiftPM can't model that at all.

madsodgaard:
@FranzBusch @glbrntt I'll change it back to a string, if that is OK with you?

FranzBusch:
Sounds good to me! Maybe you can leave a doc comment on the option explaining what we discussed here

madsodgaard:
@FranzBusch @glbrntt  Just wanted to bump this, to see if we could get this merged soon? Happy to make any changes needed.

thomasvl:
Going with `semver/minor` here since this does add another knob to the plugin config.

Also merged this up with main and let tests run again.

madsodgaard:
@thomasvl I fixed the formatting. Everything else seems to be OK. Could we merge and tag a release?

madsodgaard:
@thomasvl any chance we could get a new release tagged with this change?

PR #1922, opened 2025-12-04
Allow for the use of edition unstable in feature resolver

rgoldfinger6:
Follow-up to the integration of edition unstable in [commit f9555969](https://github.com/apple/swift-protobuf/commit/f9555969eb85b13bbcffcb233f0ea1ffcc624d00)

### DIFF
--- a/Sources/SwiftProtobufPluginLibrary/FeatureResolver.swift
+++ b/Sources/SwiftProtobufPluginLibrary/FeatureResolver.swift
@@ -85,7 +85,7 @@ package class FeatureResolver {
         // When protoc generates defaults, they are ordered, so find the last one.
         var found: Google_Protobuf_FeatureSetDefaults.FeatureSetEditionDefault?
         for d in defaults.defaults {
-            guard (d.edition <= edition || edition == unstable) else { break }
+            guard d.edition <= edition else { break }
             found = d
         }
 


Sources/SwiftProtobufPluginLibrary/FeatureResolver.swift
```diff
@@ -75,7 +75,7 @@ package class FeatureResolver {
         featureSetDefaults defaults: Google_Protobuf_FeatureSetDefaults,
         featureExtensions extensions: [any AnyMessageExtension] = []
     ) throws {
-        guard edition >= defaults.minimumEdition && edition <= defaults.maximumEdition else {
+        guard edition >= defaults.minimumEdition && (edition <= defaults.maximumEdition || edition == .unstable) else {
```
allevato, line 78: Would this be a little clearer written this way? Otherwise I think it's implicitly relying on the fact that `unstable` happens to be numerically greater than `minimumEdition`.

```suggestion
        guard (edition >= defaults.minimumEdition && edition <= defaults.maximumEdition) || edition == .unstable else {
```

rgoldfinger6: Either way works. It should never be possible for edition unstable to be less than the minimumEdition, which is why I only added it to the maximum portion. This is how I configured it for other similar cases, but I am happy to go with your change if you feel that is clearer.

allevato: If that's how it is in the other implementations (especially C++, which we usually use as our main reference), then let's leave it for consistency.

allevato (approved):

Tests/SwiftProtobufPluginLibraryTests/Test_FeatureResolver.swift
```diff
@@ -61,6 +61,19 @@ final class Test_FeatureResolver: XCTestCase {
         }
     }
 
+    func testInit_EditionUnstable() {
```
allevato: This needs a `throws` effect since it `try`s inside.

```suggestion
    func testInit_EditionUnstable() throws {
```

PR #1940, opened 2026-01-13
Update Reference/upstream when updating protobuf.

thomasvl:
Since the repo no longer shadows all the protos, make the update process also update the generated files for the tests.


thomasvl:
Realized the other day this is needed because of https://github.com/apple/swift-protobuf/pull/1932

FranzBusch (approved):

PR #1941, opened 2026-01-13
Also set the Swift image to use.

thomasvl:
Now that Reference is updated, ensure a know swift version is being used.

### DIFF
--- a/.github/workflows/update_protobuf.yml
+++ b/.github/workflows/update_protobuf.yml
@@ -13,6 +13,9 @@ on:
 jobs:
   check-and-update:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/update_protobuf.yml
```diff
@@ -13,6 +13,9 @@ on:
 jobs:
   check-and-update:
     runs-on: ubuntu-latest
+    # Ensure we have the lastest Swift incase so Reference can be updated.
+    container:
+      image: swift:latest
```
thomasvl, line 18: When I manually triggered the protobuf/abseil flow to test the new Reference update, I realized I had no idea how the Swift build was going to work/was working version-wise.

PR #1944, opened 2026-01-14
[table-driven-branch] Implement text format encoding and decoding.

allevato:
The only missing features here are extensions (which are not yet
supported anywhere in the table-driven implementation) and enums
(which need some further refactoring to set up an abstract "enum
layout" concept that we can adopt to give `_MessageStorage` access
to enum case names without the concrete enum types.

### DIFF
--- a/Sources/SwiftProtobuf/_MessageStorage+BinaryEncoding.swift
+++ b/Sources/SwiftProtobuf/_MessageStorage+BinaryEncoding.swift
@@ -94,7 +94,7 @@ extension _MessageStorage {
             case .bool:
                 let values = assumedPresentValue(at: offset, as: [Bool].self)
                 if isPacked {
-                    serializePackedTrivialField(values, for: fieldNumber, into: &encoder) {
+                    serializePackedFixedField(values, for: fieldNumber, into: &encoder) {
                         $1.putBoolValue(value: $0)
                     }
                 } else {
@@ -112,7 +112,7 @@ extension _MessageStorage {
             case .double:
                 let values = assumedPresentValue(at: offset, as: [Double].self)
                 if isPacked {
-                    serializePackedTrivialField(values, for: fieldNumber, into: &encoder) {
+                    serializePackedFixedField(values, for: fieldNumber, into: &encoder) {
                         $1.putDoubleValue(value: $0)
                     }
                 } else {
@@ -127,7 +127,7 @@ extension _MessageStorage {
             case .fixed32:
                 let values = assumedPresentValue(at: offset, as: [UInt32].self)
                 if isPacked {
-                    serializePackedTrivialField(values, for: fieldNumber, into: &encoder) {
+                    serializePackedFixedField(values, for: fieldNumber, into: &encoder) {
                         $1.putFixedUInt32(value: $0)
                     }
                 } else {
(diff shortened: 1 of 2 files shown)


Sources/SwiftProtobuf/_MessageStorage+TextDecoding.swift
```diff
@@ -0,0 +1,290 @@
...
+                        }
+                    }
+                    return true
+                }
+            }
+
+        case .array:
+            switch fieldType {
+            case .bool:
+                try scanPossibleArray(from: &reader) { reader in
```
thomasvl: Q: is it worth putting the `switch` inside the `scanPossibleArray`?  Just wondering about final code size/etc.

Sources/SwiftProtobuf/_MessageStorage+TextDecoding.swift
```diff
@@ -0,0 +1,290 @@
...
+                    appendValue(try reader.scanner.nextUInt(), to: field)
+                }
+
+            case .float:
+                try scanPossibleArray(from: &reader) { reader in
+                    appendValue(try reader.scanner.nextFloat(), to: field)
+                }
+
+            case .group, .message:
+                try scanPossibleArray(from: &reader) { reader in
```
thomasvl: Don't arrays not do the `[...]` syntax for repeated?

Sources/SwiftProtobuf/_MessageStorage+TextEncoding.swift
```diff
@@ -0,0 +1,347 @@
...
+
+            func emitRepeatedField<Value>(_ emitValue: (Value) -> Void) {
+                let values = assumedPresentValue(at: offset, as: [Value].self)
+                if isPacked {
+                    // Use the shorthand representation, "fieldName: [...]".
+                    emitName(ofFieldNumber: fieldNumber, into: &encoder)
+                    encoder.startRegularField()
+                    encoder.startArray()
+                    for value in values {
+                        emitValue(value)
```
thomasvl, line 84: I'm missing something, where does the comma get added?

Sources/SwiftProtobuf/_MessageStorage+TextEncoding.swift
```diff
@@ -0,0 +1,347 @@
...
+
+            case .string:
+                precondition(!isPacked, "a packed string field should not be reachable")
+                emitRepeatedField { encoder.putStringValue(value: $0) }
+
+            default: preconditionFailure("Unreachable")
+            }
+
+        case .scalar:
+            func emitScalarField<Value>(_ emitValue: (Value) -> Void) {
```
thomasvl: Just curious, why do this helper vs. just call the before things before the switch and the after one after the switch?  seems like that would ensure smaller codegen?

Sources/SwiftProtobuf/_MessageStorage+TextDecoding.swift
allevato: Done.

Sources/SwiftProtobuf/_MessageStorage+TextEncoding.swift
allevato: Good catch. We actually didn't have any coverage of packed encodings since I'm using the proto2 test protos. I fixed this bug and one that I discovered with packed binary encoding of varints.

Sources/SwiftProtobuf/_MessageStorage+TextEncoding.swift
allevato: Mainly because I'm used to writing them functionally composed this way.

Groups/messages are the odd ones out because they have different delimiters. I pulled them out into a separate switch so that I could rewrite the scalar cases like you described. I don't know if it's _cleaner_ but it likely is more efficient size-wise now.

PR #1946, opened 2026-01-16
[table-driven-branch] Encode/decode enums in text format.

allevato:
This PR defines an `EnumLayout` that is similar to `_MessageLayout` (I plan on dropping the underscore from the latter eventually since it's SPI-guarded) but for enums. Right now it only serves to provide access to the name map in the same way `_MessageLayout` does, but eventually we could put other information here that's useful for reflection, like:

* whether it's an open or closed enum
* the list of defines cases, using a compact/sparse representation like upb has

### DIFF
--- a/Sources/SwiftProtobuf/_MessageLayout.swift
+++ b/Sources/SwiftProtobuf/_MessageLayout.swift
@@ -166,7 +166,7 @@ import Foundation
         _ storage: _MessageStorage,
         _ operation: TrampolineFieldOperation,
         _ perform: (EnumLayout, inout Int32) throws -> Bool,
-        _ onInvalidValue: (Int32) -> Void
+        _ onInvalidValue: (Int32) throws -> Void
     ) throws -> Void
 
     /// The function type for the generated function that is called to retrieve the "message" layout
--- a/Sources/SwiftProtobuf/_MessageStorage+TextEncoding.swift
+++ b/Sources/SwiftProtobuf/_MessageStorage+TextEncoding.swift
@@ -109,8 +109,7 @@ extension _MessageStorage {
                 emitRepeatedField { encoder.putDoubleValue(value: $0) }
 
             case .enum:
-                // TODO: Support enums.
-                break
+                emitRepeatedEnumField(field, into: &encoder)
 
             case .fixed32, .uint32:
                 emitRepeatedField { (value: UInt32) in encoder.putUInt64(value: UInt64(value)) }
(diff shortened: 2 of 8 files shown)


Sources/SwiftProtobuf/EnumLayout.swift
```diff
@@ -0,0 +1,57 @@
...
+    ///
+    /// TODO: This is public so that it can be read by generated enums to satisfy the
+    /// `ProtoNameProviding` requirement. Make it internal once that's no longer necessary.
+    public let nameMap: _NameMap
+
+    /// Creates a new enum layout from the given values.
+    ///
+    /// This initializer is public because generated enums need to call it.
+    public init(layout: StaticString, names: StaticString) {
+        self.layout = UnsafeRawBufferPointer(start: layout.utf8Start, count: layout.utf8CodeUnitCount)
```
thomasvl, line 54: Comment said layout was unused, but you still assign it?

Sources/SwiftProtobuf/EnumLayout.swift
```diff
@@ -0,0 +1,57 @@
...
+        self.nameMap = names.utf8CodeUnitCount != 0 ? _NameMap(bytecode: names) : _NameMap()
```
thomasvl, line 55: I was going to say does it make sense to have a second initializer without the `names` argument to shrink code gen, but I an enum with no cases doesn't seem worth it.

Sources/SwiftProtobuf/_MessageStorage+TextEncoding.swift
```diff
@@ -109,8 +109,7 @@ extension _MessageStorage {
                 emitRepeatedField { encoder.putDoubleValue(value: $0) }
 
             case .enum:
-                // TODO: Support enums.
-                break
+                emitRepeatedEnumField(field, into: &encoder)
```
thomasvl, line 112: Is it at all worth a comment about why `emitRepeatedField` isn't used for enums?

Sources/SwiftProtobuf/_MessageLayout.swift
```diff
@@ -166,7 +166,7 @@ import Foundation
         _ storage: _MessageStorage,
         _ operation: TrampolineFieldOperation,
         _ perform: (EnumLayout, inout Int32) throws -> Bool,
-        _ onInvalidValue: (Int32) -> Void
+        _ onInvalidValue: (Int32) throws -> Void
```
thomasvl, line 169: I guess binary hadn't needed to throw because it could have things end up in unknown values?

Sources/SwiftProtobuf/EnumLayout.swift
allevato: I guess I could drop the stored property and just have the initializer do nothing with the argument (so that we don't have to deal with a signature change for the generated code) but we'll probably end up using this in the near future so I figure it's less churn later.

Sources/SwiftProtobuf/_MessageLayout.swift
allevato: Right.

I could have kept this as non-throwing and have the caller just store a boolean value that it checks after the callback returns, but that seemed more obnoxious than just letting this thing throw.

Sources/SwiftProtobuf/_MessageStorage+TextEncoding.swift
allevato: I think it's sort of self-evident by navigating the code that enums need different treatment. The binary encoding code is similar in how it has a different helper function for repeated enums.

PR #1947, opened 2026-01-20
[table-driven-branch] Add JSON encoding/decoding.

allevato:
This is complete except for extensions (not yet supported
anywhere), and it does not yet handle well-known types with
custom JSON representations. The latter will require hoisting
the fully-qualified message/enum name into the layout so that
we can check it during encoding/decoding.

### DIFF
--- a/Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
+++ b/Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
@@ -98,7 +98,7 @@ extension _MessageStorage {
                     try scanEnumValue(field, from: &reader, operation: .append)
 
                 case .fixed32, .uint32:
-                    let n = try reader.scanner.nextSInt()
+                    let n = try reader.scanner.nextUInt()
                     if n > UInt64(UInt32.max) {
                         throw TextFormatDecodingError.malformedNumber
                     }
@@ -153,9 +153,10 @@ extension _MessageStorage {
         from reader: inout JSONReader,
         requireQuotedBool: Bool = false
     ) throws {
+        let isNull = reader.scanner.skipOptionalNull()
         switch field.rawFieldType {
         case .bool:
-            if reader.scanner.skipOptionalNull() {
+            if isNull {
                 clearValue(of: field, type: Bool.self)
                 break
             }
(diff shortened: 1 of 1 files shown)


Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
```diff
@@ -0,0 +1,426 @@
...
+                    appendValue(try reader.scanner.nextBytesValue(), to: field)
+
+                case .double:
+                    appendValue(try reader.scanner.nextDouble(), to: field)
+
+                case .enum:
+                    try scanEnumValue(field, from: &reader, operation: .append)
+
+                case .fixed32, .uint32:
+                    let n = try reader.scanner.nextSInt()
```
thomasvl: nextUInt?

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
```diff
@@ -0,0 +1,426 @@
...
+    ///   - requireQuotedBool: If true and the field's type is `bool`, the value is expected to be
+    ///     quoted. This is used when scanning map keys.
+    private func scanSingularValue(
+        of field: FieldLayout,
+        from reader: inout JSONReader,
+        requireQuotedBool: Bool = false
+    ) throws {
+        switch field.rawFieldType {
+        case .bool:
+            if reader.scanner.skipOptionalNull() {
```
thomasvl: Future optimization:

```
let skippedNull = reader.scanner.skipOptionalNull()
switch field.rawFieldType {
  case ...:
    if skippedNull {
}
```

To maybe shrink the code size by always having a single call for the skipOptionalNull.

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
allevato: Oops, done.

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
allevato: Future? I'll do it now.

PR #1948, opened 2026-01-21
Use URL APIs in SwiftProtobufPlugin

Gogopex:
## Summary
- migrate SwiftProtobufPlugin from deprecated Path APIs to URL-based APIs
- avoid storing non-Sendable Target in PluginError by capturing target type description
- update documentation comments for URL parameters

## Testing
- swift build
- ran swift test locally, too

### DIFF
--- a/Plugins/SwiftProtobufPlugin/plugin.swift
+++ b/Plugins/SwiftProtobufPlugin/plugin.swift
@@ -6,7 +6,7 @@ struct SwiftProtobufPlugin {
     /// Errors thrown by the `SwiftProtobufPlugin`
     enum PluginError: Error, CustomStringConvertible {
         /// Indicates that the target where the plugin was applied to was not `SourceModuleTarget`.
-        case invalidTarget(Target)
+        case invalidTarget(String)
         /// Indicates that the file extension of an input file was not `.proto`.
         case invalidInputFileExtension(String)
         /// Indicates that there was no configuration file at the required location.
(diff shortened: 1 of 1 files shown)


thomasvl:
Seems like this might overlap #1912?  We're waiting for oss-fuzz to land a PR that updates what version of Swift is used there so we can drop 5.10 and thus not having to make the plugin updates conditional on the verison.

Gogopex:
crap, thank @thomasvl, I did not see #1912, my bad for not checking beforehand. I was just getting annoyed at the warnings while rebuilding my project every 30mn. 

Feel free to close this (or tell me to) if this ends up not needed, else we can park it there just in case.

Ty!

thomasvl:
> crap, thank @thomasvl, I did not see #1912, my bad for not checking beforehand. I was just getting annoyed at the warnings while rebuilding my project every 30mn.
> 
> Feel free to close this (or tell me to) if this ends up not needed, else we can park it there just in case.
> 
> Ty!

No worries, just wanted to be sure it was for the same concerns as the other.

Sadly, we've been blocked on getting oss-fuzz updated for a while now (early December?), we'd rather not lose that safety net so we've had to keep 5.10 support around as a result. And to "fix" the warnings you mention would mean some conditional code to deal with the different versions and we just keep hoping that we'll be able to drop 5.10 _soon_.

thomasvl:

Gogopex:
> fyi - the 5.10 support is now gone on `main`.

Thank you! Closing this and will bump deps. Appreciate it

thomasvl:
> > fyi - the 5.10 support is now gone on `main`.
> 
> Thank you! Closing this and will bump deps. Appreciate it

Not sure exactly what you were expecting - it is only on 'main', os if "will bump deps" implies you expect to see the warnings go away, that hasn't happened and hasn't made a release.  We've just updated `main` to remove the 5.10 support, so something like this PR should be easier to move forward with now.

PR #1954, opened 2026-01-23
[table-driven-branch] Add support for custom JSON representations used by well-known types.

allevato:


Sources/SwiftProtobuf/NameMap.swift
```diff
@@ -291,6 +297,9 @@ public struct _NameMap: ExpressibleByDictionaryLiteral {
     /// support TextFormat's requirement to skip these numbers in all cases.
     private var reservedRanges: [Range<Int32>] = []
 
+    /// The fully-qualified name of the owning message or enum.
+    @_spi(ForGeneratedCodeOnly) public private(set) var fullyQualifiedName: String = ""
```
thomasvl, line 301: I wonder if this should just be a getter with private storage so getting can confirm the opcode was found and thus the name was set.

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -350,6 +352,7 @@ class MessageGenerator {
 
     private func generateMessageLayout(printer p: inout CodePrinter) {
         var writer = ProtoNameInstructionWriter()
+        writer.writeFullyQualifiedName(descriptor.fullName)
```
thomasvl, line 355: Not sure if it is in a followup commit, I'm going through in order, but should where we capture the full name for messages also get updated to instead call through to this data?

Sources/SwiftProtobuf/CustomJSONWKTClassification.swift
```diff
@@ -0,0 +1,84 @@
...
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+///
+/// Classification of well-known types based on custom JSON encodings.
+///
+// -----------------------------------------------------------------------------
+
+/// Classifies a subset of the well-known types based on the nature of their custom JSON encodings.
+package enum CustomJSONWKTClassification {
```
thomasvl, line 16: Descriptor also has information about if it was a WKT, it's worth doing an opcode to say it is a wkt or not to avoid atleast the prefix lookup on messages when starting up the name info.

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
```diff
@@ -37,31 +37,117 @@ extension _MessageStorage {
...
-                // (depending on the decoding options), so any field we get back from this reader
-                // should always exist.
-                preconditionFailure("unreachable")
-            }
-            // It is an error in JSON if we encounter values for more than one member of the same
-            // `oneof`.
-            if case .oneOfMember(let oneOfOffset) = field.presence {
-                if populatedOneofMember(at: oneOfOffset) != 0 {
-                    throw JSONDecodingError.conflictingOneOf
+        switch CustomJSONWKTClassification(messageLayout: layout) {
```
thomasvl, line 50: I wonder if messageLayout should cache the enum value vs have to keep looking it up?

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
```diff
@@ -37,31 +37,117 @@ extension _MessageStorage {
...
+        case .boolValue:
+            try disallowingNull {
```
thomasvl, line 52: I wonder if a local with the WKTClassification and then pick off all the disallowingNull's at once before doing the switch for parsing makes much different in code size.

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
```diff
@@ -424,4 +509,147 @@ extension _MessageStorage {
...
+    /// - Precondition: The receiver must be the storage for `google.protobuf.Timestamp`.
+    private func parseAsTimestamp(from reader: inout JSONReader) throws {
+        let timestampString = try reader.scanner.nextQuotedString()
+        let (seconds, nanos) = try parseTimestamp(s: timestampString)
+        updateValue(of: layout[fieldNumber: 1]!, to: seconds)
+        updateValue(of: layout[fieldNumber: 2]!, to: nanos)
+    }
+
+    /// Parses the next value from the input and interprets it as the JSON representation of an
+    /// arbitrary value.
```
thomasvl: "Value" ?  (since it's the name)

Sources/SwiftProtobuf/_MessageStorage+JSONEncoding.swift
```diff
@@ -310,4 +368,138 @@ extension _MessageStorage {
         default: preconditionFailure("Unreachable")
         }
     }
+
+    /// Emits a double-quoted string to the encoder that is the JSON representation of the receiver
+    /// as a well-known-type `Duration`.
+    ///
+    /// - Precondition: The receiver must be the storage for `google.protobuf.Duration`.
+    private func emitAsDuration(into encoder: inout JSONEncoder) throws {
+        let seconds = value(of: layout[fieldNumber: 1]!) as Int64
```
thomasvl, line 377: Random question - would this work?

```swift
let seconds: Int64 = value(of: layout[fieldNumber: 1]!)
```

Is there any preference to one vs. the other? In my head, it feels like "as Int64" could match something and to coercion or something vs. ensuring the right overload exists.

Sources/SwiftProtobuf/_MessageStorage+JSONEncoding.swift
```diff
@@ -310,4 +368,138 @@ extension _MessageStorage {
...
+            }
+            encoder.putDoubleValue(value: numberValue)
+
+        case 3:
+            encoder.putStringValue(value: value(of: field))
+
+        case 4:
+            encoder.putNonQuotedBoolValue(value: value(of: field))
+
+        case 5, 6:
```
thomasvl, line 490: The explicit "1" for most of the wkts doesn't bug me, but I do sorta wonder if for these we should have some constants to between encode/decode for the different values and then maybe a package test place the looks up the field and asserts the right name/type/etc. just as an added safety net.

Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -294,8 +294,32 @@ extension _MessageStorage {
...
+                    case .hasBit(let hasByteOffset, let hasMask):
+                        _ = updatePresence(hasBit: (hasByteOffset, hasMask), willBeSet: true)
+                    case .oneOfMember(let oneofOffset):
+                        _ = updatePopulatedOneofMember((oneofOffset, field.fieldNumber))
+                    }
+                } else {
+                    pointer.pointee._protobuf_ensureUniqueStorage(accessToken: _MessageStorageToken())
+                }
+                pointer.pointee.nullValue = .nullValue
+                return true
```
thomasvl, line 300: I don't think you need this since the other path relies on the `return true` after the whole if block.

Sources/SwiftProtobuf/_MessageStorage.swift
```diff
@@ -294,8 +294,32 @@ extension _MessageStorage {
...
+            } else {
+                clearValue(of: field, type: type)
```
thomasvl: given how short this is, I sorta wonder if the first thing in this `case` should be:

```swift
guard type == Google_Protobuf_Value.self else {
  clearValue(of: field, type: type)
  return true
}
```

And then remove the intent on all the complexity needed with it is the wkt.  it's a little odd since the "common" case will be the `else` of the `guard`, but might make the code a little more readable to dropping the indent.

Anyway, up to you style wise.

Sources/SwiftProtobuf/CustomJSONWKTClassification.swift
allevato: Once the rest of the layout settles, we can investigate if we have a free bit that we can use for that.

Sources/SwiftProtobuf/NameMap.swift
allevato: I think the entire `NameMap` abstraction as it is today is going to be completely rewritten when it gets folded into a `Reflection`-like thing and the bytecode will change drastically as a result, so I don't think we need to do that kind of validation now.

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
allevato: Right now `MessageLayout` is completely immutable (and static data). We'd have to introduce some sort of synchronization to cache values.

We could look into storing this information in the layout directly, but it's 5 bits for something that would only be used for WKTs.

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
allevato: Might be worth looking at later, at the cost of making the code structured a lot worse (since we'd have to break the switch into those that disallow nulls and those that don't).

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
allevato: Done.

Sources/SwiftProtobuf/_MessageStorage+JSONEncoding.swift
allevato: `: Int64` would be equivalent.

`as` doesn't coerce unlike types; for example, `let x: Int = 5; let y = x as Int64` isn't valid. The coercions it supports are upcasts, existential wrapping, and Foundation bridging. Since none of those apply here, `as Int64` is equivalent to the type annotation on the variable (they both provide the constraint for the type system).

Sources/SwiftProtobuf/_MessageStorage+JSONEncoding.swift
allevato: Yeah, I'm mainly wary of introducing say, an enum for namespacing (which adds type metadata) and sometimes Swift even generates unnecessary code for global constants. We can revisit this once I'm absolutely sure the optimizer does the right thing here.

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: Done.

Sources/SwiftProtobuf/_MessageStorage.swift
allevato: Yeah, that `guard` reads awkwardly to me. I'll write it as `if != Value` instead and just `return` early inside it for the common case, and then dedent the special case.

Sources/protoc-gen-swift/MessageGenerator.swift
allevato: Yeah, that's some later cleanup.

PR #1955, opened 2026-01-23
Share the .build directory for all the builds.

thomasvl:
Looks like right now each of the CI test steps end up rebuilding the core library/etc. This seems like it will hopefully let them share artifacts.

Until something like
https://forums.swift.org/t/rfc-file-hashes-for-swift-incremental-compilation/80334
lands, github's actions/cache can't really help SwiftPM builds since it is all
timestamp based, but maybe this will help share a few things and save a little
time off the build.


thomasvl:
Looking at the `core` configs from `build.yml` for the last few PRs and how the actions did on this PR, it seems like only the 6.0 build might be showing drop in times, the 6.1 and 6.2 don't really show a change but maybe VM speed variations hide all of this anyways.

thomasvl:
Closing this out, doesn't seem worth it.

PR #1957, opened 2026-01-28
Fixed Sendable warnings and Path deprecation warnings.

rafaelcepeda:
**Motivation**
Anyone trying to build this project, or that uses `swift-protobuf` as a dependency, are presented with a bunch of compilation warnings when trying to compile the project.

**Change**
This PR fixes all of the SwiftPlugin warnings currently related to Sendable conformance and the deprecation of Path.

**Result**
Now builds are free from any compilation warnings.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
       image: swift:${{ matrix.swift.version }}
     steps:
     - name: Checkout
-      uses: actions/checkout@v4
+      uses: actions/checkout@v6
       with:
         submodules: true
     - name: Update and install dependencies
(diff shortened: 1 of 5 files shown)


Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -119,40 +119,40 @@ struct SwiftProtobufPlugin {
     ///   - tool: The tool method from the context.
     /// - Returns: The build commands configured based on the arguments.
     func createBuildCommands(
-        pluginWorkDirectory: PackagePlugin.Path,
+        pluginWorkDirectory: URL,
```
FranzBusch, line 122: This won't work as seen with the CI since the URL based APIs are too new. There is another PR that tried the same #1948

thomasvl:
As mentioned, see #1948 for details. We're waiting on oss-fuzz to update so we can generally drop 5.10 support and make updating easier

thomasvl:

thomasvl:
We've landed #1912 which should fix these issues. So we'll close this out.

PR #1958, opened 2026-01-28
Fix swift-tools-version in Package@swift-6.0.swift

pepicrft:
## Summary

The `Package@swift-6.0.swift` file incorrectly declares `swift-tools-version:5.10` instead of `swift-tools-version:6.0`. This causes Swift Package Manager to fail when resolving the package for projects using Swift 6.x tools, particularly when the package is fetched from the Swift Package Registry.

## Problem

When a project with `swift-tools-version: 6.1` (or similar) tries to use swift-protobuf from the Swift Package Registry, SPM selects `Package@swift-6.0.swift` based on the filename convention. However, the declared tools version inside the file is `5.10`, creating a mismatch that causes manifest compilation to fail with:

```
error: Invalid manifest (compiled with: [...-swift-version 5...])
Missing or empty JSON output from manifest compilation for apple.swift-protobuf
```

## Fix

Changed the first line of `Package@swift-6.0.swift` from:
```swift
// swift-tools-version:5.10
```
to:
```swift
// swift-tools-version:6.0
```

This aligns the declared tools version with what the filename convention implies.

### DIFF
--- a/.github/workflows/update_protobuf.yml
+++ b/.github/workflows/update_protobuf.yml
@@ -12,6 +12,8 @@ on:
 
 jobs:
   check-and-update:
+    # Only do this job on the main repo (and not forks of it).
+    if: github.repository == 'apple/swift-protobuf'
     runs-on: ubuntu-latest
     # Ensure we have the lastest Swift incase so Reference can be updated.
     container:


gjcairo (approved):
Thanks @pepicrft !

thomasvl:
Looks like this causes some things to fail because changes the compile mode for some things.

pepicrft:
@thomasvl pushed a fix addressing the issues.

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -1,5 +1,5 @@
 import Foundation
-import PackagePlugin
+@preconcurrency import PackagePlugin
```
thomasvl, line 2: Hopefully this can eventually come out when we update the apis within the plugin?

PR #1959, opened 2026-01-29
[table-driven-branch] Implement hashing.

allevato:


Sources/SwiftProtobuf/_MessageStorage+Hashing.swift
```diff
@@ -0,0 +1,141 @@
...
+//
+// Copyright (c) 2014 - 2026 Apple Inc. and the project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+///
+/// Manages the in-memory storage of the fields of a message.
```
thomasvl: Update comment?

Sources/SwiftProtobuf/duration.pb.swift
```diff
@@ -211,4 +211,7 @@ extension Google_Protobuf_Duration: Message, _MessageImplementationBase, _ProtoN
   public static func ==(lhs: Google_Protobuf_Duration, rhs: Google_Protobuf_Duration) -> Bool {
     return lhs._storage.isEqual(to: rhs._storage)
   }
+  public func hash(into hasher: inout Swift.Hasher) {
```
thomasvl, line 214: Long term, do these need to be generated vs have default impls or anything to shrink codegen?

Sources/SwiftProtobuf/_MessageStorage+Hashing.swift
allevato: Done.

Sources/SwiftProtobuf/duration.pb.swift
allevato: I think we can get them as default impls. The trick is that I want to avoid exposing the `_MessageStorage` type as public-non-SPI, but to make them default impls there needs to be a protocol requirement on `Message` that can get the storage. I've hacked that in today using `AnyObject` as the return type (so the runtime has to cast it back) and a token that people outside the runtime can't create.

In the future we might be able to make this cleaner by using the reflection APIs. If there's a protocol requirement to get a concrete `ReflectedMessage` from an arbitrary `Message`-conforming type, then the runtime could just reach into that for the storage without it being exposed to clients.

Issue #1943, opened 2026-01-14
Unable to use well-known types with SwiftPM plugin


madsodgaard:
I am trying to use the well known types from `google/protobuf`, but I am running into a issue when building. I am pretty sure I had this working at some point, but am I missing something? I am using the SwiftPM build plugin.

Using the latest 1.33.3 release.

The build error is:
```
Import "google/protobuf/timestamp.proto" was not found or had errors.
```

**swift-protobuf-config.json**
```json
{
    "invocations": [
        {
            "protoFiles": [
                "v1/transport.proto",
                "v1/messages.proto",
            ],
            "visibility": "package"
        }
    ]
}
```

**transport.proto**
```proto
syntax = "proto3";

import "google/protobuf/timestamp.proto";

message MyMessage {
  google.protobuf.timestamp my_date   = 1;
}
```

thomasvl:
Can you upload a sample project that shows the problem?  Normally the WKTs should be with protoc and found via that, but it's possible something with the new plugin support isn't right.

@FranzBusch fyi

madsodgaard:
@thomasvl This project shows the problem: https://github.com/madsodgaard/protobuf-example

If you clone and run
```
swift build
```
then the error shows:

```
google/protobuf/timestamp.proto: File not found.
/Users/mads/dev/protobuf-example/Sources/protobuf-example/v1/messages.proto:3:1: Import "google/protobuf/timestamp.proto" was not found or had errors.
/Users/mads/dev/protobuf-example/Sources/protobuf-example/v1/messages.proto:6:3: "google.protobuf.timestamp" is not defined.
```

thomasvl:
@FranzBusch @gjcairo - off the top of my head, this may be because the invocation provides`-I` directives, so you also need to provide one to the directory `protoc` came form so it can find the bundled WKTs. Assuming the code is similar, the grpc plugin might also have this issue and need the same fix (what ever it ends up being).

thomasvl:
@gjcairo @FranzBusch - I poked at this briefly today, and it is what expected. `protoc` can't find the WKTs that are normally bundled with it, I think basically a change like this is needed:

```diff
diff --git a/Plugins/SwiftProtobufPlugin/plugin.swift b/Plugins/SwiftProtobufPlugin/plugin.swift
index e8f4907d..3b40b702 100644
--- a/Plugins/SwiftProtobufPlugin/plugin.swift
[...]

PR #1979, opened 2026-02-11
Example using a WKT

thomasvl:
This does not work with the protoc built locally because the WKTs aren't found. If you try building using the env variable to provide a full `protoc` install, then it works since that bundles all the WKTs and the proto compiler auto finds the bundled files.

See #1943 for some of the research into options in this space.

### DIFF
--- a/PluginExamples/Package.swift
+++ b/PluginExamples/Package.swift
@@ -56,6 +56,15 @@ let package = Package(
                 .plugin(name: "SwiftProtobufPlugin", package: "swift-protobuf")
             ]
         ),
+        .target(
+            name: "UsesWKTs",
+            dependencies: [
+                .product(name: "SwiftProtobuf", package: "swift-protobuf")
+            ],
+            plugins: [
+                .plugin(name: "SwiftProtobufPlugin", package: "swift-protobuf")
+            ]
+        ),
         .testTarget(
             name: "ExampleTests",
             dependencies: [
@@ -64,6 +73,7 @@ let package = Package(
                 .target(name: "Import"),
                 .target(name: "AccessLevelOnImport"),
                 .target(name: "CustomProtoPath"),
+                .target(name: "UsesWKTs"),
             ]
         ),
     ],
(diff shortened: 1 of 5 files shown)


FranzBusch:
> This does not work with the protoc built locally because the WKTs aren't found

Would be great if we can fix that as well. Is it just about building them and adding a `-I`?

thomasvl:
> > This does not work with the protoc built locally because the WKTs aren't found
> 
> Would be great if we can fix that as well. Is it just about building them and adding a `-I`?

Either we provide a `-I` directive to point to all the files, or we bundle the files next to the binary the [startup code](https://github.com/protocolbuffers/protobuf/blob/acddf37316bfd24d388bdc2302e2628c14f42b0a/src/google/protobuf/compiler/command_line_interface.cc#L261) can auto find them.

FranzBusch:
> > > This does not work with the protoc built locally because the WKTs aren't found
> > 
> > 
> > Would be great if we can fix that as well. Is it just about building them and adding a `-I`?
> 
> Either we provide a `-I` directive to point to all the files, or we bundle the files next to the binary the [startup code](https://github.com/protocolbuffers/protobuf/blob/acddf37316bfd24d388bdc2302e2628c14f42b0a/src/google/protobuf/compiler/command_line_interface.cc#L261) can auto find them.

I don't think we can move the files next to the binary since when `protoc` is built from source the binary is somewhere in the build directory. It seems like the best path forward is for the Swift PM plugin to add a `-I` directive into the submodule where the WKTs are.

thomasvl:
> > > > This does not work with the protoc built locally because the WKTs aren't found
> > > 
> > > 
> > > Would be great if we can fix that as well. Is it just about building them and adding a `-I`?
> > 
> > 
> > Either we provide a `-I` directive to point to all the files, or we bundle the files next to the binary the [startup code](https://github.com/protocolbuffers/protobuf/blob/acddf37316bfd24d388bdc2302e2628c14f42b0a/src/google/protobuf/compiler/command_line_interface.cc#L261) can auto find them.
> 
> I don't think we can move the files next to the binary since when `protoc` is built from source the binary is somewhere in the build directory. It seems like the best path forward is for the Swift PM plugin to add a `-I` directive into the submodule where the WKTs are.

The resources support seems like it should do this (not the exact structure we need), but I couldn't figure out how to make SwiftPM resources work at all for the plugin binary.

And I didn't see any way to get a path back to where the sources for the plugin exist on disk.

PR #1997, opened 2026-03-06
Add `Foundation` linkage test and last missing pieces

madsodgaard:
Adds linkage test to make sure we don't regress importing Foundation.

Also, the `.copy` resources in SwiftPM imports full Foundation, so we guard that behind Darwin.

Also fixes the last remaining Foundation part, which was `String.contains`.

## Result
`swift-protobuf` can now be built with `FoundationEssentials` only.


thomasvl:
Maybe put the LinkageTest in the CompileTests directory?  Not sure if other do, but I tend to expect `Sources/*` and `Tests/*` to map to things in the main `Package.swift` file.

Why do you need the `chmod +x`?  If the file was adding to git already executable shouldn't it come back down as executable?

thomasvl:
ps - the `contains` part is odd to me.  So that was didn't need an import of Foundation, but still caused it to get linked?

allevato:
> ps - the `contains` part is odd to me. So that was didn't need an import of Foundation, but still caused it to get linked?

If you have the following code:

```swift
_ = "foo".contains("f")
```

and you **don't** import Foundation, then the function that gets called is `Swift.(file).Sequence extension.contains [with Self -> String]`.

If you import Foundation, or—because Swift extensions are leaky—if a file in the same module imports Foundation or Foundation is imported somewhere in the transitive closure of your imports—then Foundation also defines a compatible `contains` overload and it gets picked instead: `Foundation.(file).StringProtocol extension.contains [with Self -> String T -> String]`, presumably because `StringProtocol` is narrower than `Sequence`.

It's gross. Compiling with `MemberImportVisibility` might avoid the problem and would be a good idea in the future, but we probably don't want to do that just yet, at least not as part of this change.

madsodgaard:
> Maybe put the LinkageTest in the CompileTests directory? Not sure if other do, but I tend to expect `Sources/*` and `Tests/*` to map to things in the main `Package.swift` file.

Sure! This is just what we did for other repos like openapi-generator and async-http-client, but I can move it there if you think its a better place for this repo 👍 

> Why do you need the chmod +x? If the file was adding to git already executable shouldn't it come back down as executable?

You are probably right yeah.

FranzBusch (approved):

madsodgaard:
thanks 🙏  - is it possible to get a new release tagged as well?😄

0xpablo:
Thanks for the FoundationEssentials adoption! I was trying it out in WASM (cross compiling from a Darwin host) and noticed that the full Foundation was still being linked.
I dug a bit deeper and it might be related to the `resource_bundle_accessor.swift` generated by:
```swift
#if canImport(Darwin)
let resources: [Resource] = [
    .copy("PrivacyInfo.xcprivacy")
]
#else
let resources = [Resource]()
#endif
```

Is there a better way or workaround for this? 
Thanks in advance

madsodgaard:
@0xpablo Unfortunately not, the issue is also tracked in SwiftPM: https://github.com/swiftlang/swift-package-manager/issues/9792

thomasvl:
I guess the other "fix" could be there was a way to say that resource is only needed when building *for* Apple platforms instead of the current gate, which matches building "on" Apple platforms.  i.e. - I'll if you did your WASM cross compiling from a Linux host, I'll bet things would have been fine.

PR #2019, opened 2026-04-07
[table-driven-branch] Support `google.protobuf.Any`.

allevato:
This change discards the previous "lazy" implementation of `Any` due to the complexity it would introduce to the new table-driven layout that's shared by all protos. No other language implementation supports this kind of unresolved-JSON preserving strategy and it introduces odd inconsistencies in how messages are serialized and parsed, so IMO we should simply retire it while we're considering breaking changes.

Because of this change, JSON and text encoding now allow an extension map to be passed in to support the case of an `Any` that wraps a message that has extensions set. This is necessary because the `Any` stores the binary-serialized blob of the message in its `value` field, so serializing that has to _deserialize_ it first in order to determine which extension fields are relevant and get their names. The Java implementation also does this, so this is not without precedent. (The C++ implementation is able to use a global extension registry via static initialization, so the same constraint doesn't apply. One day we might also be able to do that.)


Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
```diff
@@ -494,6 +497,100 @@ extension _MessageStorage {
...
+        try reader.scanner.skipRequiredObjectStart()
+
+        // An empty JSON object is allowed.
+        if reader.scanner.skipOptionalObjectEnd() {
+            clearValue(of: typeURLField, type: String.self)
+            clearValue(of: valueField, type: Data.self)
+            return
+        }
+
+        var typeURL = ""
```
thomasvl: Would optionals for these make more sense?

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
```diff
@@ -494,6 +497,100 @@ extension _MessageStorage {
...
+                    throw SwiftProtobufError.JSONDecoding.emptyAnyTypeURL()
+                }
+                break
+            }
+            try reader.scanner.skipRequiredComma()
+        }
+        jsonEncoder.endObject()
+
+        guard let messageSchema = Google_Protobuf_Any.messageSchema(forTypeURL: typeURL) else {
+            throw SwiftProtobufError.JSONDecoding.invalidAnyTypeURL(type_url: typeURL)
```
thomasvl: This feels wrong. It isn't that the url is "invalid" the problem is the type wasn't registered. So we probably need to add a new error. We didn't have this because it was hidden in other transcoding cases.

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
```diff
@@ -494,6 +497,100 @@ extension _MessageStorage {
...
+                try parseJSONBuffer(UnsafeRawBufferPointer(buffer))
+            }
+        }
+
+        updateValue(of: typeURLField, to: typeURL)
+        updateValue(of: valueField, to: try messageStorage.serializedBytes(partial: true, options: BinaryEncodingOptions()))
+    }
+
+    /// Returns a value indicating whether the `Any` type URL is valid.
+    func isTypeURLValid(_ typeURL: String) -> Bool {
```
thomasvl: Probably should pull the full comment over from AnyMessageStorage.

There is a note there about some languages also doing lookup, I'd still hold since we have a manual registration we should do "invalid" and "unknown" as different errors.

Sources/SwiftProtobuf/_MessageStorage+TextDecoding.swift
```diff
@@ -291,6 +306,33 @@ extension _MessageStorage {
...
         }
     }
+
+    /// Parses the next object from the input and interprets it as the expanded form of the
+    /// well-known type `Any` containing a message with the given type URL.
+    ///
+    /// - Precondition: The receiver must be the storage for `google.protobuf.Any`.
+    private func parseAsExpandedAny(from reader: inout TextFormatReader, typeURL: String) throws {
+        guard let messageSchema = Google_Protobuf_Any.messageSchema(forTypeURL: typeURL) else {
+            throw SwiftProtobufError.TextFormatDecoding.invalidAnyTypeURL(type_url: typeURL)
```
thomasvl, line 316: Same, unknown vs. invalid

Sources/SwiftProtobuf/_MessageStorage+TextEncoding.swift
```diff
@@ -28,16 +28,80 @@ extension _MessageStorage {
     /// A recursion helper that serializes the fields in the storage into the given text format encoder.
     func serializeText(into encoder: inout TextFormatEncoder, options: TextFormatEncodingOptions) {
         var mapEntryWorkingSpace = MapEntryWorkingSpace(ownerSchema: schema)
-        for field in schema.fields {
-            guard isPresent(field) else { continue }
-            serializeField(field, into: &encoder, mapEntryWorkingSpace: &mapEntryWorkingSpace, options: options)
+
+        switch CustomJSONWKTClassification(messageSchema: schema) {
```
thomasvl, line 32: Any preferences for `switch` of `if ... else` when there's only two "cases"?

Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
```diff
@@ -23,91 +23,57 @@ extension Google_Protobuf_Any {
...
         partial: Bool = false,
-        typePrefix: String = defaultAnyTypeURLPrefix
+        typeURLPrefix: String = defaultAnyTypeURLPrefix
     ) throws {
-        if !partial && !message.isInitialized {
-            throw BinaryEncodingError.missingRequiredFields
-        }
+        // Since we're passing a concrete message type here, we can auto-register it for
+        // convenience.
+        Google_Protobuf_Any.register(messageType: M.self)
```
thomasvl, line 43: this api has the ability to report if there was a conflict, I have no idea if we should bother to do so here.

Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
```diff
@@ -23,91 +23,57 @@ extension Google_Protobuf_Any {
...
+    @available(*, deprecated, renamed: "init(packing:partial:typeURLPrefix:)")
     public init(
-        textFormatString: String,
-        options: TextFormatDecodingOptions = TextFormatDecodingOptions(),
-        extensions: (any ExtensionMap)? = nil
+        message: any Message,
+        partial: Bool = false,
+        typePrefix: String = defaultAnyTypeURLPrefix
     ) throws {
+        // Since we're passing a concrete message type here, we can auto-register it for
```
thomasvl, line 70: Do you want to just call the other one?

Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
```diff
@@ -23,91 +23,57 @@ extension Google_Protobuf_Any {
...
+    ///   - partial: If `false` (the default), this method will check
+    ///     ``Message/isInitialized-6abgi`` before encoding to verify that all required
+    ///     fields are present. If any are missing, this method throws
+    ///     ``BinaryEncodingError/missingRequiredFields``.
+    ///   - typePrefix: The prefix to be used when building the `type_url`.
+    ///     Defaults to "type.googleapis.com".
+    /// - Throws: ``BinaryEncodingError/missingRequiredFields`` if
+    /// `partial` is false and `message` wasn't fully initialized.
+    @available(*, deprecated, renamed: "init(packing:partial:typeURLPrefix:)")
     public init(
```
thomasvl, line 65: Want to keep a deprecated shim for this for the time being?

Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
```diff
@@ -153,6 +153,14 @@ extension Google_Protobuf_Any {
         return result
     }
 
+    /// Returns the message schema expected for the given type URL.
+    public static func messageSchema(forTypeURL url: String) -> MessageSchema? {
```
thomasvl, line 157: public for reflection apis eventually?  Seems like it could be package/internal otherwise.

Sources/SwiftProtobuf/AnyMessageStorage.swift
thomasvl, line 1: Was the generator already updated to remove all the special casing of Any?  i.e. - nothing commented out left to delete/etc.

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
allevato: Right, I was trying to reproduce the current behavior here, but I'll add a new case for it instead.

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
allevato: I guess for `typeURL` it makes sense if we want to differentiate between "invalid" URL and "unregistered". For the cached value JSON I think it just makes it harder to use below.

thomasvl:
I should have said, its nice to see the hashing/equality issues in there going away.

Sources/SwiftProtobuf/_MessageStorage+TextEncoding.swift
allevato: Entirely mood based 🤷🏻‍♂️ 

I guess I'll update the other place where I used an `if` to use a `switch`, since it "feels better" to use a `switch` with an `enum`.

Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
allevato: I'm not following, what do you mean that isn't being done here?

Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
allevato: We can't since that one expects a generic type at compile time and this one just takes `any Message`, which being an existential doesn't conform to `Message`.

Sources/SwiftProtobuf/Google_Protobuf_Any+Registry.swift
allevato: Yeah, while we'll probably keep the concrete type versions of registration for the common case, I can imagine a situation where a reflection user wants to dynamically create a schema and then register that and maybe get it back out later.

Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
allevato: Can think about it as follow-up.

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
allevato: Done.

Sources/SwiftProtobuf/_MessageStorage+TextDecoding.swift
allevato: Done.

Sources/SwiftProtobuf/AnyMessageStorage.swift
allevato: I'm going to do cleanup all at once so that I can actually see the big picture of how much code we can delete once the old implementation is no longer in use. Right now, I'm only deleting code that largely won't compile.

Sources/SwiftProtobuf/Google_Protobuf_Any+Extensions.swift
thomasvl: I don't see a `public init(textFormatString:options:extesnions:)` which used to exist, no?

allevato: Because we can use the default implementation that's in `extension Message` in `Message+TextFormatAdditions` now; we don't need to override it with custom behavior.

thomasvl: ah!

thomasvl (approved):

PR #2020, opened 2026-04-07
[table-driven-branch] Implement JSON for `FieldMask`.

allevato:
I haven't brought over the full `FieldMask` tests because the vast majority of them depend on visitor/decoder infrastructure that's going to be deleted. When I do that deletion, this branch will be in a transitional state where those APIs are no longer supported. The plan will be to have them use the future reflection API instead, since that will be the only way to generically operate on a message.

### DIFF
--- a/Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
+++ b/Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
@@ -607,9 +607,10 @@ extension _MessageStorage {
     ///
     /// - Precondition: The receiver must be the storage for `google.protobuf.FieldMask`.
     private func parseAsFieldMask(from reader: inout JSONReader) throws {
+        let pathsField = schema[fieldNumber: 1]!
         let fieldMaskString = try reader.scanner.nextQuotedString()
         try parseFieldMask(fieldMaskString) { name in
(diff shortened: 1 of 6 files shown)


Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
```diff
@@ -603,6 +602,17 @@ extension _MessageStorage {
...
     }
 
+    /// Parses the next quoted string from the input and interprets it as the JSON representation
+    /// of a well-known type `FieldMask`.
+    ///
+    /// - Precondition: The receiver must be the storage for `google.protobuf.FieldMask`.
+    private func parseAsFieldMask(from reader: inout JSONReader) throws {
+        let fieldMaskString = try reader.scanner.nextQuotedString()
+        try parseFieldMask(fieldMaskString) { name in
+            appendValue(name, to: schema[fieldNumber: 1]!)
```
thomasvl: Is it worth looking up the schema in advance?

thomasvl (approved):

Sources/SwiftProtobuf/_MessageStorage+JSONDecoding.swift
allevato: The optimizer miiiiiiight take care of it but yes, there's no reason not to precompute it. Done.

PR #2021, opened 2026-04-09
Replace protobuf/abseil submodules with vendored sources

kateinoigakukun:
Fixes apple/swift-protobuf#2000.

SwiftPM recursively checks out git submodules during package resolution. The protobuf and abseil submodules are large C++ repositories, but every consumer has to pay the cost of fetching them for every checkout because SwiftPM cannot cache submodules as their bare repositories.

This PR replace protobuf/abseil submodules with vendored sources.

## Commits

1. Add protobuf vendor update automation
3. Remove protobuf and abseil submodule
4. Vendor protobuf and abseil

Reviewers can focus on the first commit; commits 2 and 3 are mechanical.

## Alternative Considered

git subtree was also considered but ruled out: it does not support selective path filtering across multiple directories (we only need a subset of files from each repo).


FranzBusch:
Marking this PR as a `minor` even though it doesn't change public API just to indicate that there is a larger internal change.

kateinoigakukun:
@FranzBusch I forgot to include .proto files used in tests. Would you mind kicking the CI again?

FranzBusch (approved):

allevato (approved):

FranzBusch:
Thanks @kateinoigakukun. I updated the branch and if everything passes this should go in. @thomasvl could you help us get a release out with this?

thomasvl:
> Thanks @kateinoigakukun. I updated the branch and if everything passes this should go in. @thomasvl could you help us get a release out with this?

Sure, once it is in, I can look at doing a release within a day or so.

kateinoigakukun:
Thanks!

.github/workflows/update_protobuf.yml
```diff
@@ -15,152 +15,56 @@ jobs:
...
-      if: steps.check_release.outputs.needs_update == 'true'
-      id: abseil_info
-      run: |
-        cd Sources/protobuf/abseil
-        ABSEIL_TAG=$(git describe --tags --abbrev=0 2>/dev/null || echo "commit ${{ steps.check_abseil.outputs.abseil_commit }}")
-        echo "abseil_version=$ABSEIL_TAG" >> $GITHUB_OUTPUT
-        cd ../../..
+    - name: Update vendored dependencies
+      id: update
+      run: python3 scripts/UpdateProtobufSubtrees.py --github-output "$GITHUB_OUTPUT"
```
thomasvl, line 32: Looks like this doesn't work on CI: https://github.com/apple/swift-protobuf/actions/runs/24355578742/job/71121508459

kateinoigakukun: Sorry, just opened a follow-up fix https://github.com/apple/swift-protobuf/pull/2025

carlos-santos-anchor:
~~Are we planning a new release after this change?~~

Apologies, I missed this PR: https://github.com/apple/swift-protobuf/pull/2024

<details><summary>Old question that does not make sense anymore</summary>
<p>

We rely on the version tag and we usually update submodules to get `Sources/protobuf/protobuf`.
Because this PR removed `.gitmodules` we have no way to get to the submodule files.

I am assuming the next tag will include the files, since they were moved in, right?

Also, is #2025 necessary for the fix?

</p>
</details>

PR #2024, opened 2026-04-13
Version bump for a release.

thomasvl:


carlos-santos-anchor:
Hi, 

we were hoping to see a new `1.37.0` coming out because this was a change that affected our internal workflows (as submodules were removed - our swift codegen process broke).

We are adapting our workflow to use the files in the repo and not in submodules, but we would need this version out.

Any idea when is this tag going to be released?

thomasvl:
@FranzBusch gentle ping.

thomasvl:
I want to confirm this is the right version bump for the change in how abseil and protobuf are pulled in.

FranzBusch (approved):

PR #2027, opened 2026-04-15
[table-driven-branch] Add new "reflection table" for names and default values and update generator.

allevato:
Add the new "reflection table" concept, which will hold the text/JSON names of fields/enum cases as well as default values (the latter is not yet implemented). This will replace the current `_NameMap`, and we'll have APIs on the schemas to look these names up.

The reflection table is generated but *not* yet hooked up. We're still generating the classic `_NameMap` as well, and that's still being used for the text/JSON serialization. `_NameMap` is tightly coupled with the encoders/scanners so I need to be able to delete all that code before migrating things cleanly over.

See the first commit description for details/rationale on the compression approach we're using. For expedience, I'm implemented a simple `NSLock`-guarded state to lazily decompress the tables on first access. We may want to iterate on this in the future, but many other recommended approaches are either platform-specific (`OSAllocatedUnfairLock`) or require too high of a minimum OS deployment version for our clients (`Synchronization.Mutex`) so it'll require extra care. The `DispatchQueue` approach we're using for the `Any` registry may not be suitable here because requesting the reflection data for a message/enum could potentially happen on a hot path. When we do encoding/decoding inside the library we can make sure to grab it once and pass it around as needed, but the reflection table itself won't be exposed to clients so the barrier could easily be hit inside a tight loop.

With the final commit, we're now generating descriptor.proto and the plugin protos with the new implementation, so the table driven implementation is now being used end-to-end and we can start removing old runtime support.


Sources/protoc-gen-swift/Compression+Compressing.swift
```diff
@@ -0,0 +1,297 @@
+import SwiftProtobuf
```
thomasvl, line 19: New files need the boilerplate.

Sources/SwiftProtobuf/Compression+Decompressing.swift
```diff
@@ -0,0 +1,227 @@
...
+        var reader = Bit7Reader(input: compressed)
+        guard let b0 = reader.nextByte(),
+            let b1 = reader.nextByte(),
+            let b2 = reader.nextByte(),
+            let b3 = reader.nextByte()
+        else {
+            preconditionFailure("Compressed data too short to contain size header")
+        }
+
+        let size = Int(UInt32(b0) | (UInt32(b1) << 8) | (UInt32(b2) << 16) | (UInt32(b3) << 24))
```
thomasvl: &<< etc? (and anywhere else in this related code)

Sources/SwiftProtobuf/Compression+Decompressing.swift
```diff
@@ -0,0 +1,227 @@
...
+
+        // The most significant bit of the size is reserved for future
+        // expansion.
+        guard (size & 0x8000_0000) == 0 else {
```
thomasvl, line 37: Not required, but does it make sense to reserve >1 bit?  i.e. - do we think we need that full size range in this support?

Sources/SwiftProtobuf/Compression+Decompressing.swift
```diff
@@ -0,0 +1,227 @@
...
+    /// Range coding allows us to represent symbols using fractional bits and to
+    /// do so adaptively so that we achieve good compression ratios for
+    /// arbitrary protos instead of trying to build static tables that may only
+    /// be optimal for some specific subset.
+    struct RangeDecoder: ~Copyable {
+        /// The lower bound of the current range.
+        private var low: UInt32 = 0
+
+        /// The size of the current range.
+        private var size: UInt32 = 0xFFFF_FFFF
```
thomasvl, line 173: Worth a comment about it being a marker can can't ever happen in reality?

Sources/SwiftProtobuf/Compression+Decompressing.swift
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    /// A helper structure that reads 7-bit encoded bytes from a stream and
+    /// reconstructs full 8-bit bytes.
+    struct Bit7Reader: ~Copyable {
+        /// The input data stream.
+        var input: UnsafeRawBufferPointer
+
+        /// An internal buffer to hold pending bits.
+        var buffer: UInt64 = 0
```
thomasvl: Why not UInt32 or UInt16, doesn't seem you can only have 7 to start and add 8 bits, hence a max of 15 bits would ever be temporarily needed?

Sources/SwiftProtobuf/Compression+Decompressing.swift
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        /// Reads the next 8-bit byte from the stream.
+        ///
+        /// - Returns: The next decoded 8-bit byte, or `nil` if the end of the
+        ///   stream is reached and not enough bits are available.
+        mutating func nextByte() -> UInt8? {
+            while bitCount < 8 {
+                guard let byte = input.first else {
+                    return nil
```
thomasvl, line 147: Is it acceptable to reach the end with some bits buffered or should that be some sort of error condition?  i.e. - are they just padding and should always be zero or something like that?

Sources/SwiftProtobuf/Compression+Decompressing.swift
```diff
@@ -0,0 +1,227 @@
...
+        /// Initializes a new range decoder and preloads the first 4 bytes of
+        /// code.
+        init(input: consuming Bit7Reader) {
+            self.input = input
+
+            for _ in 0..<4 {
+                if let byte = self.input.nextByte() {
+                    code = (code << 8) | UInt32(byte)
+                } else {
+                    code = code << 8
```
thomasvl: Interesting, so even if we run out of bytes, we keep shifting it over?

Sources/SwiftProtobuf/Compression+Decompressing.swift
```diff
@@ -0,0 +1,227 @@
...
+        ///   - total: The total frequency of all symbols in the model.
+        mutating func remove(cumulativeFrequency: UInt32, frequency: UInt32, total: UInt32) {
+            let r = size / total
+            low += cumulativeFrequency * r
+            code -= cumulativeFrequency * r
+            size = frequency * r
+
+            // Renormalization: If the top 8 bits of 'low' and 'low + size' are
+            // identical, they are fixed. We shift them out and read a new byte.
+            while (low ^ (low + size)) < 0x1000000 {
```
thomasvl: I wonder if `((low ^ (low + size)) & 0xFFFFFF) == 0)` is any more clear about the check. This took me a second, I guess I'm more used to bit tests using masking.

Sources/SwiftProtobuf/Compression+Decompressing.swift
```diff
@@ -0,0 +1,227 @@
...
+        /// state.
+        ///
+        /// - Parameters:
+        ///   - cumulativeFrequency: The cumulative frequency of the decoded
+        ///     symbol.
+        ///   - frequency: The individual frequency of the decoded symbol.
+        ///   - total: The total frequency of all symbols in the model.
+        mutating func remove(cumulativeFrequency: UInt32, frequency: UInt32, total: UInt32) {
+            let r = size / total
+            low += cumulativeFrequency * r
```
thomasvl: `&` on any of these through out?

Sources/SwiftProtobuf/Compression.swift
```diff
@@ -0,0 +1,255 @@
+/// A namespace for a custom compression algorithm designed for short strings,
```
thomasvl, line 16: header.

Sources/SwiftProtobuf/Compression+Decompressing.swift
```diff
@@ -0,0 +1,227 @@
+import Foundation
```
thomasvl, line 15: header.

Tests/protoc-gen-swiftTests/Test_Compression.swift
```diff
@@ -0,0 +1,66 @@
+import SwiftProtobuf
```
thomasvl, line 15: header

Sources/SwiftProtobuf/Compression.swift
```diff
@@ -0,0 +1,255 @@
...
+                self.code = char &- UInt8(ascii: "0") &+ 54
+            default:
+                return nil
+            }
+        }
+
+        /// Creates a new `Symbol` from its numeric value.
+        @inline(__always)
+        package init(code: UInt8) {
+            self.code = code
```
thomasvl, line 121: So this allows values in that aren't allowed?

Sources/protoc-gen-swift/EnumSchemaCalculator.swift
```diff
@@ -0,0 +1,42 @@
...
+
+    /// The Swift string literals (without surrounding quotes) that encode the enum schema in
+    /// the generated source.
+    var schemaLiteral: String { schemaWriter.schemaCode }
+
+    /// Creates a new enum schema calculator for an enum with the given name.
+    init(fullyQualifiedName: String) {
+        self.schemaWriter = .init()
+
+        schemaWriter.writeBase128Int(0, byteWidth: 1)
```
thomasvl, line 39: Maybe a reference to this payload some place?  (version, length, name, etc.)

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -339,19 +338,22 @@ class MessageGenerator {
...
-        messageLayoutCalculator.layoutLiterals.printConditionalBlocks(to: &p) { value, _, p in
+        messageSchemaCalculator.schemaLiterals.printConditionalBlocks(to: &p) { value, _, p in
             p.print(
                 "@_alwaysEmitIntoClient @inline(__always)",
                 #"private static var _protobuf_messageSchemaString: StaticString { "\#(value)" }"#
             )
         }
-
+        p.print(
+            "@_alwaysEmitIntoClient @inline(__always)",
```
thomasvl, line 351: Maybe I'm forgetting something?  Why does the schema go into the client vs. live in the generated code module?

Sources/SwiftProtobuf/ReflectionTable.swift
```diff
@@ -0,0 +1,259 @@
...
+                UInt8(ascii: "e"), 0,
+                // Alignment padding
+                0, 0,
+            ]
+        )
+    }()
+}
+
+/// A reflection table that is lazily decompressed when first accessed.
+final class LazyReflectionTable {
```
thomasvl, line 219: This isn't used yet, right?  Or am I missing it?

Sources/protoc-gen-swift/Compression+Compressing.swift
allevato: Done.

Sources/protoc-gen-swift/MessageGenerator.swift
allevato: The idea was to ensure that this always ends up becoming a simple pointer access and not a call to some accessor (and indeed, no accessor generated). But I should double check that this doesn't cause the string itself to be duplicated into the client. Even if the linker merged them, it'd be annoying.

I'll double check this when I start doing other cleanup.

Sources/SwiftProtobuf/Compression+Decompressing.swift
allevato: Done.

Sources/SwiftProtobuf/Compression+Decompressing.swift
allevato: Done.

Sources/SwiftProtobuf/Compression+Decompressing.swift
allevato: My thinking was that if we need more data later one, we can just place it after the size, so we only need 1 bit to say whether there's more stuff coming later before the compressed blob.

Sources/SwiftProtobuf/Compression+Decompressing.swift
allevato: You're right, this can be a `UInt16` (in both compression and decompression). I guess the model was being defensive here.

Sources/SwiftProtobuf/Compression+Decompressing.swift
allevato: Since we're packing bytes into 7 bit chunks, any data that's not a multiple of 56 bits would have padding left over. It's safe to ignore that.

Sources/SwiftProtobuf/Compression+Decompressing.swift
allevato: Done.

Sources/SwiftProtobuf/Compression+Decompressing.swift
allevato: This mirrors the range encoder, which always emits four bytes to flush the final state. But that means that we should only hit the `else` block on truncated compressed data, which shouldn't happen, so we can make it a precondition failure instead.

Sources/SwiftProtobuf/Compression+Decompressing.swift
allevato: Done.

Sources/SwiftProtobuf/Compression+Decompressing.swift
allevato: How about just `(low & 0xFF000000) == ((low &+ size) & 0xFF000000)` and trust the compiler to optimize it?

Sources/SwiftProtobuf/Compression.swift
allevato: Done.

Sources/SwiftProtobuf/Compression.swift
allevato: Added a `precondition`.

Sources/SwiftProtobuf/ReflectionTable.swift
allevato: Right, it'll be added into the schemas once I can remove the hard dependencies on `_NameMap`. Right now things are still too tightly coupled.

Tests/protoc-gen-swiftTests/Test_Compression.swift
allevato: Done.

Sources/protoc-gen-swift/EnumSchemaCalculator.swift
allevato: Done.

Sources/SwiftProtobuf/Compression+Decompressing.swift
thomasvl: That works, thanks.

thomasvl (approved):

PR #2028, opened 2026-04-15
[table-driven-branch] Remove all `Visitor`/`Decoder` APIs and support code.

allevato:
Now that the generator is working end-to-end with table-driven protos, we can start deleting the bits that are no longer in use. Likewise, the extension map support has been unified; `NewExtensionMap` is now just named `ExtensionMap` and that is the single concrete type for these things—there is no more protocol.

There are some things that I've commented out and will need to fix or have deleted and will need to bring back, such as

* Message set support
* JSON top-level arrays of messages
* The conformance test target
* Various field mask helper APIs

This leaves off `NameMap` cleanup/migration to the new reflection table, since that requires designing the right cache for those uncompressed tables.


Sources/SwiftProtobuf/Message+BinaryAdditions.swift
```diff
@@ -33,54 +33,11 @@ extension Message {
...
-            // an internal error in the library and should never happen.
-            assert(visitor.encoder.remainder.count == 0)
-        }
-        return data
-    }
-
-    /// Returns the size in bytes required to encode the message in binary format.
-    /// This is used by `serializedData()` to precalculate the size of the buffer
-    /// so that encoding can proceed without bounds checks or reallocation.
     internal func serializedDataSize() throws -> Int {
```
thomasvl: Didn't want to remove it yet?

Sources/SwiftProtobuf/Message.swift
```diff
@@ -102,7 +102,7 @@ public protocol Message: Sendable, CustomDebugStringConvertible {
     /// `traverse` methods.  The details get a little involved due to
     /// the need to allow particular messages to override particular
     /// behaviors for specific encodings, but the general idea is quite simple.
-    func traverse<V: Visitor>(visitor: inout V) throws
+    // func traverse<V: Visitor>(visitor: inout V) throws
```
thomasvl: delete

Sources/SwiftProtobuf/Message.swift
```diff
@@ -162,11 +149,13 @@ extension Message {
     }
 
     /// A hash based on the message's full contents.
-    public func hash(into hasher: inout Hasher) {
-        var visitor = HashVisitor(hasher)
-        try? traverse(visitor: &visitor)
-        hasher = visitor.hasher
-    }
+    /// TODO: Delete this when all of the other test protos are generated with
```
thomasvl: Has that not happened?

Sources/SwiftProtobuf/Message.swift
```diff
@@ -206,6 +195,17 @@ extension Message {
...
     }
+
+    // TODO: None of the generated protos use these anymore. Once we've removed
+    // code in the runtime that depends on these requirements existing in the
+    // `Message` protocol, delete these default implementations and the protocol
+    // requirements.
+    public mutating func decodeMessage<D: Decoder>(decoder: inout D) throws {
+        fatalError()
+    }
+    // public func traverse<V: Visitor>(visitor: inout V) throws {
```
thomasvl: delete?

Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_Any.swift
```diff
@@ -1169,19 +1169,12 @@ struct ConflictingImportMessage:
...
-    }
-
     static let _protobuf_nameMap: SwiftProtobuf._NameMap = SwiftProtobuf._NameMap()
 
     static func == (lhs: ConflictingImportMessage, rhs: ConflictingImportMessage) -> Bool {
         if lhs.unknownFields != rhs.unknownFields { return false }
         return true
     }
+
+    func hash(into hasher: inout Hasher) {}
```
thomasvl: Guess this whole things need to be regenerated and checked copied here?

Sources/SwiftProtobuf/Message+BinaryAdditions.swift
allevato: It was still needed at some point during the transition period, but I can remove it now.

Sources/SwiftProtobuf/Message.swift
allevato: Done.

Sources/SwiftProtobuf/Message.swift
allevato: Yeah, this can go now.

Sources/SwiftProtobuf/Message.swift
allevato: Done.

Tests/ExperimentalTableDrivenSwiftProtobufTests/Test_Any.swift
allevato: We should actually be able to provide default impls for this and `==` now. Updated the generator to do that.

thomasvl (approved):

PR #2032, opened 2026-04-16
[table-driven-branch] Stop generating legacy name maps.

allevato:
This PR stops generating the `_NameMap`s in the generated protos and removes all code pertaining to that type.

To migrate to the new reflection table quickly, I've set up `MessageSchema` and `EnumSchema` to just immediately decompress them when they're created (which would be the first time the schema of a generated message is accessed). That's not what we want long term—we can delay it even longer until such a message is parsed or serialized in text/JSON format. That will require some synchronization/caching, so I'll do it in a separate change.

These name tables operate on `String`s for expediency, which means I've dropped some optimizations in parsing/serialization where we would operate on raw buffers instead of materializing strings. I think we can get those back later (would be nice if we could use `UTF8Span` but it's too new), but since there's a likelihood that we look at replacing the scanners with something new, I figure it's better to wait for that to pan out.


Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -105,7 +105,11 @@ class MessageGenerator {
 
         self.messageSchemaCalculator = MessageSchemaCalculator(
             fullyQualifiedName: descriptor.fullName, fieldsSortedByNumber: fieldsSortedByNumber)
-        self.compressedReflectionData = ReflectionTableCalculator(fields: fieldsSortedByNumber).stringLiteral()
+        self.compressedReflectionData = ReflectionTableCalculator(
+            fields: fieldsSortedByNumber,
+            reservedRanges: descriptor.reservedRanges,
```
thomasvl, line 112: you might want to look at using `sortAndMergeContinuous` on this like we do for extension ranges. This could also help compact the data down.

Sources/protoc-gen-swift/ReflectionTableCalculator.swift
```diff
@@ -164,43 +178,98 @@ package struct ReflectionTableCalculator {
...
+            for (number, textOffset) in numberToTextOffset.sorted(by: { $0.key < $1.key }) {
+                appendInteger(number)
+                appendInteger(textOffset)
+            }
 
-        // JSON format name offset to field/case number (UInt32 -> UInt32), in alphabetical order.
-        for (offsetAndName, number) in jsonOffsetToNumber.sorted(by: { $0.key.name < $1.key.name}) {
-            appendInteger(offsetAndName.offset)
-            appendInteger(number)
+            // Text format name offset to field/case number (UInt32 -> UInt32), in alphabetical order.
```
thomasvl, line 235: Nit: you might want to do s/Text format/TextFormat/, I try to be consistent to make searching/etc. easier.

Sources/protoc-gen-swift/ReflectionTableCalculator.swift
```diff
@@ -107,14 +119,16 @@ package struct ReflectionTableCalculator {
 
         /// Reserves space for an aligned integer of the given type and returns the index where it
         /// should be written later by calling `updateInteger`.
-        func reserveInteger<T: BinaryInteger & FixedWidthInteger>(type: T.Type) -> Int {
-            align(to: T.self)
-            let index = result.endIndex
-            for _ in 0..<MemoryLayout<T>.size {
+        func appendSection(body: () -> Void) {
+            align(to: UInt32.self)
```
thomasvl: Is this needed or should it assert that the current offset is already aligned?

Sources/protoc-gen-swift/ReflectionTableCalculator.swift
```diff
@@ -164,43 +178,98 @@ package struct ReflectionTableCalculator {
...
+                """
+                The current metadata format only supports a maximum of 2^16 reserved names, \
+                2^16 single-element reserved ranges, and 2^16 multi-element reserved ranges.
+                """)
+        }
+        singleElementReservedRanges.sort()
+        multipleElementReservedRanges.sort { $0.lowerBound < $1.lowerBound }
+
+        // This section contains bidirectional mappings between numbers and names.
+        appendSection {
```
thomasvl, line 218: Given the section gets a length, does it make sense to not write anything in this section if there are no reserved numbers/names?  i.e. - when reading if the section size was zero, then you know there are no entries in either list?

Sources/SwiftProtobuf/JSONScanner.swift
```diff
@@ -1281,29 +1281,22 @@ internal struct JSONScanner {
...
+                if let fieldNumber = messageSchema.fieldNumber(forJSONName: key) {
                     return fieldNumber
                 }
-                if let s = utf8ToString(bytes: key.baseAddress!, count: key.count) {
-                    fieldName = s
-                } else {
-                    throw JSONDecodingError.invalidUTF8
-                }
+                fieldName = key
+//                throw JSONDecodingError.invalidUTF8
```
thomasvl: delete?

Sources/SwiftProtobuf/ReflectionTable.swift
```diff
@@ -64,46 +64,150 @@ extension ReflectionTable {
 
     /// Returns the field number with the given text name.
     package func fieldNumber(forTextName name: String) -> UInt32? {
-        number(for: name, inTableStartingAt: 8 + fieldCount * 8, count: fieldCount)
+        let num = rawNumber(forTextName: name)
+        // We might hit a reserved name with a number of 0, so convert that to nil.
```
thomasvl, line 84: Would callers ever need to know is it reserved vs. invalid?  return nil seems like it would make callers not know the difference, and I thought for parsing we ended up having to know to properly handling the option for not skipping unknown fields (but still skipping reserved names).

Sources/SwiftProtobuf/ReflectionTable.swift
```diff
@@ -64,46 +64,150 @@ extension ReflectionTable {
...
+        guard let number = rawNumber(forJSONName: name) ?? rawNumber(forTextName: name) else { return nil }
         return Int32(bitPattern: number)
     }
 }
 
+extension ReflectionTable {
+    /// Returns true if the given field number is reserved.
+    package func isNumberReserved(_ number: UInt32) -> Bool {
+        guard data.count >= 4 else { return false }
+        let section2Offset = data.withUnsafeBytes { bytes in
```
thomasvl: I may need to pull this down to trace it better, but what I'm following at the moment:

- In the api to write a section, there was that call to start out Int32 aligned, I don't see something matching here. So if that did pad, how is that accounted for here?
- In that same section writer, it pads at the end. I don't think we have to account for it because we have explicit sub section counts, so maybe we don't have to account for that and we're still good?

Sources/SwiftProtobuf/ReflectionTable.swift
```diff
@@ -64,46 +64,150 @@ extension ReflectionTable {
...
-        guard let number = fieldNumber(forJSONName: name) else { return nil }
+        guard let number = rawNumber(forJSONName: name) ?? rawNumber(forTextName: name) else { return nil }
         return Int32(bitPattern: number)
     }
 }
 
+extension ReflectionTable {
+    /// Returns true if the given field number is reserved.
+    package func isNumberReserved(_ number: UInt32) -> Bool {
+        guard data.count >= 4 else { return false }
```
thomasvl: less that 4 (and maybe no zero), could also be an error/corruption, no?  Guess it isn't worth signaling that?

Sources/SwiftProtobuf/ReflectionTable.swift
```diff
@@ -64,46 +64,150 @@ extension ReflectionTable {
...
+        let section2Offset = data.withUnsafeBytes { bytes in
+            Int(bytes.load(fromByteOffset: 0, as: UInt32.self))
+        }
+        guard section2Offset > 0 && section2Offset < data.count else { return false }
+        
+        return data.withUnsafeBytes { bytes in
+            let endOffset = Int(bytes.load(fromByteOffset: section2Offset, as: UInt32.self))
+            if endOffset - section2Offset <= 8 {
```
thomasvl: All these 8 & 4 values, should they be using UInt32's memory layout size given that's how we load them vs. just being hardcoded?

Sources/SwiftProtobuf/ReflectionTable.swift
```diff
@@ -64,46 +64,150 @@ extension ReflectionTable {
...
+extension ReflectionTable {
+    /// Returns true if the given field number is reserved.
+    package func isNumberReserved(_ number: UInt32) -> Bool {
+        guard data.count >= 4 else { return false }
+        let section2Offset = data.withUnsafeBytes { bytes in
+            Int(bytes.load(fromByteOffset: 0, as: UInt32.self))
+        }
+        guard section2Offset > 0 && section2Offset < data.count else { return false }
+        
+        return data.withUnsafeBytes { bytes in
```
thomasvl: &+, &*, etc. all though here to shrink it?

Sources/SwiftProtobuf/ReflectionTable.swift
```diff
@@ -64,46 +64,150 @@ extension ReflectionTable {
...
+            guard multiStart + multiCount * 8 <= data.count else { return false }
+            low = 0
+            high = multiCount - 1
+            var candidateIndex = -1
+
+            while low <= high {
+                let mid = (low + high) / 2
+                let offset = multiStart + mid * 8
+                let lower = bytes.load(fromByteOffset: offset, as: Int32.self)
+                let uLower = UInt32(bitPattern: lower)
```
thomasvl: Why not just load it as a UInt32, you don't seen to need `lower`

Sources/SwiftProtobuf/TextFormatEncoder.swift
```diff
@@ -46,10 +46,6 @@ internal struct TextFormatEncoder {
         data.append(contentsOf: buff)
     }
 
-    internal mutating func append(name: _NameMap.Name) {
```
thomasvl, line 49: Should there be apis to get the raw byte subanges out of the tables for both enums and fields and diectly add them to avoid the conversion to/from strings?

Sources/protoc-gen-swift/MessageGenerator.swift
allevato: Added a TODO to come back to that since it doesn't require a breaking change in the reflection data layout.

Sources/protoc-gen-swift/ReflectionTableCalculator.swift
allevato: I guess it's not needed since we also align at the end of the section as well. Replaced with an assert.

Sources/protoc-gen-swift/ReflectionTableCalculator.swift
allevato: Do you mean the next section (this is the first section with regular fields + reserved names, the second section contains reserved ranges).

Each section isn't preceded by its length, but rather the offset of the next section, so a section is empty if its next section offset is the current offset + 4.

But yes, we could have a fast path that just exits early when writing the second section if there's no data. Then the runtime could check the offset+4 condition and exit early as well.

Regenerated after that and it shaves off the expected 4 bytes from the WKTs.

Sources/protoc-gen-swift/ReflectionTableCalculator.swift
allevato: Yeah, there are a couple functions where I use `text` vs `textFormat` as well, I'll try to do a cleanup pass later.

Sources/SwiftProtobuf/JSONScanner.swift
allevato: Done.

Sources/SwiftProtobuf/ReflectionTable.swift
thomasvl: Thinks some more about this over night, should we just hard code things asserts in the generator and runtime that check that sizes are what we expect?  Even if we coded everything using memory layout, that doesn't ensure generator sizes match the runtime sizes. Cause we're making binary payloads potentially on different platforms since you can generate on a different OS/chip then where you run.

Sources/SwiftProtobuf/TextFormatEncoder.swift
allevato: Yeah, I mention that in the PR description. If we're looking at replacing the scanners/tokenizer as well, then it's worth waiting until that lands so that we can have the right shape for both decoding and encoding.

Sources/SwiftProtobuf/ReflectionTable.swift
thomasvl: This is do deal with endian issues?  Maybe comment that?  Should it then be explicitly using little endian initializers any place?  Did we do that when writing them out also?

Sources/SwiftProtobuf/ReflectionTable.swift
allevato: The current flow matches what the decoder was doing previously: asking for the field number with the given name, and if it found nothing, checking if it's reserved. So this returns nil and then it makes the reserved check.

We could potentially fold those into a single operation by returning an enum that's shaped like `case reserved; case number(UInt32)`, but I'd have to make sure that doesn't make it more annoying to use the API for callers who don't care about the reserved fields.

Sources/SwiftProtobuf/ReflectionTable.swift
allevato: I did a bunch of refactoring of this method and now it has a precondition.

Sources/SwiftProtobuf/ReflectionTable.swift
```diff
@@ -64,46 +64,150 @@ extension ReflectionTable {
...
     package func enumCase(forJSONName name: String) -> Int32? {
-        guard let number = fieldNumber(forJSONName: name) else { return nil }
+        guard let number = rawNumber(forJSONName: name) ?? rawNumber(forTextName: name) else { return nil }
         return Int32(bitPattern: number)
     }
 }
 
+extension ReflectionTable {
+    /// Returns true if the given field number is reserved.
+    package func isNumberReserved(_ number: UInt32) -> Bool {
```
thomasvl, line 109: Looking back at https://github.com/apple/swift-protobuf/blob/main/Sources/SwiftProtobuf/NameMap.swift#L534-L542, that was a linear search. They values are sorted, so we could even stop search the moment the end is > the value being checked. So the complexity of the binary searches seems overkill given how in frequently this is going to be hit. So I sorta wonder if we should do a really simple linear search and be done with it. We can always add this if we really come up with it being a hot spot, but smaller/simple for this sorta edge case seems like it might be the better tradeoff since the driver for all this is smaller.

Sources/SwiftProtobuf/ReflectionTable.swift
allevato: PTAL at the new refactor that hopefully makes this a lot clearer.

Sources/SwiftProtobuf/ReflectionTable.swift
allevato: Done. (There are a couple `+` that had to remain because they're pointer math and pointers only support `+`, not `&+`.)

Sources/SwiftProtobuf/ReflectionTable.swift
allevato: Yeah, I don't know what happened here. Gemini's fault 🙃

But thanks for mentioning the endianness issues, I had overlooked that here. Updated all these loads to use helper functions that auto-convert to the correct endianness.

Sources/SwiftProtobuf/ReflectionTable.swift
allevato: Done. Left a comment that we can dynamically choose a binary search in the future if this ever becomes a bottleneck.

Sources/SwiftProtobuf/ReflectionTable.swift
allevato: I've updated these to use `MemoryLayout` to be clearer.

Long-term, I think there are a few places in these layouts where it might be a good idea to have a `checkInvariants` function that's only called from preconditions so that debug builds will exercise them but they get compiled out in release builds. We could do that at the time that the data is decompressed (or in the case of message schemas, when it's created). I can look at that as follow-up work.

Sources/protoc-gen-swift/ReflectionTableCalculator.swift
thomasvl: Yea, sorry, comment wasn't to clear.  Generally, any place we can exploit zero "length" seems like it might be worth doing so.

Sources/SwiftProtobuf/ReflectionTable.swift
thomasvl: ack

Sources/SwiftProtobuf/ReflectionTable.swift
thomasvl: Sounds good.  I the fact that we lean into binary packing makes this seem more important from a safty pov.

thomasvl (approved):

PR #2037, opened 2026-04-21
Update FuzzTesting Package.swift.

thomasvl:
Now that https://github.com/google/oss-fuzz/pull/14342 has landed, we can properly updated FuzzTesting/Package.swift to also be 6.x based.

This pulls over what @glbrntt's had attempted in #1913 but couldn't be landed because of issues in 6.0/6.1, but now that we only do fuzz testing on 6.2+ it is safe.

It also sets the explicit `traits` for SwiftProtobuf since the FieldMask additions aren't needed.


glbrntt (approved):
Nice!

PR #2040, opened 2026-04-23
[table-driven-branch] Migrate everything over to table-driven protos.

allevato:
Complete the migration and fix up various unit tests as needed.

As of this PR, all the original unit tests pass with the table-driven approach except for functionality that hasn't been re-implemented yet (JSON arrays, message sets, field masks).

Commits have been made small and targeted to individual bugs that were discovered while running the full test suite.

Conformance tests still need work as well; will do that next.


Sources/SwiftProtobufPluginLibrary/NamingUtils.swift
```diff
@@ -38,22 +38,15 @@ private let reservedTypeNames: Set<String> = {
...
     // Basic Message properties we don't want to shadow:
     names.insert("isInitialized")
     names.insert("unknownFields")
+    names.insert("messageSchema")
 
     // Standard Swift property names we don't want
     // to conflict with:
     names.insert("debugDescription")
     names.insert("description")
-    names.insert("dynamicType")
```
thomasvl, line 56: I wonder if any of these will result in other breaks on folks.

Sources/SwiftProtobuf/MessageStorage+JSONDecoding.swift
```diff
@@ -197,7 +197,11 @@ extension MessageStorage {
                     appendValue(try reader.scanner.nextDouble(), to: field)
 
                 case .enum:
-                    try scanEnumValue(field, from: &reader, operation: .append)
+                    do {
+                        try scanEnumValue(field, from: &reader, operation: .append)
+                    } catch JSONDecodingError.unrecognizedEnumValue where reader.options.ignoreUnknownFields {
```
thomasvl, line 202: Do these work correctly if it is a repeated field of enums and in the middle there is an unknown one?  i.e. - do we properly set the before/after ones?

Part of me wonders if there is a better way to move the error down a level. If we wanted to overhaul the json error handling to use a more generic set of errors, we might lose this specific one as it isn't really useful from a public api pov.

Sources/SwiftProtobuf/MessageStorage+JSONDecoding.swift
```diff
@@ -513,6 +527,15 @@ extension MessageStorage {
             }
             hasSeenValue = true
 
+            if reader.scanner.skipOptionalNull() {
```
thomasvl, line 540: Can this go after the check for quoted strings and raw numbers?  i.e. - those are likely to be way more common, so it seems like we should do this check after those.

Sources/SwiftProtobuf/MessageStorage+TextEncoding.swift
```diff
@@ -121,11 +121,17 @@ extension MessageStorage {
...
                 mapEntryWorkingSpace.storage(for: field.submessageIndex),
                 .read,
-                false  // useDeterministicOrdering
+                true  // useDeterministicOrdering
             ) {
                 emitName(ofFieldNumber: fieldNumber, into: &encoder)
                 encoder.startMessageField()
-                $0.serializeText(into: &encoder, options: options)
+                let mapEntrySchema = $0.schema
+                if let keyField = mapEntrySchema[fieldNumber: 1] {
```
thomasvl: Can these lets fail?  i.e. - could we still leave off one of these? (force unwrap?)

Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
```diff
@@ -135,14 +136,29 @@ extension MessageStorage {
...
+                var subReader = WireFormatReader(buffer: slice, recursionBudget: reader.recursionBudget)
+                try workingSpace.merge(
+                    byReadingFrom: &subReader,
+                    extensions: extensions,
+                    partial: partial,
+                    discardUnknownFields: false
+                )
+                // If we saw any unknown enum cases, they will end up in unknown fields, so return
+                // that to indicate that the entire map entry should be moved there.
+                return workingSpace.unknownFields.data.isEmpty
```
thomasvl: If there was field 3 within the map entry wouldn't it also end up here?  i.e. - is this really limited to unknown enum entries?

Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
```diff
@@ -135,14 +136,29 @@ extension MessageStorage {
...
+            }
+            if !success {
+                // Re-parse the entire map entry as unknown fields.
+                let fieldTag = FieldTag(fieldNumber: tag.fieldNumber, wireFormat: .lengthDelimited)
+                let bodySize = slice.count
+                let fieldSize = fieldTag.encodedSize + Varint.encodedSize(of: Int64(bodySize)) + bodySize
+                var fieldData = Data(count: fieldSize)
+                fieldData.withUnsafeMutableBytes { body in
```
thomasvl: If you wanted to reduce memory usage here, you could only write the tag and the size in fieldData, append that, and then append `slice`.  i.e. - skip one of the copies of `slice`.

Sources/SwiftProtobuf/MessageSchema.swift
```diff
@@ -463,7 +463,21 @@ extension MessageSchema {
...
     func fieldNumber(forTextName name: String) -> UInt32? {
-        reflection.table.fieldNumber(forTextName: name)
+        // Fast path: Binary search in the reflection table.
+        if let number = reflection.table.fieldNumber(forTextName: name) {
+            return number
+        }
+        // Slow path: If it wasn't found, check if it's a group name spelled in
+        // lowercase form.
+        let lowercaseName = name.lowercased()
+        for field in fields where field.rawFieldType == .group {
```
thomasvl, line 473: I guess the other choice would be to use something like the enum "alias" support to have these names. Not sure if that's worth it though.

allevato: Yeah, I mainly wanted to avoid bloating every group-containing message with an additional name that is trivially computable. It sucks that we can't do binary search with the lowercased form but it should be rare enough to not matter.

Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
allevato: Done.

Sources/SwiftProtobuf/MessageStorage+JSONDecoding.swift
allevato: It should, because this is nested inside the `scanArray` call, so this particular piece of code is called once per tokenized array element. I'm pretty sure we have tests for this case.

Sources/SwiftProtobuf/MessageStorage+JSONDecoding.swift
allevato: We don't have a method in the scanner like `skipOptionalSInt64`; we can only attempt to scan an integer and then fail by throwing. I've moved the null check under the string check, but if we want to go all the way down, we might want to revisit that with the new tokenizer.

Sources/SwiftProtobuf/MessageStorage+TextEncoding.swift
allevato: Yeah, these should never fail. Done.

Sources/SwiftProtobufPluginLibrary/NamingUtils.swift
allevato: I guess it's possible that it would turn a `dynamicType_p` generation back to just `dynamicType` (and vice versa for `messageSchema`, which is unavoidable).

Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
allevato: Yeah, this got changed by a later commit where we start throwing an error for invalid enum cases to distinguish it. We have tests that are checking that extra fields are ignored and they correctly pass.

thomasvl (approved):

PR #2041, opened 2026-04-24
[table-driven-branch] Fixes to get all conformance tests passing.

allevato:
As of this PR, all conformance tests pass except for one involving message sets, which haven't been implemented yet in the new runtime.


Sources/protoc-gen-swift/MessageFieldGenerator.swift
```diff
@@ -97,7 +97,9 @@ class MessageFieldGenerator: FieldGeneratorBase, FieldGenerator {
             trampolineFieldKind = .message(swiftSingularType, isArray: descriptor.isRepeated)
         case .message:
             if let mapKeyAndValue = descriptor.messageType!.mapKeyAndValue {
-                trampolineFieldKind = .map(swiftType, valueIsMessage: mapKeyAndValue.value.type == .message)
+                let keyType = RawFieldType(fieldDescriptorType: mapKeyAndValue.key.type)
+                let valueType = RawFieldType(fieldDescriptorType: mapKeyAndValue.value.type)
+                trampolineFieldKind = .map(swiftType, keyType: keyType, valueType: valueType)
```
thomasvl, line 102: Do you really need the swift type since it came from the key/value pair?

thomasvl (approved):

Sources/protoc-gen-swift/MessageFieldGenerator.swift
allevato: Yeah, because if the value is just `message` or `group`, we also need the dictionary type to contain the actual Swift name of the message type used in the value.

allevato: (Although I guess for a map, it'll only ever be message, not group.)

PR #2043, opened 2026-04-24
[table-driven-branch] Improve docs about schema layouts and remove some stale TODOs.

allevato:


Sources/SwiftProtobuf/ReflectionTable.swift
```diff
@@ -16,6 +16,90 @@
...
+///
+/// The reflection table consists of a sequence of sections. The first 4 bytes of the table contain
+/// the offset to the next section (Section 1).
+///
+/// ### Global Layout
+///
+/// To support future extensions without breaking ABI, the layout is divided into sections. Each
+/// section starts with a 4-byte little-endian integer indicating the offset to the next section.
+/// It is possible for a section to be empty (the offset to the next section will simply be the
+/// current offset + 4).
```
thomasvl: Not sure if it is worth calling out: This can serve as shorthand for any section that has "lists" where they would be empty, i.e. - don't have to encode the "zero" length also. And/or we note this on the sections where we make use of it.

Sources/SwiftProtobuf/ReflectionTable.swift
```diff
@@ -16,6 +16,90 @@
...
+/// | Bytes 0-1           | Bytes 2-3           | Bytes 4-5       | Bytes 6-7   | Bytes 8...      |
+/// | Distinct JSON count | Reserved name count | Text name count | (Padding)   | Tables and text |
+/// +---------------------+---------------------+-----------------+-------------+-----------------+
+/// ```
+///
+/// *   **Distinct JSON count**: The number of fields whose JSON names differ from their text format
+///     names.
+/// *   **Reserved name count**: The number of reserved names.
+/// *   **Text name count**: The total count of entries in the text name to number table.
+///
```
thomasvl, line 69: Do we assert that padding is zero when reading?  i.e - to potentially allow it to be used in the future?

Sources/SwiftProtobuf/ReflectionTable.swift
```diff
@@ -16,6 +16,90 @@
...
+/// into UTF-8 `StaticString`s, reflection tables are packed *after* being compressed. This means
+/// that the uncompressed table layout, which is described below, can use standard little-endian
+/// fixed-size integers.
+///
+/// The reflection table consists of a sequence of sections. The first 4 bytes of the table contain
+/// the offset to the next section (Section 1).
+///
+/// ### Global Layout
+///
+/// To support future extensions without breaking ABI, the layout is divided into sections. Each
```
thomasvl, line 32: Aren't sections also always 4 byte aligned? If yes, is the last one also padded, or does the alignment only happen on starting a section?

thomasvl (approved):

Sources/SwiftProtobuf/ReflectionTable.swift
allevato: Added a mention of the alignment. Section _starts_ are aligned, but we don't bother padding the last section since it would be wasted space (size vs. stride, basically).

Sources/SwiftProtobuf/ReflectionTable.swift
allevato: I think that would only matter if the sole section contents were a list and didn't need any other header information, so I don't know if it's useful to call out.

Sources/SwiftProtobuf/ReflectionTable.swift
allevato: The generator guarantees it by writing zeroes for alignment. I don't think it's that useful to assert it at runtime; we'd have to do it at the time that we create the table.

PR #2044, opened 2026-04-25
Support well known proto types in Swift PM plugin

### LINKED ISSUE #1943: Unable to use well-known types with SwiftPM plugin

FranzBusch:
Fixes #1943 and supersedes #1979.

This PR adds support for well known types in the Swift PM plugin by either:
- Looking for the types next to the binary when the env variable or a direct protoc path is specified
- Looking for the source directory of our vendored protobuf repo

### DIFF
--- a/Plugins/SwiftProtobufPlugin/plugin.swift
+++ b/Plugins/SwiftProtobufPlugin/plugin.swift
@@ -145,7 +145,8 @@ struct SwiftProtobufPlugin {
     func createBuildCommands(
         pluginWorkDirectory: URL,
         sourceFiles: FileList,
-        tool: (String) throws -> PackagePlugin.PluginContext.Tool
+        tool: (String) throws -> PackagePlugin.PluginContext.Tool,
+        sourceTreeWellKnownTypesPath: URL? = nil
     ) throws -> [Command] {
         guard
             let configurationFilePath = sourceFiles.first(
@@ -162,15 +163,23 @@ struct SwiftProtobufPlugin {
 
         // We need to find the path of protoc and protoc-gen-swift
         let protocPath: URL
+        let protocIncludeDir: URL?
         if let configuredProtocPath = configuration.protocPath {
             // The user set the config path in the file. So let's take that
             protocPath = URL(fileURLWithPath: configuredProtocPath)
+            // Standard protoc installs auto-find their bundled include directory
+            protocIncludeDir = nil
         } else if let environmentPath = ProcessInfo.processInfo.environment["PROTOC_PATH"] {
             // The user set the env variable. So let's take that
             protocPath = URL(fileURLWithPath: environmentPath)
+            // Standard protoc installs auto-find their bundled include directory
+            protocIncludeDir = nil
         } else {
             // The user didn't set anything so let's try see if SPM can find a binary for us
             protocPath = try tool("protoc").url
+            // Where the tool gets built won't be where the sources are, they will be in
+            // the source tree.
+            protocIncludeDir = sourceTreeWellKnownTypesPath
         }
         let protocGenSwiftPath = try tool("protoc-gen-swift").url
 
@@ -179,6 +188,7 @@ struct SwiftProtobufPlugin {
                 directory: configurationFilePath.deletingLastPathComponent(),
                 invocation: invocation,
                 protocPath: protocPath,
+                protocIncludeDir: protocIncludeDir,
                 protocGenSwiftPath: protocGenSwiftPath,
                 outputDirectory: pluginWorkDirectory
             )
@@ -191,6 +201,7 @@ struct SwiftProtobufPlugin {
     ///   - directory: The plugin's target directory.
     ///   - invocation: The `protoc` invocation.
     ///   - protocPath: The path to the `protoc` binary.
+    ///   - protocIncludeDir: The path for where the WKTs can be found.
     ///   - protocGenSwiftPath: The path to the `protoc-gen-swift` binary.
     ///   - outputDirectory: The output directory for the generated files.
     /// - Returns: The build command configured based on the arguments.
@@ -198,6 +209,7 @@ struct SwiftProtobufPlugin {
         directory: URL,
         invocation: Configuration.Invocation,
         protocPath: URL,
+        protocIncludeDir: URL?,
         protocGenSwiftPath: URL,
         outputDirectory: URL
     ) -> Command {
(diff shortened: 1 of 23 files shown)


FranzBusch:
@thomasvl I think I got this to work by using the protobuf directory as the location for the WKT. Let know what you think

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -292,12 +317,45 @@ extension SwiftProtobufPlugin: BuildToolPlugin {
...
+            sourceTreeWellKnownTypesPath: wellKnownTypesIncludePath
         )
     }
+
+    /// Finds the well-known proto types include path from the swift-protobuf
+    /// package's vendored protobuf sources.
+    private static func findWellKnownTypesIncludePath(in package: Package) -> URL? {
+        func protocIncludePath(for package: Package) -> URL? {
+            let candidate = package.directoryURL
+                .appending(path: "Sources/protobuf/protobuf/src")
```
glbrntt: `package.targets.first(where:)` is probably a slightly safer way to find the `protobuf` dir (e.g. in case the package structure ever changes.)

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -162,15 +163,31 @@ struct SwiftProtobufPlugin {
...
+        let protocIncludeDir: URL?
         if let configuredProtocPath = configuration.protocPath {
             // The user set the config path in the file. So let's take that
             protocPath = URL(fileURLWithPath: configuredProtocPath)
+            // Standard protoc installs have: <prefix>/bin/protoc and <prefix>/include/google/...
+            protocIncludeDir =
+                protocPath
+                .deletingLastPathComponent()
+                .deletingLastPathComponent()
+                .appending(path: "include")
```
glbrntt: Should this be set when the user is providing `protoc`?

FranzBusch: I believe so. When they provide `protoc` we still want to use the include dir for finding the WKT. @thomasvl WDYT?

thomasvl: See https://github.com/apple/swift-protobuf/issues/1943#issuecomment-3862417847, a protoc distribution will auto find the files on it's own. So all we need is for the case we built/package it in the non standard way.

Plugins/SwiftProtobufPlugin/plugin.swift
thomasvl: So looking in the zip files distributed vs. their tree layout, there are some additional files that get bundles that aren't at this fixed path. So this will get the WKTs, but some of the bundled language specific editions files could still trip things up. We might need a step during the protobuf sources update where we collect all of the files into a know location in the tree.

FranzBusch: Is there a place that lists all the WKT? I was really struggling to find that.

Plugins/SwiftProtobufPlugin/plugin.swift
FranzBusch: I see. I will update this once I have a better idea how to find all the WKTs.

Plugins/SwiftProtobufPlugin/plugin.swift
thomasvl: So the WKTs are documented here: https://protobuf.dev/reference/protobuf/google.protobuf/ But what we're really after is _all the .proto files bundled with `protoc` so that developers don't have to provide them where another .proto file references them_.

Historically, I've just looked inside the released .zip files on the github page, but it looks like all the packaging for their `bazel` build might be in https://github.com/protocolbuffers/protobuf/tree/main/pkg. It appears that the [`wkt_protos_files`](https://github.com/protocolbuffers/protobuf/blob/2893999a3c4fff2444cdcab0f278374579861efb/pkg/BUILD.bazel#L15-L27) is the correct target.  It includes all of the _classic_ WKTs, along with `descriptor.proto` and then all of the Edition features `.proto` files for the language generators that are bundled within in the main repo (C#, Go, Java, CPP).

Rather that try to duplicate the collection of those, I wonder if it might be easier to make the script that updates our copy to simply download one of the release, and extract the directory and land it also in distinct directory of `Sources/protobuf` and just reference it there?

Aside: `wkt_protos_files` also seems to exist/be referenced in `src/file_lists.cmake` and `cmake/install.cmake`, but I don't know enough _cmake_ to see if there is something to reuse there.

FranzBusch:
@thomasvl @glbrntt I addressed all your comments and I updated the vendoring script to copy the WKTs to a new `include` directory. We just have to update the script if a new WKT is introduced.

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -292,12 +309,47 @@ extension SwiftProtobufPlugin: BuildToolPlugin {
...
+            if FileManager.default.fileExists(
+                atPath: candidate.appending(path: "google/protobuf/timestamp.proto").fileSystemPath
+            ) {
+                return candidate
+            }
+            return nil
+        }
+
+        // Check if this package itself is swift-protobuf
+        if let path = protocIncludePath(for: package) {
```
thomasvl, line 347: Any reason here and in the lower loop to not check the displayName on the package to confirm it is "SwiftProtobuf"?

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -292,12 +309,47 @@ extension SwiftProtobufPlugin: BuildToolPlugin {
...
     }
+
+    /// Finds the well-known proto types include path from the swift-protobuf
+    /// package's vendored protobuf sources.
+    private static func findWellKnownTypesIncludePath(in package: Package) -> URL? {
+        func protocIncludePath(for package: Package) -> URL? {
+            guard let protocTarget = package.targets.first(where: { $0.name == "protoc" }) else {
+                return nil
+            }
+            let candidate = protocTarget.directoryURL.appending(path: "include")
```
thomasvl, line 334: Looks like this needs some conditional code to deal with `directoryURL` vs. `directory` for the 6.0 support until we drop that.

thomasvl (commented):
Nice!

glbrntt (approved):
Looks great modulo the CI failure!

thomasvl (approved):

PR #2046, opened 2026-05-04
[table-driven-branch] Refactor trampoline functions as "message/enum witnesses".

allevato:
Since we store concrete message types in the message storage buffer (and in the case of repeated enums, the concrete enum types), the runtime needs a way to perform operations on those types without concrete knowledge of them.

The first attempt at doing this use "trampoline functions", where the runtime would call a generated function on the message schema, and that function would call back into the runtime an equivalent function that took the concrete type as a parameter. These generated functions took basically an integer token that said "which submessage/subenum" and had large switch statements to call with the right context.

These generated functions could be large in code size. Worse, they were not often deduplicable by the compiler/linker, because they were multiple 1:N mappings from a message to its submessages/subenums. Each generated message would basically have unique copies of these large functions. The operations being performed were very high-level, like "execute an arbitrary closure on the submessages" or "loop over all of the enum raw values".

## The new approach

This new approach flips it on its head, adopting a technique used by Swift itself. Each Swift value type defines a "value witness table" that allows the runtime to dispatch primitive operations on value types without knowing the concrete type, like "initialize", "destroy", "copy", etc.

Generated messages and enums can do basically the same thing. Now the large high-level operations being performed by the runtime are replaced by smaller operations like "initialize a new message into this raw memory" or "append a new message to the array stored in this raw memory", and the runtime calls those with pointers into the message's storage buffer.

This provided a number of advantages:

* The code in the runtime that manipulates messages/enums is much less obtuse and easier to follow.
* Now, instead of generating multiple trampoline functions per message, we only generate a single method: an accessor that returns the correct message or enum schema or a submessage or enum field.
* Despite operating on concrete types, the witness function itself can be generic and live exclusively in the runtime. The generated code forms a reference to it with a specific concrete type, which means we only generate a small shim that loads the appropriate type metadata and delegates to the runtime function. (In Embedded, we would specialize the entire function, but that's unavoidable in that environment.)

This provided another ~5% reduction in code size for the generated WKTs compared to the trampoline approach, for an overall reduction of ~27% compared to `main`.


Sources/protoc-gen-swift/ExtensionSetGenerator.swift
```diff
@@ -104,30 +102,33 @@ class ExtensionSetGenerator {
...
 
                 // Since an extension is just a single field, there will be either zero or one of
                 // these.
                 if let field = extensionSchemaCalculator.trampolineFields.first {
-                    p.print(",")
-                    p.print("performNontrivialExtensionOperation: { operation, ext, storage in")
-                    p.printIndented("storage.performNontrivialExtensionOperation(operation, extension: ext, type: \(field.kind.name).self)")
-                    p.print("},")
-
+                    let result: String
```
thomasvl: Nit: Maybe "resolver"? Since that's what seems to be going in the value and it isn't the _result_ of the function?

Sources/protoc-gen-swift/ExtensionSetGenerator.swift
```diff
@@ -104,30 +102,33 @@ class ExtensionSetGenerator {
...
+                    case .enum(let name):
+                        result = ".enum(\(name).enumSchema)"
                     case .map:
                         preconditionFailure("unreachable; extensions cannot be map fields")
                     }
-                    p.print("}", newlines: false)
-                } else {
+                    p.print(
+                        ",",
+                        "submessageOrEnumResolver: { \(result) }", newlines: false)
```
thomasvl: Nit: Why not just put ", " on the start of the string since this is also `newline: false`

Sources/protoc-gen-swift/MapEntryGenerator.swift
```diff
@@ -25,8 +25,18 @@ class MapEntryGenerator {
...
+    /// containing message for the map entry.
+    let entrySchemaName: String
+
+    private let keyParticipantType: String
+    private let valueParticipantType: String
+
+    /// Computes the name of the static variable of type `MessageSchema` that will be generated into the
+    /// containing message for the map entry with the given descriptor.
+    static func schemaName(for descriptor: Descriptor) -> String {
+        "_protobuf_mapEntrySchema_\(descriptor.name)"
```
thomasvl, line 38: Why do we need the name on this vs. using a fixed name?  Is these some collision potential here? If so, why is "name" good enough vs some transform of "full_name"?

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -274,110 +261,41 @@ class MessageGenerator {
...
-                "private static func _protobuf_performOnSubmessageStorage(for token: SwiftProtobuf.MessageSchema.TrampolineToken, field: SwiftProtobuf.FieldSchema, storage: SwiftProtobuf.MessageStorage, operation: SwiftProtobuf.TrampolineFieldOperation, perform: (SwiftProtobuf.MessageStorage) throws -> Swift.Bool) throws -> Swift.Bool {"
+                "private static func _protobuf_resolveSubmessageOrEnum(for token: SwiftProtobuf.MessageSchema.TrampolineToken) -> SwiftProtobuf.SubmessageOrEnumSchema {"
             )
             p.withIndentation { p in
                 p.print("switch token.index {")
-                var nonMessageMapFieldIndices = [Int]()
                 for field in trampolineFields {
-                    // Only submessage fields and map fields with message values need this storage
-                    // trampoline.
+                    let result: String
```
thomasvl: Nit: Same general comment on "result", but atleast here it is being returned by the _case_ of the generated code. But maybe "schema" would be better?"

Sources/protoc-gen-swift/MessageGenerator.swift
```diff
@@ -388,63 +306,20 @@ class MessageGenerator {
...
-                            name: name,
-                            keyType: keyType,
-                            valueType: valueType
-                        )
-                        let entryGenerator = mapEntries[key]
-                        entryGenerator?.generateSchemaReturnStatement(printer: &p)
+            // Generate map entry schemas, if any.
+            for field in trampolineFields {
+                switch field.kind {
+                case .map(let schemaName):
```
thomasvl: Nit: if case or something smaller given you only look at one type of `kind`?

Might even be able to roll in the second let making the whole things a guard?

Sources/SwiftProtobuf/ExtensionStorage+BinarySize.swift
```diff
@@ -69,7 +69,16 @@ extension ExtensionStorage {
 
             case .group:
                 precondition(!isPacked, "a packed message/group field should not be reachable")
-                return serializedByteSize(ofGroupExtension: schema, fieldNumber: fieldNumber)
+                var totalSize = 0
+                forEachMessage(inAssumedPresentRepeatedMessageField: schema) { groupStorage in
+                    let messageSize = groupStorage.serializedBytesSize()
+                    totalSize += messageSize
+                        // Include the size of the length-delimited tag.
+                        + FieldTag.encodedSize(ofTagWithFieldNumber: fieldNumber)
```
thomasvl: The tag size seems like we could compute it once (and maybe see totalSize with that times the number of items). (like done below for message?)

ps - and shouldn't there be a start and end tag for groups?

Sources/SwiftProtobuf/ExtensionStorage.swift
```diff
@@ -143,22 +140,55 @@ public final class ExtensionStorage {
...
+    }
 }
 
 /// An operation that the runtime requests to be performed on nontrivial extension fields that
 /// require trampolining through a generated function to propagate the correct concrete type.
-@_spi(ForGeneratedCodeOnly)
-public enum NontrivialExtensionOperation {
-    /// The value of the extension field should be deinitialized.
-    case deinitialize
+// @_spi(ForGeneratedCodeOnly)
```
thomasvl: Delete?

Sources/SwiftProtobuf/ExtensionStorage.swift
```diff
@@ -231,72 +261,72 @@ extension ExtensionStorage {
...
-
-        case .copy(destination: let destination):
-            destination.updateValue(of: valueStorage.schema, to: valueStorage.value(as: type))
-            return true
-
-        case .isEqual(other: let other):
-            return isExtensionValue(valueStorage, equalToSameValueIn: other, type: type)
-        }
-    }
+    // public func performNontrivialExtensionOperation<T: Equatable>(
```
thomasvl: Delete? (and the test of the commented out code in here)

Sources/SwiftProtobuf/ExtensionStorage.swift
```diff
@@ -492,12 +522,48 @@ extension ExtensionStorage {
...
+                        break
+                    }
+                    // Both are present; compare them.
+                    let lhsCount = elementCount(forAssumedPresentRepeatedEnumField: ext)
+                    let rhsCount = other.elementCount(forAssumedPresentRepeatedEnumField: ext)
+                    guard lhsCount == rhsCount else {
+                        equalSoFar = false
+                        break
+                    }
+                    for i in 0..<lhsCount {
```
thomasvl, line 284: I guess we can't just do memcmp on these, can we?

Sources/SwiftProtobuf/ExtensionValueStorage.swift
```diff
@@ -110,11 +146,11 @@
...
     ///
     /// - Precondition: The type of the extension field must be a message or group.
-    func unsafePointerToSubmessageWithUniqueStorage<Value: Message>(
-        as type: Value.Type,
-    ) -> UnsafeMutablePointer<Value> {
-        let pointer = UnsafeMutablePointer<Value>(bitPattern: Int(truncatingIfNeeded: Int64(bitPattern: storage)))!
-        pointer.pointee._protobuf_ensureUniqueStorage(accessToken: MessageStorageToken())
-        return pointer
-    }
+    // func unsafePointerToSubmessageWithUniqueStorage<Value: Message>(
```
thomasvl: delete?

Sources/SwiftProtobuf/MapEntryWorkingSpace.swift
```diff
@@ -40,15 +40,22 @@ struct MapEntryWorkingSpace {
     /// Returns the `MessageStorage` used to encode/decode map entries with the given trampoline
     /// index, creating it if necessary.
     mutating func storage(for trampolineIndex: Int) -> MessageStorage {
-        if let storage = entryStorage[trampolineIndex] {
-            return storage
-        }
+        // if let storage = entryStorage[trampolineIndex] {
```
thomasvl: delete?

Sources/SwiftProtobuf/MessageSchema.swift
```diff
@@ -223,110 +157,79 @@ public struct MessageSchema: @unchecked Sendable {
...
-                preconditionFailure("This should have been unreachable; this is a generator bug")
-            },
-            performOnMapEntry: { _, _, _, _, _, _, _ in
-                preconditionFailure("This should have been unreachable; this is a generator bug")
+            invokeWitness: MapEntryWitnesses<K, ProtobufMapMessageField<M>>.perform,
+            submessageOrEnumResolver: { token in
+                switch token.index {
+                // TODO: Introduce a `GeneratedMessage` protocol with a `static var messageSchema`
+                // requirement and use that here.
+                case 1: return .message(M().messageSchema)
```
thomasvl: Nit: if vs. switch to shorten the code?

thomasvl (approved):
Mostly nits, one "group" question that suck out.

Sources/protoc-gen-swift/ExtensionSetGenerator.swift
allevato: Done.

Sources/protoc-gen-swift/ExtensionSetGenerator.swift
allevato: There are three parameters passed to the extension initializer, all of which are on their own lines: schema, extendedMessageResolver, and submessageOrEnumResolver. This puts the trailing comma at the end of extendedMessageResolver and then puts submessageOrEnumResolver on its own line.

Sources/protoc-gen-swift/MapEntryGenerator.swift
allevato: If a message has map fields, we generate a schema for each map field that represents the message layout for the map entry. Now that the schema includes the name of the message, that means they're all unique, and we generate them using the autogenerated name given by the descriptor:

```protobuf
package foo.bar;

message M {
  map<int32, int32> map1 = 1;
  map<string, string> map2 = 2;
}
```

```swift
extension M {
  let _protobuf_mapEntrySchema_Map1Entry_string = "...foo.bar.M.Map1Entry"
  let _protobuf_mapEntrySchema_Map2Entry_string = "...foo.bar.M.Map2Entry"
  // ...
}
```

We don't use the name for anything so we could consider leaving it out, but I left it in in case some future reflection task needed it.

Sources/protoc-gen-swift/MessageGenerator.swift
allevato: Done.

Sources/protoc-gen-swift/MessageGenerator.swift
allevato: Done.

Sources/SwiftProtobuf/ExtensionStorage+BinarySize.swift
allevato: Ah, this is the same mistake I had in message subgroup fields too; I swapped the message and group implementations. Odd that no conformance test caught this...

Sources/SwiftProtobuf/ExtensionStorage.swift
allevato: Done.

Sources/SwiftProtobuf/ExtensionStorage.swift
allevato: Done.

Sources/SwiftProtobuf/ExtensionStorage.swift
allevato: It's technically not safe to do so. (One could imagine an enum having a size != stride and there not being guarantees about the padding bytes between two different arrays of otherwise equal values.)

Sources/SwiftProtobuf/ExtensionValueStorage.swift
allevato: Done (along with another unused method).

Sources/SwiftProtobuf/MessageSchema.swift
allevato: Done.

Sources/SwiftProtobuf/MapEntryWorkingSpace.swift
allevato: Ah, I had commented out the caching part of this because I the code flow changed in a way that it made it awkward to do the field clearing that we were doing previously, and forgot to revisit it.

Since I'm about to do some runtime sampling, I'm going to leave the non-caching version in place and see if it's a hot enough spot that it's worth adding the caching back to avoid the allocations. Then I'll decide whether this needs to exist at all.

Sources/SwiftProtobuf/ExtensionStorage+BinarySize.swift
```diff
@@ -99,13 +97,15 @@ extension ExtensionStorage {
...
             case .message:
                 precondition(!isPacked, "a packed message/group field should not be reachable")
-                let count = elementCount(forAssumedPresentRepeatedMessageField: schema)
-                // It's slightly more efficient to pre-calculate the total size of the tags here
-                // instead of adding it inside the loop, so we don't use `forEach...` here.
-                var totalSize = count * (2 * FieldTag.encodedSize(ofTagWithFieldNumber: fieldNumber))
-                for i in 0..<count {
-                    totalSize +=
-                        messageStorage(at: i, inAssumedPresentRepeatedMessageField: schema).serializedBytesSize()
+                var totalSize = 0
```
thomasvl, line 100: Now I'm confused, you moved the field calc inside the loop here, you had it outside.

allevato: The code to handle group and message calculation was accidentally swapped the first time. For messages it goes inside because the length depends on the varint of the computed message size. For groups it's outside because it's 2 fixed tags.

PR #2048, opened 2026-05-04
[table-driven-branch] A handful of runtime performance improvements.

allevato:


Sources/SwiftProtobuf/UTF8Name.swift
```diff
@@ -0,0 +1,106 @@
...
+/// used to avoid string materialization overhead.
+///
+// -----------------------------------------------------------------------------
+
+/// An immutable UTF-8 view of a message name, field name, or enum case.
+///
+/// Since the strings owned by the schemas and reflection tables are effectively immortal once
+/// defined and/or decompressed, we can pass around UTF-8 buffers directly without concern for the
+/// lifetimes of these values. This type is very similar in spirit to the recently added `UTF8Span`
+/// (which we can't use yet since it's too new).
```
thomasvl, line 21: Is this a case where when the toolchain moves past a point it could be use, or does it depend on the min runtime version to use it?  i.e. - is it worth capture the specific version and what exactly has to move past that to adopt the new type?

Sources/SwiftProtobuf/UTF8Name.swift
```diff
@@ -0,0 +1,106 @@
...
+                return false
+            case (nil, nil):
+                return false
+            }
+        }
+    }
+
+    /// Consumes the given prefix from the receiver and returns the remaining suffix,
+    /// or nil if the receiver does not start with the prefix.
+    package func consumePrefix(_ prefix: String) -> UTF8Name? {
```
thomasvl: Should `prefix` be a `StaticString`? Does that buy you anything?

thomasvl (approved):

Sources/SwiftProtobuf/UTF8Name.swift
allevato: Yeah, we'd still have minimum deployment issues with a new type. We'd have to drop everything before appleOS 26.0, which probably isn't realistic for a very long time.

Sources/SwiftProtobuf/UTF8Name.swift
allevato: My thinking was "the prefix is a small string so it should optimize well" but it turns out that `"google.protobuf."` is exactly one character longer than the small string optimization. I don't think this is a huge hotspot (it's only called during JSON serialization) but it can't hurt to do the obvious optimization here.

PR #2050, opened 2026-05-05
[table-driven-branch] Implement caching for uncompressed reflection tables.

allevato:
Rather than have a global table that would require keeping track of a mapping between compressed pointers and uncompressed data, I've chosen to just make each table reference self-caching. This has the advantage of each reference being able to maintain an independent lock, instead of a lock around the whole mapping (or something more complex that would allow multiple readers/one writer).

`NSLock` is chosen as a lowest-common denominator since we don't want to be too limiting about supported minimum OS versions. It may be worth exploring different options on different platforms in the future, but I expect the cost of the current lock would be small compared to the actual encoding/decoding operations that use this table. I didn't notice any performance issues when introducing this compared to the original version that just decompressed the table on schema creation.


thomasvl (approved):
Grabbing a lock on every lookup worries me a little, but as you say, we can revisit/optimize as needed in the future.

PR #2051, opened 2026-05-05
[table-driven-branch] Optimize `==` and `hash` for messages with only trivial fields.

allevato:


Sources/SwiftProtobuf/MessageStorage+Equality.swift
thomasvl: Update comment?

Sources/SwiftProtobuf/MessageStorage+Equality.swift
```diff
@@ -29,22 +29,30 @@ extension MessageStorage {
...
+            return false
+        }
+
+        // If the message contains only trivial fields, we are done since we compared them above.
+        guard firstNontrivialStorageOffset != schema.storageSize else {
+            return true
+        }
+
         var equalSoFar = true
         for field in schema.fields {
```
thomasvl, line 53: Does it make sense to have something like `fields(after: firstNontrivialStorageOffset)`, that can internally skip to the first place that is actually needed?

Sources/SwiftProtobuf/MessageStorage+Hashing.swift
```diff
@@ -24,24 +24,26 @@ extension MessageStorage {
...
+        // If the message contains only trivial fields, we are done since we hashed them above.
+        guard firstNontrivialStorageOffset != schema.storageSize else {
+            return
+        }
 
-        // Loops through the fields, combining the hashes for any with non-trivial types. We ignore
-        // the trivial ones here, instead tracking the byte offset of the first non-trivial field
-        // so that we can hash them as a contiguous byte buffer slice afterwards.
-        var firstNontrivialStorageOffset = schema.storageSize
         for field in schema.fields {
```
thomasvl, line 40: Same comment about a helper to skip to the point of the first thing needed.

thomasvl (approved):

Sources/SwiftProtobuf/MessageStorage+Equality.swift
allevato: Done.

Sources/SwiftProtobuf/MessageStorage+Equality.swift
allevato: Fields are stored and iterated in the schema in field number order, not in storage order, so there's no way to skip them as a batch. For example,

```protobuf
message M {
  string a = 1;
  int32 b = 2;
  bytes c = 3;
  bool d = 4;
}
```

This would be laid out in memory in the order 4, 2, 1, 3.

Sources/SwiftProtobuf/MessageStorage+Hashing.swift
allevato: Ditto.

Sources/SwiftProtobuf/MessageStorage+Equality.swift
thomasvl: Ah, right, storage vs. iteration order.

PR #2052, opened 2026-05-05
[table-driven-branch] Support message tests.

allevato:


Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
```diff
@@ -635,4 +657,152 @@ extension MessageStorage {
...
+    ) throws -> Bool {
+        // This is loosely based on the C++:
+        //   ExtensionSet::ParseMessageSetItem()
+        //   WireFormat::ParseAndMergeMessageSetItem()
+        // And more implementation seem to draw from:
+        //   upb_Decoder_DecodeMessageSetItem()
+
+        var extensionSchema: ExtensionSchema?
+        var messageBytes: UnsafeRawBufferPointer?
+        let success =  try reader.withReaderForNextGroup(withFieldNumber: 1) { subReader in
```
thomasvl: Maybe `WireFormat.MessageSet.FieldNumbers.item` instead of the raw `1`?

Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
```diff
@@ -108,8 +108,30 @@ extension MessageStorage {
             return true
         }
 
+        if schema.extensibilityMode == .messageSet && tag.fieldNumber == WireFormat.MessageSet.FieldNumbers.item {
```
thomasvl, line 113: This doesn't look right compared to `decodeExtensionFieldsAsMessageSet` on main.

The startGroup tracks, but if the library that did the encoding doesn't know about message set, the typeId will be the field number and the message will be the length encoded value. This will just read field number 1 as group or length encoded. And then decodeMessageSetItemFromLengthDelimited seems to still expect the two distinct fields instead. So I'm sorta surprised this is passing tests.

Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
```diff
@@ -635,4 +657,152 @@ extension MessageStorage {
...
+            byReadingFrom: &subReader,
+            extensions: extensions,
+            partial: partial,
+            discardUnknownFields: discardUnknownFields
+        )
+        return true
+    }
+
+    /// Decodes a message set item from a length-delimited slice.
+    private func decodeMessageSetItemFromLengthDelimited(
```
thomasvl: See comment above, this doesn't seem correct.

Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
allevato: Done.

Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
allevato: You're right, the length-delimited code path wasn't actually being hit. Those are handled below, when we look up extensions if the message is either a regular extensible message or a message set.

Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
allevato: Deleted.

thomasvl (approved):

PR #2054, opened 2026-05-06
[table-driven-branch] Support field mask APIs.

allevato:
This is _mostly_ the same as the original APIs, except that some of the methods that take message metatypes are now constrained by `GeneratedMessage` instead of `Message` so that they can access the schema.


Sources/SwiftProtobuf/Message+FieldMask.swift
```diff
@@ -13,45 +13,17 @@
 // -----------------------------------------------------------------------------
 
 // TODO: Re-implement these using a reflection API.
```
thomasvl, line 15: Did you want to keep these comments?

Sources/SwiftProtobuf/Message+FieldMask.swift
```diff
@@ -74,67 +45,20 @@ extension Message {
...
-        //     self = newExtensible as? Self ?? newMessage
-        // } else {
-            self = newMessage
-        // }
-        self.unknownFields = message.unknownFields
-    }
-}
-*/
\ No newline at end of file
+}
```
thomasvl: Missing a trailing newline it seems.

Sources/SwiftProtobuf/MessageStorage+FieldMask.swift
```diff
@@ -0,0 +1,367 @@
...
+            }
+            let path = String(decoding: name.buffer, as: UTF8.self)
+            if fieldMask.contains(path) {
+                continue
+            }
+            
+            // Check if the current field is a prefix of any path in the mask.
+            let isPrefix = fieldMask.paths.contains { $0.hasPrefix("\(path).") }
+            if isPrefix {
+                if field.rawFieldType == .message || field.rawFieldType == .group {
```
thomasvl, line 101: Is it at all worth helpers for things like `isMessageField` or does this optimize enough that it isn't worth worrying about the two lookups? (Since we do a few of these)

Sources/SwiftProtobuf/MessageStorage+FieldMask.swift
```diff
@@ -0,0 +1,367 @@
...
+                    if isPresent(field) {
+                        let subStorage = messageStorage(forAssumedPresentSingularMessageField: field)
+                        let subMask = removingPrefix(path, from: fieldMask)
```
thomasvl: I wonder if it makes sense to make the public api call a helper that passes in `prefix`, and then ever time a field is looked up you can prepend the prefix, that way you don't have to make new fieldmasks as you recurse through the graph.

Tests/SwiftProtobufTests/Test_FieldMask.swift
```diff
@@ -821,5 +820,4 @@ final class Test_FieldMask: XCTestCase, PBTestHelpers {
         try m1.merge(from: m2, fieldMask: .with({ $0.paths = ["singulargroup"] }))
         XCTAssertEqual(m1.singularGroup, m2.singularGroup)
     }
-}
-*/
\ No newline at end of file
+}
```
thomasvl: Trailing newline missing

thomasvl (approved):

Sources/SwiftProtobuf/Message+FieldMask.swift
allevato: Yeah, because I think we'll still want to look at that once we've built that reflection API.

Sources/SwiftProtobuf/Message+FieldMask.swift
allevato: Done.

Sources/SwiftProtobuf/MessageStorage+FieldMask.swift
allevato: I think it's nicer to be explicit when we support both at any given call site, IMO.

Sources/SwiftProtobuf/MessageStorage+FieldMask.swift
allevato: Yeah, that's cleaner. Done.

Tests/SwiftProtobufTests/Test_FieldMask.swift
allevato: Done.

PR #2055, opened 2026-05-07
Generate nonisolated code

### LINKED ISSUE #1797: Generated Pb files fail to compile with the new Default Actor Isolation flag set to MainActor
After watching the WWDC 2025, Embracing Swift concurrency, I try to play with the new concurrency flags which you can find on the Xcode 26 beta.

Following is the recommended flag settings from the tutorial video:

Image

I find, for my existing project that contains generated *.pb.swift files, if I set the Default Actor Isolation to the MainActor. All these files will have compile time errors (about 200 errors in my case). 

Some typical errors are like: Conformance of "Pb_Person" to protocol 'Enum' crosses into main actor-isolated code and can cause data races.

The fix is to set the flag to nonisolated. 

This makes sense because the generated code is created with the nonisolated as default isolation. 
Since Apple has recommended the new default isolation, the current generated code will be a road block.
What is the solution here for the existing projects and new projects?

glbrntt:
Motivation:

When setting the default actor isolation to 'MainActor', generated
protobuf code will fail to compile.

Modifications:

- Add 'nonisolated' to generated declaration
- Add a compile test

Result:

- Resolves https://github.com/apple/swift-protobuf/issues/1797


Sources/protoc-gen-swift/GeneratorOptions.swift
```diff
@@ -120,6 +125,7 @@ package class GeneratorOptions {
         var swiftProtobufModuleName: String? = nil
         var implementationOnlyImports: Bool = false
         var useAccessLevelOnImports = false
+        var nonisolatedDeclarations: Bool = false
```
FranzBusch: I'm wondering if we should default this to `true`. We do expect many folks to run this code generator for iOS apps where it is recommend and I think the default to have main actor isolation on. This requires them to hunt for this option to make it work. `nonisolated` is arguably the right thing for our generated types anyways and as far as I can see doesn't hurt even in the cases where there is no default isolation right?

glbrntt: Won't it fail to compile on Swift 6.1 and Swift 6.0?

FranzBusch: I thought that `nonisolated` declarations were valid syntax before that already but we can guard the option to `true` only on 6.2 and above as well.

glbrntt: Oh right, default actor isolation was added in 6.2 but `nonisolated` has been around for a while.

glbrntt: It seems that 6.1 is fine but 6.0 is not.

> I thought that `nonisolated` declarations were valid syntax before that already but we can guard the option to `true` only on 6.2 and above as well.

I think we should only do this in the build plugin as `protoc-gen-swift` may be used to generate code ahead of time on 6.2+ for a target that still supports 6.0.

FranzBusch: > It seems that 6.1 is fine but 6.0 is not.

We should probably just drop 6.0 then since that is out of policy anyways.

> I think we should only do this in the build plugin as `protoc-gen-swift` may be used to generate code ahead of time on 6.2+ for a target that still supports 6.0.

I think our policy so far has been that we don't support this configuration really. If the generator is used with 6.2 even for ahead of time generation but the generated code is compiled on older out of policy Swift then it might work but we don't guarantee it. Otherwise the support matrix becomes way too complex.

allevato: +1 to dropping 6.0. It was last released in Xcode 16.2, and Apple just stopped allowing app submissions below Xcode 26. I presume we don't have a need to keep Swift 6.0 users supported on non-Apple platforms either?

glbrntt: > > I think we should only do this in the build plugin as `protoc-gen-swift` may be used to generate code ahead of time on 6.2+ for a target that still supports 6.0.
> 
> I think our policy so far has been that we don't support this configuration really. If the generator is used with 6.2 even for ahead of time generation but the generated code is compiled on older out of policy Swift then it might work but we don't guarantee it. Otherwise the support matrix becomes way too complex.

The suggestion assumed Swift 6.0 still being supported (given the current tools version). I completely agree we shouldn't support this if we drop 6.0 though.

> +1 to dropping 6.0. It was last released in Xcode 16.2, and Apple just stopped allowing app submissions below Xcode 26. I presume we don't have a need to keep Swift 6.0 users supported on non-Apple platforms either?

Not that I'm aware.

glbrntt: Drop Swift 6 support in: https://github.com/apple/swift-protobuf/pull/2056

thomasvl:
What happens if the code gets compiled into a library/target that hasn't adopted swift 6 language mode?  Would it fail to compile?

FranzBusch (approved):

FranzBusch:
> What happens if the code gets compiled into a library/target that hasn't adopted swift 6 language mode? Would it fail to compile?

It shouldn't since `nonisolated` is the default.

glbrntt:
I believe the API breaks are false positives:

```
  💔 API breakage: struct Google_Protobuf_Any is now with nonisolated
  💔 API breakage: var Google_Protobuf_Any.typeURL is now with nonisolated
  💔 API breakage: var Google_Protobuf_Any.value is now with nonisolated
  💔 API breakage: constructor Google_Protobuf_Any.init() is now with nonisolated
  💔 API breakage: constructor Google_Protobuf_Any.init(message:partial:typePrefix:) is now with nonisolated
  💔 API breakage: constructor Google_Protobuf_Any.init(textFormatString:extensions:) is now with nonisolated
  💔 API breakage: constructor Google_Protobuf_Any.init(textFormatString:options:extensions:) is now with nonisolated
  💔 API breakage: func Google_Protobuf_Any.isA(_:) is now with nonisolated
  💔 API breakage: func Google_Protobuf_Any.hash(into:) is now with nonisolated
  💔 API breakage: func Google_Protobuf_Any.decodeMessage(decoder:) is now with nonisolated
  💔 API breakage: func Google_Protobuf_Any.traverse(visitor:) is now with nonisolated
... (etc)
```

@FranzBusch do you agree?

thomasvl:
> I believe the API breaks are false positives:
> [...]

Worth a bug report against the tooling?

FranzBusch:
Yeah I think those are in our case non-API breaking. They are probably just ABI breaks.

glbrntt:
> > I believe the API breaks are false positives:
> > [...]
> 
> Worth a bug report against the tooling?

Yes: https://github.com/swiftlang/swift/issues/89263

PR #2057, opened 2026-05-07
[table-driven-branch] Various runtime/generator tweaks, mostly to reduce binary size.

allevato:
Each commit describes a specific change; all regeneration is held to the end.

Overall, this amounts to about a 170KB reduction in the `.o` files in the runtime, or about 4.4% when built in release without debug info. Not too shabby.


Sources/SwiftProtobuf/MessageSchema.swift
```diff
@@ -708,16 +708,16 @@ package struct FieldMode: RawRepresentable, Equatable, Hashable, Sendable {
...
     /// The message is a map entry and is thus not extensible.
-    package static let mapEntry = Self(rawValue: 0b100_0000)
+    package static var mapEntry: Self { .init(rawValue: 0b100_0000) }
 
     /// The message is extensible and its extensions fields use message set wire format.
-    package static let messageSet = Self(rawValue: 0b110_0000)
+    package static var messageSet: Self { .init(rawValue: 0b110_0000) }
 
     /// The bitmask used to isolate the extensibility mode in the schema representation.
     package static let bitMask: UInt8 = 0b0110_0000
```
thomasvl: Why not this one? Should we search for any other `.* let \w+(\s*:\s? \w+)\s*=\s*.+` to sweep for others?

thomasvl (approved):
I can't help but wonder if it is worth doing some scriptiin in the Makefile that we wire in to `check` that uses regexes over the swift sources to try and catch when some of the things that should be avoid are done.

Sources/SwiftProtobuf/MessageSchema.swift
allevato: Just forgot this one.

Might be worth doing a separate pass later, yeah.

PR #2060, opened 2026-05-12
Add WKT import path directly to protoc binary

glbrntt:
Motivation:

The recent changes to the build plugin only add import paths to the
SwiftPM entrypoint. The Xcode plugin entrypoint doesn't have appropriate
support to add an import for the WKTs.

Another limitation of the current approach is that users building and
using `protoc` from the pacakge (i.e. `swift build --product protoc`)
won't get the WKT include paths.

This changes takes a different approach by rewriting the entrypoint to
protoc to unconditionally add the WKT include path. This means WKTs work
for the SwiftPM plugin, Xcode plugins, and when using `protoc` directly.
It also obviates the need for downstream pacakges depending on `protoc`
in their plugins (e.g. gRPC) to scan for the include paths in their
plugins.

Modifications:

- Add a new 'main.cc' which does what 'main_no_generators.cc' does but
  adds the WKT include path to the invocation.
- Remove 'main_no_generators.cc' and update the vendoring script to
  prune it out.
- Revert recent changes to the SwiftPM plugin which scanned for protobuf
  sources for the include path

Result:

WKTs work with protoc in more places


Sources/protobuf/main.cc
```diff
@@ -19,15 +19,19 @@ std::string WellKnownTypesIncludeFlag() {
 int main(int argc, char* argv[]) {
   absl::InitializeLog();
 
+  // Append (don't prepend) the bundled WKT include path. protoc searches
```
thomasvl: Nit: this make it sounds like your decision to have them last. It might make sense to instead reference CommandLineInterface::Run(), which calls `ParseArguments` thus collecting the values passed, then can call `InitializeDiskSourceTree()` which does the other builtin search.

having said that, by adding this, you are changing what protoc will do at startup. If you look at `Run()` it uses `proto_path_` as a gate to decided if it is only working on descriptor set or not, and thus won't initialize disk access. So if you want to support the normal `--descriptor_set_in` flow, you need to scan the arguments to see if there are any `-I` or `--proto_path` arguments and only add our path if there are any.

glbrntt: That's a great point, we should keep the `--descriptor_set_in` behaviour to avoid any surprises.

Sources/protobuf/main.cc
```diff
@@ -0,0 +1,86 @@
...
+// Matches "--descriptor_set_in" / "--descriptor_set_in=<value>".
+//
+// See also CommandLineInterface::ParseArgument()
+bool IsDescriptorSetInFlag(const char* arg) {
+  return std::strcmp(arg, "--descriptor_set_in") == 0 ||
+         std::strncmp(arg, "--descriptor_set_in=", 20) == 0;
+}
+}  // namespace
+
+int main(int argc, char* argv[]) {
```
thomasvl, line 40: Nit: Maybe put a pointer to the code this needs to follow upstream for what it does.

Sources/protobuf/main.cc
```diff
@@ -0,0 +1,86 @@
...
+  //
+  // Run() also gates disk access on proto_path_ being non-empty. After
+  // ParseArguments() the only mode that leaves proto_path_ empty, and
+  // therefore skips disk initialization, is "--descriptor_set_in with no
+  // -I". Adding -I in would force disk initialization and change the behavior
+  // of Run().
+  bool has_proto_path = false;
+  bool has_descriptor_set_in = false;
+
+  for (int i = 1; i < argc; ++i) {
```
thomasvl: Nit: you could just do these checks while copying the arg over into the new array since you don't skip that in the no addition case.

thomasvl (approved):

PR #2061, opened 2026-05-14
[table-driven-branch] Reimplement TextFormat/JSON parsing with a common tokenizer.

allevato:


Sources/SwiftProtobuf/TextualParser.swift
```diff
@@ -0,0 +1,165 @@
...
+
+    /// Returns a value indicating whether the current token is one of the given types, without
+    /// consuming it.
+    func at(_ kinds: Token.Kind...) -> Bool {
+        kinds.contains(tokenizer.current.kind)
+    }
+
+    /// Consume the next token if its type is one of the given kinds.
+    ///
+    /// - Returns: `true` if the end of the input has been reached, otherwise `false`.
```
thomasvl: This doesn't seem right, it never returns false, it throws, right?

Sources/SwiftProtobuf/TextualParser.swift
```diff
@@ -0,0 +1,165 @@
...
+            let expectedDescription = kinds.map(\.errorDescription).joined(separator: ", ")
+            throw tokenizer.makeError(
+                "Expected one of \(expectedDescription); but found \(tokenizer.current.kind.errorDescription)"
+            )
+        }
+        _ = try tokenizer.next()
+        return true
+    }
+
+    /// Consume the next token if its type is one of the given kinds.
```
thomasvl: s/kinds/kind/ (only accepts one)

Same issue on returns and 'false'

Sources/SwiftProtobuf/Tokenizer.swift
```diff
@@ -0,0 +1,902 @@
...
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+///
+/// A low-level tokenizer shared by the TextFormat and JSON parsers.
+///
+// -----------------------------------------------------------------------------
+
+import Foundation
```
thomasvl, line 15: FoundationEssentials?

Sources/SwiftProtobuf/Tokenizer.swift
```diff
@@ -0,0 +1,902 @@
...
+
+/// Converts a stream of UTF-8 code units into tokens that can be parsed by a TextFormat or JSON
+/// parser.
+///
+/// This is largely based on the `Tokenizer` class from the C++ protobuf implementation.
+package struct Tokenizer {
```
thomasvl: Unless you are going to add some tests, you can likely drop the package visibility.

Sources/SwiftProtobuf/Tokenizer.swift
```diff
@@ -0,0 +1,902 @@
...
+
+    /// Returns the string value of the token.
+    ///
+    /// - Throws: An error if the token is not a valid UTF-8 string of if it contained surrogates
+    ///   when they were not allowed.
+    package func stringValue(allowSurrogates: Bool) throws -> String {
+        guard text.count >= 2 else { return "" }
+        let content = UnsafeBufferPointer(rebasing: text[1..<text.count - 1])
+
+        func utf8Error() -> TextualParsingError {
```
thomasvl: I think you only call this from one place, so could likely inline it there.

Sources/SwiftProtobuf/TextFormatReader.swift
```diff
@@ -33,49 +33,479 @@ struct TextFormatReader {
...
-        scanner: TextFormatScanner,
-        messageSchema: MessageSchema,
-        terminator: UInt8?
-    ) {
-        self.scanner = scanner
+    ) throws {
+        precondition(buffer.baseAddress != nil, "buffer.baseAddress must not be nil")
+
+        var tokenizer = Tokenizer(buffer: buffer)
+        try tokenizer.next()
```
thomasvl, line 47: I guess there's no way to hide the try so the initializer can't fail.

Sources/SwiftProtobuf/TextFormatReader.swift
```diff
@@ -33,49 +33,479 @@ struct TextFormatReader {
...
+
+    /// Consumes a Boolean value from the tokenizer, which may be represented as an identifier or as
+    /// a numeric value.
+    mutating func consumeBool() throws -> Bool {
+        switch tokenizer.current.kind {
+        case .integer:
+            return try consumeUnsignedInteger(upperBound: 1) == 1
+        case .identifier:
+            switch try consumeIdentifier() {
+            case "t", "true", "True":
```
thomasvl, line 70: The single chars was out of spec right?  Should probably comment about that.

Sources/SwiftProtobuf/TextFormatReader.swift
```diff
@@ -33,49 +33,479 @@ struct TextFormatReader {
...
+            default:
+                throw parsingError(expected: "a Boolean value", at: tokenizer.previous)
+            }
+        default:
+            throw parsingError(expected: [.integer, .identifier])
+        }
+    }
+
+    /// Consumes an identifier from the tokenizer.
+    mutating func consumeIdentifier() throws -> String {
```
thomasvl, line 83: Does this have to construct a String?  I guess there's no good way to check against the raw bytes since we know there is no escaping, etc.

Sources/SwiftProtobuf/TextFormatReader.swift
```diff
@@ -33,49 +33,479 @@ struct TextFormatReader {
...
+            }
+            if options.ignoreUnknownFields || messageSchema.isFieldNameReserved(identifier) {
+                return .unknown
+            }
+            throw parsingError(reason: "Unknown field '\(identifier)' on '\(messageSchema.messageName)'")
+        }
+
+        if at(.integer) {
+            // Look up the regular field or extension field by number.
+            let fieldNumber = UInt32(try consumeUnsignedInteger(upperBound: 1 << 29 - 1))
```
thomasvl: I wonder if 1 << 29 - 1 should be a constant some place.

Sources/SwiftProtobuf/TextFormatReader.swift
```diff
@@ -33,49 +33,479 @@ struct TextFormatReader {
...
+    /// - Parameter wasNameAlreadyConsumed: Whether the field name has already been consumed (and
+    ///   thus should not be skipped by this function).
+    mutating func skipField(wasNameAlreadyConsumed: Bool) throws {
+        if !wasNameAlreadyConsumed {
+            if at(.leftBracket) {
+                _ = try consumeAnyTypeURLOrExtensionName()
+            } else if at(.identifier) {
+                _ = try consumeIdentifier()
+            } else if at(.integer) {
+                _ = try consumeUnsignedInteger(upperBound: UInt64.max)
```
thomasvl, line 416: In theory, field number max, right?  (cpp just doesn't do this)

Sources/SwiftProtobuf/TextFormatReader.swift
```diff
@@ -33,49 +33,479 @@ struct TextFormatReader {
...
 
-    
-    /// Parses and returns the number of the next field in the text format string, or nil if the
-    /// end of the input was reached.
-    mutating func nextFieldNumber() throws -> UInt32? {
-        if hasSeenField {
-            scanner.skipOptionalSeparator()
+    /// Skips a field value which is not a message.
+    mutating func skipFieldValue() throws {
+        try decrementRecursionBudget()
```
thomasvl, line 450: I guess I wasn't expecting this for a field, why count it? Does this mean a field that is a message gets a depth of 2?

Tests/SwiftProtobufTests/Test_TextFormatDecodingOptions.swift
```diff
@@ -336,12 +338,10 @@ final class Test_TextFormatDecodingOptions: XCTestCase {
         assertDecodeIgnoringUnknownsSucceeds("bool", "true")
         assertDecodeIgnoringUnknownsSucceeds("bool", "True")
         assertDecodeIgnoringUnknownsSucceeds("bool", "t")
-        assertDecodeIgnoringUnknownsSucceeds("bool", "T")
```
thomasvl, line 339: why keep the lower but not the upper?

Sources/SwiftProtobuf/JSONReader.swift
```diff
@@ -15,63 +15,726 @@
...
+        }
+
+        // When parsing a map key, the value must be a string (required by JSON).
+        if asQuotedString {
+            guard at(.string, .stringWithEscapes) else {
+                throw valueError()
+            }
+            let value = try tokenizer.current.stringValue(allowSurrogates: false)
+            _ = try tokenizer.next()
+            switch value {
```
thomasvl: If you use an `else` for the not quoted string case, you might be able to put the next() and switch after the if...else and collapse this some. No idea it codegen gets better with that or not.

Sources/SwiftProtobuf/JSONReader.swift
```diff
@@ -15,63 +15,726 @@
...
+        return try decodeBase64(from: string)
+    }
+
+    /// Consumes a signed integer from the tokenizer.
     ///
-    /// When the closure has completed, the receiver's current position will be located after the
-    /// closing of the object that was read.
+    /// - Parameter upperBound: The maximum value of the integer to consume. If the integer is
+    ///   greater than `upperBound`, an error will be thrown.
+    mutating func consumeSignedInteger(upperBound: Int64) throws -> Int64 {
```
thomasvl, line 131: It feels like some of the integers handlings are pretty close, I wonder if there's a worth while utility that would do this given a tokenizer to collapse the impls?

Sources/SwiftProtobuf/TextualParser.swift
allevato: Fixed both of these to return `Void`.

Sources/SwiftProtobuf/TextualParser.swift
allevato: Done.

Sources/SwiftProtobuf/Tokenizer.swift
allevato: Since I branched before those changes were made, I was planning to go back through the whole branch in the future and back-patch the `FoundationEssentials` imports in a single PR. There are a bunch of other imports that need to be updated too.

Sources/SwiftProtobuf/Tokenizer.swift
allevato: Done. Can revisit if we want separate tests for it later.

Sources/SwiftProtobuf/Tokenizer.swift
allevato: That's true here, but in a later commit that refactors this method, it ends up being called in two places.

Sources/SwiftProtobuf/TextFormatReader.swift
allevato: Not if we keep it modeled like the C++ tokenizer, where `next` populates the current token state instead of being called with the expectation that it returns it.

I don't think it's a problem in practice; this is called exclusively from throwing methods anyway.

Sources/SwiftProtobuf/TextFormatReader.swift
allevato: The single uppercase `T` and `F` were out of spec, according to https://protobuf.dev/reference/protobuf/textformat-spec/, and the C++ parser only handles the following:

> True values: "True", "true", "t", 1
> False values: "False", "false", "f", 0

Added a comment (and a link to the spec in the main struct doc comment).

Sources/SwiftProtobuf/TextFormatReader.swift
allevato: We might be able to revisit this, but we'd need APIs on the reflection table/schemas to look up based on the raw bytes (wrapped in `UTF8Name`) instead of `String`. It was a bit too much to bite off in this PR.

Sources/SwiftProtobuf/TextFormatReader.swift
allevato: Done.

Sources/SwiftProtobuf/TextFormatReader.swift
allevato: `skipFieldValue` and `skipFieldMessage` are mutually exclusive; they get called after the field name (and possible colon) has already been read, and then the parser calls `skipFieldMessage` if the next token is `{` or `<`, or it calls `skipFieldValue` otherwise.

But this matches the behavior of the C++ parser, which decrements the budget for regular fields as well as submessages. I've verified the behavior with a C++ test harness to make sure that our tests behave the same.

Tests/SwiftProtobufTests/Test_TextFormatDecodingOptions.swift
allevato: See above; only lower is to spec.

Sources/SwiftProtobuf/JSONReader.swift
allevato: Done.

Sources/SwiftProtobuf/JSONReader.swift
allevato: I think we can look at it in the future. There are enough conceptual differences that I didn't want to try to blend them yet, like text format supports different radixes but JSON doesn't, and JSON supports string literals and exponential notation but text format doesn't.

Sources/SwiftProtobuf/TextFormatReader.swift
allevato: Yeah, I don't know why the C++ code doesn't check for this, but since I was aiming to follow their lead I implemented it the same way they did.

thomasvl (approved):

PR #2062, opened 2026-05-14
Add an experimental generator option to suppress field/enum case/message names.

allevato:
If you're only using binary format protos or protos as values in memory, shedding the TextFormat/JSON support can be useful to reduce binary size.


Tests/protoc-gen-swiftTests/Test_GeneratorOptions.swift
```diff
@@ -0,0 +1,51 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+
+import SwiftProtobufPluginLibrary
+import XCTest
+@testable import protoc_gen_swift
```
thomasvl: use `package` to avoid this?

Sources/protoc-gen-swift/FileGenerator.swift
```diff
@@ -225,9 +225,10 @@ class FileGenerator {
                 "// MARK: - Code below here is support for the SwiftProtobuf runtime."
             )
             if needsProtoPackage {
+                let packageStr = generatorOptions.experimentalHiddenNames.contains(.types) ? "" : protoPackage
```
thomasvl: This shouldn't be needed, just leave it out. MessageGenerator shouldn't be referencing it at all.

Tests/protoc-gen-swiftTests/Test_GeneratorOptions.swift
allevato: Done.

Sources/protoc-gen-swift/FileGenerator.swift
allevato: Done.

thomasvl (approved):

PR #2068, opened 2026-05-25
Reject invalid sub-second precision in Timestamp/Duration JSON decoding

adityasingh2400:
The JSON decoders for `google.protobuf.Timestamp` and `google.protobuf.Duration` accept fractional-second values that the format does not permit.

In `parseTimestamp`, the loop that reads the fraction after the `.` consumes any number of digits. Once nine digits have been read, `digitValue` has already divided down to zero, so any further digits contribute nothing and are silently discarded. A `.` with no digits after it falls straight through the loop and is treated as zero nanos rather than an error. As a result these all decode successfully today:

- `"1970-01-01T00:00:00.1234567890Z"` (ten fractional digits, the tenth silently dropped)
- `"1970-01-01T00:00:00.123456789012345Z"` (fifteen digits)
- `"1970-01-01T00:00:00.Z"` (empty fraction)

`parseDuration` has the same issue from the opposite side: it pads a short fraction out to nine digits and truncates anything longer with a `removeLast` loop, so `"1.s"` and `"1.1234567890s"` are accepted too.

The protobuf JSON spec and the reference parser (`TakeNanosAndAdvance` in `json/internal/parser.cc`) require the fraction to contain between one and nine digits and reject anything outside that range. Both decoders now count the fractional digits and throw `malformedTimestamp` / `malformedDuration` when the count is zero or greater than nine. This also removes the silent precision loss on over-long input, since such input is now an explicit error instead of being quietly rounded.

The Duration change drops the now-dead `removeLast` truncation loop, which only ran for inputs that are now rejected up front.

Verification: added decode tests for one-digit and nine-digit fractions (still accepted, with the expected nanos) plus empty and over-nine-digit fractions (now rejected) in `Test_Timestamp` and `Test_Duration`. The new cases fail before the change and pass after. The full `swift test` suite passes (960 tests, 0 failures), including the existing wide-range timestamp round-trip test.

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Timestamp+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Timestamp+Extensions.swift
@@ -152,14 +152,21 @@ private func parseTimestamp(s: String) throws -> (Int64, Int32) {
         var digitValue = 100_000_000
         var fractionalDigits = 0
         while pos < value.count && value[pos] >= zero && value[pos] <= nine {
+            fractionalDigits += 1
+            // The fraction may have at most 9 digits. Reject as soon as a
+            // tenth digit appears so we fail fast and never accumulate past
+            // nanosecond precision, which also keeps the Int32 math below from
+            // trapping on pathological input.
+            if fractionalDigits > 9 {
+                throw JSONDecodingError.malformedTimestamp
+            }
             nanos += Int32(digitValue * (value[pos] - zero))
             digitValue /= 10
-            fractionalDigits += 1
             pos += 1
         }
-        // The fraction must have between 1 and 9 digits; matches the
-        // RFC 3339 grammar and the reference protobuf JSON parser.
-        if fractionalDigits < 1 || fractionalDigits > 9 {
+        // The fraction must have at least one digit, matching the RFC 3339
+        // grammar and the reference protobuf JSON parser.
+        if fractionalDigits < 1 {
             throw JSONDecodingError.malformedTimestamp
         }
     }
--- a/Sources/SwiftProtobuf/Google_Protobuf_Timestamp+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Timestamp+Extensions.swift
@@ -153,10 +153,10 @@ private func parseTimestamp(s: String) throws -> (Int64, Int32) {
         var fractionalDigits = 0
         while pos < value.count && value[pos] >= zero && value[pos] <= nine {
             fractionalDigits += 1
-            // The fraction may have at most 9 digits. Reject as soon as a
-            // tenth digit appears so we fail fast and never accumulate past
-            // nanosecond precision, which also keeps the Int32 math below from
-            // trapping on pathological input.
+            // Protobuf encodes sub-second time as nanoseconds, so the fraction
+            // is limited to 9 digits. Reject as soon as a tenth digit appears
+            // so we fail fast, and so the Int32 accumulation below cannot
+            // overflow on pathological input.
             if fractionalDigits > 9 {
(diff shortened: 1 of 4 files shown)


Sources/SwiftProtobuf/Google_Protobuf_Timestamp+Extensions.swift
```diff
@@ -150,11 +150,18 @@ private func parseTimestamp(s: String) throws -> (Int64, Int32) {
...
+        var fractionalDigits = 0
         while pos < value.count && value[pos] >= zero && value[pos] <= nine {
             nanos += Int32(digitValue * (value[pos] - zero))
             digitValue /= 10
+            fractionalDigits += 1
             pos += 1
         }
+        // The fraction must have between 1 and 9 digits; matches the
+        // RFC 3339 grammar and the reference protobuf JSON parser.
+        if fractionalDigits < 1 || fractionalDigits > 9 {
```
thomasvl: We likely needs the <1 check still here, but the check for >9 should go inside the loop otherwise to catch the failure sooner. It also ensure no input could cause Swift to trap on any of the loop match because of over/underflow.

Tests/SwiftProtobufTests/Test_Duration.swift
```diff
@@ -95,6 +95,22 @@ final class Test_Duration: XCTestCase, PBTestHelpers {
         assertJSONDecodeFails("\"100.001sXXX\"")
     }
 
+    func testJSON_subsecondDigits() throws {
```
thomasvl, line 98: Similar comment as the other PR, might make sense to send protobuf a PR to add some tests for these to the conformance tests in general to benefit all languages.

Sources/SwiftProtobuf/Google_Protobuf_Timestamp+Extensions.swift
adityasingh2400: Done in ccbc5152. I moved the digit check inside the loop so it throws on the tenth fractional digit before the Int32 accumulation, and kept the at-least-one-digit check after the loop. Failing fast there also means an overlong fraction is rejected immediately instead of walking the whole string.

Tests/SwiftProtobufTests/Test_Duration.swift
adityasingh2400: Good idea, same as the other PR. I will look at adding cases for the invalid sub-second precision to the upstream protobuf conformance suite as a follow-up so the other language implementations get the same coverage.

Sources/SwiftProtobuf/Google_Protobuf_Timestamp+Extensions.swift
```diff
@@ -152,14 +152,21 @@ private func parseTimestamp(s: String) throws -> (Int64, Int32) {
...
+            if fractionalDigits > 9 {
+                throw JSONDecodingError.malformedTimestamp
+            }
             nanos += Int32(digitValue * (value[pos] - zero))
             digitValue /= 10
-            fractionalDigits += 1
             pos += 1
         }
-        // The fraction must have between 1 and 9 digits; matches the
-        // RFC 3339 grammar and the reference protobuf JSON parser.
```
thomasvl: Quick follow up - I thought the sub-second digit limit was a protobuf thing and isn't actually in the RFC. We shouldn't say it is part of the RFC if it isn't.

adityasingh2400: Good catch, you are right. RFC 3339 allows arbitrary fractional precision, the 9-digit cap is purely protobuf encoding nanoseconds. Fixed in 85a38a9c: the comments now credit protobuf and the reference parser, with no RFC claim for the digit limit.

thomasvl (approved):
Thank you!

PR #2070, opened 2026-05-26
[table-driven-branch] Compiler releated tweaks

thomasvl:
Happened to try a build of the branch while building something else locally, and I guess I put just enough load on my machine to trip up time limit on the type checker, and then also got some warnings from Xcode. So some attempted fixes for those.


allevato (approved):
Thanks, I actually saw these occasionally (command line would succeed, VS Code's SourceKit would get hung up) so I think it's specific to type checker improvements in certain versions of the compiler.

PR #2073, opened 2026-05-27
Keep destination message fields when merging an unset source field

### LINKED ISSUE #1875: Merging two protobuffers discards fields of original protobuffer.
When using `merge(from:fieldMask:mergeOption:)`, fields of `self` are cleared when they're not present in `source`, but it seems to depend at least partly on the field type. I added a few unit tests to `Test_FieldMask.swift` to find out what's going on, and this is what I found:

This test succeeds:
```
    func testMergeUnsetFieldsOfMessage() throws {
        var message = SwiftProtoTesting_TestAllTypes.with { model in
            model.optionalInt32 = 0
            model.optionalInt64 = 1
            model.repeatedInt32 = [3, 4]
        }
        let secondMessage = SwiftProtoTesting_TestAllTypes.with { model in
            model.optionalInt32 = 0
            model.optionalInt64 = 2
        }
        let fieldMask = Google_Protobuf_FieldMask(allFieldsOf: SwiftProtoTesting_TestAllTypes.self)
        try message.merge(from: secondMessage, fieldMask: fieldMask)
        XCTAssertEqual(message.optionalInt32, 0)
        XCTAssertEqual(message.optionalInt64, 2)
        XCTAssertEqual(message.repeatedInt32, [3, 4])
    }
```

This test fails:
```
    func testMergeUnsetFieldsOfMessage() throws {
        var message = SwiftProtoTesting_TestAllTypes.with { model in
            model.optionalInt32 = 0
            model.optionalInt64 = 1
            model.optionalString = "blah"
        }
        let secondMessage = SwiftProtoTesting_TestAllTypes.with { model in
            model.optionalInt32 = 0
            model.optionalInt64 = 2
        }
        let fieldMask = Google_Protobuf_FieldMask(allFieldsOf: SwiftProtoTesting_TestAllTypes.self)
        try message.merge(from: secondMessage, fieldMask: fieldMask)
        XCTAssertEqual(message.optionalInt32, 0)
        XCTAssertEqual(message.optionalInt64, 2)
        XCTAssertEqual(message.optionalString, "blah")  // <-- this field is missing
    }
```
I would expect both tests to succeed. Note that if I serialize `secondMessage` and use `merge(serializedData:)`, both tests succeed.

**Steps to reproduce**
1. Clone repository
2. Add the above code to `Tests/SwiftProtobufTests/Test_FieldMask.swift`
3. Run `make test` in the root

adityasingh2400:
Fixes #1875.

### Problem

`merge(from:fieldMask:)` discards data on the destination. When the mask names a singular message field but the source has no value for it, the field is cleared on the destination instead of being left alone. The reporter saw this as inconsistent behavior across field types: a repeated field survived the merge while a message field (and a singular string) did not.

Walking through it, `PathDecoder.setMessageValue` always called `setValue(&value, defaultValue: nil)` for a message field named directly by the mask. With no source value that writes `nil`, clearing the destination.

### Fix

The default FieldMask merge behavior in the upstream C++ and Java implementations skips an unset singular message field, so the destination keeps whatever it already holds. This change does the same: the message field is only assigned when the source actually provides a value.

Singular primitives are intentionally left unchanged. The documented default for primitive fields is to copy the source value, including its default when the source is unset, so that part of the report is already spec-conformant.

### Verification

Added two tests to `Test_FieldMask.swift`:

- `testMergeUnsetMessageFieldOfMessage` checks that a message field in the mask is preserved when the source leaves it unset, and still replaced when the source sets it.
- `testMergeAllFieldsKeepsUnsetSourceFields` reproduces the all-fields mask from the report.

Both fail before the change and pass after. The full `SwiftProtobufTests` suite passes (881 tests, 0 failures), and `swift format lint` is clean on the touched files.

### DIFF
--- a/Sources/SwiftProtobuf/PathDecoder.swift
+++ b/Sources/SwiftProtobuf/PathDecoder.swift
@@ -156,6 +156,14 @@ struct PathDecoder<T: Message>: Decoder {
         _ value: inout M?
     ) throws {
         if nextPath.isEmpty {
+            // A singular message field named directly by the mask. When the
+            // source has no value for it, leave the destination untouched
+            // rather than clearing it. This matches the default merge
+            // behavior of the reference implementations, where an unset
+            // singular message field in the source is skipped.
+            guard self.value != nil else {
+                return
+            }
             try setValue(&value, defaultValue: nil)
             return
         }
--- a/Tests/SwiftProtobufTests/Test_FieldMask.swift
+++ b/Tests/SwiftProtobufTests/Test_FieldMask.swift
@@ -140,6 +140,70 @@ final class Test_FieldMask: XCTestCase, PBTestHelpers {
         XCTAssertEqual(message.optionalNestedMessage.bb, 3)
     }
 
+    // A singular message field named directly by the mask should not be
+    // cleared when the source does not set it. The default merge behavior
+    // skips an unset singular message field, leaving the destination intact.
+    func testMergeUnsetMessageFieldOfMessage() throws {
+        var message = SwiftProtoTesting_TestAllTypes.with { model in
+            model.optionalInt32 = 1
+            model.optionalNestedMessage = .with { nested in
+                nested.bb = 2
+            }
+        }
+
+        let secondMessage = SwiftProtoTesting_TestAllTypes.with { model in
+            model.optionalInt32 = 5
+        }
+
+        // The mask covers the message field, but the source leaves it unset,
+        // so the destination keeps its existing value.
+        try message.merge(
+            from: secondMessage,
+            fieldMask: .init(protoPaths: "optional_int32", "optional_nested_message")
+        )
+        XCTAssertEqual(message.optionalInt32, 5)
+        XCTAssertTrue(message.hasOptionalNestedMessage)
+        XCTAssertEqual(message.optionalNestedMessage.bb, 2)
+
+        // When the source does set the message field, it still replaces the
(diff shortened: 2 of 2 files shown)


thomasvl:
Is there a c++ test that covers this case we can reference? (And/or a reference to their code doing this?)

adityasingh2400:
Yes. The canonical behavior is in the C++ FieldMaskUtil. In src/google/protobuf/util/field_mask_util.cc, FieldMaskTree::MergeMessage handles a singular message field by first clearing the destination only when options.replace_message_fields() is set, and then merging from the source only when source_reflection.HasField is true for that field. So when the source does not set the singular message field, the destination is left untouched, because replace_message_fields defaults to false (the MergeOptions constructor in field_mask_util.h initializes replace_message_fields_ to false). That is exactly what this PR does.

The contrast is intentional. The scalar COPY_VALUE macro just above clears the destination when the source is unset, whereas the message case does not, and that difference is the inconsistency the reporter saw in #1875.

For a test reference, TEST(FieldMaskUtilTest, MergeMessage) in src/google/protobuf/util/field_mask_util_test.cc exercises MergeMessageTo across both the replace_message_fields false and true paths.

thomasvl:
Thanks for the pointers. I guess at some point we should have the option for replacing message fields also for completeness, but we can always wait until someone asks for it.

thomasvl (approved):
Thank you again!

PR #2080, opened 2026-05-27
[table-driven-branch] Add `allowPartial` to `BinaryDecodingOptions`.

thomasvl:
- Add new apis that don't take the `partial` argument for decoding.
- Deprecate the old apis that took `partial` and remap them to the new apis.
- Update the tests to all use the new apis.
- Update some of the internal apis over.


thomasvl:
Note: MessageStorage currently is passing the options as individual parameters to some apis. Do we want to move the options more though the internals?

allevato:
> Note: MessageStorage currently is passing the options as individual parameters to some apis. Do we want to move the options more though the internals?

Yeah, we should just pass the options through. The original idea was that the options were the "public facing API" and the internal methods could break it down to just what's necessary, but if we move it to `borrowing` then it's about the same anyway. And if we start moving other things like `ExtensionMap` and the `Any` registry into options, it'll clean up all of those functions.

thomasvl:
Ok, I'll queue that up for followup once you are good with this part.

allevato (approved):

PR #2082, opened 2026-05-27
[table-driven-branch] Remove `partial` from internal apis also.

thomasvl:
Update the rest of the apis in `MessageStorage` to use the `BinaryDecodingOptions` instead of passing the distinct values around with the one footnote being some unknown field handling on helpers.

### DIFF
--- a/Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
+++ b/Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
@@ -36,24 +36,18 @@ extension MessageStorage {
         try merge(
             byReadingFrom: &reader,
             extensions: extensions,
-            partial: options.allowPartial,
-            discardUnknownFields: options.discardUnknownFields)
+            options: options)
     }
 
     /// Decodes field values from the given wire format reader into this storage class.
     ///
     /// - Parameters:
     ///   - buffer: The binary-encoded message data to decode.
-    ///   - partial: If `false` (the default), this method will verify that all required
-    ///     fields are present before encoding. If any are missing, this method throws
-    ///     ``BinaryDecodingError/missingRequiredFields``.
-    ///   - discardUnknownFields: If true, unknown fields will be discarded during
-    ///     parsing.
+    ///   - options: The ``BinaryDecodingOptions`` to use.
     func merge(
         byReadingFrom reader: inout WireFormatReader,
         extensions: ExtensionMap?,
-        partial: Bool,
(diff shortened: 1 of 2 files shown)


Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
```diff
@@ -153,11 +143,12 @@ extension MessageStorage {
                 let isEnumValue = valueField.rawFieldType == .enum
 
                 var subReader = WireFormatReader(buffer: slice, recursionBudget: reader.recursionBudget)
+                var subOptions = options
+                subOptions.discardUnknownFields = !isEnumValue
```
thomasvl: This was the one thing that gave me pause.

Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
```diff
@@ -205,14 +196,14 @@ extension MessageStorage {
                         from: &reader,
                         fieldNumber: tag.fieldNumber,
                         isRepeated: true,
-                        discardUnknownFields: discardUnknownFields
+                        discardUnknownFields: options.discardUnknownFields
```
thomasvl, line 194: And these were the cases I didn't try to force the options through, since it didn't seem right at first glance.

Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
allevato: Hmm, can we push this down a level? Instead of doing this in the message storage that contains the map field, we could check it when reading a singular/repeated enum field in the map entry's own storage. Within a `MessageStorage`, `self.schema.extensibilityMode == .mapEntry` should tell us that we _are_ a map entry, and then you can conditionalize the discard-unknown logic based on that.

arifhosenbelal42-a11y (approved):

Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
thomasvl: Looks like you already did this, and it is left over from before we moved to the tokenizer.

Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
allevato: Unless we need something else from the options there later, this seems fine.

allevato (approved):

PR #2085, opened 2026-05-28
[table-driven-branch] Optimize required field validation during decoding

thomasvl:
Like we did in encoding, do the validation while working on the message graph. However, there are two issues that prevent directly raising when we find a missing required field:

1.  It would throw a false positive failure if an incomplete message is seen on the wire but is later completed when the sub-message occurs again. This could happen with say binary blobs concatenated together.

2.  It would return a false success if you are decoding into a message that already has some sub-message fields present (merging). If the sub-message does not occur in the binary payload, we will never visit it and discover the incomplete sub-message.

To deal with these, we do the validation along the way and expose that issue to the caller as a hint. Then the top most layers can use that information along with the operation being done (merge vs. new load) to decided if `isInitialized` still needs to be called.

Since some of the decoding apis already have return result to indicate if things should go into unknown fields, we use a boolean `inout` parameter. This also make things slightly more clear then a boolean return result from a "merge" that isn't some sort of success/failure.

TextFormat decode of an Any doesn't bother with the validation since the expanded output would be better.

### DIFF
--- a/Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
+++ b/Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
@@ -25,30 +25,47 @@ extension MessageStorage {
     ///
     /// - Parameters:
     ///   - buffer: The binary-encoded message data to decode.
+    ///   - extensions: The extension map to use.
     ///   - options: The ``BinaryDecodingOptions`` to use.
+    ///   - isInitializedShallow: An `inout` boolean used to track if required fields are
+    ///     shallowly present across the entire message tree parsed so far. It is passed as `inout`
+    ///     because deep recursion and multi-pass partial updates (e.g. out-of-order fields)
+    ///     require accumulative status tracking across nested submessages and extensions. The
+    ///     caller is assumed to have initialized it to true before the first call, and it will
+    ///     be set to false if any message fails the initialization check while decoding.
     /// - Throws: ``BinaryDecodingError`` if decoding fails.
     func merge(
         byReadingFrom buffer: UnsafeRawBufferPointer,
         extensions: ExtensionMap?,
-        options: BinaryDecodingOptions
+        options: BinaryDecodingOptions,
+        isInitializedShallow: inout Bool
     ) throws {
         var reader = WireFormatReader(buffer: buffer, recursionBudget: options.messageDepthLimit)
         try merge(
             byReadingFrom: &reader,
             extensions: extensions,
-            options: options
+            options: options,
+            isInitializedShallow: &isInitializedShallow
         )
     }
 
     /// Decodes field values from the given wire format reader into this storage class.
     ///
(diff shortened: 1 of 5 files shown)


Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
```diff
@@ -69,16 +87,28 @@ extension MessageStorage {
             // group tracking state. If that didn't happen, then we ran out of data too early.
             throw BinaryDecodingError.truncated
         }
-        if !options.allowPartial && !isInitialized {
```
thomasvl, line 72: In working on this, I realized this was probably bad. Doesn't this get called for each sub message along the way?  i.e. - since it calls `isInitialized`, it would check the transitive graph, then as we recurse up to the message this message was a field of, it when that finished decoding, it would do the recursive check again…

Sources/SwiftProtobuf/MessageStorage+JSONEncoding.swift
```diff
@@ -392,11 +392,19 @@ extension MessageStorage {
         let bytes = assumedPresentValue(at: valueField.offset) as Data
         try bytes.withUnsafeBytes { buffer in
             let messageStorage = MessageStorage(schema: messageSchema)
+            // TODO: Should we keep the original behavior of failing for required fields in JSON?
```
thomasvl, line 395: @allevato do you happen to remember what cpp did when you were just revisiting all the parsing?

Sources/SwiftProtobuf/MessageStorage+TextEncoding.swift
```diff
@@ -61,10 +61,15 @@ extension MessageStorage {
             let bytes: Data = value(of: valueField)
             do {
                 try bytes.withUnsafeBytes { buffer in
+                    // We still want the expanded verbose output even if the submessage is partial.
```
thomasvl, line 64: Called this out in the commit message also, but these seems generally worth while, thoughts?

Sources/SwiftProtobuf/MessageStorage+BinaryDecoding.swift
allevato: Yeah, I wonder if this was supposed to be `isInitializedShallow` and I just mistyped it.

Sources/SwiftProtobuf/MessageStorage+TextEncoding.swift
allevato: Yeah, that makes sense.

Sources/SwiftProtobuf/MessageStorage+JSONEncoding.swift
allevato: I'm not sure; I'd have to dig through the C++ code again.

allevato (approved):

PR #2090, opened 2026-06-02
Always use PathToUnderscores naming for SwiftPM plugin outputs

### LINKED ISSUE #1761: Duplicate proto filenames fail build even with fileNaming=PathToUnderscores
When using PathToUnderscores, we are now getting an error that the generated file does not exist. However, the path to the non-existant file is the full path version (e.g. `some/path/to/my.pb.swift`). I believe the issue is in plugin.swift where the proto input file is mapped to an output file

```
            let protobufOutputPath = outputDirectory.appending(file)

            // Add the outputPath as an output file
            outputFiles.append(protobufOutputPath)
```

That sets up a dependency from `some/path/to/my.proto` to `some/path/to/my.pb.swift` but the latter doesn't exist since the generator actually generated `some_path_to_my.pb.swift`.

* MacOS 15.3
* Xcode 16.2
* Swift 6.0
* SwiftProtobuf 1.29.0

adityasingh2400:
The SwiftProtobufPlugin build tool plugin declared the path of every generated file by taking the input proto path and swapping the .proto suffix for .pb.swift while keeping the full relative path, and it passed the configured fileNaming straight through to protoc-gen-swift.

That breaks the build whenever the declared output name does not match what protoc-gen-swift actually writes. With PathToUnderscores or DropPath the generator writes a differently named file than the plugin declared, and two protos that share a file name in different subdirectories collide. The build system then fails because the declared output does not exist. This is the problem reported in #1761.

Per the discussion on this PR: the generated files go into the build directory, so the file name on disk is never observed and there is no value in giving users a choice for file naming when using the build plugin. The plugin now always uses PathToUnderscores, the same approach grpc-swift-protobuf takes.

What changed:

- The protoc invocation is pinned to FileNaming=PathToUnderscores.
- The declared output names are derived with the same transformation, dropping the .proto suffix and replacing the directory separators with underscores. Because both sides use one strategy they can never diverge.
- The mirrored copy of FileGenerator.outputFilename is gone, so there is no logic to keep in sync with protoc-gen-swift and the maintenance concern raised in the thread disappears.

On the existing fileNaming config field: it already exists on main, so I kept decoding it for backwards compatibility. It may be omitted or set to PathToUnderscores. Any other value is now a clear configuration error rather than a silently ignored setting, since dropping a user's explicit choice on the floor would be worse.

Verification: added a PathToUnderscores example target whose two protos share the file name Duplicate.proto in different subdirectories. This only builds because of the underscore naming, which is the case from #1761. swift build and swift test --package-path PluginExamples pass, including the new testPathToUnderscores case, and the existing examples stay green under the new default. I also confirmed by hand that a DropPath or FullPath setting now produces the configuration error.

Fixes #1761

### DIFF
--- a/Plugins/SwiftProtobufPlugin/plugin.swift
+++ b/Plugins/SwiftProtobufPlugin/plugin.swift
@@ -11,8 +11,6 @@ struct SwiftProtobufPlugin {
         case invalidInputFileExtension(String)
         /// Indicates that there was no configuration file at the required location.
         case noConfigFound(String)
-        /// Indicates that the configuration set a `fileNaming` option the build plugin does not support.
-        case unsupportedFileNaming(String)
 
         var description: String {
             switch self {
@@ -24,13 +22,6 @@ struct SwiftProtobufPlugin {
                 return """
                     No configuration file found named '\(path)'. The file must not be listed in the \
                     'exclude:' argument for the target in Package.swift.
-                    """
-            case let .unsupportedFileNaming(value):
-                return """
-                    The 'fileNaming' option '\(value)' is not supported by the build plugin. The build \
-                    plugin always generates files using the 'PathToUnderscores' naming because the \
-                    generated files go into the build directory and the name is never observed. Remove \
-                    the 'fileNaming' option or set it to 'PathToUnderscores'.
                     """
             }
         }
@@ -298,10 +289,16 @@ struct SwiftProtobufPlugin {
                     throw PluginError.invalidInputFileExtension(protoFile)
                 }
             }
-            // The build plugin always uses PathToUnderscores naming. Reject any other explicit
-            // value so the setting is never silently ignored.
+            // The build plugin always uses PathToUnderscores naming. Warn on any other explicit
+            // value so the setting is never silently ignored, without breaking existing builds.
             if let fileNaming = invocation.fileNaming, fileNaming != .pathToUnderscores {
-                throw PluginError.unsupportedFileNaming(fileNaming.rawValue)
+                Diagnostics.warning(
+                    """
+                    The 'fileNaming' option '\(fileNaming.rawValue)' is ignored by the build plugin. The build \
+                    plugin always generates files using the 'PathToUnderscores' naming because the \
+                    generated files go into the build directory and the name is never observed.
+                    """
+                )
             }
         }
     }
--- a/Plugins/SwiftProtobufPlugin/plugin.swift
+++ b/Plugins/SwiftProtobufPlugin/plugin.swift
@@ -228,10 +228,23 @@ struct SwiftProtobufPlugin {
             protocArgs.append("--swift_opt=Visibility=\(visibility.rawValue)")
         }
 
+        // The build plugin always uses PathToUnderscores naming. Warn once per invocation on any
+        // other explicit value so the setting is never silently ignored, without breaking existing
+        // builds.
+        if let fileNaming = invocation.fileNaming, fileNaming != .pathToUnderscores {
+            Diagnostics.warning(
+                """
+                The 'fileNaming' option '\(fileNaming.rawValue)' is ignored by the build plugin. The build \
+                plugin always generates files using the 'PathToUnderscores' naming because the \
+                generated files go into the build directory and the name is never observed.
+                """
+            )
+        }
+
         // Always generate with PathToUnderscores naming. The declared output paths below are
         // derived the same way, so the names the build system expects can never drift from the
-        // names protoc-gen-swift actually writes. The configured fileNaming, if any, was already
-        // validated to be PathToUnderscores.
+        // names protoc-gen-swift actually writes. The configured fileNaming, if any, only triggers
+        // the warning above and does not change the naming used.
         protocArgs.append("--swift_opt=FileNaming=PathToUnderscores")
 
         // Add the implementation only imports flag if it was set
@@ -288,17 +301,6 @@ struct SwiftProtobufPlugin {
                 if !protoFile.hasSuffix(".proto") {
                     throw PluginError.invalidInputFileExtension(protoFile)
                 }
-            }
-            // The build plugin always uses PathToUnderscores naming. Warn on any other explicit
-            // value so the setting is never silently ignored, without breaking existing builds.
-            if let fileNaming = invocation.fileNaming, fileNaming != .pathToUnderscores {
-                Diagnostics.warning(
-                    """
-                    The 'fileNaming' option '\(fileNaming.rawValue)' is ignored by the build plugin. The build \
-                    plugin always generates files using the 'PathToUnderscores' naming because the \
-                    generated files go into the build directory and the name is never observed.
-                    """
-                )
             }
         }
     }
--- a/Plugins/SwiftProtobufPlugin/plugin.swift
+++ b/Plugins/SwiftProtobufPlugin/plugin.swift
@@ -241,7 +241,7 @@ struct SwiftProtobufPlugin {
         // derived the same way, so the names the build system expects can never drift from the
         // names protoc-gen-swift actually writes. The configured fileNaming, if any, was already
         // validated to be PathToUnderscores.
-        protocArgs.append("--swift_opt=FileNaming=\(Configuration.Invocation.FileNaming.pathToUnderscores.rawValue)")
+        protocArgs.append("--swift_opt=FileNaming=PathToUnderscores")
 
         // Add the implementation only imports flag if it was set
         if let implementationOnlyImports = invocation.implementationOnlyImports {
--- a/Plugins/SwiftProtobufPlugin/plugin.swift
+++ b/Plugins/SwiftProtobufPlugin/plugin.swift
@@ -11,6 +11,8 @@ struct SwiftProtobufPlugin {
         case invalidInputFileExtension(String)
         /// Indicates that there was no configuration file at the required location.
         case noConfigFound(String)
+        /// Indicates that the configuration set a `fileNaming` option the build plugin does not support.
+        case unsupportedFileNaming(String)
 
         var description: String {
             switch self {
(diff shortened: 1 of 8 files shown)


Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -272,6 +272,49 @@ struct SwiftProtobufPlugin {
...
+    /// `FileGenerator.outputFilename`, so the names declared as build command
+    /// outputs match the files that are actually written. The input is always a
+    /// `.proto` file because that is validated up front.
+    ///
+    /// - Parameters:
+    ///   - protoFile: The relative path of the input `.proto` file.
+    ///   - fileNaming: The file naming strategy, or `nil` for the default
+    ///     (`FullPath`).
+    /// - Returns: The relative path of the generated `.pb.swift` file.
+    static func outputFileName(
```
thomasvl: @FranzBusch @glbrntt - Should we consider moving the options into the plugin target or making a PluginOptions target that just have these enums and the backing code for them so it can be shared between the SwiftPM plugin and protoc generator (and then maybe also used by the grpc, etc.) to avoid having duplicate copies of the names/functionality?

FranzBusch: Sadly it isn't possible to share code between a plugin and another target directly. Swift PM doesn't support dependencies in plugins. The only thing we could do is copy or symlink

thomasvl: Wait, can plugins not depend on [other libraries to share code](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/writingbuildtoolplugin/#Build-tool-target-dependencies)?  Guess that means we do have to duplicate stuff.  We could share it between the protoc generators, but looks like SwiftPM plugins might have to be self contained?  Ugh.

thomasvl: Ha, looks like we just both replied at the same time.

Symlink seems likely to break when making edits based on what someone's editor does for saves.

Copy has the risk of drifting, we could have the `Makefile` check that they stay in sync, but not great. Think either is worth it?

thomasvl: And even if we share it via the symlink/copy here, that wouldn't help something like grpc (the SwiftPM plugin). As there the copy/symlink becomes even more of a maintenance issue.

adityasingh2400:
For what it is worth on the drift concern: the new `outputFileName` in the plugin is a deliberate line-for-line mirror of `FileGenerator.outputFilename` in protoc-gen-swift. Same three cases, same `dir + base + ext` for `fullPath`, same `/`-to-`_` replacement for `pathToUnderscores`, same `base + ext` for `dropPath`, and the dir/base split matches `splitPath`. I duplicated it precisely because, as you note, the plugin target cannot depend on the generator's library code, and the build breaks if the predicted output name does not match what the generator writes.

I am happy to follow whichever sharing strategy you land on. If you go with a copied source file plus a `Makefile` drift check, I can move the naming logic into a single small file and add that check in this PR. If the conclusion is to leave it duplicated for now given it would not help grpc anyway, I can instead add a short comment on both sides pointing at each other as the source of truth so a future edit to one is a clear signal to update the other. Just let me know which you prefer and I will adjust.

Plugins/SwiftProtobufPlugin/plugin.swift
glbrntt: I don't think there's actually any value in giving users a choice for file naming when using the build plugin as the outputs go into the build dir. 

We should just always use path-to-underscores (this is what gRPC does). This also means the transformation is straightforward which removes the maintenance problem.

adityasingh2400: Agreed, this is the right call. I reworked the PR to always use PathToUnderscores.

The protoc invocation is now pinned to FileNaming=PathToUnderscores, and the declared output names are derived with the same transformation (drop .proto, replace the directory separators with underscores). Because both the generator invocation and the declared outputs use one strategy, they can never diverge. The mirrored copy of FileGenerator.outputFilename is gone, so there is nothing left to keep in sync and the maintenance concern goes away.

On the config field: fileNaming already exists on main, so I kept decoding it for backwards compatibility. It may be omitted or set to PathToUnderscores, and any other value is now a clear configuration error rather than a silently ignored setting, since dropping a user's explicit choice would be worse than telling them. The example target now builds two protos that share the name Duplicate.proto in different subdirectories, which only works because of the underscore naming and covers the original #1761 case. The PR description is updated to match.

thomasvl:
@adityasingh2400 lets not worry about sharing the code immediately in this PR, it's something we can consider in the future. Swift gRPC has a close duplicate of this, and it would be nice to share that, but given the limitation that SwiftPM plugins can't have dependencies, there isn't an obvious good way to deal with it. 😦 

Also, sorry for the delay here, with WWDC next week, Apple folks are little a little busy, and I've had a bunch of my plate, but I'll take a look now.

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -222,10 +237,11 @@ struct SwiftProtobufPlugin {
...
 
-        // Add the file naming if it was set
-        if let fileNaming = invocation.fileNaming {
-            protocArgs.append("--swift_opt=FileNaming=\(fileNaming.rawValue)")
-        }
+        // Always generate with PathToUnderscores naming. The declared output paths below are
+        // derived the same way, so the names the build system expects can never drift from the
+        // names protoc-gen-swift actually writes. The configured fileNaming, if any, was already
+        // validated to be PathToUnderscores.
+        protocArgs.append("--swift_opt=FileNaming=\(Configuration.Invocation.FileNaming.pathToUnderscores.rawValue)")
```
thomasvl: If we are really just doing one mode, I'd likely just hardcode it so the option could eventually go away.

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -280,6 +298,11 @@ struct SwiftProtobufPlugin {
                     throw PluginError.invalidInputFileExtension(protoFile)
                 }
             }
+            // The build plugin always uses PathToUnderscores naming. Reject any other explicit
+            // value so the setting is never silently ignored.
+            if let fileNaming = invocation.fileNaming, fileNaming != .pathToUnderscores {
```
thomasvl: @glbrntt you all good with making this error if folks had another value set? Or should print a diagnostic of some kind about ignoring it?

Plugins/SwiftProtobufPlugin/plugin.swift
adityasingh2400: Done in a9163ddd, the protoc arg is now the hardcoded string so nothing references the FileNaming enum on this path anymore.

adityasingh2400:
Sounds good on deferring the code sharing, and no worries about the timing. I hardcoded the protoc option string per the inline note. For the existing fileNaming config values I will follow whichever way you and @glbrntt land, keeping the clear error or switching to a printed diagnostic that says the value is ignored. Both are a small change on my side.

Plugins/SwiftProtobufPlugin/plugin.swift
glbrntt: Rather than throwing an error we should emit a `Diagnostic.warning()` (I _think_ that's the API) to avoid breaking builds.

adityasingh2400: Done in 673b124f. validateConfiguration now emits Diagnostics.warning saying the value is ignored and PathToUnderscores is always used, and the unsupportedFileNaming error case is gone. Existing configs keep building.

glbrntt: The validation and diagnostic should come earlier: it's currently in the input file loop so users would see the diagnostic repeated N times. I think it should come just before `--swift_opt=FileNaming=PathToUnderscores` is added to the args.

adityasingh2400:
Good call. I moved the validation and diagnostic out of the per-file validation loop. The ignored-fileNaming warning now fires once per invocation in invokeProtoc, right before the FileNaming=PathToUnderscores arg is added, so it can no longer be repeated N times. Behavior is otherwise unchanged. Also rebased on latest main.

glbrntt (approved):
LGTM, thank you!

Plugins/SwiftProtobufPlugin/plugin.swift
```diff
@@ -244,17 +264,19 @@ struct SwiftProtobufPlugin {
...
-            file.removeLast(5)
-            file.append("pb.swift")
-            let protobufOutputPath = outputDirectory.appending(path: file)
+            // The output file name has to match exactly what protoc-gen-swift writes, otherwise
+            // the build system looks for a file that does not exist and the build fails. We always
+            // generate with PathToUnderscores naming, so the relative proto path becomes a single
+            // file name with the directory separators replaced by underscores. We validated up
+            // front that every file has the .proto suffix, which is dropped for .pb.swift.
+            let outputName = String(file.dropLast(".proto".count))
+                .replacingOccurrences(of: "/", with: "_") + ".pb.swift"
```
glbrntt: The formatter is unhappy with this:

```diff
--- a/Plugins/SwiftProtobufPlugin/plugin.swift
+++ b/Plugins/SwiftProtobufPlugin/plugin.swift
@@ -274,7 +274,8 @@ struct SwiftProtobufPlugin {
             // generate with PathToUnderscores naming, so the relative proto path becomes a single
             // file name with the directory separators replaced by underscores. We validated up
             // front that every file has the .proto suffix, which is dropped for .pb.swift.
-            let outputName = String(file.dropLast(".proto".count))
+            let outputName =
+                String(file.dropLast(".proto".count))
                 .replacingOccurrences(of: "/", with: "_") + ".pb.swift"
             let protobufOutputPath = outputDirectory.appending(path: outputName)
```

adityasingh2400:
Rebased onto main and fixed the two reds.

swift-format wanted the chained call on the output name broken across lines. Rather than take the wrap it inserts, which reads awkwardly, I pulled the base name into its own local. I confirmed a re-run of `swift format format --in-place` over the tree leaves that form untouched, so the check should stay green.

The semantic version label check failed because the run predates the `semver/patch` label being applied. The label is on now, so that one should clear on the fresh run.

PR #2092, opened 2026-06-05
Standardize depth enforcement.

thomasvl:
Use `defer` to avoid accidentally missing a call on a `return` path.

This also counts "Value" as an object; that effectively halves the depth of some raw JSON parsed into a "Struct", but it was allowing what might be a stack bomb attempt to be parsed in some cases.


thomasvl:
@Lukasa @FranzBusch @tbkka - I'm putting this up in _draft_ for thoughts. The Protobuf JSON spec doesn't formally say what counts for _depth_, but we already had "Struct" and "List" counting, "Value" is the potential gap here. Since this would have impact on things that current parse, but could be dangerous, I'm not sure if we should change things, or just let it go.  Thoughts?

FranzBusch:
Looks in general fine to me.

thomasvl:
@Lukasa @tbkka Any thoughts/concerns on what I flagged in that first comment?

thomasvl:
@Lukasa ok to land? Please a look at my first comment about the one concern I had.

FranzBusch (approved):

PR #2099, opened 2026-06-10
[table-driven-branch] Fix memory leaks.

allevato:
This fixes three leaks that were in the table-driven implementation:

* When deinitializing `ExtensionStorage`, we were correctly deinitializing any submessages/repeated fields that were in extension fields, but we were not deallocating the heap-allocated buffer that held them.
* `updatePresence` had a bit-masking bug that was causing it to return false when a field was overwritten, thus failing to deinitialize the old value.
* When a `oneof` member is overwritten by another `oneof` member, we update the presence _before_ deinitializing the old value, but `deinitializeField` exits early if the field isn't present, which it would not be in this case.

Since running `leaks` on `xctest` bundles is nontrivial, I've added a test case that does this at process exit when an environment variable is set.

### DIFF
--- a/Tests/SwiftProtobufTests/Test_Leaks.swift
+++ b/Tests/SwiftProtobufTests/Test_Leaks.swift
@@ -1,4 +1,4 @@
-// Tests/SwiftProtobufTests/Test_Leaks.swift - Proto3 coding/decoding
+// Tests/SwiftProtobufTests/Test_Leaks.swift - Memory leak detection
 //
 // Copyright (c) 2014 - 2026 Apple Inc. and the project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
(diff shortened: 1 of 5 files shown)


Tests/SwiftProtobufTests/Test_Leaks.swift
```diff
@@ -0,0 +1,90 @@
+// Tests/SwiftProtobufTests/Test_Leaks.swift - Proto3 coding/decoding
```
thomasvl: Fix comment on the line?

thomasvl (approved):

Tests/SwiftProtobufTests/Test_Leaks.swift
allevato: Done.

PR #2109, opened 2026-06-13
Add explicit watchOS 9 platform minimum for Xcode 27

kenjichanhkg:
Xcode 27 infers watchOS 8.0 for Swift packages without an explicit watchOS deployment target in Package.swift, which is below the supported range (9.0–27.x) and triggers deprecated armv7k architecture errors.

Declare watchOS 9 as the package minimum so SPM uses a supported deployment target. This matches the workaround recommended on Apple Developer Forums and fixes watchOS builds without changing runtime behavior for apps already targeting watchOS 9+.

Refs:
- https://github.com/swiftlang/swift-package-manager/issues/10187
- https://developer.apple.com/forums/thread/831681

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -21,6 +21,9 @@ let resources = [Resource]()
 
 let package = Package(
     name: "SwiftProtobuf",
+    platforms: [
+        .watchOS(.v9),
+    ],
     products: [
         .executable(
             name: "protoc-gen-swift",


Lukasa:
Unfortunately, this would be a breaking change for all earlier Swift releases, as this minimum will be higher than the inferred minimum by Xcode. There is not really a suitable workaround for this other than a new Xcode beta release.

kenjichanhkg:
Thanks for explaining — I wasn’t aware that an explicit `platforms` minimum could behave differently on older Xcode releases than the inferred default when `platforms` is omitted. That makes sense as a semver/manifest concern.

For our case (watchOS app, Xcode 27), the failure was SPM inferring watchOS 8.0 for SwiftProtobuf targets, which is below the supported range and triggers the deprecated armv7k errors. Adding `watchOS(.v9)` locally fixes that; we’ll keep a fork/submodule patch until the SPM inference fix lands in a released Xcode (swift-package-manager#10188).

Happy to close this if you’d rather wait for the toolchain fix than bump package minimums project-wide — just wanted to flag the issue in case others hit the same Xcode 27 watchOS build errors.

thomasvl:
@Lukasa should we close this then since it won't be merged?

Lukasa:
I think so.

PR #2112, opened 2026-06-18
reject unescaped control characters in json strings

jmestwa-coder:
Two JSON string scanners accept raw control characters that RFC 8259 and the upstream parser reject:
- parseOptionalQuotedString: unescaped U+0000-U+001F bytes pass straight into decoded string values and map keys
- nextOptionalKey repeats the same scan for field and enum names
Reject bytes below 0x20 inside a string in both; the escaped forms stay valid.


thomasvl (approved):
Thank you!

PR #2114, opened 2026-06-19
reject illegal characters in field mask json paths

jmestwa-coder:
The field mask JSON codec passed unrecognized characters straight through instead of rejecting them, unlike the upstream parser which only allows path characters in [A-Za-z0-9.]:
- JSONToProto accepted segments like "foo-bar", "foo/bar", "foo bar" into path values on decode
- ProtoToJSON did the same on encode, so a path holding the "," separator round-trips back into two separate paths

Both default branches now return nil, so decode throws malformedFieldMask and encode throws fieldMaskConversion. Legal paths and the camelCase/snake_case mapping are unchanged.

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_FieldMask+Extensions.swift
@@ -47,10 +47,7 @@ private func ProtoToJSON(name: String) -> String? {
         case "a"..."z", "0"..."9", ".", "(", ")":
             jsonPath.append(c)
         default:
-            // TODO: Change this to `return nil`
-            // once we know everything legal is handled
-            // above.
-            jsonPath.append(c)
+            return nil
         }
     }
     return jsonPath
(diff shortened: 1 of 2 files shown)


thomasvl:
Can you check if there is C++ coverage for this for reference?

jmestwa-coder:
checked the C++ side. the behavior is enforced in the JSON codec, but I don't see a dedicated test for it:

- decode: `ParseFieldMask` in json/internal/parser.cc returns "unexpected character in FieldMask" for anything outside [a-z0-9.] (uppercase mapped to snake_case), unless `allow_legacy_nonconformant_behavior` is set (parser.cc:1017-1020).
- encode: `WriteFieldMask` in json/internal/unparser.cc does the same (unparser.cc:753-754).

that legacy flag is the analog of the old passthrough these TODOs described. on the test side, json_test.cc only has `TestFieldMask` (valid round-trip) and `TestFieldMaskSnakeCase`; the "unexpected character in FieldMask" string doesn't appear in any test, so the rejection path itself isn't covered upstream. the cases I added here mirror the conformant C++ behavior.

thomasvl (approved):
Thank you and thank you for the test cases!

PR #2116, opened 2026-06-22
[table-driven-branch] Add the ExperimentalHiddenNames generator option.

allevato:
This is https://github.com/apple/swift-protobuf/pull/2062 but on the table-driven branch.

As with the empty `_NameMap` on main, the implementation falls out fairly gracefully if we simply don't generate anything for the relevant sections of the `ReflectionTable`s.

I've also removed the `_protobuf_packageName` variable that was still being generated, since that's unused in the new implementation.


Sources/protoc-gen-swift/EnumGenerator.swift
```diff
@@ -79,12 +79,14 @@ class EnumGenerator {
...
         self.enumSchemaCalculator = EnumSchemaCalculator(
-            fullyQualifiedName: enumDescriptor.fullName,
+            fullyQualifiedName: generatorOptions.experimentalHiddenNames.contains(.types)
+                ? "" : enumDescriptor.fullName,
             enumValues: mainEnumValueDescriptorsSorted
         )
         self.compressedReflectionData = ReflectionTableCalculator(
             enumValues: mainEnumValueDescriptorsSorted,
-            aliasInfo: aliasInfo
+            aliasInfo: aliasInfo,
```
thomasvl, line 88: Rather than checking the options for both of these and changing what we pass, would it make sense to pass the options to EnumSchemaCalculate and ReflactionTableCalculator (and the field on below), and them them inspect the option and adapt?  Noticed it because of your TODO later on that might be releated.

allevato: The TODO is still only relevant w.r.t. `suppressNames`; I don't think there's any need to pass the entire option down. This keeps the API boundary a little cleaner by only passing what's necessary to the subcomponents.

thomasvl (approved):

PR #2124, opened 2026-07-03
reject duplicate keys in json maps

jmestwa-coder:
JSON map decoding kept the last of a repeated key instead of rejecting it, unlike the reference parser (json/internal/parser.cc ParseMap, which tracks keys in a set):
- decodeMapField overwrote silently for the scalar, enum, and message maps
- Struct decodeJSON did the same, since a Struct is a map<string, Value>
- each now tracks the keys seen in the current object and rejects a repeat

Distinct keys still decode, and a map field repeated at the message level still merges.

### DIFF
--- a/Sources/SwiftProtobuf/Google_Protobuf_Struct+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Struct+Extensions.swift
@@ -45,8 +45,12 @@ extension Google_Protobuf_Struct: _CustomJSONCodable {
         if decoder.scanner.skipOptionalObjectEnd() {
             return
         }
+        var keysSeen = Set<String>()
         while true {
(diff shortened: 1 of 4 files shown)


thomasvl (approved):
I don't think it really matters, but I'll note it as I look at the PR:

The Struct case will detect the duplicate key *before* reading the value. The other code paths will read the value *before* detecting the duplicate key. So in some case, a duplicate key could get reported as some other failure as a side effect of reading the value. Since it is just a debate of what type of failure those cases trigger, it likely doesn't matter.

thomasvl:
Thank you!

jmestwa-coder:
fair point on the check ordering, the map paths reuse the existing value decode so the set check lands after it. thanks for merging this.

PR #2126, opened 2026-07-07
[table-driven-branch] Use the type & accessor vs. raw strings.

thomasvl:
Reduces the change of errors if these ever need to evolve in the future.

### DIFF
--- a/Sources/SwiftProtobufPluginLibrary/CodeGenerator.swift
+++ b/Sources/SwiftProtobufPluginLibrary/CodeGenerator.swift
@@ -125,15 +125,15 @@ extension CodeGenerator {
                 // sites as well.
                 let extendedMessageName = e.extendedMessage.messageName
                 precondition(
-                    extendedMessageName.utf8CodeUnitsEqual("google.protobuf.EnumOptions")
-                        || extendedMessageName.utf8CodeUnitsEqual("google.protobuf.EnumValueOptions")
-                        || extendedMessageName.utf8CodeUnitsEqual("google.protobuf.ExtensionRangeOptions")
-                        || extendedMessageName.utf8CodeUnitsEqual("google.protobuf.FieldOptions")
-                        || extendedMessageName.utf8CodeUnitsEqual("google.protobuf.FileOptions")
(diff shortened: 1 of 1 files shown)


Sources/SwiftProtobufPluginLibrary/CodeGenerator.swift
```diff
@@ -125,15 +125,17 @@ extension CodeGenerator {
...
-                    extendedMessageName.utf8CodeUnitsEqual("google.protobuf.EnumOptions")
-                        || extendedMessageName.utf8CodeUnitsEqual("google.protobuf.EnumValueOptions")
-                        || extendedMessageName.utf8CodeUnitsEqual("google.protobuf.ExtensionRangeOptions")
-                        || extendedMessageName.utf8CodeUnitsEqual("google.protobuf.FieldOptions")
-                        || extendedMessageName.utf8CodeUnitsEqual("google.protobuf.FileOptions")
-                        || extendedMessageName.utf8CodeUnitsEqual("google.protobuf.MessageOptions")
-                        || extendedMessageName.utf8CodeUnitsEqual("google.protobuf.MethodOptions")
-                        || extendedMessageName.utf8CodeUnitsEqual("google.protobuf.OneofOptions")
-                        || extendedMessageName.utf8CodeUnitsEqual("google.protobuf.ServiceOptions"),
+                    extendedMessageName.utf8CodeUnitsEqual(Google_Protobuf_EnumOptions.protoMessageName)
```
allevato: Can you use `.messageSchema.messageName` instead of `.protoMessageName`? I'd like to deprecate and eventually remove the latter.

thomasvl: That works, done.

I was originally thinking the comment/todo you have about `String` interfaces meant it was a better path, but this can always be revised.

allevato: Yeah, since the left-hand-side is already a `UTF8Name` and forced to use `utf8CodeUnitsEqual`, using the `package`-visible schema property here doesn't bother me.

allevato (approved):

Issue #2132, opened 2026-08-02
Upstream has some new conformance tests that need a look


thomasvl:
```
  Required.Editions_Proto3.JsonInput.TimestampJsonInputNonLeapFeb29 # Should have failed to parse, but didn't.
  Required.Editions_Proto3.JsonInput.TimestampJsonInputSecondTooLarge # Should have failed to parse, but didn't.
  Required.Proto3.JsonInput.TimestampJsonInputNonLeapFeb29 # Should have failed to parse, but didn't.
  Required.Proto3.JsonInput.TimestampJsonInputSecondTooLarge # Should have failed to parse, but didn't.
```

PR #2135, opened 2026-08-03
Ensure seconds are <60 not <62.

thomasvl:
Git history doesn't really explain if this was a typo or by design, but a new upstream conformance test is testing it. timestamp.proto also makes a comment that all minutes only have 60 seconds, and leap seconds should be smeared instead, so that seems to be the only other thing to point to here.

Progress on #2132

### DIFF
--- a/Sources/Conformance/failure_list_swift.txt
+++ b/Sources/Conformance/failure_list_swift.txt
@@ -5,4 +5,3 @@ Recommended.*.JsonInput.FieldNameDuplicateDifferentCasing2 # Should have failed 
 
 # Tracking in #2132
 Required.*.JsonInput.TimestampJsonInputNonLeapFeb29 # Should have failed to parse, but didn't.
-Required.*.JsonInput.TimestampJsonInputSecondTooLarge # Should have failed to parse, but didn't.
--- a/Sources/SwiftProtobuf/Google_Protobuf_Timestamp+Extensions.swift
+++ b/Sources/SwiftProtobuf/Google_Protobuf_Timestamp+Extensions.swift
@@ -107,7 +107,7 @@ private func parseTimestamp(s: String) throws -> (Int64, Int32) {
 
     // Second: 2 digits (following char is checked below)
     let second = try fromAscii2(value[17], value[18])
-    if second > Int(61) {
+    if second > Int(59) {
         throw JSONDecodingError.malformedTimestamp
     }
 


thomasvl:
@tbkka - git history seems to say this was always testing this way, do you recall a reason for that?

allevato (approved):

tbkka:
UTC allows up to 62 seconds per minute (for two specific minutes, the last minute of June and the last minute of December).  If protobuf disagrees with UTC on this point, we should change it.

thomasvl:
> UTC allows up to 62 seconds per minute (for two specific minutes, the last minute of June and the last minute of December). If protobuf disagrees with UTC on this point, we should change it.

Is there a spec for that?  Most everything I've found so far when talking about leap seconds talks about 1 in June and/or December. So 2 on the year, but never more than 61 seconds in a minute. But I'm not finding an actual spect/rfc for reference (and I'm probably just not looking in the right place), its all just writeup by folks with an spec.

I did find: https://en.wikipedia.org/wiki/Coordinated_Universal_Time which does talk about more:

> Some time in the 22nd century, two leap seconds will be required every year. The current practice of only allowing leap seconds in June and December will be insufficient to maintain a difference of less than 1 second, and it might be decided to introduce leap seconds in March and September. In the 25th century, four leap seconds are projected to be required every year, so the current quarterly options would be insufficient.

FYI: The [protobuf change](https://github.com/protocolbuffers/protobuf/commit/f8dfd63322a1b107bb3b386fef293a075388e7da) which references [rfc3339](https://datatracker.ietf.org/doc/html/rfc3339).

tbkka:
Apparently, I was wrong.  According to NIST:

https://www.nist.gov/pml/time-and-frequency-division/time-realization/leap-seconds
> Leap seconds are used to keep the difference between UT1 and UTC to within ±0.9 s.

which cannot happen with a 2-second adjustment.  So UTC only allows minutes of 59-61 seconds.

thomasvl:
Oops, meant to include the timestamp.proto link about leap seconds and smear: https://github.com/apple/swift-protobuf/blob/main/Protos/Sources/SwiftProtobuf/google/protobuf/timestamp.proto#L49-L51

Thanks.  Given the timestamp.proto comments on leap seconds and the new conformance test, I think we likely want this change then. Does that make sense to you?

tbkka:
We should certainly follow protobuf conventions here.

tbkka (approved):

PR #2140, opened 2026-08-04
Reduce closure overhead in generated message equality

karim-alweheshy:
## Summary

- generate message equality without wrapping all field comparisons in a capturing `withExtendedLifetime` closure
- keep the storage identity fast path, field order, unknown-field comparison, extension comparison, and specialized storage equality unchanged
- regenerate the checked-in Swift sources and reference output

The generated equality function now holds both storage objects in strong local variables and uses deferred `withExtendedLifetime` barriers. This preserves the copy-on-write storage lifetime across every early return without materializing a tuple, closure context, and closure result around the field comparisons.

## Motivation

Generated equality is emitted once per message type. For messages with many fields, the closure form adds repeated setup and control-flow scaffolding that survives optimized builds. A reduced eight-field probe changed as follows with the same compiler and linker inputs:

| Measure | Existing generator | This change | Delta |
|---|---:|---:|---:|
| Equality symbol | 848 B | 796 B | -52 B (-6.1%) |
| Call instructions | 36 | 25 | -11 (-30.6%) |
| Executable `__text` | baseline | baseline -96 B | -96 B |
| Gzip-compressed executable | baseline | baseline -173 B | -173 B |

These are reduced-probe code-generation measurements, not an application-wide size or runtime claim.

## Correctness

The comparisons themselves are unchanged, including their order and Swift equality semantics. In particular, this does not replace equality with serialized-byte or hash equality, which would be incorrect for floating-point values, map ordering, unknown fields, and extensions.

- storage identity still short-circuits field comparison
- each field still uses the existing generated comparison
- unknown fields and extensions are still compared after storage fields
- specialized `storageProvidesEqualTo` messages still call the existing storage equality implementation
- deferred lifetime barriers run on success and on every early return

## Validation

- `make build test-runtime test-plugin` — 967 tests passed
- `make compile-tests` — all compile-test projects passed
- `make reference` — regenerated reference output matches test generation
- `swift format lint --strict Sources/protoc-gen-swift/MessageGenerator.swift`
- mechanical generated-diff audit: 199 tuple-closure equality bodies became 199 strong-local/deferred-lifetime bodies; the three specialized storage equality cases only lost the redundant temporary Boolean
- `git diff --check`

**Decision:** the current design is not merge-ready. I am keeping this PR draft/open while redesigning around a production-shaped reproducer that preserves the closure form's favorable full-graph code shape.


karim-alweheshy:
I am keeping this PR draft/open because the current design is not merge-ready. I will replace the reduced probe with a public, production-shaped reproducer before requesting review.

PR #2141, opened 2026-08-13
docs updates

heckj:
Broad sweep of documentation cleanup

- capitalization consistency - Git, GitHub, Protocol Buffers, and protobuf for shorthand
- reworking 'will' to present tense sentences
- converting passive voice sentences to active voice, present tense
- removing latin abbreviations, replacing with english prose
- use sentence casing for headings, and use backticks for code voice consistently across markdown docs (aka in Documentation/)
- fix symbol links, ArrayValue -> ListValue updates
- fix out of date references, clean up style nits
- grammar fixes for markdown details
- consolidate deltas in API and PLUGIN docs using Documentation/ as the most recently updated source of truth, fill in a missing section in API.md in Documentation/ that was in the DocC catalog
- adds missing abstracts (where not sourced from upstream .proto files)
- expand SwiftPackageIndex hosting of docs content for both libraries provided by the package


tbkka:
This looks great!

heckj:
state at the moment, rendered locally:

SwiftProtobuf

SwiftProtobufPluginLibrary

In terms of coverage, it's fairly significantly expanded:

### SwiftProtobuf

| | Types | Members | Globals |
|---|---|---|---|
| Before | 54% (126/234) abstract / 58% curated | 76% (3469/4573) abstract / 54% curated | 0.40% (1/253) abstract / 52% curated |
| After | 84% (196/234) abstract / 76% curated | 87% (3972/4573) abstract / 91% curated | 32% (81/253) abstract / 72% curated |

Overall abstract coverage: **71% → 84%**

### SwiftProtobufPluginLibrary

| | Types | Members | Globals |
|---|---|---|---|
| Before | 80% (33/41) abstract / 68% curated | 44% (294/672) abstract / 34% curated | 14% (6/42) abstract / 45% curated |
| After | 89% (39/44) abstract / 77% curated | 53% (361/678) abstract / 59% curated | 25% (11/44) abstract / 52% curated |

Overall abstract coverage: **44% → 54%**

The main libraries have organiztion/curation for the types to make them easier for someone scanning/reading online as well as through IDEs that expose DocC content/abstracts during development (both Xcode and VSCode with the Swift extension, perhaps others - I only checked the first two)

Some of the doc content exposed comes from code generation from upstream sources (the `.proto` files) and in those cases, I didn't attempt to override or overload, and used what's sourced from upstream faithfully.

.jazzy.yaml
thomasvl, line 1: I think there are still jazzy references in the Makefile, might as well drop those if removing the config.

thomasvl (commented):
Is there any automation that can be used for some of this?  especially the pages that list all the types and methods? It seems like those can easily go stale if there isn't a good way to ensure new things get added and/or updates are made if things get deprecated.

heckj:
@thomasvl - I stripped out the Jazzy references in the Makefile. For the automation, none that's dramatically effective. There's some capabilities within DocC to provide that default "curation" - the list of all the types, methods, etc - but it's not very forward and what's there is set up primarily for someone who has a large library with a lot of symbols (like this), but doesn't have anything in place. In that case, there's an experimental command in DocC to generate the DocC catalog.

But more generally, DocC does the work to make sure nothing "silently drops" when API is added - it automatically includes the content under generic sections as a sort of fall back while keeping the reference content complete, if not as nicely organized. At this point, the automation I've been reaching for has been LLM-based to review and look for gaps, and help resolve them. DocC does support some analytics output (doc coverage, etc) - which is where those numbers came from above. I temporarily added the `swift-docc-plugin` (intentionally not added with this PR)  and used the command `swift package generate-documentation` with `--experimental-documentation-coverage` and then `--coverage-summary-level detailed` for a rather absurdly long and detailed list, but which is really helpful to spot what's covered vs. not in terms of curation, and then I dig based on that - or have an LLM help me do so, for where more needs to be written or something updated. It's by no means perfect, but it's done a pretty good job.

This particular repo was a smidge challenging in the content that's replicated in `docs/` and into the DocC catalog under `sources`, so actually that was my largest concern. Unfortunately, we don't have a great way to "include" markdown from an external location into DocC super easily, but it looked like earlier contributions tended to update what was in the README/GFM markdown in the `docs/` directory - so I used that as the primary source of truth when the two diverged (checking against the code and implementation content), and tried to make sure the two were in sync.

heckj:
rebased over main, and conflicting (new) property resolved

thomasvl:
I'm deferring to the Apple folks on most of this, I haven't done much with docc, so don't really know what to double check.

glbrntt (approved):
Thanks Joe!

PR #2143, opened 2026-08-19
[table-driven-branch] Add support for future `nil` schema returns for enum/message fields.

allevato:
In the future to support weak linking between protos, schema lookup may involve dynamic symbol lookup, with a `nil` result if the linker has dropped unused protos. How we deal with this in the runtime depends on the operation being performed:

*   For binary encoding/decoding, dropped schemas fold neatly into unknown fields. This is the easy case.
*   For operations that expect a value of the concrete type to exist in memory (copy, deinit, text/JSON encoding), for that value to exist, the schema could *not* have been dropped by the linker. So a missing schema in these cases is a contradiction and therefore a precondition failure.
*   For text and JSON decoding, a missing schema means we don't know what enum value names/submessage field names would be expected by the value being decoded, so we have to throw an error.


thomasvl (approved):
Questions:

If a message has a _direct_ required field, should it some how force a reference for the enum/message schema of that field to ensure it is there for decoding/validation?

If a message has a _transitive_ field that reaches *required*, should it force a reference to its field that is the path to that transitive field? That would then chain, and ensure that if you use a message, all transitive messages needed to get that required field parse/validation are there.

Thought:

I do wonder if we could eventually generate something as part of JSON/TextFormat apis that would force a reference to all the direct types of fields (that would then chain), so if you did use the JSON/TextFormat apis, it would force the references so their respective closures would be forced in to support the parsing.

allevato:
> If a message has a _direct_ required field, should it some how force a reference for the enum/message schema of that field to ensure it is there for decoding/validation?

That seems reasonable, and would be easy enough to handle during generation. My local draft just updates the generator to generate a call to a runtime helper function that manages the `dlsym` lookup without considering presence, but we could tweak that behavior to say that if a field is required, then we just emit the direct symbol reference for the submessage/enum instead.

> If a message has a _transitive_ field that reaches _required_, should it force a reference to its field that is the path to that transitive field? That would then chain, and ensure that if you use a message, all transitive messages needed to get that required field parse/validation are there.

Do you mean if you have something like this:

```protobuf
message Foo {
  optional Bar bar = 1;
}
message Bar {
  required int64 baz = 2;
}
```

then `Foo` should have a strong reference to `Bar` because `Bar` has a required field? I would think no, because a client of `Foo` might never read or set the `bar` field, so they'd want it to be stripped. But you're right that we'd potentially lose presence validation since a present `bar` would be shunted into unknown fields. Do any other implementations say anything about this?

> I do wonder if we could eventually generate something as part of JSON/TextFormat apis that would force a reference to all the direct types of fields (that would then chain), so if you did use the JSON/TextFormat apis, it would force the references so their respective closures would be forced in to support the parsing.

I'm planning on having weak-linking be a generator option for now. My gut feeling is that if you're relying on text format/JSON heavily, you'd need to just not use that option. But for build systems where you may need to decide that at a more granular level, we'd need to think of a better approach that doesn't require users to manually manage lists of message metatype symbols to force a strong link. I'll have to think about that some more.

thomasvl:
> Do any other implementations say anything about this?

Some internal (to Google) tooling does this.

> ... My gut feeling is that if you're relying on text format/JSON heavily, you'd need to just not use that option. ...

My fear is in the complex app, someone might not realize something they depend on use JSON requests vs. binary. So it will be a lurking runtime failure for them that only shows up when the exercise the feature that needs it.

thomasvl:
Looks like upstream has a problem, triggering the upstream conformance on `main` has the same problem: https://github.com/apple/swift-protobuf/actions/runs/32294163672/job/96201526432

allevato:
> Some internal (to Google) tooling does this.

I dug into what Java and C++ do.

Java (on Android, with AppReduce) is conservative and won't prune anything that transitively reaches a `required` field. So `isInitialized` validation still works there.

C++ with weak fields goes for maximum size reduction, replacing pruned messages with placeholders and putting the data in unknown fields, bypassing `isInitialized` validation.

Since our goals here are also for maximum size reduction, and since required fields are strongly discouraged these days, and since weak linkage will be an opt-in feature, I'm comfortable pointing to the C++ precedent and doing the same.

> > ... My gut feeling is that if you're relying on text format/JSON heavily, you'd need to just not use that option. ...
> 
> My fear is in the complex app, someone might not realize something they depend on use JSON requests vs. binary. So it will be a lurking runtime failure for them that only shows up when the exercise the feature that needs it.

Yeah, I'll keep thinking about ways to thread this needle better. Maybe some sort of registry is needed.

PR #2144, opened 2026-08-20
[table-driven-branch] Initial check-in of a compile test for weak imports.

allevato:
This checks in a multi-module example that we can build on to test inter-module weak imports when that support is added to the generator and runtime.

Also, add a small harness to do symbol linkage verification. We'll need to verify that symbols are either present or stripped from the binary, so the script in the second commit uses an LLVM FileCheck-inspired convention to let us annotate the source file and then run `nm` on the binary to check the symbols.


thomasvl:
https://github.com/apple/swift-protobuf/blob/main/.github/workflows/build.yml#L63

You'll likely need something like:

```yml
  run: apt-get install -y python3 && make compile-tests
```

thomasvl (approved):

allevato:
> https://github.com/apple/swift-protobuf/blob/main/.github/workflows/build.yml#L63
> 
> You'll likely need something like:
> 
> ```yaml
>   run: apt-get install -y python3 && make compile-tests
> ```

Now that the compile tests are in a separate job, I added it to the `apt-get install` line for those.

PR #2148, opened 2026-08-21
Run CompileTests in separate GitHub CI jobs.

allevato:
These tests are separate "sub-packages" in the repo, so when they're built, they each independently rebuild the SwiftProtobuf runtime (and in some cases, protoc and the plugin). Move them into their own separate checks so they run in parallel with the main build.


thomasvl:
@glbrntt @FranzBusch @Lukasa - one of you all will need likely need to update the required checks to include these new checks.

thomasvl (approved):

glbrntt:
> @glbrntt @FranzBusch @Lukasa - one of you all will need likely need to update the required checks to include these new checks.

Done!

PR #2153, opened 2026-08-21
[table-driven-branch] Add our own `Lock` that doesn't require Foundation.

allevato:
This is adapted from `NIOLock` in swift-nio, but using `os_unfair_lock` on Apple platforms instead of `pthread_mutex`.

Also re-enable the no-Foundation linkage test.

### DIFF
--- a/Sources/SwiftProtobuf/ReflectionTableReference.swift
+++ b/Sources/SwiftProtobuf/ReflectionTableReference.swift
@@ -51,14 +51,7 @@ final class ReflectionTableReference: @unchecked Sendable {
     /// Calls the given body with the reflection table, decompressing it on the
     /// first call if needed.
     func withTable<R>(_ body: (borrowing ReflectionTable) throws -> R) rethrows -> R {
-        // Fast path (lock-free): if we already have the decompressed table, use it.
-        if let table = cachedTable {
-            return try body(table)
-        }
-
-        return try lock.withLock {
-            // Second chance: another thread may have decompressed the table while
-            // we were waiting for the lock.
+        try lock.withLock {
             if let table = cachedTable {
                 return try body(table)
             }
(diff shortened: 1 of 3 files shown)


Sources/SwiftProtobuf/ReflectionTableReference.swift
```diff
@@ -13,60 +13,66 @@
...
 
     /// Calls the given body with the reflection table, decompressing it on the
     /// first call if needed.
     func withTable<R>(_ body: (borrowing ReflectionTable) throws -> R) rethrows -> R {
-        try lock.withLock {
-            switch state {
-            case .direct(let table):
-                return try body(table)
+        // Fast path (lock-free): if we already have the decompressed table, use it.
+        if let table = cachedTable {
```
thomasvl: How is this safe?  without an atomic, I didn't this variable assigned/fetching was atomic, and since this is an optional, doesn't that make an an enum with associated value under the hood or something like that so not even just a pionter?

Sources/SwiftProtobuf/ReflectionTableReference.swift
```diff
@@ -13,60 +13,66 @@
...
+            return try body(table)
+        }
 
-            case .compressed(let buffer, let fieldCount):
-                let table = ReflectionTable(
-                    fieldCount: fieldCount,
-                    data: Compression.decompress(buffer)
-                )
-                state = .direct(table)
+        return try lock.withLock {
```
thomasvl: Maybe use a local to fetch the result of `.withLock` and return the table so you can invoke `body` after releasing the lock instead of holding the lock longer?

Sources/SwiftProtobuf/ReflectionTableReference.swift
allevato: Yeah, it turns out it very much wasn't safe. I was hoping I could get away with a fast-path for the read since it's immutable once allocated, but this could have experienced tearing unless we want to introduce a stdatomic.h shim or depend on swift-atomics (since we need to support Apple OS versions before `Synchronization` was introduced).

I'm just taking the lock unconditionally now.

Sources/SwiftProtobuf/ReflectionTableReference.swift
allevato: Done.

thomasvl (approved):

PR #2154, opened 2026-08-22
Add support for strict memory safety

### LINKED ISSUE #2087: Strict memory safety warnings
When built with [strict memory safety](https://docs.swift.org/compiler/documentation/diagnostics/strict-memory-safety/), a `.pb.swift` file containss code that triggers memory safety warnings. The code is:

``` swift
fileprivate var _storage = _StorageClass.defaultInstance
```

Each instance of that code produces this warning:
> - warning: expression uses unsafe constructs but is not marked with 'unsafe' [#StrictMemorySafety]
> - note: reference to nonisolated(unsafe) static property 'defaultInstance' is unsafe in concurrently-executing code

protoc can prevent the warning by being explicit that the code is unsafe:

``` swift
fileprivate var _storage = unsafe _StorageClass.defaultInstance
```

Owain94:
Closes #2087


thomasvl:
Normally tests like this go in CompileTests instead so they aren't something most users of the Package won't ever have to deal with, can you look at moving it there like some of the other compile tests?

Owain94:
> Normally tests like this go in CompileTests instead so they aren't something most users of the Package won't ever have to deal with, can you look at moving it there like some of the other compile tests?

My bad, that makes a lot more sense yeah

allevato (commented):
So it looks like the only place where memory unsafety leaks into the generated code is in the reference to the `nonisolated(unsafe)` default instance of storage classes?

I think it would be even better if we could make the generated code compile without any changes. One way to do that would be to stop using the singleton default and instead just have the generated message call the initializer directly (its initializer would have to be made `fileprivate`).

But that would incur a minor runtime performance hit because creating a new message will incur that allocation instead of just retaining the shared default.

I'm not sure which is better, but I think we should think about the implications of both choices here.

glbrntt:
> But that would incur a minor runtime performance hit because creating a new message will incur that allocation instead of just retaining the shared default.

Only if that message is never written to. If it is written to the allocation is just moved from the first CoW to construction time. 

You've mentioned large messages before with few fields being touched, so the additional allocations could have an impact in those cases.

> I'm not sure which is better, but I think we should think about the implications of both choices here.

It'd be nice to not have `unsafe`, but as it's not in the user-facing API I would bias towards not regressing the runtime perf. Also, this shared-empty-instance pattern is widely used so I'm not terribly concerned about the use of `unsafe` here.

thomasvl:
@glbrntt do you want to move this forward?

Owain94:
Happy to look into different solutions as well if preferred @glbrntt

glbrntt:
> @glbrntt do you want to move this forward?

Yeah, I think this approach is fine. 

One thing to keep in mind: when Swift 6.4 is released we can drop support for 6.1 which means we can drop the conditional in the generated code. 

Should we just wait until then rather than churn the generated code twice in relatively short order?

thomasvl:
> Should we just wait until then rather than churn the generated code twice in relatively short order?

Have quickly after release is nio likely to drop 6.4, since I think we've lagged that a little here.

glbrntt:
> > Should we just wait until then rather than churn the generated code twice in relatively short order?
> 
> Have quickly after release is nio likely to drop 6.4, since I think we've lagged that a little here.

Typically adding a new version is faster than dropping the old one but there's no reason that needs to be the case. New CI tends to be added pretty quickly, within a week or so.

thomasvl:
Since it's been a while since we did a release, and there is a new major version of protoc we maybe should include in a release, I'll go ahead and trigger the test to make sure things are good.

thomasvl:
Since it isn't that much work to come back and drop the conditionals, should we just go forward with this to get it in a release?

@FranzBusch @Lukasa

Owain94:
> Since it isn't that much work to come back and drop the conditionals, should we just go forward with this to get it in a release?
> 
> @FranzBusch @Lukasa

I can come back when swift 6.1 is dropped to remove the conditionals. Keeping this open until then is also fine of course :)

PR #2155, opened 2026-08-25
[table-driven-branch] Add option to generate `@_weakLinked import`.

allevato:


thomasvl (approved):

allevato:
Interesting that two of the symbol checks behaved differently on macOS vs. Linux, but I guess that's to be expected. Removed those checks for now; once the implementation is further along, I'll decide if we need any platform-conditional constraints in the checker script.

PR #2158, opened 2026-08-28
[table-driven-branch] Add runtime support for looking up weak schemas and generate those lookups.

allevato:
More progress toward weak protos.

I've punted on map fields for the time being because those need a little bit more care; the interaction with `MessageWitness` today means we have to put the map entry's value type into it as a generic argument, which blows away the savings. Options there may include either wedging in a placeholder type (like the "implicit weak message" that C++ uses).

### DIFF
--- a/CompileTests/WeakImports/Sources/Client/Client.swift
+++ b/CompileTests/WeakImports/Sources/Client/Client.swift
@@ -27,6 +27,7 @@ struct Main {
             // HAS-SYMBOL: type metadata accessor for ModuleA.Test_MessageA
             // HAS-SYMBOL: type metadata for ModuleA.Test_MessageA
             // HAS-SYMBOL: ModuleA.Test_MessageA.init() -> ModuleA.Test_MessageA
+            // HAS-SYMBOL: _test_MessageA_getMessageSchema
             var msg = Test_MessageA()
             msg.title = "Hello Weak Imports"
             expect(msg.hasTitle)
@@ -40,6 +41,7 @@ struct Main {
             // HAS-SYMBOL: nominal type descriptor for ModuleC.Test_MessageC
             // HAS-SYMBOL: type metadata accessor for ModuleC.Test_MessageC
             // HAS-SYMBOL: type metadata for ModuleC.Test_MessageC
+            // HAS-SYMBOL: _test_MessageC_getMessageSchema
             msg.nestedC.id = 12345
             expect(msg.hasNestedC)
             expect(msg.nestedC.id == 12345)
@@ -62,7 +64,8 @@ struct Main {
         // particular pattern end up in the final linkage.
         //
         // Protobuf runtime support:
-        //   HAS-SYMBOL: ModuleB.Test_MessageB.messageSchema.unsafeMutableAddressor : SwiftProtobuf.MessageSchema
+        //   HAS-SYMBOL: _test_MessageB_getMessageSchema
+        //   HAS-SYMBOL-NOT: ModuleB.Test_MessageB.messageSchema.unsafeMutableAddressor : SwiftProtobuf.MessageSchema
         //   HAS-SYMBOL: static ModuleB.Test_MessageB.messageSchema : SwiftProtobuf.MessageSchema
         //
         // Protocol conformance support:
--- a/CompileTests/WeakImports/Sources/Client/Client.swift
+++ b/CompileTests/WeakImports/Sources/Client/Client.swift
@@ -27,7 +27,7 @@ struct Main {
             // HAS-SYMBOL: type metadata accessor for ModuleA.Test_MessageA
             // HAS-SYMBOL: type metadata for ModuleA.Test_MessageA
             // HAS-SYMBOL: ModuleA.Test_MessageA.init() -> ModuleA.Test_MessageA
-            // HAS-SYMBOL: _test_MessageA_getMessageSchema
+            // HAS-SYMBOL: _test_DMessageA_getMessageSchema
             var msg = Test_MessageA()
             msg.title = "Hello Weak Imports"
             expect(msg.hasTitle)
@@ -41,7 +41,7 @@ struct Main {
             // HAS-SYMBOL: nominal type descriptor for ModuleC.Test_MessageC
             // HAS-SYMBOL: type metadata accessor for ModuleC.Test_MessageC
             // HAS-SYMBOL: type metadata for ModuleC.Test_MessageC
-            // HAS-SYMBOL: _test_MessageC_getMessageSchema
+            // HAS-SYMBOL: _test_DMessageC_getMessageSchema
             msg.nestedC.id = 12345
             expect(msg.hasNestedC)
             expect(msg.nestedC.id == 12345)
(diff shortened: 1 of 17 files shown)


Sources/SwiftProtobuf/WeakLinkageSupport.swift
```diff
@@ -0,0 +1,77 @@
...
+
+#if canImport(Darwin)
+private var defaultDynamicLibraryHandle: UnsafeMutableRawPointer? {
+    UnsafeMutableRawPointer(bitPattern: -2)
+}
+#elseif canImport(Glibc) || canImport(Musl) || canImport(Bionic)
+private var defaultDynamicLibraryHandle: UnsafeMutableRawPointer? {
+    UnsafeMutableRawPointer(bitPattern: 0)
+}
+#endif
```
thomasvl, line 34: else error?

Sources/SwiftProtobuf/WeakLinkageSupport.swift
```diff
@@ -0,0 +1,77 @@
...
+@preconcurrency import Glibc
+#elseif canImport(Musl)
+@preconcurrency import Musl
+#elseif canImport(Bionic)
+@preconcurrency import Bionic
+#endif
+
+#if canImport(Darwin)
+private var defaultDynamicLibraryHandle: UnsafeMutableRawPointer? {
+    UnsafeMutableRawPointer(bitPattern: -2)
```
thomasvl: Seems like `-2` might be magic enough to be worthy of a comment?

allevato:
Aside from the obvious `make reference` and format failures, I need to figure out why the WeakImports compile test is crashing at runtime on Linux. The linker might be doing something different there, since I don't see it on macOS.

Sources/SwiftProtobuf/WeakLinkageSupport.swift
```diff
@@ -0,0 +1,77 @@
...
+    /// Dynamically looks up a message schema based on the name of a generated
+    /// accessor function.
+    ///
+    /// Returns nil if the symbol is not found (i.e., if it has been dropped by the
+    /// linker).
+    @_spi(ForGeneratedCodeOnly)
+    public static func resolveLazy(named symbolName: String) -> MessageSchema? {
+        // TODO: Put a cache around this.
+        #if canImport(Darwin) || canImport(Glibc) || canImport(Musl) || canImport(Bionic)
+        guard let symbol = dlsym(defaultDynamicLibraryHandle, symbolName) else { return nil }
```
thomasvl: Will this work for the support platforms if they link to a framework/shared library?

Sources/protoc-gen-swift/EnumGenerator.swift
```diff
@@ -119,6 +119,26 @@ class EnumGenerator {
             )
         }
         p.print("}")
+
+        if generatorOptions.experimentalWeakImports {
+            let getterSymbol = namer.dynamicSymbolName(
+                forProtoFullName: enumDescriptor.fullName,
+                suffix: "_getEnumSchema"
+            )
+            let spiSnippet = generatorOptions.visibilitySourceSnippet.contains("public") ? "@_spi(ForGeneratedCodeOnly)\n" : ""
```
thomasvl: Seems like just checking `visibility` for the enum value might be safer?

Sources/protoc-gen-swift/SwiftProtobufNamer+Extensions.swift
```diff
@@ -43,4 +43,10 @@ extension SwiftProtobufNamer {
             return aliasInfo.aliases(aliasOf)![firstAlias!] === $0
         }
     }
+
+    /// Returns the symbol name to use for dynamic schema lookups
+    // (`@_silgen_name`) for the given proto full name and suffix.
+    package func dynamicSymbolName(forProtoFullName protoFullName: String, suffix: String) -> String {
```
thomasvl, line 49: I worry a little here about collisions given we see lots of proto packages like "foo.bar.baz" & "foo_bar_baz".

The Namer does with this by doing different things on dots vs. underscores, I wonder if to help reduce collisions we should so something like the namer as the prefix?  (Extension could take more thought though).

Sources/SwiftProtobuf/WeakLinkageSupport.swift
allevato: Added a comment. Also went ahead and put the full matrix of values across multiple platforms, since FreeBSD and OpenBSD actually also use -2, not 0. (This matches what swift-testing has in their package.)

Sources/SwiftProtobuf/WeakLinkageSupport.swift
allevato: That would make this a compiler error in the runtime on any platform where we don't support weak linking at all (like WASM). Right now the functions just gracefully return nil (and you shouldn't use this experimental flag on those platforms anyway).

Sources/SwiftProtobuf/WeakLinkageSupport.swift
allevato: `RTLD_DEFAULT` should scan all loaded images.

Sources/protoc-gen-swift/SwiftProtobufNamer+Extensions.swift
allevato: Since these functions only exist to be called at runtime, the symbol names don't have to look "pretty", so I've made a completely unambiguous mangling for these: replacing `_` with `__` and `.` with `_D`.

Sources/protoc-gen-swift/EnumGenerator.swift
allevato: Didn't realize that was exposed directly! Done.

allevato (commented):
Still need to figure out the runtime crash on Linux, but pushing all the other changes first.

allevato:
🙏🏻  Finally. Had to fix up the expected mangling for C functions (no leading underscore on Linux) and add `-export-dynamic` when linking the Client binary. This should be good now.

CompileTests/WeakImports/Sources/Client/Client.swift
```diff
@@ -27,7 +27,7 @@ struct Main {
             // HAS-SYMBOL: type metadata accessor for ModuleA.Test_MessageA
             // HAS-SYMBOL: type metadata for ModuleA.Test_MessageA
             // HAS-SYMBOL: ModuleA.Test_MessageA.init() -> ModuleA.Test_MessageA
-            // HAS-SYMBOL: _test_DMessageA_getMessageSchema
+            // HAS-SYMBOL: {{_?}}test_DMessageA_getMessageSchema
```
thomasvl, line 34: Is it worth a comment about why you need the leading underscore to be optional?

CompileTests/WeakImports/Sources/Client/Client.swift
```diff
@@ -64,8 +64,7 @@ struct Main {
         // particular pattern end up in the final linkage.
         //
         // Protobuf runtime support:
-        //   HAS-SYMBOL: _test_DMessageB_getMessageSchema
-        //   HAS-SYMBOL-NOT: ModuleB.Test_MessageB.messageSchema.unsafeMutableAddressor : SwiftProtobuf.MessageSchema
```
thomasvl: Why drop this one?

thomasvl (approved):

CompileTests/WeakImports/Sources/Client/Client.swift
allevato: It differed on Linux and macOS (it was present on Linux, not present on macOS). I'm less worried about this one as the schema itself, so I'll decide if it's relevant when I start whittling these down to broader regex matches.

`unsafeMutableAddressor` is basically just a small function that returns the address of a global/static stored property, so checking for the parent property ought to be sufficient; if the parent property isn't there, then the addressor wouldn't be either.

CompileTests/WeakImports/Sources/Client/Client.swift
allevato: Done.

thomasvl (approved):

PR #2159, opened 2026-08-29
Add useDeterministicOrdering to TextFormatEncodingOptions

### LINKED ISSUE #1938: Add `useDeterministicOrdering` to `TextFormatEncodingOptions`
While re-implementing text format serialization in the table-driven branch, I noticed that `TextFormatEncodingOptions` doesn't support `useDeterministicOrdering`; binary and JSON formats do.

For completeness, we should add it to text format encoding as well.

shubhransh-gupta:
### Motivation:

Currently, `BinaryEncodingOptions` and `JSONEncodingOptions` support `useDeterministicOrdering: Bool = false` to deterministically sort map entries in lexicographical key order during serialization, whereas `TextFormatEncodingOptions` lacked this property (Fixes #1938).

### Modifications:

1. **`TextFormatEncodingOptions`**:
   - Added `public var useDeterministicOrdering: Bool = false` with doc comments explaining deterministic serialization semantics.
2. **`TextFormatEncodingVisitor`**:
   - Updated `iterateAndEncode(map:fieldNumber:isOrderedBefore:encode:)` to conditionally sort map entries when `options.useDeterministicOrdering` is `true`, and iterate over the map directly when `false` (aligning with Binary and JSON visitors).
3. **`FuzzCommon/Options.swift`**:
   - Added `.boolean(\.encoding.useDeterministicOrdering)` to `TextFormatFuzzingOptions.fuzzOptionsList`.
4. **Tests**:
   - Added `Test_TextFormatEncodingOptions.swift` testing deterministic key ordering across string, message, and enum maps, as well as `printUnknownFields`.
   - Updated test helpers and tests to pass `options.useDeterministicOrdering = true` where deterministic map ordering is explicitly asserted.

### Result:

`TextFormatEncodingOptions` now supports `useDeterministicOrdering`. All 969 unit tests pass with zero errors.


thomasvl:
Since we are sorting currently, I was pretty sure there was a reason for it.

Looks like there is a conformance test for it: https://github.com/protocolbuffers/protobuf/blob/5a2f1f78d073e79703d086caf18e58422a812f2d/conformance/text_format_conformance_suite.cc#L755-L770

And the support landed in https://github.com/apple/swift-protobuf/pull/1076 (with a reference to the original issue about it).

So the option likely needs to default to `True` to stay inline with the spec.

shubhransh-gupta:
Good catch! You're completely right — the Protocol Buffers conformance test suite and spec require deterministic map key ordering for TextFormat by default.

I have updated `TextFormatEncodingOptions.useDeterministicOrdering` to default to `true` (and updated the doc comments, fuzzing options, and unit tests accordingly). Changes have been pushed!

thomasvl (approved):
Thank you

PR #2160, opened 2026-08-31
[table-driven-branch] Define `ImplicitWeakMessage` as a placeholder for messages that are decoded from binary but dropped by the linker.

allevato:
This is similar to the C++ implementation which uses a message of the same name as its weak-linked placeholder. Unlike C++, we need to generate ours, because hand-writing the schema/reflection `StaticString`s would be fragile.

I opted not to use `google.protobuf.Empty` as the placeholder, because we may want the flexibility to remove the WKTs from the runtime library one day.

This ends up reverting a bunch of work that was done to make the schema resolver function return an optional, because now we return the placeholder instead and it's never `nil`. Honestly, that's better, because the natural implementation falls out for everything (i.e., text format and JSON will try to look up names in an empty reflection table and throw an error when they aren't found, which is the behavior we want).

Since we store concrete enum values for repeated enums (but not singular), I'm going to need to introduce a similar `ImplicitWeakEnum` that we define as an open enum with just a zero case, and then any others will get tossed into the `UNRECOGNIZED` payload. But I wanted to make sure things were sound with the message version first.

I've added tests in Client.swift to verify the round-tripping and hazzer behavior for a dropped message. Of course, they pass today because the message isn't being dropped yet, but that will be coverage for the future work.


Protos/Sources/SwiftProtobuf/implicit_weak_message.proto
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See LICENSE.txt for license information:
+// https://github.com/apple/swift-protobuf/blob/main/LICENSE.txt
+//
+// -----------------------------------------------------------------------------
+
+edition = "2023";
+
+package swift_protobuf;
```
thomasvl, line 13: We could also use google.protobuf.Empty.  But I guess if we want to split off the WKTs it gets a little fuzzy

thomasvl (approved):

Protos/Sources/SwiftProtobuf/implicit_weak_message.proto
thomasvl: Oops, missed the PR write up.

PR #2162, opened 2026-09-01
Fix the Parameter name on Enum.init?(rawUTF8:)

darkdi:
Reading `Enum.swift` and the callout above `init?(rawUTF8:)` names `name`, which is not an argument there. DocC drops a `- Parameter` it cannot match against the signature, so the line quietly disappears from the rendered page instead of documenting the buffer.


thomasvl (approved):
Thank you!

PR #2163, opened 2026-09-01
Support multiple instances of ProtoPathModuleMappings option

### LINKED ISSUE #1067: Enhancement Request: Add support for multiple instances of the --swift_opt=ProtoPathModuleMappings=PATH flag
## Background

The `--swift_opt=ProtoPathModuleMappings=PATH` flag allows you to specify a file containing the mapping of proto file to Swift module that includes it, so that the module dependency graph can be represented in the generated proto sources.

One pattern we've found ourselves doing is having a multi layer dependency graph of Swift packages where each package contains protos that get generated into a module, but are also available for usage by downstream dependents. The way we do this is by adding `-I=.build/checkouts/DEPENDENCY/Protos` include paths in the proto generation step (sample [Makefile](https://github.com/apple/swift-llbuild2/blob/master/Makefile#L58)). 

When there's a need for a single mapping file, this works great, but as you add more and more dependencies, we've needed to resort to concatenating all of the mappings files before passing it in the flag. 

## Enhancement Request

Ideally, the generator should accept multiple instances of the `ProtoPathModuleMappings` option, so that clients can avoid the external concatenation command.

Created this issue to discuss this particular idea, but also to discuss the general problem of transitive dependencies and generator and see if there's been any thought in this area (i.e. outside of bazel)

shubhransh-gupta:
### Description
Fixes #1067.

Currently, the `ProtoPathModuleMappings=PATH` generation option in `protoc-gen-swift` only supports a single mapping file. In complex or layered multi-package architectures (such as `swift-llbuild2`), developers often have multiple module mapping files across packages and dependencies. Previously, passing multiple `--swift_opt=ProtoPathModuleMappings=...` flags would cause `GeneratorOptions` to overwrite `moduleMapPath` with the last value provided, forcing users to manually concatenate mapping files before running `protoc`.

This PR:
1. **Adds Multi-Path Initializers to `ProtoFileToModuleMappings`**:
   - Adds `init(paths: [String], swiftProtobufModuleName: String? = nil)` to load and merge module mappings from multiple files on disk.
   - Adds `init(moduleMappingsProtos: [SwiftProtobuf_GenSwift_ModuleMappings], swiftProtobufModuleName: String? = nil)` to merge multiple mapping message structures in-memory.
   - Preserves all collision detection and error handling (e.g. `LoadError.duplicateProtoPathMapping` if a proto file is assigned to different modules across files).
2. **Updates `GeneratorOptions` in `protoc-gen-swift`**:
   - Accumulates multiple `ProtoPathModuleMappings` arguments into `moduleMapPaths: [String]` rather than overwriting.
3. **Updates Documentation**:
   - Documents multi-mapping support in `Documentation/PLUGIN.md` and `Sources/protoc-gen-swift/Docs.docc/index.md`.
4. **Adds Unit Tests**:
   - Adds test cases for multiple proto/file loading, disjoint merge, identical mapping overlap (allowed), and conflicting mappings (`duplicateProtoPathMapping`) in `Test_ProtoFileToModuleMappings.swift`.
   - Adds test case in `Test_GeneratorOptions.swift` verifying multiple `ProtoPathModuleMappings` parameters.

### Testing
- Ran `swift test --filter Test_ProtoFileToModuleMappings` (all 9 tests passed).
- Ran `swift test --filter Test_GeneratorOptions` (all 2 tests passed).

### DIFF
--- a/Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
+++ b/Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
@@ -52,26 +52,59 @@ public struct ProtoFileToModuleMappings {
     /// Loads and parses the given module mapping from disk.  Raises LoadError
     /// or TextFormatDecodingError.
     public init(path: String) throws {
-        try self.init(path: path, swiftProtobufModuleName: nil)
+        try self.init(paths: [path], swiftProtobufModuleName: nil)
     }
 
     /// Loads and parses the given module mapping from disk.  Raises LoadError
     /// or TextFormatDecodingError.
     public init(path: String, swiftProtobufModuleName: String?) throws {
-        let content: String
-        do {
-            content = try String(contentsOfFile: path, encoding: String.Encoding.utf8)
-        } catch {
-            throw LoadError.failToOpen(path: path)
-        }
+        try self.init(paths: [path], swiftProtobufModuleName: swiftProtobufModuleName)
+    }
 
-        let mappingsProto = try SwiftProtobuf_GenSwift_ModuleMappings(textFormatString: content)
-        try self.init(moduleMappingsProto: mappingsProto, swiftProtobufModuleName: swiftProtobufModuleName)
+    /// Loads and parses the given module mappings from disk.  Raises LoadError
+    /// or TextFormatDecodingError.
+    public init(paths: [String]) throws {
+        try self.init(paths: paths, swiftProtobufModuleName: nil)
+    }
+
+    /// Loads and parses the given module mappings from disk.  Raises LoadError
+    /// or TextFormatDecodingError.
+    public init(paths: [String], swiftProtobufModuleName: String?) throws {
+        var merged = SwiftProtobuf_GenSwift_ModuleMappings()
+        for path in paths {
+            let content: String
+            do {
+                content = try String(contentsOfFile: path, encoding: String.Encoding.utf8)
+            } catch {
+                throw LoadError.failToOpen(path: path)
+            }
+
+            let mappingsProto = try SwiftProtobuf_GenSwift_ModuleMappings(textFormatString: content)
+            merged.mapping.append(contentsOf: mappingsProto.mapping)
+        }
+        try self.init(moduleMappingsProto: merged, swiftProtobufModuleName: swiftProtobufModuleName)
     }
 
     /// Parses the given module mapping.  Raises LoadError.
     public init(moduleMappingsProto mappings: SwiftProtobuf_GenSwift_ModuleMappings) throws {
         try self.init(moduleMappingsProto: mappings, swiftProtobufModuleName: nil)
+    }
+
+    /// Parses the given module mappings.  Raises LoadError.
+    public init(moduleMappingsProtos mappingsList: [SwiftProtobuf_GenSwift_ModuleMappings]) throws {
+        try self.init(moduleMappingsProtos: mappingsList, swiftProtobufModuleName: nil)
+    }
+
+    /// Parses the given module mappings.  Raises LoadError.
+    public init(
+        moduleMappingsProtos mappingsList: [SwiftProtobuf_GenSwift_ModuleMappings],
+        swiftProtobufModuleName: String?
+    ) throws {
+        var merged = SwiftProtobuf_GenSwift_ModuleMappings()
+        for mappings in mappingsList {
+            merged.mapping.append(contentsOf: mappings.mapping)
+        }
+        try self.init(moduleMappingsProto: merged, swiftProtobufModuleName: swiftProtobufModuleName)
     }
 
     /// Parses the given module mapping.  Raises LoadError.
--- a/Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
+++ b/Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
@@ -24,13 +24,41 @@ public struct ProtoFileToModuleMappings {
         /// Raised if the path wasn't found.
         case failToOpen(path: String)
         /// Raised if an mapping entry in the protobuf doesn't have a module name.
-        /// mappingIndex is the index (0-N) of the mapping.
-        case entryMissingModuleName(mappingIndex: Int)
+        /// mappingIndex is the index (0-N) of the mapping within the source file or proto.
+        case entryMissingModuleName(mappingIndex: Int, path: String?)
         /// Raised if an mapping entry in the protobuf doesn't have any proto files listed.
-        /// mappingIndex is the index (0-N) of the mapping.
-        case entryHasNoProtoPaths(mappingIndex: Int)
+        /// mappingIndex is the index (0-N) of the mapping within the source file or proto.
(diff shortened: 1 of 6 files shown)


thomasvl (commented):
The validation in the final initializer sorta was built around the fact that that's one path, so it doesn't report a path for where the problem mapping was (missing data, or two entries that conflict).

How much work would it be to reflow things such it could report the input file that had the problem entry/entries being reported?  It might work to move from an initializer to having some sort of "add" api that adds some mappings and takes a reporting path.

thomasvl:
Thinking a little more, and _add_ type api would only be able to report that the error is in that file and maybe something before it. ☹️ So it might take doing the validation on the file itself to report self conflicting issues, and then another audit after the fact to report the overlaps via some extra tracking. Or some way when using the appended lists to pass along what ranges come from what sources and then use that during the validations.

shubhransh-gupta:
Hi @thomasvl,

That makes complete sense. Flattening all entries into a single proto upfront loses the file-origin context, so index-based errors (`entryMissingModuleName`, `entryHasNoProtoPaths`) report the flattened index rather than the file-local index, and duplicate conflicts can't point to the conflicting files.

To resolve this cleanly, we can:
1. **Validate entries per file/proto**: Validate each mapping file as it is decoded so that `entryMissingModuleName` and `entryHasNoProtoPaths` correspond to the local index within that file.
2. **Track mapping sources**: Store `[ProtoFilePath: (moduleName: String, source: String?)]` during aggregation so that when a conflict is encountered, we know exactly which two files (or proto descriptors) conflicted.
3. **Update `LoadError`**:
   - `entryMissingModuleName(mappingIndex: Int, source: String? = nil)`
   - `entryHasNoProtoPaths(mappingIndex: Int, source: String? = nil)`
   - `duplicateProtoPathMapping(path: String, firstModule: String, firstSource: String? = nil, secondModule: String, secondSource: String? = nil)`
   *(with `source` being the file path when loaded from disk, or `nil` when loaded directly from in-memory protos)*.
4. **Update `GeneratorOptions`**: Process the mapping paths sequentially so that file I/O or decoding failures immediately identify the specific `ProtoPathModuleMappings=<path>` that failed.

I will prepare and push the update with source tracking and expanded test coverage.

Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
```diff
@@ -52,60 +80,122 @@ public struct ProtoFileToModuleMappings {
...
-    /// Parses the given module mapping.  Raises LoadError.
+    /// Parses the given module mappings.  Raises LoadError.
+    public init(moduleMappingsProtos mappingsList: [SwiftProtobuf_GenSwift_ModuleMappings]) throws {
+        try self.init(moduleMappingsProtos: mappingsList, swiftProtobufModuleName: nil)
+    }
+
+    /// Parses the given module mappings.  Raises LoadError.
     public init(
-        moduleMappingsProto mappings: SwiftProtobuf_GenSwift_ModuleMappings,
+        moduleMappingsProtos mappingsList: [SwiftProtobuf_GenSwift_ModuleMappings],
```
thomasvl, line 126: Rather than having two initializer that are so close to the same code, can we have the one that takes paths, load the protos and the have both this and that one call an private/internal helper where the loaded protos and list of strings (or an empty list) are pass to do the full initialization by walking the two arrays together to tell if there was an input path? (can likely skip the "add" helper and just inline that code in that one place)

Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
```diff
@@ -52,60 +80,122 @@ public struct ProtoFileToModuleMappings {
...
+    public init(paths: [String]) throws {
+        try self.init(paths: paths, swiftProtobufModuleName: nil)
+    }
+
+    /// Loads and parses the given module mappings from disk.  Raises LoadError
+    /// or TextFormatDecodingError.
+    public init(paths: [String], swiftProtobufModuleName: String?) throws {
+        self.swiftProtobufModuleName = swiftProtobufModuleName ?? defaultSwiftProtobufModuleName
+        var builder = [String: (module: String, sourcePath: String?)]()
+        for (proto, mod) in wktMappings(swiftProtobufModuleName: self.swiftProtobufModuleName) {
```
thomasvl: The wktMappings(swiftProtobufModuleName:) helper likely isn't a good idea any more as it makes a dictionary just to turn around and loop over it and throw it away, so it should probably be dropped and just directly feed into builder.

Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
```diff
@@ -189,3 +279,41 @@ public struct ProtoFileToModuleMappings {
 private func wktMappings(swiftProtobufModuleName: String) -> [String: String] {
     SwiftProtobufInfo.bundledProtoFiles.reduce(into: [:]) { $0[$1] = swiftProtobufModuleName }
 }
+
+extension ProtoFileToModuleMappings.LoadError: CustomStringConvertible {
```
thomasvl, line 263: Why was this needed?

Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
shubhransh-gupta: Updated in `f0e4822e`! Unified both initializers into a common private `init(moduleMappings: [(mapping: SwiftProtobuf_GenSwift_ModuleMappings, path: String?)], swiftProtobufModuleName: String?)`, eliminating duplicate code and removing the static `add` helper.

Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
shubhransh-gupta: Updated in `f0e4822e`! Dropped `wktMappings` completely and seeded `builder` directly from `SwiftProtobufInfo.bundledProtoFiles`.

Sources/SwiftProtobufPluginLibrary/ProtoFileToModuleMappings.swift
shubhransh-gupta: When `GeneratorOptions` catches a `LoadError` while parsing plugin options, it wraps it in `GenerationError.wrappedError(message: ..., error: e)`. Because `GenerationError.description` formats errors via `"\(message): \(error)"`, conforming to `CustomStringConvertible` produces clean, human-readable CLI error messages (e.g. `Failed to open 'path'` or `Duplicate mapping for proto 'a.proto': 'M1' (in 'p1') vs 'M2' (in 'p2')`) rather than the raw enum debug reflection.

thomasvl: I thought it might be. Thank you, guess we never really looked at how the plugins were reporting errors.

thomasvl:
@Lukasa @FranzBusch The api break is the plugin library, and not something anyone was going to catch, are we ok with that, or is there a better way to expand the error reporting here to avoid the break?

PR #2164, opened 2026-09-06
Add the JSON option sources to protoc-gen-swift's CMake target

lenamonj:
`Sources/protoc-gen-swift/CMakeLists.txt` does not list `json_options.pb.swift` and `json_enumvalue_options.pb.swift`, added in #2111, so the CMake build of `protoc-gen-swift` has failed since then on the sources that use them. The two files are added in the position the sibling lists use.

`cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Release && cmake --build build` fails on main and links `protoc-gen-swift` with this change.


thomasvl (approved):
Thank you!

PR #2166, opened 2026-09-11
[table-driven-branch] Add support for weak enums in the generator/runtime.

allevato:
This is effectively the same strategy for messages, where we define a placeholder open proto `enum` that allows values to round-trip successfully when the enum isn't linked into the binary.

(Of course, we still aren't yet at the point where it's actually stripped.)


Sources/SwiftProtobuf/implicit_weak_message.pb.swift
```diff
@@ -30,6 +30,39 @@ fileprivate nonisolated struct _GeneratedWithProtocGenSwiftVersion: ProtobufAPIV
   typealias Version = _2
 }
 
+/// An open enum used as a placeholder when an enum's schema is not present in
+/// the linkage.
+@_spi(ForGeneratedCodeOnly)
+public nonisolated enum SwiftProtobuf_ImplicitWeakEnum: Enum, Swift.CaseIterable {
```
thomasvl, line 36: Actually, should bot of these be generate as `internal` and not `public`?

thomasvl (approved):

Sources/SwiftProtobuf/implicit_weak_message.pb.swift
allevato: Hmm, it's possible. I _thought_ there was a reason they needed to be referenced directly by generated code at one point (and SPI requires `public`), but I'm not seeing those references anymore; just the ones in the runtime. Maybe it's left over from an earlier attempt.

I'll fix that in a follow-up.

